PR #23, opened 2025-05-29
SFT Data Processor Fix

adil-a:
- Fixes a bug in the processor where only a subset of data will be loaded in.
- Fixes SFT loss calculation so it sums across all DP ranks and divides by the number of tokens in the global batch.
- Adds validation loss calculation
- WandB logging
- Functioning grad norm clipping only without TP. This needs to be fixed at some point otherwise TP is essentially not usable under correct training dynamics.

### DIFF
--- a/recipes/finetune.py
+++ b/recipes/finetune.py
@@ -159,19 +159,22 @@ class FinetuneRecipeForNextTokenPrediction(BaseRecipe):
         """
         self.dist_env = build_distributed(self.cfg.get("dist_env", {}))
 
-        self.model_wrapper = None
+        self.device_mesh = None
         distributed_sampler_kwargs = {}
+        model_wrapper = None
         if "distributed" in self.cfg:
-            self.model_wrapper = self.cfg.distributed.instantiate(
+            model_wrapper = self.cfg.distributed.instantiate(
                 world_size=self.dist_env.world_size
             )
             distributed_sampler_kwargs = {
-                "num_replicas": self.model_wrapper.device_mesh["data_parallel"].size(),
-                "rank": self.model_wrapper.device_mesh["data_parallel"].get_local_rank(),
+                "num_replicas": model_wrapper.device_mesh["data_parallel"].size(),
+                "rank": model_wrapper.device_mesh["data_parallel"].get_local_rank(),
                 "shuffle": self.cfg.dataloader.get("shuffle", True),
             }
             if "shuffle" in self.cfg.dataloader:
                 del self.cfg.dataloader.shuffle
+            if hasattr(model_wrapper, 'device_mesh'):
+                self.device_mesh = model_wrapper.device_mesh
 
         torch.manual_seed(self.cfg.get("seed", 42) + self.dist_env.rank)
 
@@ -185,7 +188,7 @@ class FinetuneRecipeForNextTokenPrediction(BaseRecipe):
             )
 
         # Build components
-        self.model = build_model(self.dist_env.device, self.model_wrapper, self.cfg.model)
+        self.model = build_model(self.dist_env.device, model_wrapper, self.cfg.model)
         self.optimizer = build_optimizer(self.cfg.optimizer, self.model)
         self.loss_fn   = build_loss_fn(self.dist_env.device, self.cfg.loss_fn)
         self.dataloader = build_dataloader(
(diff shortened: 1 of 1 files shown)


recipes/finetune.py
```diff
@@ -5,17 +5,44 @@
...
 from nemo_automodel.training.base_recipe import BaseRecipe
 from nemo_automodel.training.step_scheduler import StepScheduler
+import contextlib
 
 
 # ---------------------------
 #  Stateless helper functions
 # ---------------------------
 
+def get_sync_ctx(model, is_optim_step):
```
akoumpa: can we move this under the distributed/ dir?

recipes/finetune.py
```diff
@@ -201,28 +274,112 @@ def _run_train_step(self, batch, is_optim_step, clip_norm=1.0, num_grad_acc_step
...
+        return grad_norm
+    
+    def _run_validation_epoch(self) -> float:
+        """Run one pass over `self.val_dataloader` and return average loss per token."""
+        self.model.eval()
+
+        total_loss = 0.0
+        total_tokens = 0
+
+        with torch.no_grad():
```
akoumpa: ok to use the decorator instead?

recipes/finetune.py
```diff
@@ -201,28 +274,112 @@ def _run_train_step(self, batch, is_optim_step, clip_norm=1.0, num_grad_acc_step
...
+                total_loss += local_loss.item()
+                total_tokens += mask.sum().item()
+
+        # Aggregate across ranks if distributed is initialized
+        if dist.is_initialized():
+            tensor = torch.tensor([total_loss, total_tokens], device=self.dist_env.device)
+            dist.all_reduce(tensor, op=dist.ReduceOp.SUM)
+            total_loss, total_tokens = tensor.tolist()
+
+        self.model.train()
```
akoumpa: let's move this to the train function

recipes/llama_3_2_1b_hellaswag.yaml
```diff
@@ -15,12 +17,12 @@ model:
   _target_: transformers.AutoModelForCausalLM.from_pretrained
   pretrained_model_name_or_path: meta-llama/Llama-3.2-1B
 
-distributed:
-  _target_: nemo_automodel.distributed.fsdp2.FSDP2Manager
-  dp_size: 1
-  tp_size: 1
-  cp_size: 1
-  sequence_parallel: false
+# distributed:
```
akoumpa: can we keep it?

recipes/llama_3_2_1b_hellaswag.yaml
```diff
@@ -46,3 +70,20 @@ optimizer:
   weight_decay: 0
   foreach: true # foreach is not compatible with TP
   # min_lr: 1.0e-5
+
+logger:
+  _target_: nemo_automodel.utils.common_utils.LoggerConfig
```
akoumpa: do we need this?

recipes/finetune.py
```diff
@@ -220,32 +271,115 @@ def _run_train_step(self, batch, is_optim_step, clip_norm=1.0, num_grad_acc_step
...
+
+        with get_sync_ctx(self.model, is_optim_step):
+            local_loss.backward()
 
+        grad_norm = None
         if is_optim_step:
-            grad_params = []
+            world_size = get_world_size_safe()
+            num_tokens_for_grad_scaling = self.total_num_tokens.clone().detach()
+            dist.all_reduce(num_tokens_for_grad_scaling)
```
akoumpa: does this work on single gpu runs?

adil-a: Yes, tested on single gpu. Need to speak to @BoxiangW during standup to ask what changes need to be done so we only do this over DP

adil-a: cc @BoxiangW can you also verify why grad norm clipping won't work on TP > 1?

BoxiangW: Try use `model_wrapper.device_mesh["data_parallel"].get_group()` and grad_clip has DTensor error that we can dive into later

nemo_automodel/utils/dist_utils.py
```diff
@@ -245,3 +245,87 @@ def dump_dataclass_to_yaml(obj: Any, filename: Optional[str] = None) -> Optional
...
+    Returns:
+        A context manager that synchronizes the model.
+    """
+    # Use `no_sync` on DDP models when we are *not* on the final micro-batch for
+    # this gradient update (i.e., when `is_grad` is False). This avoids an
+    # all-reduce for every micro-batch and greatly improves throughput.
+    if isinstance(model, dist.fsdp._fully_shard._fully_shard.FSDPModule):
+        model.set_requires_gradient_sync(is_optim_step)
+        sync_ctx = nullcontext()
+    elif isinstance(model, torch.nn.parallel.DistributedDataParallel) and not is_optim_step:
```
akoumpa: if is ddp then / else

PR #48, opened 2025-06-05
Add CLI options

akoumpa:
Enable parameter passing and parameter overriding via CLI.

The special parameter `--config` specifies the yaml file, and options within the file can be overrode with `--param=val`. To handle sectioned yaml config options, please use a dotted expression `--section.param=val`

### DIFF
--- a/nemo_automodel/datasets/llm/squad.py
+++ b/nemo_automodel/datasets/llm/squad.py
@@ -1,0 +1,125 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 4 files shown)


nemo_automodel/config/cli.py
```diff
@@ -0,0 +1,89 @@
...
+            else:
+                # case B) --key val
+                # if next token exists and isn't another --..., take it as the value
+                if i + 1 < len(argv) and not argv[i + 1].startswith("--"):
+                    val = argv[i + 1]
+                    overrides.append(f"{key}={val}")
+                    i += 2
+                else:
+                    # case C) bare --flag => True
+                    overrides.append(f"{key}=True")
```
adil-a, line 63: What happens in the case where a user mistakenly does a --flag but the user made a mistake and that flag doesn't actually take a bool value? As an example maybe something like setting adam's params and users just does --lr

akoumpa: we don't do strict input validation and therefore it won't be caught at the argument parsing level, instead the program most likely will crash later on. Maybe we could use a combination of annotation during the parsing to figure out type-related errors.

PR #52, opened 2025-06-07
feat: DCP checkpointing

adil-a:
This PR adds base checkpointing for HF safetensors.

- Adds run-time checkpointing logic
- Adds all of the accompanying files required for checkpointing
- Ability to save with DCP or .safetensors

### DIFF
--- a/recipes/llm/finetune.py
+++ b/recipes/llm/finetune.py
@@ -75,7 +75,7 @@ def build_checkpoint_config(cfg_ckpt, cache_dir, model_repo_id):
     from transformers.utils import TRANSFORMERS_CACHE
 
     ckpt_kwargs = dict(  
-        enabled=True,
+        enabled=False,
         checkpoint_dir="checkpoints/",
         model_save_format="safetensors",
         model_repo_id=model_repo_id,


nemo_automodel/checkpoint/stateful_wrappers.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        # this line automatically manages FSDP FQN's, as well as sets the default state dict type to FSDP.SHARDED_STATE_DICT
+        model_state_dict = get_model_state_dict(
+            self.model,
+            options=StateDictOptions(
+                cpu_offload=True,
+                full_state_dict=True if self.serialization_format == SerializationFormat.SAFETENSORS else False,
+            ),
+        )
+        if self.remove_lm_head:
```
akoumpa: let's remove this?

nemo_automodel/training/base_recipe.py
```diff
@@ -55,25 +62,119 @@ def __setattr__(self, key, value):
...
         if not '__state_tracked' in self.__dict__:
             self.__dict__['__state_tracked'] = set()
-        if isinstance(value, (nn.Module, Optimizer)) or has_load_restore_state(value):
-            assert not key in self.__dict__['__state_tracked']
+        # Track stateful objects unless they are validation/eval components.
+        should_track = (
+            isinstance(value, (nn.Module, Optimizer)) or has_load_restore_state(value)
+        )
+
+        if should_track and not any(substr in key.lower() for substr in ("val", "eval", "test")):
```
akoumpa, line 70:
```
not any(substr in key.lower() for substr in ("val", "eval", "test")):
```
I would say let's remove this, WDYT?

recipes/llm/finetune.py
```diff
@@ -263,18 +300,16 @@ def run_train_validation_loop(self):
...
-        for epoch in self.step_scheduler.epochs:
+        for _ in self.step_scheduler.epochs:
             for batch_idx, batch in enumerate(self.dataloader):
                 is_optim_step, is_ckpt_step, is_val_step = self.step_scheduler.update(batch_idx)
                 self._run_train_step(batch, is_optim_step, 1.0)
-                if is_ckpt_step and self.dist_env.is_main:
-                #     self._save_checkpoint()
-                    pass
+                if is_ckpt_step:
+                    self.save_checkpoint()
```
akoumpa: I'd prefer to pass the current epoch and step to save_checkpoint's args, WDYT?

recipes/llm/finetune.py
```diff
@@ -249,10 +279,17 @@ def setup(self):
...
+        self.checkpoint_config = build_checkpoint_config(
+            self.cfg.get('checkpoint', None),
+            self.cfg.get('model.cache_dir', None),
+            self.cfg.model.pretrained_model_name_or_path,
+        )
 
         # Optionally resume
-        if (path := self.cfg.get("restore_from")) is not None:
-            raise NotImplemented("TODO resume from {}".format(path))
+        self.load_checkpoint(restore_from)
```
akoumpa, line 300:
```
if (path := self.cfg.get("restore_from")) is not None:
    self.load_checkpoint(path)
```
would this work?

nemo_automodel/checkpoint/stateful_wrappers.py
```diff
@@ -0,0 +1,136 @@
...
+        set_model_state_dict(
+            self.model,
+            state_dict,
+        )
+
+
+class OptimizerState(Stateful):
+    """Helper class for tracking optimizer state in distributed checkpointing.
+
+    This class is compliant with the Stateful protocol, allowing DCP to automatically
```
akoumpa, line 90: I'm thinking whether we need this one, I was under the impression that any Optimizer is compliant with the Stateful protocol

nemo_automodel/checkpoint/stateful_wrappers.py
adil-a: If we remove it, it will also checkpoint the LM head (which increases the checkpoint size by considerable amount depending on model size. Even for a 3B, the LM head is ~500MB). I decided to keep this for some memory savings since if the weights are tied, we don't need the lm head checkpointed.

nemo_automodel/training/base_recipe.py
adil-a: If we remove this, it also checkpoints things like the validation dataloader. I don't think this needs to be checkpointed right? If we remove this, I think we should have some functionality to dictate what to exclude from ckpting.

recipes/llm/finetune.py
adil-a: Sure, will make the change.

recipes/llm/finetune.py
adil-a: Current functionality is that if `restore_from` is None, then the latest checkpoint will be found and used (if found). See: https://github.com/NVIDIA-NeMo/NeMo-Automodel/pull/52/files/5bcda8d89d374164eb3db06543b92b7beea7ba28#diff-90aa71521cec91d1ea394d4e8bbaa6989bd9bc6d84dc566b7ba3f8fe867caca5R126-R130

nemo_automodel/checkpoint/stateful_wrappers.py
adil-a: The problem isn't whether or not there state_dict or load_state_dict methods, the problem is that under distributed conditions we need specific state_dict and load_state_dict methods. We [wrap](https://github.com/NVIDIA-NeMo/NeMo-Automodel/blob/adil/checkpointing/nemo_automodel/checkpoint/checkpointing.py#L197) our optimizer with this class during saving and loading so the appropriate methods are called. This is also the case in [NeMo-RL](https://github.com/NVIDIA/NeMo-RL/blob/c0d42e63044995ab6be00c192a23ee780ac0a2ba/nemo_rl/utils/native_checkpoint.py#L164) for reference.

nemo_automodel/checkpoint/stateful_wrappers.py
```diff
@@ -0,0 +1,146 @@
...
+    call state_dict/load_state_dict as needed in the dcp.save/load APIs.
+
+    Args:
+        model: The PyTorch model to track.
+    """
+
+    def __init__(self, model: torch.nn.Module, serialization_format: SerializationFormat):
+        self.model = model
+        self.remove_lm_head = model.config.tie_word_embeddings
```
akoumpa: getattr(getattr(model, 'config', {}), 'tie_word_embeddings', False)

nemo_automodel/checkpoint/checkpointing.py
```diff
@@ -0,0 +1,278 @@
...
+import torch.distributed.checkpoint as dcp
+from nemo_automodel.checkpoint.hf_storage import (
+    HuggingFaceStorageWriter,
+    HuggingFaceStorageReader,
+    get_fqn_to_file_index_mapping,
+)
+from nemo_automodel.checkpoint.stateful_wrappers import ModelState, OptimizerState
+import glob
+import json
+from torch.distributed.checkpoint.filesystem import SerializationFormat
```
akoumpa: is this in pytorch 2.7?

PR #81, opened 2025-06-16
feat: Add Triton Kernel for LoRA

torsli:
Adds Triton kernel for LoRA.

To enable the triton kernel please update the `peft` section in your YAML file as follows:
```
peft:
  peft_fn: nemo_automodel.components._peft.lora.PeftConfig
  match_all_linear: True
  dim: 8
  alpha: 32
  use_triton: True
```

In particular, the **`use_triton`** option needs to be included and be enabled.

For a complete YAML config, please see [here](https://github.com/NVIDIA-NeMo/Automodel/blob/main/examples/llm/llama_3_2_1b_hellaswag_peft.yaml).

### DIFF
--- a/nemo_automodel/_peft/lora.py
+++ b/nemo_automodel/_peft/lora.py
@@ -307,7 +307,7 @@ class LoRATritonFunction(torch.autograd.Function):
         """
         Stores context for LoRA backward pass.
         """
-        x, lora_A, lora_B, scale, _, _ = inputs
+        x, lora_A, lora_B, scale, _ = inputs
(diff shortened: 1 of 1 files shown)


recipes/llm/llama_3_2_1b_hellaswag_peft.yaml
```diff
@@ -22,6 +22,7 @@ peft:
   match_all_linear: True
   dim: 8
   alpha: 32
+  use_triton: True
```
akoumpa, line 25: nice

PR #101, opened 2025-06-19
fix: Fix nvfsdp usage in automodel

BoxiangW:
- Fix nvfsdp usage in automodel
- Merge `build_model` and `build_optimizer` function in `fintune.py`

### DIFF
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -141,7 +141,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - script: L2_HF_Transformer_SFT
+            runner: linux-amd64-gpu-rtxa6000-latest-2-nemo
+          - script: L2_HF_Transformer_SFT_nvfsdp
+            runner: linux-amd64-gpu-rtxa6000-latest-2-nemo
           - script: L2_HF_Transformer_PEFT
+            runner: linux-amd64-gpu-rtxa6000-latest-2-nemo
+          - script: L2_HF_Transformer_PEFT_nvfsdp
             runner: linux-amd64-gpu-rtxa6000-latest-2-nemo
           - script: L2_DCP_Checkpoint
             runner: linux-amd64-gpu-rtxa6000-latest-2-nemo


nemo_automodel/distributed/parallelizer.py
```diff
@@ -237,18 +233,27 @@ def nvfsdp_strategy_parallelize(
...
+        dp_mesh_name="data_parallel",
+        cp_mesh_name="context_parallel",
+        tp_mesh_name="tensor_parallel",
+        dp_cp_mesh_name=dp_cp_mesh_name,
         data_parallel_sharding_strategy=data_parallel_sharding_strategy,
-        init_nvfsdp_with_meta_device=init_nvfsdp_with_meta_device,
+        init_model_with_meta_device=init_nvfsdp_with_meta_device,
         grad_reduce_in_fp32=grad_reduce_in_fp32,
         preserve_fp32_weights=preserve_fp32_weights,
         overlap_grad_reduce=overlap_grad_reduce,
```
cspades, line 259: @BoxiangW Oh, before I forget, `fully_shard` has an argument that you want to turn off for performance reasons:
```
sync_grads_each_step = False,  # default True
```
This will synchronize gradients every step, which is undesirable in the case of gradient accumulation > 1. It's a very nice QoL feature for beginners so they don't need to manually sync but NeMo Automodel doesn't need it since we handle all the logic explicitly. In this case, it will be triggered by `optimizer.step()` which is precisely when we want to sync:
```
sync_grad_before_optimizer_step = kwargs.pop(
                # If sync_grads_each_step is enabled, gradients are synchronized automatically
                # during the post-backward hook, so we do not need to synchronize them here.
                "sync_grad_before_optimizer_step",
                not sync_grads_each_step,
            )
```

BoxiangW: I see, thanks for finding this issue

BoxiangW: In that case we should just set this to False as default value for `fully_shard` on nvFSDP side and add assertion for `optimizer is not None and sync_grads_each_step = False`

cspades: Forgot to mention, if you do gradient clipping and norms, you want it to be `True`. Cause it needs to install the optimizer params before `model.parameters()` will point to the optimizer parameters + grads.

`fully_shard` should definitely be `sync_grads_each_step=True` since it makes the training loop very clean, and users will be able to see the grads after backwards pass, just like when you peek into DTensor while FSDP2 is working.

You can also do:
```
model.finish_grad_sync()

# INSERT GRAD CLIPPING AND NORMS HERE

optimizer.step(
    sync_grad_before_optimizer_step=False,
    install_optimized_model_weights=True,
)
```
which is performance-optimal and also you have access to the gradients. I suggest this for NeMo Automodel since you do gradient clipping:
```
            if HAVE_NVFSDP and isinstance(self.model, nvFSDP):
                # If the model uses nvFSDP, wait for all sharded gradients to be reduced and unsharded.
                # Necessary because the post-backward reduce-scatter is asynchronous, so gradients and backward
                # computations are concurrent, but the gradients of the final layer may not be available yet.
                self.model.finish_grad_sync()
            
            # Clip gradients **after** any rescaling.
            # TODO(@boxiangw): Fix TP gradient clipping
            if not self.device_mesh or self.device_mesh["tensor_parallel"].size() == 1:
                grad_norm = clip_gradients(self.model, clip_norm)
            else:
                # TODO: TP WAR
                grad_norm = 0.0
```

I was just debugging this for `no_shard` haha
```
> /workspace/Automodel/nemo_automodel/utils/dist_utils.py(299)clip_gradients()
-> torch.distributed.barrier()
(Pdb) c
INFO:root:step 32 | epoch 0 | loss 3.0050 | grad_norm 8497.2305 | mem: 19.27 GiB | steps_per_second: 11.5045 it/s
> /workspace/Automodel/nemo_automodel/utils/dist_utils.py(299)clip_gradients()
-> torch.distributed.barrier()
(Pdb) c
INFO:root:step 48 | epoch 0 | loss 2.8791 | grad_norm 8016.0654 | mem: 19.27 GiB | steps_per_second: 11.4891 it/s
> /workspace/Automodel/nemo_automodel/utils/dist_utils.py(299)clip_gradients()
```

PR #112, opened 2025-06-25
test: unit tests for training/loggers/utils

akoumpa:
Adds unit tests for:
- training
- loggers
- utils

### DIFF
--- a/tests/functional_tests/L2_DCP_Checkpoint.sh
+++ b/tests/functional_tests/L2_DCP_Checkpoint.sh
@@ -17,17 +17,15 @@ set -xeuo pipefail # Exit immediately if a command exits with a non-zero status
 
 export PYTHONPATH=${PYTHONPATH:-}:$(pwd)
 export CUDA_VISIBLE_DEVICES="0,1"
-# MODEL=/home/TestData/akoumparouli/hf_mixtral_2l/
-MODEL=/home/TestData/akoumparouli/hf_gemma_38m/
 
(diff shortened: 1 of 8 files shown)


nemo_automodel/training/rng.py
```diff
@@ -55,10 +56,13 @@ def __init__(self, seed: int, ranked: bool = False):
...
             ranked (bool): Adjust seed based on process rank.
         """
-        init_all_rng(seed, ranked)
+        self._init_state = self.state_dict()
+        self._saved_state = None
         self.seed = seed
         self.ranked = ranked
-        self._saved_state = None
+
+    def __del__(self):
```
adil-a, line 64: could you clarify why this goes inside the __del__ special method?

nemo_automodel/training/step_scheduler.py
```diff
@@ -43,19 +49,35 @@ def __init__(self,
...
+
+        # len() may be missing for IterableDataset
+        self.epoch_len: int | None = (
+            len(dataloader) if dataloader is not None and hasattr(dataloader, "__len__") else None
+        )
 
+        # per-batch cached flags
+        self._optim_flag = False
+        self._val_flag   = False
+        self._ckpt_flag  = False
```
adil-a: imo we should either type annotate all class variables or none for style consistency. right now we're annotating some public and private variables

nemo_automodel/training/step_scheduler.py
adil-a, line 1: nice, I like this clean up :D

PR #121, opened 2025-06-27
fix: vlm dcp loading

HuiyingLi:


nemo_automodel/checkpoint/stateful_wrappers.py
```diff
@@ -46,6 +46,27 @@ def _add_outer_prefix(sd: dict[str, Any], prefix: str = _PREFIX) -> None:
...
+        "model.lm_head.weight"
+    ]
+    for path in possible_paths:
+        try:
+            obj = model
+            for attr in path.split('.'):
+                obj = getattr(obj, attr)
+            return obj, path
+        except AttributeError:
+            continue
```
adil-a: since all of them include `lm_head`, can we instead iterate over `model.named_parameters(remove_duplicate=False)`? It should give you the name and weight. You can just then check if `lm_head` is present in the name string.

HuiyingLi: @adil-a Thank you so much for the suggestion! Just pushed a fix could you take a look? Thank you!

PR #143, opened 2025-06-30
fix: cfg.packed_sequence -> cfg.get('packed_sequence', None)

akoumpa:


nemo_automodel/config/loader.py
```diff
@@ -120,15 +120,26 @@ class ConfigNode:
...
             k: self._wrap(k, v) for k, v in d.items()
         }
+        self.raise_on_missing_attr = raise_on_missing_attr
+
+    def __getattr__(self, key):
+        try:
+            return self.__dict__[key]
+        except:
+            if self.raise_on_missing_attr:
+                raise AttributeError
```
adil-a, line 140: this fix was needed because the try block will raise a KeyError, but we want to send back an AttributeError (for methods like `hasattr`)

nemo_automodel/config/loader.py
```diff
@@ -120,15 +120,26 @@ class ConfigNode:
     This class allows nested dictionaries and lists to be accessed as attributes and
     provides functionality to instantiate objects from configuration.
     """
-    def __init__(self, d):
+    def __init__(self, d, raise_on_missing_attr=True):
```
adil-a, line 123: setting default to True. Otherwise method calls of `hasattr` return True because when `raise_on_missing_attr=False`, we just return `None`.

nemo_automodel/config/loader.py
```diff
@@ -173,7 +184,7 @@ def instantiate(self, *args, **kwargs):
         # Prepare kwargs from config
         config_kwargs = {}
         for k, v in self.__dict__.items():
-            if k == '_target_':
+            if k == '_target_' or k == 'raise_on_missing_attr':
```
adil-a, line 187: needed for filtering otherwise downstream calls receive an unexpected key

nemo_automodel/config/loader.py
```diff
@@ -212,7 +223,7 @@ def to_dict(self):
             dict: A dictionary representation of the configuration node.
         """
         return {
-            k: self._unwrap(v) for k, v in self.__dict__.items()
+            k: self._unwrap(v) for k, v in self.__dict__.items() if k != 'raise_on_missing_attr'
```
adil-a, line 226: needed for filtering otherwise downstream calls receive an unexpected key

PR #162, opened 2025-07-03
fix: fixing generate file

adil-a:


HuiyingLi (approved):
Tested working with trained gemma3 peft ckpt. Thank you so much!

recipes/vlm/generate.py
```diff
@@ -193,25 +215,44 @@ def main():
...
 
-    logging.info(
-        f"Loading model type base_model:{args.base_model} from checkpoint_path:{args.checkpoint_path}"
-    )
+    logging.info(f"Loading model type base_model:{args.base_model} from checkpoint_path:{args.checkpoint_path}")
 
     model = load_model_from_checkpoint(
         args.checkpoint_path,
         args.base_model,
         args.is_peft,
```
akoumpa: I'm thinking we may want to have a `--peft-ckpt` argument that's either `None` or a path. If it's a path, then it loads the adapter, and that would be all that's needed to restore the lora checkpoint. I know the `is_peft` was not introduced in this PR.

adil-a: Are you saying that we strictly only need `--peft-ckpt`? Because we still need the path to the base model to load a peft ckpt (since the ckpt only holds adapter weights).

PR #163, opened 2025-07-06
fix: checkpoint logging

adil-a:
- Adds warning for peft checkpoint and torch_save format
- Removes hardcoded lm head module name

### DIFF
--- a/nemo_automodel/checkpoint/checkpointing.py
+++ b/nemo_automodel/checkpoint/checkpointing.py
@@ -15,20 +15,20 @@
 """Checkpoint management utilities for HF models."""
 
 import glob
+import json
+import logging
 import os
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Optional
-import json
 
 import torch
 import torch.distributed
 import torch.distributed.checkpoint as dcp
 import torch.nn as nn
-
-from transformers import PreTrainedModel
-from safetensors.torch import save_file
 from safetensors import safe_open
+from safetensors.torch import save_file
+from transformers import PreTrainedModel
 
 from nemo_automodel.checkpoint._backports.filesystem import SerializationFormat
 from nemo_automodel.checkpoint._backports.hf_storage import (
@@ -44,6 +44,7 @@ class CheckpointingConfig:
     """
     Configuration for checkpointing.
     """
+
     enabled: bool
     checkpoint_dir: str | Path
     model_save_format: SerializationFormat | str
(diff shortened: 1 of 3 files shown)


nemo_automodel/checkpoint/checkpointing.py
```diff
@@ -15,20 +15,20 @@
...
 import torch
 import torch.distributed
 import torch.distributed.checkpoint as dcp
 import torch.nn as nn
-
-from transformers import PreTrainedModel
-from safetensors.torch import save_file
 from safetensors import safe_open
+from safetensors.torch import save_file
+from transformers import PreTrainedModel
```
akoumpa: WDYT about dropping this completely and instead checking if it's an `nn.Module`?

nemo_automodel/_peft/lora.py
```diff
@@ -29,15 +30,37 @@
...
+    dim: int = 8
+    alpha: int = 32
+    dropout: float = 0.0
+    dropout_position: Literal["pre", "post"] = "post"
+    lora_A_init: str = "xavier"
+    lora_dtype: Optional[torch.dtype] = None
+    use_triton: bool = False
+
+    def to_dict(self):
+        return self.__dict__
```
akoumpa:
```suggestion
        return self.__dict__.copy()
```

nemo_automodel/checkpoint/checkpointing.py
akoumpa: can we drop `PretrainedModel`?

nemo_automodel/checkpoint/checkpointing.py
```diff
@@ -77,12 +95,14 @@ def save_model(
         weights_path: Path to save model weights
         checkpoint_config: Checkpointing configuration
     """
-    # TODO(@adil-a): Need to add support for PEFT.
+    if checkpoint_config.is_peft and checkpoint_config.model_save_format == SerializationFormat.TORCH_SAVE:
```
akoumpa: I would throw an exception when the checkpoint_config is created instead of logging here.

nemo_automodel/checkpoint/checkpointing.py
```diff
@@ -298,3 +319,59 @@ def _get_safetensors_index_path(cache_dir: str, repo_id: str) -> str:
...
+
+def _save_peft_adapters(
+    model_state: ModelState,
+    peft_config: PeftConfig,
+    model_path: str,
+):
+    """
+    Save PEFT adapters to a weights path.
+    """
+    hf_peft_config = _get_hf_peft_config(peft_config, model_state)
```
akoumpa, line 295: I would pass `model_state.model` instead of `model_state`

akoumpa: actually, I would remove the model_state function argument for `_save_peft_adapters` and go with `model`, would that work?

recipes/llm/finetune.py
```diff
@@ -92,10 +93,10 @@ def build_model_and_optimizer(
             if isinstance(m, nn.Embedding):
                 m.weight.requires_grad_(False)
         # Optionally apply PEFT (e.g., LoRA/DoRA, etc)
+        peft_config = None
         if cfg_peft is not None:
-            opts = cfg_peft.to_dict()
-            peft_fn = opts.pop("peft_fn")
-            peft_fn(model, **opts)
+            peft_config = cfg_peft.peft_params.instantiate()
```
akoumpa: I would substitute the `cfg_peft` arg for `peft_config` arg and pass the `peft_config` as a parameter to `build_model_and_optimizer`

akoumpa (approved):

PR #164, opened 2025-07-06
fix: llm finetune cleanup

adil-a:
- Fixed outdated docstrings
- Fixed duplicate imports
- Added clarification/renamed variables tracking 1) the valid tokens in a batch for TPS calculation and 2) the tokens that are for calculating the loss + bwd
- Formatting + linting with ruff


recipes/llm/finetune.py
```diff
@@ -382,7 +397,7 @@ def run_train_validation_loop(self):
         """
         self.model.train()
         self.timestamp = time.perf_counter()
-        self.num_tokens = 0
+        self.num_valid_tokens = 0
```
akoumpa: I'm somewhat hesitant to use the `valid` as part of the `num_tokens` variable, since I want to avoid any association with the validation process often occurring in ML training recipes. I'm wonder if there's any alternatives, WDYT?

adil-a: Agreed, changing to `num_nonpad_tokens`

akoumpa (approved):

PR #171, opened 2025-07-07
feat: support file-paths as _target_ in configs

akoumpa:
Enables `_target_` to be set to a path of a python file. In particular, it supports the following format:
```
<absolute-file-path>:<function-name>
```
where the first part is as described above (relative or absolute path of a python file), and the second part (after column) specifies the function name to load from the file found at the `<absolute-file-path>`. Therefore, the syntax after the column allows to specify which function to instantiate, for example, by calling the [`.instantiate`](https://github.com/NVIDIA-NeMo/Automodel/blob/b5fcbc27883ed23d6ce683ec094b654c59ac5cb7/nemo_automodel/config/loader.py#L163) function on it.

Previously, the value of `_target_` could be a dotted path, for example:
```
dataset:
  _target_: nemo_automodel.datasets.llm.hellaswag.HellaSwag
```

Now, we can pass a file path as is, for example:
```
dataset:
   _target_: /mnt/4tb/auto/nemo_automodel/datasets/llm/mock_packed.py:build_packed_dataset
   num_blocks: 111
```

This is very convenient for cases where you want to define your own dataset and pass it via yaml. In the above example, it will call the `build_packed_dataset` with the rest of the parameters (i.e., `num_blocks`) that are under the `dataset` section.

While the above example uses a file (i.e., `mock_packed.py`) that is included in the `Automodel` repo, this is not a requirement, the file can present in any location.

### DIFF
--- a/nemo_automodel/config/loader.py
+++ b/nemo_automodel/config/loader.py
@@ -87,8 +87,8 @@ def _resolve_target(dotted_path: str):
         parts = dotted_path.split(':')
         assert parts[0].endswith('.py'), "Expected first part to be a python script"
         assert Path(parts[0]).exists(), "Expected python script to exist"
-        spec = load_module_from_file(parts[0])
-        return getattr(spec, parts[1])
+        module = load_module_from_file(parts[0])
+        return getattr(module, parts[1])
 
     parts = dotted_path.split(".")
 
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
LGTM

PR #173, opened 2025-07-08
fix: add lora_dtype str to torch.dtype translator

akoumpa:
when setting the following in the yaml config:
```
peft:
  peft_fn: nemo_automodel._peft.lora.apply_lora_to_linear_modules
  target_modules: '*_proj'
  dim: 8
  alpha: 32
  lora_dtype: torch.bfloat16
```

will use `torch.bfloat16` as `dtype` for the lora weights; if there's no `lora_dtype` entry will use the underlying weight's dtype.


HuiyingLi (approved):
LGTM

PR #179, opened 2025-07-09
fix: Delete medpix.jpg

adil-a:
I think this was pushed by mistake


akoumpa:
Hi, I think it's intentional (reference [here](https://github.com/NVIDIA-NeMo/Automodel/blob/d7b92061350aa3450bb4f66c479974f1e124cb7a/docs/guides/gemma3_3n.md?plain=1#L292)), but we'd want to find a better solution long-term to avoid having a binary blobs contained in the repo.

PR #182, opened 2025-07-10
feat: MCore Dataset

adil-a:
- Adds functionality to load an MCore preprocessed dataset


nemo_automodel/recipes/base_recipe.py
```diff
@@ -61,6 +65,19 @@ def has_load_restore_state(object):
...
+    """
+    Checks whether object is a dataloader.
+
+    Args:
+        object (any): the object to check.
+
+    Returns:
+        bool: returns True if object is a dataloader.
+    """
+    return isinstance(object, (StatefulDataLoader, IterableDataset, Dataset)) and has_load_restore_state(object)
```
hemildesai: Why does this mix DataLoader with Dataset?

tests/functional_tests/checkpoint/test_dcp.py
```diff
@@ -815,6 +815,8 @@ def test_dcp_checkpoint():
         "step_scheduler.pt",
         "config.yaml",
     ]
+    if trainer._get_dp_group_size() > 1:
```
hemildesai, line 820: OOC is the dataloader state independent of rank? What if user wants to change dp size on resume?

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -313,7 +320,17 @@ def build_dataloader(
         kwargs = {}
         if tokenizer is not None:
             kwargs["tokenizer"] = tokenizer
-        ds = cfg_ds.instantiate(**kwargs)
+
+        # Megatron specific kwargs
+        if cfg_ds._target_ == MegatronPretraining:
```
hemildesai, line 343: Is there a way to avoid this here and just modify the args in the `__init__` of `MegatronPretraining`?

hemildesai (commented):
Just a few small comments

nemo_automodel/recipes/llm/train_ft.py
adil-a: So things like gbs, max steps, val check interval are all passed in from the config of the StepScheduler, so I need to explicitly pass it into the constructor here. I don't see a workaround to it in the current configuration state we have.

Am I understanding you correctly?

nemo_automodel/recipes/base_recipe.py
adil-a: Thanks for catching, it's from a previous PR. I'll update it to only capture dataloader here.

tests/functional_tests/checkpoint/test_dcp.py
adil-a: It captures primarily number of times the iterator has yielded so on resume it will iterate through the same number of times before resuming. I think if we change DP size this will be problematic.

Say you checkpoint on dp=2 and you've yielded 4 times/dp rank so far. If you switch to DP=1, you need to iterate 8 times to reach the same point (assuming everything else stays the same). Not sure what the right workaround for this is currently.

nemo_automodel/components/datasets/llm/megatron_dataset.py
```diff
@@ -0,0 +1,334 @@
...
+        num_dataset_builder_threads: int = 1,
+        num_train_samples: Optional[int] = None,
+        num_val_samples: Optional[int] = None,
+        num_test_samples: Optional[int] = None,
+        trainer_max_steps: Optional[int] = None,
+        trainer_val_check_interval: int = 1000,
+        trainer_limit_val_batches: Union[int, float] = 1,
+        trainer_limit_test_batches: Union[int, float] = 1,
+        mmap_bin_files: bool = True,
+        dataloader_type: Optional[Literal["single", "cyclic", "batch"]] = "single",
```
akoumpa:
```suggestion
```

nemo_automodel/components/datasets/llm/megatron_dataset.py
```diff
@@ -0,0 +1,334 @@
...
+        self.tokenizer = tokenizer
+        self.create_attention_mask = create_attention_mask
+        self.seed = seed
+        self.split = split
+        self.index_mapping_dir = index_mapping_dir
+        self.num_dataset_builder_threads = num_dataset_builder_threads
+        self.num_train_samples = num_train_samples
+        self.num_val_samples = num_val_samples
+        self.num_test_samples = num_test_samples
+        self.dataloader_type = dataloader_type
```
akoumpa:
```suggestion
```

docs/guides/llm/mcore_pretraining.md
```diff
@@ -0,0 +1,738 @@
...
+  min_lr: 0.0
+
+# Uncomment and configure for W&B logging
+# wandb:
+#   project: <your_wandb_project>
+#   entity: <your_wandb_entity>
+#   name: <your_wandb_exp_name>
+#   save_dir: <your_wandb_save_dir>
+
+```
```
akoumpa, line 299: * please include instructions for dataset blending

PR #192, opened 2025-07-14
fix: checkpoint test additional testing

adil-a:
For all tests:
- Adds an extra layer of protection by loading the saved checkpoint into a new (NeMo Automodel) model and running a `torch.allclose` on the CE loss for a validation batch.
*Note:* This is actually necessary for nvFSDP because there are hooks that get set off when weights are loaded into a model. To ensure checkpoint reloading correctness, we need the above check. Checking the state dict on its own is not sufficient.

For HF (full-model) tests:
- Loads the model into `AutoModelForCausalLM` from HF API and verifies the weights are the same

For HF (PEFT) tests:
- Loads the model using the `PEFT` module and verifies the weights are correctly loaded in.

### DIFF
--- a/tests/functional_tests/checkpoint/test_dcp.py
+++ b/tests/functional_tests/checkpoint/test_dcp.py
@@ -22,10 +22,27 @@ from pathlib import Path
 import torch
 import torch.distributed.checkpoint as dcp
 import torch.distributed.tensor
+import torch.nn as nn
 
+from nemo_automodel.checkpoint.checkpointing import load_model
 from nemo_automodel.checkpoint.stateful_wrappers import ModelState, OptimizerState
 from nemo_automodel.config.cli import parse_args_and_load_config
-from recipes.llm.finetune import FinetuneRecipeForNextTokenPrediction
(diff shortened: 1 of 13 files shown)


tests/functional_tests/checkpoint/test_dcp.py
```diff
@@ -763,6 +797,18 @@ def test_dcp_checkpoint():
...
+    restored_model = get_new_model(trainer)
+    load_model(
+        restored_model,
+        Path(trainer.checkpoint_config.checkpoint_dir) / "epoch_0_step_10",
+        trainer.checkpoint_config,
+    )
+    source_model_loss = get_validation_loss(trainer.model, val_batch, trainer.loss_fn, trainer.dist_env.device)
+    restored_model_loss = get_validation_loss(restored_model, val_batch, trainer.loss_fn, trainer.dist_env.device)
+    assert torch.allclose(source_model_loss, restored_model_loss), "Model loss mismatch"
+
```
adil-a: cc @cspades

cspades: Yup looks good, this will catch the bug I found as well!

cspades (approved):
Nice, thanks!

PR #197, opened 2025-07-15
ci: Update cicd main logic

thomasdhc:


.github/workflows/cicd-main.yml
```diff
@@ -392,3 +386,13 @@ jobs:
           path: |
             .coverage
           include-hidden-files: true
+
+  codecov-placeholder:
+     name: codecov/patch
+     needs: [pre-flight]
+     if: always()
```
chtruong814: I don't think we want always here right?  I think just case where docs_only is true?

thomasdhc: Good point!

chtruong814 (approved):

PR #201, opened 2025-07-15
fix: test_peft for NaN mismatch

torsli:
Update arguments for torch.allclose in test to allow for equal nans.


adil-a:
Should the tensors have NaNs at all? I didn't have this problem when testing the non-Triton implementation of PEFT

akoumpa:
Since we updated the test, I'm closing this.

PR #204, opened 2025-07-16
fix: Improve root repo detection for interactive jobs

akoumpa:
The `automodel` cli has the capability to detect whether it's invoked from within an `Automodel` git repository or from any other directory.  If it is invoked from within a git repo, it will use repo's source code (instead of the one installed by the  pypi package). 

Issue: when running interactive jobs, the repo "rebasing" was missing. 
Fix: function `get_automodel_repo_root` is used to determine whether it's running from inside a git repo and launch the jobs accordingly, for both, interactive and batch jobs. 

----

Changed the syntax from
```
automodel llm **finetune** -c config.yaml
```

to 
```
automodel **finetune** llm -c config.yaml
```

### DIFF
--- a/nemo_automodel/_cli/app.py
+++ b/nemo_automodel/_cli/app.py
@@ -163,16 +163,16 @@ def build_parser() -> argparse.ArgumentParser:
 
     # Two required positionals (cannot start with "--")
     parser.add_argument(
-        "domain",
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
LGTM

PR #205, opened 2025-07-16
feat: Migrate custom optimized plans and refactor parallelizer.

ffrujeri:
This PR starts migrating functionality from nemo-rl in terms of the parallezation strategies. In specific it migrates the an util helper, the customized tensor parallel plans and refactor the parallelizer.py module to receive some functionality from the [_parallelize_model](https://github.com/NVIDIA-NeMo/RL/blob/d158cbcec1b0d14c708c6c70860455fea69fe423/nemo_rl/models/dtensor/parallelize.py#L357) in nemo-rl

This PR is associated with this other one in [nemo-rl](https://github.com/NVIDIA-NeMo/RL/pull/649).


nemo_automodel/components/distributed/parallelizer.py
```diff
@@ -39,105 +49,176 @@
...
-    dp_mesh = device_mesh[
-        ("dp_cp" if "dp_cp" in _mesh_resources.root_to_flatten_mapping.get(device_mesh, {}) else "data_parallel")
-    ]
 
-    if dp_mesh.size() > 1:
-        assert dp_mesh.ndim == 1, "Hybrid-sharding not supported"
+def get_hf_tp_shard_plan(model):
+    """Get the Hugging Face tensor parallel plan from the model.
 
-    tp_mesh = device_mesh["tensor_parallel"]
```
ffrujeri, line 100: @akoumpa Is there any reason that you used the key "tensor_parallel" here instead of "tp", are there any other assumption in the codebase? From nemo-rl version they use "tp" which would be on parity with the "dp_cp" abreviations as well. What do you think? Is there enough test coverage?

nemo_automodel/components/distributed/parallelizer.py
```diff
@@ -39,105 +49,176 @@
...
+        for layer_id, transformer_block in enumerate(module):
+            # As an optimization, do not reshard after forward for the last
+            # transformer block since FSDP would prefetch it immediately
+            reshard_after_forward = int(layer_id) < len(module) - 1
+            fully_shard(
+                transformer_block,
+                mesh=mesh,
+                mp_policy=mp_policy,
+                reshard_after_forward=reshard_after_forward,
+                offload_policy=offload_policy,
```
ffrujeri, line 92: @akoumpa Notice that I am adding that option here, is there any reason not too? Also please refer to [this document ](https://docs.google.com/document/d/14eo_px7XY4ERqDTjEOJv30Uw-_H2PJHaNK6uT-kGH7U/edit?tab=t.0#heading=h.bh1vlq4vbq3j)to check the differences from what was in parallelize.py in nemo-rl

nemo_automodel/components/distributed/parallelizer.py
BoxiangW: This is actually defined during device_mesh init. You need to align here

nemo_automodel/components/distributed/optimized_tp_plans.py
```diff
@@ -0,0 +1,266 @@
...
+    """
+    if model.__class__ == Gemma3ForConditionalGeneration:
+        model_prefix = "language_model"
+    else:
+        model_prefix = "model"
+
+    # For gemma3 models, we don't include the model.embed_tokens and lm_head in the
+    # parallelization plans because they have tied weights.
+    base_model_tp_plan = {
+        f"{model_prefix}.layers.*.self_attn.q_proj": ColwiseParallel(),
```
BoxiangW, line 99: I think tp_plan can also match "layers.*.self_attn.q_proj" even if there is no `{model_prefix}` presenting

nemo_automodel/components/distributed/parallelizer.py
```diff
@@ -263,46 +344,6 @@ def nvfsdp_strategy_parallelize(
...
-        hf_tp_shard_plan.update(model._tp_plan)
-    if hasattr(model.model, "_tp_plan") and model.model._tp_plan is not None:
-        hf_tp_shard_plan.update({f"model.{k}": v for k, v in model.model._tp_plan.items()})
-
-    hf_tp_shard_plan = {k: translate_to_torch_parallel_style(v) for k, v in hf_tp_shard_plan.items()}
-    return hf_tp_shard_plan
-
-
-@lru_cache
-def translate_to_torch_parallel_style(style: str):
```
BoxiangW, line 281: This function seems have no changes at all.

nemo_automodel/components/distributed/optimized_tp_plans.py
```diff
@@ -0,0 +1,266 @@
...
+    sequence_parallel: bool = False,
+):
+    """Parallelizes a Gemma3ForCausalLM model across data parallel dimensions.
+
+    Tensor parallelism is not supported for Gemma3 models because of tied word embeddings.
+    """
+    if model.__class__ == Gemma3ForConditionalGeneration:
+        model_prefix = "language_model"
+    else:
+        model_prefix = "model"
```
adil-a, line 94: I don't think this is extendible. This issue of model_prefix affects many other models (https://github.com/search?q=repo%3Ahuggingface%2Ftransformers%20_checkpoint_conversion_mapping&type=code)

instead of hardcoding I'd recommend you use something like `getattr(model, "_checkpoint_conversion_mapping", None)` and work with that instead

This is how HF deals with it: https://github.com/huggingface/transformers/blob/b85ed49e0a5f1bd9fd887f497d055b22b9319a12/src/transformers/modeling_utils.py#L4989-L5047

ffrujeri: Hey @adil-a, I thought this condition was related to what you were describing but I think the use case is slightly different isn't it?

nemo_automodel/components/distributed/optimized_tp_plans.py
ffrujeri: As discussed offline, for readability we can keep the complete pattern for the regex

nemo_automodel/components/distributed/parallelizer.py
ffrujeri: Yeah, I will revert it back

nemo_automodel/components/distributed/parallelizer.py
ffrujeri: Ok. I decided to make it parameterizable. Let me know what you think.

nemo_automodel/components/distributed/parallelizer.py
ffrujeri: @BoxiangW What is your opinion about this?

BoxiangW: I'm sorry what do you mean? We had this offload_policy before. Also I am not able to access the doc

nemo_automodel/components/distributed/parallelizer.py
BoxiangW: It make sense, but could bring unnecessary complexity by offering more parameters right?

nemo_automodel/components/distributed/optimized_tp_plans.py
BoxiangW: I think this only for TP_sharding, they are doing regex matching on the layer names. I think it is different use case since TP_sharding do require user's prior knowledge on TP and model structure because of the rowparallel and colparallel

ffrujeri: Yeah, agreed. I will resolve the conversation for now. Please we can discuss offline @adil-a

nemo_automodel/components/distributed/parallelizer.py
ffrujeri: You are right. The compromise I found in this case was to leave the default parameter value as the one you had "tensor_parallel" and then if the caller has another mesh parameter name I let it to be passed in. What do you think?

nemo_automodel/components/distributed/parallelizer.py
ffrujeri: Hey @BoxiangW, my intention was to highlight the differences in the [reshard_after_foward](https://github.com/NVIDIA-NeMo/RL/blob/ae54cbd4d65c5cefe7c1638c51262942957a5643/nemo_rl/models/dtensor/parallelize.py#L495) that is set to true in the nemo-rl version and False in the current version. 

Also there are some changes in the for loop vs the recursive helper function (in which I was also adding the offload_policy argument). I gave you access to the document. We can also iterate through a call offline if it is more efficient.

terrykong: re: reshard_after_forward. The pytorch fsdp team told us we should have that arg given changes in upstream incoming.

https://github.com/NVIDIA-NeMo/RL/pull/464

ffrujeri: I will also check and get a sign-off from @gshennvm here

BoxiangW: Discussed offline, yeah, I think the reason here is to set reshard_after_forward=False at the final `fully_shard` function call. This does not affect ZeRO-2 and ZeRO-3. resolve this conversation

nemo_automodel/components/distributed/parallelizer.py
BoxiangW: Agreed, we need to support this after integration with RL. Resolved

BoxiangW (dismissed):
LGTM

PR #208, opened 2025-07-16
feat: save tokenizer

adil-a:
- Adds functionality to save tokenizer/VLM processor when doing consolidated checkpointing
- Modifies existing tests to ensure tokenizer/VLM processor is saved during consolidated checkpointing (both PEFT and dense)

*NOTE*: we don't save/load the tokenizer in any other ckpt save/restore. The tokenizer saving during consolidation step is only to ensure smooth downstream usage.


nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -27,6 +27,7 @@
 import torch.nn as nn
 from safetensors import safe_open
 from safetensors.torch import save_file
+from transformers.tokenization_utils import PreTrainedTokenizerBase
```
akoumpa: can we move the import to the `if TYPE_CHECKING` section?

PR #211, opened 2025-07-17
fix: VLM legacy FQN fix

adil-a:
We base the final saving file setup of the consolidated files based on the FQN sharding scheme in the base model e.g. https://huggingface.co/google/gemma-3-4b-it/blob/main/model.safetensors.index.json

However, VLMs now get their FQNs overwritten at load time (https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma3/modeling_gemma3.py#L974-L980) which leads to FQN mismatch in the mapping. This PR fixes the issue using the same utility that HF uses to resolve the problem (https://github.com/huggingface/transformers/blob/b85ed49e0a5f1bd9fd887f497d055b22b9319a12/src/transformers/modeling_utils.py#L4989-L5047)


akoumpa (dismissed):
LGTM, thank you.

adil-a:
@NVIDIA-NeMo/automation please approve

PR #219, opened 2025-07-18
fix: allow passing extra args to cli

akoumpa:
Allow extra args that are used to overwrite settings in the YAML config, for example,
```
automodel finetune llm -c examples/llm/llama_3_2_1b_squad.yaml --distributed.backend gloo --model.attn_implementation eager
```
Would overwrite:
```
distributed:
   backedn: gloo

model:
   attn_implementation: eager
```


HuiyingLi (approved):
LGTM

PR #229, opened 2025-07-22
ci: Move cut-cross-entropy as dev dep

thomasdhc:


docker/common/install.sh
```diff
@@ -86,6 +86,7 @@ main() {
     bash docker/common/install_conflicting_deps.sh
 
     # Install the package
+    uv pip install "cut-cross-entropy @ git+https://github.com/apple/ml-cross-entropy.git@87a86ab"
```
ko3n1g: This works, but it makes it hard to users to install. We could alos move cut-cross-entry to the dev-dependencies. Dev-dependencies are not tied to pypi, but installed with `uv sync`. This might be better for users that uv sync our project. 

This still leaves pip users out, so we push to for a proper fix asap

thomasdhc: Ah, nice. Let's do that instead

pyproject.toml
```diff
@@ -108,6 +107,8 @@ linting = [
 ]
 test = ["coverage", "pytest", "peft"]
 nvfsdp = ["nvfsdp"]
+dev = ["cut-cross-entropy @ git+https://github.com/apple/ml-cross-entropy.git@87a86ab"]
```
adil-a, line 110: If a user does a `uv run ...` would that auto-install this?

thomasdhc: uv sync should

adil-a: > [Locking and syncing are automatic in uv. For example, when uv run is used, the project is locked and synced before invoking the requested command. This ensures the project environment is always up-to-date.](https://docs.astral.sh/uv/concepts/projects/sync/#automatic-lock-and-sync)

I think `uv run` calls `uv sync` under the hood so we should be good.

adil-a (dismissed):
lgtm

pyproject.toml
ko3n1g: yeah, as long as the dependency group is part of the default groups it should get installed

.github/workflows/build-test-publish-wheel.yml
```diff
@@ -33,10 +33,8 @@ jobs:
       python-package: nemo_automodel
       python-version: "3.12"
       packaging: uv
-      no-publish: ${{ !(github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/r')) }}
     secrets:
       TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
```
ko3n1g: Why did we remove this?

thomasdhc: I had ran into this error: https://github.com/NVIDIA-NeMo/Automodel/actions/runs/16445839576

thomasdhc: Ah I think the testing branch I was using was an older one, i see that in main those variables are present. Let me add it back

ko3n1g (approved):

PR #266, opened 2025-07-23
fix: add kwargs argument to mock tokenizers

torsli:
Adds `**kwargs` to mock tokenizers. This ensures that the tokenizer will be compatible with `recipes.llm.finetune.FinetuneRecipeForNextTokenPrediction`, which assumes they will accept a `tokenizer` keyword, but also with any keyword that might be required by another recipe.

### DIFF
--- a/nemo_automodel/components/datasets/llm/mock.py
+++ b/nemo_automodel/components/datasets/llm/mock.py
@@ -44,7 +44,7 @@ def build_unpacked_dataset(
     vocab_size: int = 100,
     max_sentence_len: int = 64,
     seed: int = 0,
-    **kwargs,
+    tokenizer=None,
 ):
     """
     Build a dataset where each example is one sentence (variable length).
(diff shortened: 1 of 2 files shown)


nemo_automodel/components/datasets/llm/mock.py
```diff
@@ -44,6 +44,7 @@ def build_unpacked_dataset(
     vocab_size: int = 100,
     max_sentence_len: int = 64,
     seed: int = 0,
+    **kwargs,
```
akoumpa:
```suggestion
    tokenizer=None,
```

nemo_automodel/components/datasets/llm/mock_packed.py
```diff
@@ -62,6 +62,7 @@ def build_packed_dataset(
     vocab_size: int = 100,
     max_sentence_len: int = 64,
     seed: int = 0,
+    **kwargs,
```
akoumpa:
```suggestion
    tokenizer=None,
```

akoumpa (approved):
LGMT, thank you.

PR #298, opened 2025-07-28
feat: Simplify logic for mixed_precision, offload_policy and device_mesh

ffrujeri:
Also Remove assertion for tied embeddings for optimized plans.

### DIFF
--- a/nemo_automodel/components/distributed/parallelizer.py
+++ b/nemo_automodel/components/distributed/parallelizer.py
@@ -24,7 +24,7 @@ from torch import nn
 from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
     checkpoint_wrapper,
 )
-from torch.distributed.device_mesh import DeviceMesh, _mesh_resources
+from torch.distributed.device_mesh import DeviceMesh
 from torch.distributed.fsdp import (
     CPUOffloadPolicy,
     FSDPModule,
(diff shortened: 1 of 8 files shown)


nemo_automodel/components/distributed/parallelizer.py
```diff
@@ -280,13 +274,7 @@ def fsdp2_strategy_parallelize(
         num_key_value_heads = model.config.num_key_value_heads
 
     # Set FSDP sharding mesh to context parallel mesh if CP > 1, else default to the data parallel mesh.
-    dp_mesh = device_mesh[
```
BoxiangW, line 283: This is needed for Automodel also it will bring issues for FSDP + CP usecase

nemo_automodel/components/distributed/parallelizer.py
```diff
@@ -363,15 +351,11 @@ def fsdp2_strategy_parallelize(
...
     if not mp_policy:
         mp_policy = MixedPrecisionPolicy(
-            param_dtype=param_dtype,
+            param_dtype=torch.bfloat16,
             reduce_dtype=torch.float32,
             output_dtype=torch.float32,
         )
 
-    # Set up offload policy
-    if not offload_policy:
```
BoxiangW, line 372: Can we leave this offload option as it is?

nemo_automodel/components/distributed/parallelizer.py
ffrujeri: Yeah, my thought was to make the caller be responsible to pass the the mash name, given it already did the flattening and created the "dp_cp" mesh. what do you think? So in that case instead of the "data_parallel" it would pass the "dp_cp" as the dp_mesh_name

nemo_automodel/components/distributed/parallelizer.py
ffrujeri: So, notice that the logic is the same right? I am just simplifying to instead of the have the boolean and the policy object as arguments we just have one.

ffrujeri: Oh, you mean to keep the default of pin_memory=False?

nemo_automodel/components/distributed/parallelizer.py
ffrujeri: Discussed and resolved offline. We will leave as it and the caller will be responsible to pass the correct mesh name in case of "dp" or "dp_cp".

nemo_automodel/components/distributed/parallelizer.py
ffrujeri: Resolved offline as well the current logic is functionally the same as before.

PR #299, opened 2025-07-29
ci: Group cicd tests

thomasdhc:


ko3n1g (approved):
Thanks for doing this, it will be great for the infra!

The buckets look good to me, it's probably something that you discussed in the team. 

I wonder about keeping the shell scripts, since now engineers that add shell-based tests will need to remember to also inject it into the pytest file. 

I would recommend to update the contrib guide to explain the setup and how to add new tests. 

Additionally, in other projects, I mentioned that if pre-existing buckets don't work semantically, they should reach out to the automation-team to discuss the best place for their new test.

Thanks again for picking this up so quickly!

akoumpa (approved):
LGTM, thank you.

thomasdhc:
Yeah @ko3n1g thanks for the review, I wanted to maintain the shell scripts as an easy way for engineers to reproduce the tests on their own dev machines etc, let me follow up and add to contrib guide

tests/functional_tests/hf_transformer_vlm/test_hf_transformer_vlm.py
```diff
@@ -24,11 +24,11 @@
 
 
 class TestHFTransformerVLM:
-    def test_hf_transformer_vlm_fsdp2_tp2(self):
-        run_test_script(TEST_FOLDER, HF_TRANSFORMER_VLM_FSDP2_TP2_FILENAME)
+    # def test_hf_transformer_vlm_fsdp2_tp2(self):
```
ko3n1g: what do you think about the `@pytest.mark.pleasefixme` marker that we use for unit tests?

thomasdhc: oh yeah that's a good point

PR #300, opened 2025-07-29
fix: Fixes for running Baichuan

torsli:
Two fixes to make Baichuan2-7B-Chat run correctly.

1. Baichuan requires `trust_remote_code` to be True, but the finetune recipe won't pick up the flag if it's in the tokenizer config. This tweaks `finetune.py` to pick up this flag.
2. The BaichuanConfig object doesn't have a value for `num_key_value_heads` and the FSDP2 config code assumes that it does. Here we default it to the num of attention heads, which is the standard default for MHA.


nemo_automodel/components/distributed/parallelizer.py
```diff
@@ -277,7 +277,10 @@ def fsdp2_strategy_parallelize(
     else:
         layers = model.model.layers
         num_attention_heads = model.config.num_attention_heads
-        num_key_value_heads = model.config.num_key_value_heads
+        if hasattr(model.config, "num_key_value_heads"):
+            num_key_value_heads = model.config.num_key_value_heads
+        else:
+            num_key_value_heads = model.config.num_attention_heads
```
akoumpa: Hi @torsli , are you ok with moving the whole if/else block into a function (e.g., `get_num_attention_and_kv_heads`) wrapped with a try/catch statement? I recognize this issue is not introduced in this PR.

torsli: @akoumpa I moved it to its own method. I don't think it needs a try/catch unless you want to throw a different exception, and in that case I don't know what exception you'd rather throw.

nemo_automodel/recipes/llm/finetune.py
```diff
@@ -208,7 +208,10 @@ def build_dataloader(
             "rank": device_mesh["data_parallel"].get_local_rank(),
         }
     if "tokenizer" not in cfg_ds:
-        tokenizer = AutoTokenizer.from_pretrained(cfg_model.pretrained_model_name_or_path)
+        trust_remote_code = cfg_model.trust_remote_code if hasattr(cfg_model, "trust_remote_code") else False
```
akoumpa:
```suggestion
        trust_remote_code = getattr(cfg_model, "trust_remote_code", False)
```

akoumpa (dismissed):
LGTM, thank you.

akoumpa:
ping @NVIDIA-NeMo/automation to review the changes. Thank you.

torsli:
merge conflict mess-up, opening new PR

PR #303, opened 2025-08-01
fix: changes for running Baichuan

torsli:
Change tp mesh validation to support BaichuanConfig
Change finetune recipe to ensure trust-remote-code gets flagged correctly


akoumpa (approved):
LGTM, thanks.

PR #314, opened 2025-08-05
ci: Ensure coverage runs in parallel-mode

thomasdhc:


.github/workflows/cicd-main.yml
```diff
@@ -169,7 +169,7 @@ jobs:
             timeout: 40
           # - test-name: L2_HF_Transformer_LLM
           #   test-folder: hf_transformer_llm
-          #   runner: linux-amd64-gpu-rtxa6000-latest-2-nemo
+          #   runner: self-hosted-nemo
```
chtruong814: We still want to leave this test commented out?

.github/workflows/cicd-main.yml
```diff
@@ -110,7 +110,7 @@ jobs:
       matrix:
         include:
           - test-name: L0_Unit_Tests_CPU
-            runner: linux-amd64-cpu16
+            runner: self-hosted-nemo
```
chtruong814: Did you intend to make this change?

thomasdhc: Thanks ugh did not mean for that to be here

PR #315, opened 2025-08-05
feat: checkpointing config yaml and model generation config

adil-a:
Adds checkpointing config yaml and model generation config as part of the checkpointed states

### DIFF
--- a/nemo_automodel/components/checkpoint/checkpointing.py
+++ b/nemo_automodel/components/checkpoint/checkpointing.py
@@ -25,6 +25,7 @@ import torch
 import torch.distributed
 import torch.distributed.checkpoint as dcp
 import torch.nn as nn
+import yaml
 from safetensors import safe_open
 from safetensors.torch import save_file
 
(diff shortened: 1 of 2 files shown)


nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -105,6 +108,9 @@ def save_model(
             # save the config.json file
             with open(os.path.join(consolidated_model_path, "config.json"), "w") as f:
                 f.write(model.config.to_json_string())
+            # save the generation_config.json file
+            with open(os.path.join(consolidated_model_path, "generation_config.json"), "w") as f:
+                f.write(model.generation_config.to_json_string())
```
akoumpa: can we check if the model has the `generation_config` attribute and similarly the `.config` above?

nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -260,6 +266,18 @@ def load_optimizer(
     optimizer_state.load_state_dict(reinstated_state_dict)
 
 
+def save_config(config: "ConfigNode", weights_path: str):
```
akoumpa: would it be ok to use a `dict` instead of a `ConfigNode`?

PR #335, opened 2025-08-07
feat: parallel checkpoint loading

adil-a:
Loads HF model weights in parallel at init time. Also syncs HFStorageReader with PT main - no longer deserializes the full file rather uses file seeks to read the relevant bytes for the relevant tensor.

Slight loss discrepancy with meta devices. I have raised this issue with PT team [here](https://github.com/pytorch/pytorch/issues/160340).

Llama-3.2-1B:
image

Llama-3.1-8B:
image

Qwen-2.5-7B:
image


nemo_automodel/recipes/llm/finetune.py
```diff
@@ -91,6 +98,15 @@ def build_model_and_optimizer(
     Returns:
         The instantiated model on the specified device and optimizer.
     """
+    is_meta_device = False
```
akoumpa, line 101: can we make loading on meta device the default and if loading fails, retry without meta device (we can use a try/catch for the retry)?

Also, instead of having the `is_meta_device` option in yaml can we use a `no_meta_device` option which explicitly disables (the default) loading to meta?

adil-a: Copying from Slack:

> I am a bit hesitant making it the default currently especially given that meta device gives slightly off numerical outputs (an even bigger problem if other FWs like RL use AutoModel). Can we maybe make this an issue and hold off on making it the default till PT team addresses the issue? I have it open here https://github.com/pytorch/pytorch/issues/160340

Maybe we can aim to make it the default for 25.11 contingent on this issue being addressed by PT?

akoumpa: let's merge this as is, and deprecate the `is_meta_device` once there is a fix.

akoumpa (approved):
Approving, with a note to deprecate the on_meta_device flag in the near future.

PR #345, opened 2025-08-11
Add pipeline parallel support

hemildesai:


nemo_automodel/components/datasets/llm/hellaswag.py
```diff
@@ -37,13 +46,17 @@ def __init__(self, path_or_dataset, tokenizer, split="train", num_samples_limit=
             split (str, optional): Dataset split to use (e.g., 'train', 'validation'). Defaults to 'train'.
             num_samples_limit (int, optional): Maximum number of samples to load. Defaults to None.
             trust_remote_code (bool, optional): Whether to trust remote code. Defaults to True.
+            shuffle (bool, optional): Whether to shuffle the dataset. Defaults to False.
```
adil-a, line 49: do we need shuffling at the dataset level? Distributed sampler should take care of shuffling AFAIK

nemo_automodel/components/loss/masked_ce.py
```diff
@@ -68,4 +69,11 @@ def __call__(
                 del mask
         if self.fp32_upcast:
             logits = logits.float()
+
+        if isinstance(logits, DTensor):
+            logits = logits.full_tensor()
+
+        if isinstance(labels, DTensor):
+            labels = labels.full_tensor()
+
```
adil-a, line 78: does this change need to be made for the other losses as well?

nemo_automodel/components/distributed/autopipeline/hf_utils.py
```diff
@@ -0,0 +1,296 @@
...
+
+    if isinstance(model_id_or_config, str):
+        cfg = AutoConfig.from_pretrained(model_id_or_config)
+    else:
+        cfg = model_id_or_config
+
+    with torch.device("meta"):
+        model = AutoModel.from_config(cfg, **kwargs)
+
+    return model
```
adil-a, line 263: 1. We should be using the extended classes from AutoModel instead of HF directly
2. We should be initializing models in a centralized way. It will be annoying as a user experience if we have model init inside the PP class when using PP and outside in all other cases. Let's keep model init inside the main recipe? Specifically as it already is in main

nemo_automodel/recipes/llm/finetune.py
```diff
@@ -112,27 +139,55 @@ def build_model_and_optimizer(
...
-            trainable_params = list(filter(lambda x: x.requires_grad, model.parameters()))
-            assert len(trainable_params) > 0, "trainable_params cannot be empty"
-            if tp_size > 1:
-                # TP does not support foreach
-                cfg_opt.foreach = False
-            optimizer = cfg_opt.instantiate(params=trainable_params)
+    if autopipeline is not None:
+        autopipeline.build(model, loss_fn=loss_fn, parallelize_fn=parallelize_fn)
+        autopipeline.materialize(
+            init_buffers_fn=init_hf_model_buffers,
```
adil-a, line 145: I think we won't need this as part of changes in 363

nemo_automodel/components/distributed/autopipeline/hf_utils.py
hemildesai: Yeah this is just a helper I was using, I can remove it.

PR #355, opened 2025-08-14
feat: add priorities to `attn_implementation` fallback

akoumpa:
When instantiation of a model fails, the `from_pretrain` function will retry to instantiate the model with another attention implementation. Previously, the fallback would use `eager`. This PR makes use of a predefined priority priority list so that eager is only used as last resort.

### DIFF
--- a/nemo_automodel/components/_transformers/auto_model.py
+++ b/nemo_automodel/components/_transformers/auto_model.py
@@ -239,6 +239,7 @@ class _BaseNeMoAutoModelClass(_BaseAutoModelClass):
             )
 
         # load model
+        model = None
         try:
             name = cls.__name__
             if name.startswith("NeMo"):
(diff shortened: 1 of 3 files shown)


nemo_automodel/components/_transformers/auto_model.py
```diff
@@ -149,7 +177,7 @@ def from_pretrained(
         use_sdpa_patching: bool = True,
         sdpa_method: Optional[List[SDPBackend]] = None,
         torch_dtype="auto",
-        attn_implementation: str = "flash_attention_2",
+        attn_implementation: str = "sdpa",
```
HuiyingLi: why do we change the default from fa2 to sdpa?

akoumpa: fa2 is not deterministic by default, so I switched to sdpa. But i will revert and find another way to ensure determinism.

HuiyingLi (approved):
LGTM

PR #360, opened 2025-08-15
ci: Alpha changelog and package version

thomasdhc:


ko3n1g (changes requested):
Bump the workflow requires additional inputs. I shared those offline

PR #368, opened 2025-08-19
refactor: update fp8 recipe and refac config

HuiyingLi:


examples/llm/llama_3_1_8b_hellaswag_fp8.yaml
```diff
@@ -36,11 +36,12 @@ rng:
...
 model:
   _target_: nemo_automodel.NeMoAutoModelForCausalLM.from_pretrained
-  pretrained_model_name_or_path: meta-llama/Meta-Llama-3.1-8B
+  pretrained_model_name_or_path: meta-llama/Llama-3.1-8B
   torch_dtype: bf16
+  trust_remote_code: true
 
 checkpoint:
-  enabled: true
+  enabled: false
```
akoumpa:
```suggestion
  enabled: true
```

examples/llm/llama_3_1_8b_hellaswag_fp8.yaml
```diff
@@ -52,8 +53,15 @@ distributed:
...
+compile:
+  enabled: true
+  mode: "default"  # Options: "default", "reduce-overhead", "max-autotune"
+  fullgraph: false
+  dynamic: false  # Set to false for better performance with fixed shapes
+  backend: null  # Use default backend (inductor)
+
 fp8:
-  _target_: nemo_automodel.components.quantization.fp8.FP8Config
+  enabled: true
```
akoumpa, line 64:
```suggestion
```

I'm recommending to drop the `enabled` flags. If the config is present then it can be considered enabled, if it's not present then it can be considered disabled. If we want to overwrite a config section we can do `--fp8 none` that should work, if not we can update the loader.py

HuiyingLi: thank you @akoumpa ! two reasons I added the enabled flag:
- mainly checkpoint and compile both have it so to keep consistent
- I find pass in enable via cmd convenient than commenting it from the config
wdyt?

akoumpa: I'm thinking we can simplify the configuration, for example, if the relevant section is present, the feature can be considered enabled, if it's not present it fallbacks to a default value and if the user explicitly passes `--feature None` then it's disabled. I'm thinking that semantically there's some overlap between having the section present and having an enable option. We can apply this to all features in the yaml file and be consistent. I'm ok with proceeding with the `enabled` option here, and having another PR to remove it across the board (actually that's probably better).

HuiyingLi: Sounds good! Thanks for your suggestion!

thomasdhc (approved):

PR #374, opened 2025-08-20
ci: FA install configuration

thomasdhc:


.github/workflows/install-test.yml
```diff
@@ -149,7 +149,8 @@ jobs:
           PIP_ARGS=()
 
           if [[ $EXTRA == *"fa"* ]]; then
-            pip install torch setuptools psutil
+            pip install --index-url https://download.pytorch.org/whl/cu128 torch
```
ko3n1g:
```suggestion
            pip install --index-url https://download.pytorch.org/whl/cu128 "torch<=2.8.0"
```

I would recommend to pin it here too to make sure it doesn't drift

pyproject.toml
```diff
@@ -111,13 +111,26 @@ linting = [
...
+[[tool.uv.index]]
+name = "pytorch"
+url = "https://download.pytorch.org/whl/cu128"
+explicit = true
+
 [tool.uv.sources]
 nvfsdp = { git = "https://github.com/NVIDIA-NeMo/nvFSDP.git", rev = "d97136886a65d70f298a4b361571ac75c20198af" }
+torch = [
+    { index = "pytorch", marker = "sys_platform != 'darwin'" },
+    { index = "pypi", marker = "sys_platform == 'darwin'" },
```
ko3n1g, line 132: Does this mean it falls back to pypi in case pytorch is down?

ko3n1g (dismissed):
One suggested change, but looks good to me. Thanks!

pablo-garay (dismissed):

pyproject.toml
thomasdhc: no, it seems like there's no cuda whl for macos, so the macos pypi install test was failing

.github/workflows/install-test.yml
thomasdhc: Yup, this will be good to have. Thanks!

ko3n1g (approved):

PR #411, opened 2025-09-02
feat: fallback to MaskedCrossEntropy if model does not support logits_to_keep

akoumpa:
For models that lack the `logits_to_keep` in their `forward` arg, fallback to `MaskedCrossEntropy`.


HuiyingLi (approved):
LGTM

PR #423, opened 2025-09-05
fix: fix slurm launcher

HuiyingLi:
- fix run cmd
- support custom env var
- add gpus_per_node option
- switch to torchrun for launching

### DIFF
--- a/tests/unit_tests/launcher/config.py
+++ b/tests/unit_tests/launcher/config.py
@@ -1,3 +1,17 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


nemo_automodel/components/launcher/slurm/template.py
```diff
@@ -47,7 +47,7 @@
 #SBATCH -A {account}
 #SBATCH -p {partition}
 #SBATCH -N {nodes}
-#SBATCH --ntasks-per-node {ntasks_per_node}
+#SBATCH --ntasks-per-node 1{gpus_per_node_directive}
```
akoumpa, line 50:
```suggestion
#SBATCH --ntasks-per-node {gpus_per_node_directive}
```

HuiyingLi: Thank you @akoumpa  --ntasks-per-node=1 so that srun doesn't doesnt launch 8 processes, so that torchrun does it instead.

akoumpa (approved):
LGTM, thank you.

PR #429, opened 2025-09-08
feat: add Metric logger with jsonl output

akoumpa:
Changes:
- Introduces the `MetricLogger` class and the `MetricsSample` dataclass.
-  `MetricLogger` enables training loops to log metrics to jsonl files.


adil-a (approved):
LGTM thank you

PR #440, opened 2025-09-09
fix: erroneous buffer keys in ckpt (#433)

akoumpa:
* bug fix

* fixing unit test

* removing test running twice

---------


adil-a (approved):
thank you for cherry picking it

PR #442, opened 2025-09-09
feat: Add benchmarking script + cfgs for dsv3, fix: te padding mask

hemildesai:
Adds a benchmarking script to catpure different timers and MFU for llm next token prediction.
Adds configs for moonlight and dsv3 for TE + DeepEP and pure PyTorch based implementations

### DIFF
--- a/examples/benchmarking/configs/moonlight_16b_te_deepep.yaml
+++ b/examples/benchmarking/configs/moonlight_16b_te_deepep.yaml
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 distributed:
-  pp_size: 2
-  ep_size: 4
+  pp_size: 1
+  ep_size: 8
   dp_replicate_size: 1
 
 model:
@@ -33,9 +33,9 @@ model:
 parallelize_fn: nemo_automodel.components.models.deepseek_v3.parallelizer.parallelize_model
 
 training:
-  seq_len: 1024
-  global_batch_size: 128
-  local_batch_size: 32
+  seq_len: 2048
+  global_batch_size: 256
+  local_batch_size: 16
   steps: 30
   warmup_steps: 10
   seed: 1234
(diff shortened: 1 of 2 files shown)


hemildesai:
/ok to test [62ef06d](https://github.com/NVIDIA-NeMo/Automodel/pull/442/commits/62ef06dcc466dbd72ee0aa9f04c7db8d05f1f184)

hemildesai:
/ok to test [f70fd1f](https://github.com/NVIDIA-NeMo/Automodel/pull/442/commits/f70fd1f79cb052d9057dd4d836a33ac385024654)

hemildesai:
/ok to test [2e2dc4d](https://github.com/NVIDIA-NeMo/Automodel/pull/442/commits/2e2dc4df06676186cdf4a6df6e85901603018aad)

examples/benchmarking/configs/deepseek_v3_te_deepep.yaml
adil-a, line 1: do you think we should keep configs streamlined? i.e., I view it as that there shouldn't be much difference between the configs for benchmarking vs the ones for training. I see differently named flags here like `training` or `pipelining` not using AutoPipeline etc. Wdyt?

examples/benchmarking/next_token_prediction.py
```diff
@@ -0,0 +1,355 @@
...
+        if cfg.distributed.ep_size > 1:
+            mesh["moe"] = init_device_mesh(
+                device_type="cuda",
+                mesh_shape=(
+                    cfg.distributed.pp_size,
+                    cfg.distributed.dp_shard_size // cfg.distributed.ep_size,
+                    cfg.distributed.ep_size,
+                ),
+                mesh_dim_names=("pp", "ep_shard", "ep"),
+            )
```
adil-a, line 97: what's the reason for not using the functions defined in AutoModel?

To clarify, my understanding of the benchmarking scripts is that if we bring in a custom model, what is the perf like using existing AutoModel APIs? If we have divergence in how we do setups or other functions, it can end up being somewhat misrepresented

hemildesai: This benchmarking script rn is intended to be a standalone script to keep it as simple as possible for now. In the future, we should move the benchmarking specifics from this script to the main recipe itself and unify the configs.

PR #456, opened 2025-09-16
fix: erroneous import

akoumpa:
fixes minor error in the import path;  alternatively, the recipe can be used directly, e.g.:
```
torchrun --nproc-per-node=4 nemo_automodel/recipes/llm/kd.py -c ..
```
instead of the one in examples/


HuiyingLi (approved):
LGTM

PR #474, opened 2025-09-17
feat: deepseek-v2 backward compatability

torsli:
Adds support for greedy expert selection to the MoE layers, which will allow the DeepSeek V3 model to run DeepSeek V2 checkpoints.

Label r0.1.0 added to request a cherry-pick if possible, but it's not a must.


nemo_automodel/components/moe/layers.py
```diff
@@ -472,23 +475,28 @@ def forward(
...
+                else:
+                    group_scores = scores.topk(2, dim=-1)[0].sum(dim=-1)
+
+                indices = group_scores.topk(self.topk_groups, dim=-1)[1]
+                mask = torch.zeros_like(scores[..., 0]).scatter_(1, indices, True)
+                scores = (scores * mask.unsqueeze(-1)).flatten(1)
+
+            indices = torch.topk(scores, self.topk, dim=-1)[1]
+            weights = original_scores.gather(1, indices)
+        elif self.topk_method == "greedy":
```
hemildesai, line 496: I think you can achieve greedy by setting n_groups = 0 and gate_bias_update_factor=0 in the MoEConfig

PR #481, opened 2025-09-19
fix: skip TP parallelization if TP=1 in NemotronHParallelizationStrategy

akoumpa:


nemo_automodel/components/distributed/parallelizer.py
```diff
@@ -215,25 +215,23 @@ def parallelize(
...
 
+        layers: torch.nn.ModuleList = model.backbone.layers
         tp_mesh = device_mesh[tp_mesh_name]
-        dp_mesh_dim_names = (dp_replicate_mesh_name, dp_shard_cp_mesh_name)
-        dp_mesh = device_mesh[dp_mesh_dim_names]
-
-        model_tp_plan: dict[str, ParallelStyle] = {
-            "lm_head": ColwiseParallel(output_layouts=Shard(-1), use_local_output=False),
-        }
+        if tp_mesh.size() > 1:
```
ffrujeri, line 220: @gshennvm and @terrykong, is it ok to do this change?

The original version of this logic was worked [in here](https://github.com/NVIDIA-NeMo/RL/blob/main/nemo_rl/models/dtensor/parallelize.py#L404)

gshennvm: yes it's ok as long as tests pass

adil-a (approved):
lgtm

PR #491, opened 2025-09-22
fix: TE padding masks for DSv3 MLA layer

hemildesai:
Changes TE DotProduceAttention mask type to `padding_causal` for correctness.


akoumpa (approved):
LGTM, thanks.

PR #493, opened 2025-09-22
feat: Add custom implementation for GPT OSS

hemildesai:
Gpt oss 20b convergence on Hellaswag - https://wandb.ai/nvidia/automodel-pp/runs/yms2512f?nw=nwuserhemild
Screenshot 2025-09-23 at 2 34 45 PM


hemildesai:
/ok to test [4d78b11](https://github.com/NVIDIA-NeMo/Automodel/pull/493/commits/4d78b11113fb967ee2162ff3a4ed1165495a49a6)

examples/llm_finetune/gpt_oss/gpt_oss_20b.yaml
```diff
@@ -0,0 +1,118 @@
...
+rng:
+  _target_: nemo_automodel.components.training.rng.StatefulRNG
+  seed: 1111
+  ranked: true
+
+parallelizer:
+  _target_: nemo_automodel.components.moe.parallelizer.parallelize_model
+  enable_ac: false
+
+model:
```
akoumpa, line 36: things to consider for future refactoring:

1. user should be able to use `from_pretrained` as `_target_` of `NeMoAutoModelForCausalLM` class, not be constrained by the specific class.
2. `is_meta_device` was meant to be temporary, and IIRC @adil-a had the AI to remove it once an issue was resolved
3. I'd like to remove the backend altogether. If the user wants to use the vanilla implementation, they can use `_target_: AutoModelForCausalLM.from_pretrained`; if they want to use the optimized implementation, then they can use `_target_: NeMoAutoModelForCausalLM`. I'm ok with adding one option to `NeMoAutoModelForCausalLM` to switch to the baseline implementation, but by default it should return the optimized one.
4. `enable_hf_state_dict_adapter` I expect users not to know when to use this, and therefore, I feel it would be better to consume this technical debt inside automodel's code.

hemildesai: Created https://github.com/NVIDIA-NeMo/Automodel/issues/501 to track.

adil-a (approved):
approving to remove blockers for other PRs + make sure this is merged for container release.

@akoumpa there's an issue created to track your comments. I'm in agreement with them so we should address them ahead of 25.11

PR #511, opened 2025-09-25
fix: make mcore ds helper compile on local rank0 only

HuiyingLi:


nemo_automodel/components/datasets/llm/megatron_dataset.py
```diff
@@ -94,13 +96,13 @@ def __init__(
...
+                    compile_helper()
+                dist.barrier()
+            else:
                 compile_helper()
-                assert find_spec("nemo_automodel.components.datasets.llm.megatron.helpers_cpp") is not None
-            except AssertionError:
-                raise ImportError(
-                    "Could not compile megatron dataset C++ helper functions and therefore cannot import helpers python file."
-                )
+            assert find_spec("nemo_automodel.components.datasets.llm.megatron.helpers_cpp") is not None
```
adil-a: can we keep the try except block with the error message?

adil-a (changes requested):

PR #518, opened 2025-09-26
feat: add 2l moonlight functional test

akoumpa:


nemo_automodel/components/models/deepseek_v3/model.py
```diff
@@ -214,16 +217,12 @@ class DeepseekV3ForCausalLM(nn.Module):
     @classmethod
     def from_config(
         cls,
-        pretrained_model_name_or_path: str | DeepseekV3Config,
+        config: DeepseekV3Config,
```
adil-a, line 220: this change will break our current yaml configs for dsv3 and moonlight. can you make the appropriate changes for finetune and pretrain yamls?

nemo_automodel/components/models/gpt_oss/model.py
```diff
@@ -184,15 +187,11 @@ class GptOssForCausalLM(nn.Module):
     @classmethod
     def from_config(
         cls,
-        pretrained_model_name_or_path: str | GptOssConfig,
+        config: GptOssConfig,
```
adil-a, line 190: same as above

adil-a (changes requested):
please add yaml changes to our recipes. otherwise lgtm

PR #520, opened 2025-09-26
refactor: TP parallel style with LoRA

akoumpa:
Based on feedback from @soluwalana.

The new TP plan preserve requires_grad, but still causes redistribute

### DIFF
--- a/nemo_automodel/components/distributed/parallel_styles.py
+++ b/nemo_automodel/components/distributed/parallel_styles.py
@@ -22,26 +22,33 @@ from torch.distributed.tensor.parallel import (
     RowwiseParallel,
 )
 
+def _distribute_param(_module, name, device_mesh, src_data_rank, placements):
+    param = getattr(_module, name)
+    dist_param = nn.Parameter(
+        distribute_tensor(param, device_mesh, placements, src_data_rank=src_data_rank),
+        requires_grad=param.requires_grad,
+    )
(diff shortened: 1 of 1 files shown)


nemo_automodel/components/distributed/parallel_styles.py
```diff
@@ -11,53 +11,92 @@
...
 
 
+def _distribute_param(_module, name, device_mesh, src_data_rank, placements):
+    param = getattr(_module, name)
+    dist_param = nn.Parameter(
+        distribute_tensor(param, device_mesh, placements, src_data_rank=src_data_rank),
+        requires_grad=param.requires_grad,
+    )
+    assert dist_param.requires_grad == param.requires_grad
+    _module.register_parameter(name, dist_param)
```
adil-a, line 37: OOC since we create a new parameter that we distribute, do we need to delete the old parameter by any chance or will this just overwrite the attribute (so delete the parameter)?

adil-a (approved):
small comment but lgtm

nemo_automodel/components/distributed/parallel_styles.py
akoumpa: I suspect that it goes to https://github.com/pytorch/pytorch/blob/6eb8d9671b29e1aa46ce3a44cddee59fdd89ca4f/torch/nn/modules/module.py#L637 , then `_parameters` is a dictionary, therefore only holds one param with that name.

PR #548, opened 2025-10-01
feat: Add benchmarking recipe and configs

hemildesai:
Benchmark recipe subclasses train_ft.py recipe and reuses `setup` and `_forward_backward_step`


examples/benchmark/mock_dataset.py
akoumpa, line 1: can we use the mock dataset we already have?

hemildesai: That one produces all the samples at once and is very inefficient if the total number of samples is high or the vocabulary size is high. Ref: https://github.com/NVIDIA-NeMo/Automodel/blob/5d59fa6d73e7e0f65c3f13e605b680472ea1103a/nemo_automodel/components/datasets/llm/mock.py#L64-L80

hemildesai: Happy to make the [examples/benchmark/mock_dataset.py](https://github.com/NVIDIA-NeMo/Automodel/pull/548/files/d28444aadf9efc52b4c55cab66b6019da09a6117#diff-f464f8304b5e9ab15e27b058c9dbade04795ec8446f8afc4e7854654f4566810) as the default mock dataset though

hemildesai: Moved dataset to `nemo_automodel/components/datasets/llm/mock_iterable_dataset.py` so that it can be reused elsewhere.

hemildesai:
/ok to test [7b89dd0](https://github.com/NVIDIA-NeMo/Automodel/pull/548/commits/7b89dd03db2d5f6a91ffba797eb016a2af8dcdab)

examples/benchmark/mock_dataset.py
akoumpa: thanks, I think the command have been cp instead of mv :)

examples/benchmark/benchmark_recipe.py
```diff
@@ -0,0 +1,327 @@
...
+
+from nemo_automodel.components.config._arg_parser import parse_args_and_load_config
+from nemo_automodel.components.training.timers import Timers
+from nemo_automodel.components.utils.flops_utils import get_flops_formula_for_hf_config
+from nemo_automodel.recipes.llm.train_ft import TrainFinetuneRecipeForNextTokenPrediction
+
+logger = logging.getLogger(__name__)
+
+
+def calculate_mfu(tflops, world_size, time_seconds, reference_mfu=1979.0):
```
akoumpa: I would move this inside flops_utils.py?

examples/benchmark/benchmark_recipe.py
```diff
@@ -0,0 +1,327 @@
...
+            cfg: Configuration dictionary/object for benchmarking.
+        """
+        # Store and remove benchmarking-specific parameters
+        if "step_scheduler" in cfg:
+            step_cfg = cfg.step_scheduler
+            self._bench_steps = step_cfg.get("max_steps", 30)
+            self._bench_warmup_steps = step_cfg.get("warmup_steps", 10)
+
+            # Remove benchmarking-specific parameters that StepScheduler doesn't accept
+            if "warmup_steps" in step_cfg:
```
akoumpa: does it make sense to have a `benchmark:` section in the yaml with all the info that's related to benchmarking? I was thinking it could contain warmup_steps, nsys start/end and device flops?

examples/benchmark/benchmark_recipe.py
akoumpa: also can we rename 
`examples/benchmark/benchmark_recipe.py` -> `nemo_automodel/recipes/llm/benchmark.py`

examples/benchmark/benchmark_recipe.py
```diff
@@ -0,0 +1,327 @@
...
+            # Call parent setup
+            super().setup()
+
+        # Clear validation dataloader (not needed for benchmarking)
+        self.val_dataloader = None
+
+        # Get step_scheduler config
+        step_cfg = self.cfg.get("step_scheduler", {})
+        seq_len = self._bench_seq_len
+        global_batch_size = step_cfg.get("global_batch_size", 256)
```
akoumpa:
```suggestion
        global_batch_size = self.cfg.get("step_scheduler.global_batch_size")
```

examples/benchmark/benchmark_recipe.py
```diff
@@ -0,0 +1,327 @@
...
+        with timers and profiling support, similar to the original benchmarking script.
+        """
+        rank = self.dist_env.rank
+        device = self.dist_env.device
+
+        # Get benchmarking config
+        step_cfg = self.cfg.get("step_scheduler", {})
+        steps = self._bench_steps
+        warmup_steps = self._bench_warmup_steps
+        local_batch_size = step_cfg.get("local_batch_size", 4)
```
akoumpa:
```suggestion
        local_batch_size = self.cfg.get("step_scheduler.local_batch_size")
```
Same for `global_batch_size`.

I want to avoid silent bugs, for example, if the local_batch_size is not set I want it to fail instead of using some default value. Are you ok with accessing attributes directly?

For example, a workflow could be
1. check if input yaml has required config, and if not set it in __init__
2. when super().setup is called, yaml is logged with all the used options.

examples/benchmark/mock_dataset.py
hemildesai: Updated.

examples/benchmark/benchmark_recipe.py
hemildesai: Done.

examples/benchmark/benchmark_recipe.py
hemildesai: Both done.

examples/benchmark/benchmark_recipe.py
hemildesai: Done.

examples/benchmark/benchmark_recipe.py
hemildesai: Updated the recipe to avoid using defaults.

hemildesai:
/ok to test [fee6fb9](https://github.com/NVIDIA-NeMo/Automodel/pull/548/commits/fee6fb96aadee2b2e3be67e44e5db8e90450e772)

nemo_automodel/recipes/llm/benchmark.py
```diff
@@ -0,0 +1,315 @@
...
+        with self.timers("setup", log_level=1):
+            # Call parent setup
+            super().setup()
+
+        # Clear validation dataloader (not needed for benchmarking)
+        self.val_dataloader = None
+
+        # Get step_scheduler config
+        seq_len = self._bench_seq_len
+        global_batch_size = getattr(self.cfg.step_scheduler, "global_batch_size", 256)
```
akoumpa:
```suggestion
        global_batch_size = self.cfg.step_scheduler.global_batch_size
```

hemildesai: Fixed in [a678604](https://github.com/NVIDIA-NeMo/Automodel/pull/548/commits/a67860406c2d165d12828f8ab918e8c50f0e7755)

PR #591, opened 2025-10-07
fix: cp fix

adil-a:
- SDPA CP doesn't support packed sequences currently. Remove attention mask while performing CP + SDPA.
- Do reductions over the correct meshes for reporting + training loss calculations


akoumpa (approved):
LGTM, thank you

PR #592, opened 2025-10-07
ci: Update dockerfile

thomasdhc:
- Update Automodel Dockerfile to be installed from a CUDA or PyTorch base container.
- Add build arguments to allow custom automodal dependency group to be installed
- Add build arguments to allow deepep to be installed or not
- Deprecate old install scripts
- Update docker build instructions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,9 +6,15 @@
 
 ```bash
 # Set build arguments
-export BASE_IMAGE=cuda #options [cuda, pytorch]
-export INSTALL_DEEPEP=False
-docker build -f docker/Dockerfile --build-arg AUTOMODEL_BASE_IMAGE=$BASE_IMAGE --build-arg INSTALL_DEEPEP=$INSTALL_DEEPEP -t automodel --target=automodel_base .
+export AUTOMODEL_INSTALL=dev #[all, dev, deepep, fa, moe, vlm]
+export BASE_IMAGE=cuda #[cuda, pytorch]
+export INSTALL_DEEPEP=False #[True, False]
+
+docker build -f docker/Dockerfile \
+--build-arg AUTOMODEL_INSTALL=$AUTOMODEL_INSTALL \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+    git pull && \
+    git fetch origin $DEEPEP_COMMIT && \
+    git checkout FETCH_HEAD && \
+    patch -p1 < /opt/deepep.patch && \
+    rm -rf /opt/deepep.patch; fi
+
+# Install Automodel
+ARG BASE_IMAGE=cuda
+ARG AUTOMODEL_INSTALL=dev
+COPY . /opt/Automodel
```
ko3n1g, line 64: Do we want to use layer caching again? COPY only the most necessary files here (`pyproject.toml`, `uv.lock` file, `automodel/__init__.py`) and after the install the rest of the project?

docker/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+RUN cd /opt/Automodel && \
+    if [ "$BASE_IMAGE" = "pytorch" ]; then \
+        sed -i '/\[tool\.uv\]/r /opt/Automodel/docker/common/uv-pytorch.toml' pyproject.toml && \
+        mv /opt/Automodel/docker/common/uv-pytorch.lock /opt/Automodel/uv.lock; \
+    fi && \
+    uv sync --locked --extra $AUTOMODEL_INSTALL --all-groups && \
+    if [ "$INSTALL_DEEPEP" = "True" ]; then \
+        TORCH_CUDA_ARCH_LIST="9.0 10.0 12.0" \
+        uv add --no-build-isolation -v /opt/DeepEP; \
```
ko3n1g, line 73: That's cool to know! I thought it adds it only to the pyproject.toml but doesn't install it. Nice to see it does both

docker/Dockerfile
thomasdhc: Thanks for the suggestion. Let me do a follow up PR to this one!

PR #599, opened 2025-10-07
feat: add lepton launcher

HuiyingLi:


nemo_automodel/components/launcher/lepton/utils.py
```diff
@@ -0,0 +1,68 @@
+import os
+from typing import Any, Dict
+
+from nemo_run.core.execution.lepton import LeptonExecutor
```
akoumpa, line 4: let's move this inside the `create_lepton_executor`

PR #650, opened 2025-10-15
fix: set max_steps inside constructor

akoumpa:


nemo_automodel/components/training/step_scheduler.py
```diff
@@ -74,6 +74,9 @@ def __init__(
             self.epoch_len = None
         self.val_every_steps = val_every_steps
         assert val_every_steps is None or val_every_steps > 0, "val_every_steps must be greater than 0 if not None"
+        if max_steps is None:
+            assert self.epoch_len is not None, "epoch_len must be provided if max_steps is not provided"
+            max_steps = (self.num_epochs - self.epoch) * self.epoch_len
```
adil-a: do we need to subtract self.epoch? 

I think you put it because of checkpointing, but at init time `self.epoch` will be 0. Please correct me if I am misunderstanding as I don't see other cases where a user would pass in a different value of start_epoch

PR #657, opened 2025-10-16
feat: GPT-OSS for DGX Spark

adil-a:
This PR enables single GPU dequantization. PR is essential to merge for SFT + PEFT on GPT-OSS-20B on DGX Spark

### DIFF
--- a/examples/llm_finetune/phi/phi_4_squad.yaml
+++ b/examples/llm_finetune/phi/phi_4_squad.yaml
@@ -44,7 +44,7 @@ checkpoint:
 distributed:
   _target_: nemo_automodel.components.distributed.fsdp2.FSDP2Manager
   dp_size: none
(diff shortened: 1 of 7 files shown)


examples/llm_finetune/gpt_oss/gpt_oss_20b_single_gpu_peft.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+step_scheduler:
+  global_batch_size: 128
+  local_batch_size: 32
+  ckpt_every_steps: 50
+  val_every_steps: 10  # will run every x number of gradient steps
+  num_epochs: 2
+
+dist_env:
+  backend: nccl
```
akoumpa, line 23: why do we need this in this case?

PR #661, opened 2025-10-17
fix: enable checkpoint saving by default

akoumpa:
Changes:
- enable checkpoint saving by default
- enable consolidated saving by default
- make `ckpt_every_steps` optional -> infer based on `epoch_len` value.


nemo_automodel/components/training/step_scheduler.py
```diff
@@ -88,9 +89,18 @@ def __init__(
...
             max_steps = _calculate_max_steps(self.num_epochs, self.epoch_len)
+            logger.info("max_steps not provided; will run for up to {} steps".format(max_steps))
         self.max_steps = max_steps
         assert max_steps > 0, "max_steps must be greater than 0"
 
+        if ckpt_every_steps is None:
+            if self.epoch_len is None:
+                ckpt_every_steps = self.max_steps // 2
+            else:
+                ckpt_every_steps = self.epoch_len
```
adil-a, line 100: not sure if this is needed. `is_ckpt_step` will return True both at the end of epoch and end of training as defined by max step. Imo that's sufficient

PR #719, opened 2025-10-24
feat: surface truncating & padding options

akoumpa:
Changes:
- Surfaces the `padding` and `truncation` options in the formatting utils.
- For the ColumnMapped dataset, includes above options and also adds a fallback if all labels are empty in the labels.

### DIFF
--- a/nemo_automodel/components/datasets/llm/chat_dataset.py
+++ b/nemo_automodel/components/datasets/llm/chat_dataset.py
@@ -133,6 +133,8 @@ class ChatDataset(Dataset):
         split: Optional[str] = None,
         name: Optional[str] = None,
         seq_length: Optional[int] = None,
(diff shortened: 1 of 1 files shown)


nemo_automodel/components/datasets/llm/formatting_utils.py
```diff
@@ -199,6 +207,9 @@ def format_chat_template(
         tokenize=True,
         return_dict=True,
         return_assistant_tokens_mask=template_has_generation_kwd,
+        padding=padding,
+        truncation=truncation,
```
HuiyingLi, line 211: If I understand correctly we truncate regardless of the template structure here.

akoumpa: yes

PR #729, opened 2025-10-27
ci: Pip install deepep

thomasdhc:
- Install DeepEP in container as apart of system dependency
- Resolves errors resulting within container running uv commands due to DeepEP

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -60,7 +60,8 @@ RUN if [ "$INSTALL_TE" = "True" ]; then \
     git checkout FETCH_HEAD && \
     git submodule init && git submodule update && \
     pip install nvidia-mathdx==25.1.1 && \
-    env NVTE_CUDA_ARCHS="80;90;100;120" NVTE_BUILD_THREADS_PER_JOB=8 pip install --no-cache-dir --no-build-isolation -v .; \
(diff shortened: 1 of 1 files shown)


pablo-garay (approved):

docker/Dockerfile
```diff
@@ -76,24 +77,24 @@ RUN if [ "$INSTALL_DEEPEP" = "True" ]; then \
...
+    rm -rf /opt/deepep.patch && \
+    rm -rf DeepEP; \
+    fi
 
 FROM automodel_dep as automodel_final
 
 # Install Automodel
 ARG BASE_IMAGE=cuda
-ARG AUTOMODEL_INSTALL=dev
+ARG AUTOMODEL_INSTALL=vlm
```
ko3n1g, line 90: Why are we changing this?

thomasdhc: dev is deprecated as we removed that optional dependency group

thomasdhc:
@ko3n1g The behavior I saw was that during uv sync deepep fails with torch missing. To get around that issue I've added the build isolation for it. But the problem I saw after was on sync it attempted to rebuild the whl from source.  So I wanted to skip the rebuild behavior and build fails anyways because correct env vars are not supplied during a uv sync.

Issue #509, opened 2025-09-24
Annotated config for LLM finetune


akoumpa:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #732, opened 2025-10-28
feat: Support TE attention for gptoss

hemildesai:


hemildesai:
/ok to test [8aa323c](https://github.com/NVIDIA-NeMo/Automodel/pull/732/commits/8aa323c940922b74d212d729a10c4e9cac167c86)

hemildesai:
/ok to test [5941dd7](https://github.com/NVIDIA-NeMo/Automodel/pull/732/commits/5941dd73f78015c630c8706e84ee2d10cd3b38f8)

examples/llm_finetune/gpt_oss/gpt_oss_20b.yaml
adil-a, line 1: can we keep the API simplified? You can take a look at https://github.com/NVIDIA-NeMo/Automodel/blob/main/examples/llm_pretrain/deepseekv3_pretrain.yaml#L55-L73 to show how we can pass extra args, but I'd like these to be auto detected to give the best perf and keep the config API simple

nemo_automodel/shared/import_utils.py
```diff
@@ -396,3 +396,26 @@ def is_torch_min_version(version, check_equality=True):
...
+
+        if hasattr(te, "__version__"):
+            _version = str(te.__version__)
+        else:
+            from importlib.metadata import version
+
+            _version = version("transformer-engine")
+    except ImportError:
+        _version = "0.0.0"
+    return PkgVersion(_version)
```
adil-a, line 414: Is it simpler to do it this way: https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/components/checkpoint/__init__.py#L23?

adil-a (commented):
Small comments/changes

nemo_automodel/shared/import_utils.py
hemildesai: I based this on the `get_torch_version` in the same file above to keep it consistent.

examples/llm_finetune/gpt_oss/gpt_oss_20b.yaml
hemildesai: Isn't it better to expose the defaults in the config itself so its easier to change individual options. Without the backend in the config, someone will have to look at the source code for the available options then add the entire backend config along with its target to change the attention backend, whereas in the current scenario it will suffice to just do `--model.backend.attn te`

adil-a: I think the primary use case for custom models is to give the best perf as possible. We should be able to automatically pick these out for the user without them having to specify it in the config.

We are planning to provide a "master" config that has all the customizations possible, so they'd be able to view that and learn about the different changes they can make via the config. See https://github.com/NVIDIA-NeMo/Automodel/issues/509

One of our goals was to streamline the API between HF models and custom models, so this would break that design IMO. Here's what I did in the past to auto pick the right implementation https://github.com/NVIDIA-NeMo/Automodel/blob/34f73e92ed921d0802475a83385da943f474db3b/nemo_automodel/components/moe/utils.py#L28

Lmk your thoughts

hemildesai: IMO its better to lay out all the options since the user can then have control over the settings. Too many hidden settings causes confusions to the user.

Also auto picking is not always possible, for instance te and flex gptoss attn have similar perf but te is only supported if cudnn version is above 9.13.1.

hemildesai:
/ok to test [83ca413](https://github.com/NVIDIA-NeMo/Automodel/pull/732/commits/83ca413c8ae20bd8597ef12a15231a5334177173)

examples/llm_finetune/gpt_oss/gpt_oss_20b.yaml
hemildesai: Updated in [83ca413](https://github.com/NVIDIA-NeMo/Automodel/pull/732/commits/83ca413c8ae20bd8597ef12a15231a5334177173)

adil-a (approved):
LGTM, thank you!

hemildesai:
/ok to test [4cc790a](https://github.com/NVIDIA-NeMo/Automodel/pull/732/commits/4cc790a5a64f55e186feb19c15ff630e6d83bd1a)

PR #733, opened 2025-10-28
feat: Sharding Optimization for Sequence Parallelism/LoRA

ZhiyuLi-Nvidia:
## Improvement 
1.4x MFU using the off-the-shelf Automodel and 2.1x MFU with the customized model with combined projection.
* baseline MFU: [7.13% (10.69% *2 / 3)](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/gmtu4xyz/logs?nw=nwuserzhiyul)
* with the change
   - custom model with combined projection:
      - MFU: [14.97% (22.45% * 2 / 3)](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/eczz85kj/logs)
      - recipe: examples/llm_finetune/llama3_3/custom_llama3_3_70b_instruct_peft_benchmark.yaml
   - off-the-shelf Automodel 
      - MFU: [9.78% (14.67% * 2 / 3)](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/gnenmr2p/logs)
      - recipe: examples/llm_finetune/llama3_3/llama3_3_70b_instruct_peft_benchmark.yaml

### Breakdown
* baseline MFU: [7.13% (10.69% *2 / 3)](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/gmtu4xyz/logs?nw=nwuserzhiyul)
* 2b70b: MFU [8.65% (12.98% * 2 / 3)](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/bil9xvon/logs)
   - +21% rel. from Sharding Optimization 
* e69e3 MFU: [12.24% (18.36% * 2 / 3)](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/h9soracf/logs?nw=nwuserzhiyul)
   - +42% rel. from combined qkv/gate_up implemented in custom llama 
 * a1650 MFU: [14.97% (22.45% * 2 / 3)](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/eczz85kj/logs)
   - +22% rel. use sequence_parallel as False: without norm fusion, setting sequence_parallel=False is more efficient. This configuration only requires a single reduce-scatter operation between the LlamaAttention and LlamaMLP modules. In contrast, enabling sequence parallelism introduces extra overhead: it requires a reduce-scatter to redistribute data from tensor parallelism (TP) to sequence parallelism (SP), and an all-gather for the activations before the MLP.
```
uv run --extra fa torchrun --nproc-per-node=8 \
    nemo_automodel/recipes/llm/benchmark.py \
    --config examples/llm_finetune/llama3_3/llama3_3_70b_instruct_peft_benchmark.yaml

# smaller scale
uv run --extra fa torchrun --nproc-per-node=8 \
    nemo_automodel/recipes/llm/benchmark.py \
    --config examples/llm_finetune/llama3_3/llama3_3_70b_instruct_peft_benchmark.yaml --model.num_hidden_layers=4
```

Source of improvements:
* `SequenceParallelAllGatherActivation`: Perform an all_gather on the activation (from the previous norm layer) avoid redundant and reduce  projection communication
  - Original: Performing a communication operation (like all_to_all) after the QKV/MLP matrix multiplication. This is expensive when there is up scale projection like MLP
  - move the communication step to happen before the projection. And the new flow performs one single all_gather since qkv and MLP shared the same input.
  - Benefit: Since the Q, K, and V projections all share the same input tensor, this single all_gather is sufficient for all three operations. This reduces the total communication to 1/3 or 1/2 compared to communicating the result of each projection separately. For MLP, it is usually cheaper to communicate before projection. This optimization is beneficial for both pre-training and SFT workloads.

* lora allgather intermediate: reduces communication overhead LoRA by communicating the low rank intermediate activation
  - Original: compute full activation after LoRA operation, `reduce(lora_B(lora_A(x)))`. This requires communicating the final, high-dimensional activation ([B, S, E]), which is computationally expensive.
  - Optimized Approach: We perform the communication between the two LoRA projections: lora_B(all_gather(lora_A(x))).
  - Benefit: This method is significantly cheaper because it only communicates the low-rank intermediate activation ([B, S, R]). 

*  move scale to the compiling friendly postion: Apply scale before lora_B to keep lora_res as a Partial tensor. This allows both res and lora_res to remain Partial, so only one reduce-scatter is needed after addition.Multiplying after lora_B would convert Partial to Replicate, causing an extra reduce-scatter operation.
```
lora_res = self.lora_B(self.lora_A(x) * self.scale)
```
* qkv/gate up combined projection: Using LoRA with high tensor parallelism (TP=8) results in lightweight matrix multiplications or low computational intensity, decreasing hardware utilization. Combined projection greatly improve utilization with a much bigger matmul and less overhead.

### DIFF
--- a/tests/functional_tests/hf_transformer_finetune/test_hf_transformer_sft.py
+++ b/tests/functional_tests/hf_transformer_finetune/test_hf_transformer_sft.py
@@ -23,6 +23,7 @@ HF_TRANSFORMER_PEFT_MegatronFSDP_FILENAME = "L2_HF_Transformer_PEFT_MegatronFSDP
 HF_TRANSFORMER_PEFT_NO_TOKENIZER_FILENAME = "L2_HF_Transformer_PEFT_no_tokenizer.sh"
 HF_TRANSFORMER_SFT_NO_LOGITS_FILENAME = "L2_HF_Transformer_SFT_no_logits.sh"
 HF_TRANSFORMER_QWEN3_MOE_CUSTOM_FILENAME = "L2_HF_Transformer_Qwen3_MoE_custom.sh"
+HF_TRANSFORMER_LLAMA3_CUSTOM_FILENAME = "L2_HF_Transformer_PEFT_Benchmark_Llama_custom.sh"
 
 class TestHFTransformerFinetune:
     def test_hf_transformer_sft(self):
@@ -70,6 +71,13 @@ class TestHFTransformerFinetune:
     def test_hf_transformer_qwen3_moe_sdpa(self):
         try:
             run_test_script(TEST_FOLDER, HF_TRANSFORMER_QWEN3_MOE_CUSTOM_FILENAME)
+        finally:
+            # remove the checkpoint directory
+            shutil.rmtree("checkpoints/", ignore_errors=True)
+
+    def test_hf_transformer_llama3_custom(self):
+        try:
+            run_test_script(TEST_FOLDER, HF_TRANSFORMER_LLAMA3_CUSTOM_FILENAME)
         finally:
             # remove the checkpoint directory
             shutil.rmtree("checkpoints/", ignore_errors=True)


ZhiyuLi-Nvidia:
/ok to test [96d9a54](https://github.com/NVIDIA-NeMo/Automodel/pull/733/commits/96d9a549de48ccaa44d9a339815e414eef45d477)

nemo_automodel/components/models/llama.py
```diff
@@ -0,0 +1,586 @@
...
+        position_embeddings: tuple[torch.Tensor, torch.Tensor],
+        attention_mask: Optional[torch.Tensor],
+        past_key_values: Optional[Cache] = None,
+        cache_position: Optional[torch.LongTensor] = None,
+        **kwargs: Unpack[TransformersKwargs],
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        input_shape = hidden_states.shape[:-1]
+        hidden_shape = (*input_shape, -1, self.head_dim)
+
+        if self.use_fused_qkv:
```
akoumpa: I would make this class na example of using fused QKV directly (therefore remove this flag and always use fused QKV). I understand it's useful for dev purposes.

Another goal is to maintain compatibility with the Hugging Face ecosystem, therefore any checkpoint that's saved with NeMo AutoModel should be loadable with hugging face transformers and other HF-compatible tools (e.g., vLLM, llm-eval-harnness, etc). For this I would recommend to introduce a state_dict adapter similar to the one we have for deepseek etc.

ZhiyuLi-Nvidia: Thank you for the comments.
> using fused QKV directly.  

Updated and the model is pure combined qkv implementation without an option.

> For this I would recommend to introduce a state_dict adapter 

I have added implementation along with the unit tests.

akoumpa (approved):

PR #738, opened 2025-10-29
ci: Update DeepEP commit

thomasdhc:


docker/Dockerfile
```diff
@@ -66,10 +66,13 @@ RUN if [ "$INSTALL_TE" = "True" ]; then \
 
 FROM update_base_container AS automodel_dep
 
+# Install Grouped Gemm
+RUN pip install git+https://github.com/fanshiqing/grouped_gemm@v1.1.4
```
ko3n1g, line 70: Don't we need `--no-build-isolation`?

PR #739, opened 2025-10-30
feat: add qwen3 omni ootb recipe

HuiyingLi:


pyproject.toml
```diff
@@ -90,14 +90,14 @@ dependencies = [
 vlm = [
     "pillow",
     "qwen-vl-utils[decord]; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
-    "transformers<=4.57.1",
+    "qwen-omni-utils",
+    "transformers==4.57.1",
```
thomasdhc: Let's keep this as an upper-bound

pyproject.toml
```diff
@@ -90,14 +90,14 @@ dependencies = [
...
     "timm==1.0.16",
     "backoff",
-    "numpy",
+    "numpy<2",
```
thomasdhc: Can you add a comment next to this restriction to indicate which model limits this value

PR #743, opened 2025-10-30
docs: Update CONTRIB

thomasdhc:


thomasdhc:
close in favor of #849

PR #769, opened 2025-11-05
feat: fp32 lm_head and fp32 apply_rope options for MoE

hemildesai:


nemo_automodel/components/moe/parallelizer.py
```diff
@@ -175,7 +180,22 @@ def apply_fsdp(
 
     lm_head = getattr(_model, "lm_head", None) or getattr(model, "lm_head", None)
     if lm_head is not None:
-        fully_shard_default(lm_head)
+        # Use custom mixed precision policy for lm_head if lm_head_precision is specified
+        if lm_head_precision == torch.float32:
```
akoumpa, line 184: is it possible to inspect the lm_head to figure out the precision?

hemildesai: This option is to force lm head in fp32 regardless of checkpoint dtype. fp32 lm_head helps with RL stability.

adil-a (approved):
LGTM

PR #773, opened 2025-11-06
fix: custom model fixes (NeMo RL)

adil-a:
- Letting config be passed through kwargs for custom model
- NemotronH .initialize_weights() when sharded has an issue with a DTensor op being missing (likely due to Torch version). Skipping init for now. This doesn't affect finetuning but does affect pretraining.
- Flexibility to pass a custom TP plan in FSDP2Manager

### DIFF
--- a/nemo_automodel/components/checkpoint/checkpointing.py
+++ b/nemo_automodel/components/checkpoint/checkpointing.py
@@ -332,10 +332,9 @@ class Checkpointer:
         # doesn't support initialize_weights when the model is sharded. This is because Gemma's
         # initialize_weights method requires setting a row to zeros in the embedding matrix.
         # This index selection op is not supported for DTensors in the pinned torch version.
-        config = getattr(model, "config", None)
-        if config is not None:
-            model_class = getattr(config, "architectures", [""])[0]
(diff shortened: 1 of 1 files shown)


nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -334,7 +332,11 @@ def load_base_model(
         # doesn't support initialize_weights when the model is sharded. This is because Gemma's
         # initialize_weights method requires setting a row to zeros in the embedding matrix.
         # This index selection op is not supported for DTensors in the pinned torch version.
-        if not isinstance(model, Gemma3ForConditionalGeneration):
+        try:
```
akoumpa, line 335: fix for https://github.com/NVIDIA-NeMo/Automodel/actions/runs/19127576314/job/54662020970?pr=773

akoumpa (approved):
LGTM, thanks.

nemo_automodel/components/checkpoint/checkpointing.py
adil-a: Thanks!

PR #789, opened 2025-11-10
fix: Update is_hf_model to exclude custom models 

hemildesai:


nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -181,7 +182,14 @@ def build_model_and_optimizer(
     Returns:
         The instantiated model on the specified device, the state dict keys before any parallelization, the optimizer, and the loss function.
     """
-    is_hf_model = cfg_model.get("pretrained_model_name_or_path", None) is not None
+
+    is_hf_model = (
```
akoumpa: @hemildesai let's make this a function living inside the recipe.

PR #803, opened 2025-11-14
fix: Raise informative error in ``MegatronPretraining`` if no glob paths found

jrbourbeau:
I accidentally had a bad glob path and got a confusing error here

https://github.com/NVIDIA-NeMo/Automodel/blob/de7dff940b2ff214352fe8249113e2408b330242/nemo_automodel/components/datasets/llm/megatron/megatron_utils.py#L41-L43

when attempting to `zip` an empty list (the list was only empty because of the glob path not being correct).

This PR raises an informative error early on for the case that no glob paths are found. 

EDIT: The particular case I was running was the pretraining example

```
uv run torchrun --nproc-per-node=2 examples/llm_pretrain/pretrain.py --config examples/llm_pretrain/megatron_pretrain_gpt2.yaml
```

where I didn't have the fineweb dataset downloaded locally

### DIFF
--- a/tests/unit_tests/datasets/llm/test_megatron_dataset.py
+++ b/tests/unit_tests/datasets/llm/test_megatron_dataset.py
@@ -16,7 +16,7 @@ import json
 
 import pytest
 
-from nemo_automodel.components.datasets.llm.megatron_dataset import try_load_blend_from_json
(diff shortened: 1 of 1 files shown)


adil-a:
Thanks a lot for the fix @jrbourbeau !

Issue #780, opened 2025-11-07
Sequence classification reports incorrect metrics


adil-a:
**Describe the bug**

```
2025-11-07 17:20:43 | INFO | root | )
2025-11-07 17:20:43 | INFO | root | LR scheduler: <unavailable>
2025-11-07 17:20:43 | INFO | root | Step scheduler:
2025-11-07 17:20:43 | INFO | root | - Gradient accumulation steps: 1
2025-11-07 17:20:43 | INFO | root | - Checkpoint every steps: 5
2025-11-07 17:20:43 | INFO | root | - Current Epoch: 0
2025-11-07 17:20:43 | INFO | root | - Number of epochs: 2
2025-11-07 17:20:43 | INFO | root | - Validation every steps: 100
2025-11-07 17:20:43 | INFO | root | - Max train steps: 10
2025-11-07 17:20:43 | INFO | root | step 0 | epoch 0 | loss 0.8371 | grad_norm 0.0000 | lr 3.00e-04 | mem 4.63 GiB | tps 0.00(0.00/gpu) | num_label_tokens 0
2025-11-07 17:20:43 | INFO | root | step 1 | epoch 0 | loss 0.6606 | grad_norm 0.0000 | lr 3.00e-04 | mem 4.53 GiB | tps 0.00(0.00/gpu) | num_label_tokens 0
2025-11-07 17:20:43 | INFO | root | step 2 | epoch 0 | loss 0.8939 | grad_norm 0.0000 | lr 3.00e-04 | mem 4.23 GiB | tps 0.00(0.00/gpu) | num_label_tokens 0
2025-11-07 17:20:43 | INFO | root | step 3 | epoch 0 | loss 0.6635 | grad_norm 0.0000 | lr 3.00e-04 | mem 4.11 GiB | tps 0.00(0.00/gpu) | num_label_tokens 0
2025-11-07 17:20:44 | INFO | root | step 4 | epoch 0 | loss 0.6588 | grad_norm 0.0000 | lr 3.00e-04 | mem 4.13 GiB | tps 0.00(0.00/gpu) | num_label_tokens 0
```

```
python examples/llm_seq_cls/seq_cls.py -c examples/llm_seq_cls/glue/mrpc_roberta_lora.yaml
```

qiaochuz-nv:
## QA RCCA Analysis
**Fix PR:** #841 `fix: sequence classification metric and training bugs`
[...]

PR #807, opened 2025-11-16
fix: Use uv Python for MCore dataset compilation (#438)

### LINKED ISSUE #438: MCore Dataset needs to launch with uv instead of system python
The .so file that gets created contains the python version in the filename. This all works fine if your system python version matches the python version specified in the project for uv (3.12), but if for any reason it deviates, there will be a mistmatch because the subprocess doesn't use uv and ends up using the system python version. Function found [here](https://github.com/NVIDIA-NeMo/Automodel/blob/95e7de0f46ad91731310d7f4a68a3d6c89c691a5/nemo_automodel/components/datasets/llm/megatron/megatron_utils.py#L69)

yuhezhang-ai:
## Description

Changed Makefile to use `uv run python` instead of system `python3`, ensuring the compiled extension matches the uv Python environment.

Also added `-undefined dynamic_lookup` linking flag for macOS to fix 'Undefined symbols' errors during compilation.

### Testing
Verified with system Python 3.11 and uv Python 3.12 - the compiled `.so` file now correctly uses the uv Python version (3.12).

Fixes #438

---
First-time contributor here.  I'm a research engineer transitioning from edge models to LLM infrastructure and algorithms. Happy to help with more tasks in the future. Thanks for reviewing!


adil-a:
Thank you so much @yuhezhang-ai ! We really appreciate community contributions :)

@nvidia-nemo/automation @thomasdhc can you help verify the changes?

thomasdhc:
Hey @yuhezhang-ai Thanks so this update. The uv run python is actually re-installing torch when it should not be and is causing this error. This is caused by some of our testing setup incorrectly mounting another copy of Automodel. I'll need to make changes to the overall test workflow. When that PR is done I will apply those changes to this PR.

No further action needs to be taken from your side.

Thanks!

adil-a:
Hey @yuhezhang-ai thank you so much for your enthusiasm! It'd be great to have more hands on-board :) We usually file any open issues on the GitHub Issues tab so feel free to pick up anything interesting to you. https://github.com/NVIDIA-NeMo/Automodel/issues/780 might be a good and easy one to pick up

akoumpa (approved):

PR #825, opened 2025-11-18
fix: Cast norm to fp32 in clip_grad_norm

hemildesai:
all reducing grad norms in bf16 across the pp mesh causes `nan` values due to underflow/overflow so need to cast local norms to fp32 before all reduce operations.


hemildesai:
/ok to test [2d2feb9](https://github.com/NVIDIA-NeMo/Automodel/pull/825/commits/2d2feb9c24401e95373a4749614073afec028725)

PR #843, opened 2025-11-20
feat: add custom implementation for qwen3vlmoe 

HuiyingLi:
image

https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/krjmje1a

### DIFF
--- a/tests/unit_tests/models/qwen3_vl_moe/test_qwen3_vl_moe_model.py
+++ b/tests/unit_tests/models/qwen3_vl_moe/test_qwen3_vl_moe_model.py
@@ -169,9 +169,7 @@ class TestQwen3VLMoeTextModelBackend:
         for layer in model.layers:
             layer.forward = MagicMock(side_effect=lambda x, **_: x + 1)
 
-        with patch(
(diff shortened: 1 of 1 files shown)


nemo_automodel/components/attention/utils.py
```diff
@@ -142,8 +142,6 @@ def preprocess_args_and_kwargs_for_attn(
         k = k.transpose(1, 2).contiguous()
         v = v.transpose(1, 2).contiguous()
         attn_kwargs["is_causal"] = True
-        if attention_mask is not None and (attention_mask.dtype == torch.bool or attention_mask.is_floating_point()):
-            attn_kwargs["attn_mask"] = attention_mask
```
hemildesai, line 146: Why is this removed?

HuiyingLi: @hemildesai  Thanks for reviewing! They were added when I wanted to add a manually created causal mask for qwen3omni. Now I changed qwen3omni to right padding too, these are no longer needed. Restore back to how they were initially.

PR #877, opened 2025-11-24
fix: fsdp mixin and parallelizer patches for custom MoEs

hemildesai:


akoumpa:
moved to https://github.com/NVIDIA-NeMo/Automodel/pull/881

PR #878, opened 2025-11-24
fix: NeMoAutoTokenizer

akoumpa:
Previously, the code could prepend/append tokens inconsistently across items in the tokenization results (e.g., across input_ids, attention_mask, etc), causing issues in subsequence processing. 

This fix ensures any token modification is consistent across items.


adil-a (approved):
LGTM thanks!

PR #915, opened 2025-12-03
feat: add ministral3 configs and improve tie_emb detection

HuiyingLi:


nemo_automodel/components/checkpoint/stateful_wrappers.py
```diff
@@ -92,16 +94,7 @@ def __init__(
...
-
-        non_tied_lm_head_models = {
-            "Qwen3OmniMoeThinkerForConditionalGeneration",  # complicated config structure
-            "InternVLForConditionalGeneration",  # even tho config says tie_word_embeddings=True, it's not
-        }
-        for m in non_tied_lm_head_models:
-            if m in type(self.model[0]).__name__:
-                self.is_tied_lm_head = False
-                break
+        self.is_tied_lm_head = is_tied_word_embeddings(model)
```
adil-a:
```suggestion
        self.is_tied_lm_head = is_tied_word_embeddings(self.model[0])
```

nemo_automodel/components/checkpoint/utils.py
```diff
@@ -0,0 +1,36 @@
...
+        bool: True if the model's word embeddings are tied, False otherwise.
+    """
+    non_tied_lm_head_models = {
+        "Qwen3OmniMoeThinkerForConditionalGeneration",  # complicated config structure
+    }
+    for m in non_tied_lm_head_models:
+        if m in type(model[0]).__name__:
+            return False
+    config = getattr(model, "config", None)
+    text_config = getattr(config, "get_text_config", lambda: None)()
```
adil-a: If `get_text_config` is returned then wouldn't it be a type dict in which case `()` would fail?

PR #954, opened 2025-12-10
docs: Fix images not rendering in docs

jrbourbeau:
This PR fixes a few images I noticed we're rendering in the docs. 

Before this PR:

Screenshot 2025-12-10 at 1 33 03 PM

After this PR:

Screenshot 2025-12-10 at 1 33 21 PM

### DIFF
--- a/docs/guides/fp8-training.md
+++ b/docs/guides/fp8-training.md
@@ -93,7 +93,12 @@ FP8 quantization provides measurable performance improvements while maintaining 
 - **Convergence**: FP8 training achieves loss parity with BF16 training.
 - **Memory**: FP8 training achieves on par memory usage with BF16 baseline.
 
-<img src="fp8_convergence.jpg" alt="FP8 Convergence Comparison" width="600px" />
+```{image} fp8_convergence.jpg
+:alt: FP8 Convergence Comparison
+:class: bg-primary
+:width: 600px
+:align: center
+```
 
 *Figure: Loss curves comparing FP8 tensorwise scaling + torch.compile vs. BF16 + torch.compile training on 8xH100 with 8k sequence length, demonstrating virtually identical convergence behavior with 1.24x speedup*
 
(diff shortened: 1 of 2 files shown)


thomasdhc (approved):

docs/guides/fp8-training.md
```diff
@@ -93,7 +93,12 @@ FP8 quantization provides measurable performance improvements while maintaining
 - **Convergence**: FP8 training achieves loss parity with BF16 training.
 - **Memory**: FP8 training achieves on par memory usage with BF16 baseline.
 
-<img src="fp8_convergence.jpg" alt="FP8 Convergence Comparison" width="600px" />
+```{image} fp8_convergence.jpg
+:alt: FP8 Convergence Comparison
+:class: bg-primary
+:width: 600px
+:align: center
+```
```
HuiyingLi: Would something the following work?

`FP8 Convergence Comparison`

HuiyingLi, line 101: Would something the following work?

`FP8 Convergence Comparison`

jrbourbeau: I think using local files is probably better in this case (for example, there can be mismatches when doing development, versioning of docs, etc).

If you're looking for something that will render both in the docs and when viewed on GitHub, we would need to move away from myst markdown parsing. That's not always possible, but is in this case with something like this

```diff
diff --git a/docs/guides/omni/gemma3-3n.md b/docs/guides/omni/gemma3-3n.md
index ac11367..5443c93 100644
--- a/docs/guides/omni/gemma3-3n.md
+++ b/docs/guides/omni/gemma3-3n.md
@@ -218,12 +218,7 @@ peft:

 The training loss should look similar to the example below:

-```{image} medpix_peft.jpg
-:alt: Training Loss Curve
-:class: bg-primary
-:width: 400px
-:align: center
-```
+![Training Loss Curve](medpix_peft.jpg){.bg-primary w=400px align=center}

 ### Checkpointing
 ```

PR #991, opened 2025-12-18
fix: prevent hang in in download_model_weights

hemildesai:


hemildesai:
/ok to test [21cea07](https://github.com/NVIDIA-NeMo/Automodel/pull/991/commits/21cea07fc109439cba0aac6198c8d01695299e0c)

Issue #691, opened 2025-10-20
CI: ver+commit


akoumpa:
**Is your feature request related to a problem? Please describe.**
Versions should include the commit id, for example: `2.8.0a0+34c6371d24`

**Describe the solution you'd like**
- pip3 install git+https://... -> pip3 show nemo_automodel -> <ver> + <commit>
- pip3 install nemo_automodel -> pip3 show nemo_automodel -> <ver> + <commit>

**Describe alternatives you've considered**
N/A

**Additional context**
It's useful when debugging things, to be able to trace the source code that was used.

PR #1005, opened 2026-01-02
feat: support commit hash in package version

ooooo-create:
# What does this PR do ?

Adds commit hash into the package version metadata so built wheels/tarballs encode the exact git revision.

1. `uv sync`
image

2. `uv build`
image

3. `uv run python -c "import nemo_automodel; print(nemo_automodel.__version__); print(nemo_automodel.__git_version__)"`
image

# Changelog

- `build_backend.py:9-107`: new custom build backend that gathers git branch/tag/hash, writes - `nemo_automodel/_version.py`, and appends the hash to __version__ unless on a release branch or tag; exports __git_version__.
- `__init__.py:16-25`: re-exports __git_version__ alongside __version__ and __package_name__.
- `package_info.py:15-16`: switches to importing __version__ and __git_version__ from the generated _version.py.
- `pyproject.toml:15-33`: configures the custom build backend and points the dynamic version source to nemo_automodel._version.__version__.
- `.gitignore:162-164`: ignores the generated nemo_automodel/_version.py.
- `version.txt:1`: sets the base package version to 0.2.0.

**Pre checks**:

# Additional Information

- Related to #691


build_utils/build_backend.py
```diff
@@ -0,0 +1,108 @@
+import logging
```
akoumpa, line 16: Thanks a lot for the contribution @ooooo-create ,

I want to move this file out of the top-of-tree. Right now I don't see a directory that's 100% suitable, but perhaps the closest would be `test/`? @thomasdhc please opine on the matter.

ooooo-create: Apologies for the delay.
How about we create a `_custom_build` directory at the project root and move `build_backend.py` into it? I noticed this pattern in the [setuptools documentation](https://setuptools.pypa.io/en/latest/build_meta.html).
 Additionally, regarding the versioning logic: currently, we rely on branch names (starting with 'release') and git tags to determine whether to exclude the commit hash from the package name. I feel this approach might be a bit **brittle**.
What do you think about introducing an environment variable, such as `AUTOMODEL_VERSION`? If this variable is set, the build system should **respect it** and use it as the definitive version.

akoumpa: Hi @ooooo-create , thanks a lot for this.

I'm ok with having a `build` or `build_utils` directory and inside contain any util python script that are necessary for building the package.

I'm not 100% sure about introducing an environment variable for the version. I feel that we want to track the version as part of the git repo. The main outcome i want to ensure is achieved is to have the git commit on (a) the package (so that any logs of recipes includes it) and (b) (optionally) to have it on code that was git checkoutd/uv run so that we can have it in the logs there as well (i marked this as optional, since git main refuse to run on shared storage, and in this case it should say n/a instead of failing to run the program completely -- best effort).

@thomasdhc please feel free to take the lead as I feel this falls more on the automation side. 

Thanks again @ooooo-create for the contribution, hope you sending more! cheers

ooooo-create: Thank you for the clear explanation; I understand the process much better now.
I have implemented support for both scenarios:
1. Regarding wheel packages containing commit info: During the build process, the `dynamic_version_info` function in build_backend is now called to generate the `nemo_automodel/_version.py` file. This is used to determine the package name and version.
2. For git checkout and uv run scenarios: I chose to call `dynamic_version_info` within `nemo_automodel/package_info.py` to dynamically retrieve and update `_version.py`. This ensures the version is correct when the package is imported. If git operations fail, it defaults to using gitunknown.
3. Clean versions for PyPI: I added an environment variable `NO_GIT_VERSION` and a check for git tags. This allows generating package names without the git commit hash, which is more appropriate when publishing to PyPI.

PR #1008, opened 2026-01-02
feat: add configurable remote logging frequency via `step_scheduler`

oliverholworthy:
Add configurable remote logging frequency via `step_scheduler`

## Motivation

When training models with fast iteration times or long training runs, logging to remote services (WandB, MLflow) on every single step can introduce unnecessary overhead

### New Configuration Option

Added `log_remote_every_steps` to the `step_scheduler` config section with corresponding `is_remote_logging_step` property:

```yaml
step_scheduler:
  global_batch_size: 128
  local_batch_size: 4
  ckpt_every_steps: 500
  val_every_steps: 500
  log_remote_every_steps: 10  # NEW: Control remote logging frequency
  num_epochs: 1
```

**Default**: `1` (every step) - fully backward compatible.

### Design Decisions

1. **Consistent with Existing Pattern**: Follows the same pattern as `val_every_steps` and `ckpt_every_steps` with a corresponding `is_remote_logging_step` property on `StepScheduler`.

2. **Clear Naming**: `log_remote_every_steps` explicitly indicates it affects remote logging services (WandB, MLflow), not local logs.

3. **Remote vs Local Separation**: Only throttles remote service uploads (WandB, MLflow). Local JSONL logs and console output still happen every step for detailed debugging.

4. **Validation Always Logged**: Validation metrics always log to remote services since validation is already infrequent (controlled by `val_every_steps`).

## Future Work

The implementation revealed code duplication across recipes (`build_wandb()`, `log_train_metrics()`, etc.). A future refactoring could:
- Move common WandB utilities to a shared module
- Add logging methods to `BaseRecipe` with overridable behavior
- Consolidate similar patterns across recipes

**Pre checks**:


nemo_automodel/recipes/biencoder/train_biencoder.py
```diff
@@ -728,8 +735,11 @@ def log_train_metrics(self, log_data: MetricsSample):
         if not self.dist_env.is_main:
             return
 
+        # Log training metrics to wandb every N steps (configurable via wandb.log_train_every_steps)
         if wandb.run is not None:
-            wandb.log(log_data.to_dict(), step=self.step_scheduler.step)
+            log_train_every_steps = self.cfg.get("wandb", {}).get("log_train_every_steps", 1)
```
akoumpa:
```suggestion
            log_train_every_steps = self.cfg.get("wandb.log_train_every_steps", 1)
```

nemo_automodel/recipes/biencoder/train_biencoder.py
```diff
@@ -223,6 +223,13 @@ def build_wandb(cfg) -> wandb.Run:
     """
     assert cfg.get("wandb", None) is not None
     kwargs = cfg.wandb.to_dict()
+
+    # Remove custom config keys that are not wandb.init parameters
+    # These are used by our training logic but not passed to wandb
+    custom_keys = {"log_train_every_steps"}
```
akoumpa: I wonder if we want to move this to the `step_scheduler` section instead?

akoumpa: Thanks a lot for making this @oliverholworthy , and I apologize for the delayed review.

My recommendation is to introduce a `is_logging_step` property (e.g., similar to [is_val_step](https://github.com/NVIDIA-NeMo/Automodel/blob/caa5177777d0f5a2c823d3fd2ac87d6605b3a222/nemo_automodel/components/training/step_scheduler.py#L144-L151)), and have it by default log on every step (and configurable as necessary). Then the train loops can proceed with checking this property and calling the `log_*metrics` function.

oliverholworthy: Thanks for the review and suggestion. I've created a property called  `is_remote_logging_step` on the `step_scheduler` that determines if we should log to Wandb or MLFlow

akoumpa (approved):

Issue #995, opened 2025-12-19
Checkpointing on Databricks


jrbourbeau:
White attempting to run multi-node model training on Databricks, I ran into some issues around model checkpointing. 

There are a few different places one can write files to on Databricks:

- Local VM filesystem. This is a POSIX filesystem, but is both ephemeral (goes away when VM shuts down) and not globally accessible (each VM used in training writes separately to their own local filesystem). 
- Workspace files ([docs](https://docs.databricks.com/aws/en/files/workspace)). This is persistent and globally accessible (🎉 ) but has a 500 MB file size limit, so can't really be used for saving model weights in practice. 
- Unity catalog volumes ([docs](https://docs.databricks.com/aws/en/volumes/volume-files)). This is persistent and globally accessible (🎉 ) but has some relevant limiations:
    - Doesn't support symlinks (which are used in checkpointing) because it's backed by cloud blob storage
    - Direct-append or non-sequential (random) writes aren't supported ([docs](https://docs.databricks.com/aws/en/volumes/volume-files#limitations-of-working-with-files-in-volumes))

A couple different options come to mind. The first is to see if there's a way to avoid some of the Databricks-specific limitations (e.g. symlinks) while also not breaking things on other systems. The second is to have each worker VM write to their own local filesystem and then have some final step that aggregates those results and saves them to some persistent location. 

I'm sure there are other options too, those are just the first things that come to mind for me

cc @HuiyingLi for visibility

HuiyingLi:
@adil-a  for viz

qiaochuz-nv:
## QA RCCA Analysis

**Status:** N/A — Platform-specific issue (Databricks filesystem). Not a code defect — Databricks POSIX filesystem limitations affect checkpoint symlinks. Explained by adil-a as expected behavior for async checkpointing on non-standard filesystems.
**Conclusion:** Not applicable for coverage analysis.

PR #1023, opened 2026-01-08
fix: Update checkpointing to support Databricks unity catalog

jrbourbeau:
This PR has a couple of updates to our checkpointing logic that enable things to run on Databricks (and other platforms that have similar file-related limitations) https://github.com/NVIDIA-NeMo/Automodel/issues/995. This PR should not be merged as is -- it's just to illustrate the sort of small-ish changes that could be made to address https://github.com/NVIDIA-NeMo/Automodel/issues/995. 

cc @HuiyingLi @adil-a for visibility

### DIFF
--- a/nemo_automodel/components/checkpoint/_backports/consolidate_hf_safetensors.py
+++ b/nemo_automodel/components/checkpoint/_backports/consolidate_hf_safetensors.py
@@ -261,7 +261,14 @@ def _process_output_file(
 
     sorted_tensors = sorted(output_data.fqn_data.items(), key=lambda x: x[1].offset_in_file)
 
-    with open(output_file, "wb") as output_stream:
+    # `_write_metadata()` already created/truncated the file and wrote the safetensors header.
+    # Here we only need to append the raw tensor bytes after that header.
+    if not os.path.exists(output_file):
+        raise FileNotFoundError(
+            f"Expected output file {output_file} to exist (header written in _write_metadata()), but it does not."
+        )
+
+    with open(output_file, "ab") as output_stream:
         # Process each tensor in sequential output order
         for tensor_fqn, tensor_fqn_data in sorted_tensors:
             full_tensor_mv = memoryview(


nemo_automodel/components/checkpoint/_backports/consolidate_hf_safetensors.py
```diff
@@ -261,8 +256,9 @@ def _process_output_file(
 
     sorted_tensors = sorted(output_data.fqn_data.items(), key=lambda x: x[1].offset_in_file)
 
-    with open(output_file, "r+b") as output_stream:
-        output_stream.seek(0, os.SEEK_END)
+    with open(output_file, "wb") as output_stream:
+    # with open(output_file, "r+b") as output_stream:
+        # output_stream.seek(0, os.SEEK_END)
```
jrbourbeau: `r+b` isn't supported on DBRX but I wasn't sure if it's actually necessary. From looking at the code here, it appears maybe `"wb"`, which is supported, is sufficient (I could be missing something though)

nemo_automodel/recipes/base_recipe.py
```diff
@@ -299,6 +288,7 @@ def _update_checkpoint_symlink(self, link_name: str, target_dir: str) -> None:
         that points to `target_dir`.
         Assumes caller ensures rank 0 if needed.
         """
+        return
```
jrbourbeau: This just skips using symlinks altogether (so not an actual solution). One alternative could be to have `LATEST` be an actual file that points to the latest checkpoint (as opposed to a symlink).

nemo_automodel/components/checkpoint/_backports/consolidate_hf_safetensors.py
akoumpa: from looking at the code, I'm not sure why r+b was even working, since at the end it does write. wb should be fine.

nemo_automodel/recipes/base_recipe.py
akoumpa: I assume DBRX does not support symlinks? If that's the case is there a case to detect the environment and guard the return based on it? Symlink is convenient because it's a fast disk operation (nearly atomic), which is useful in cases of multi-GB checkpoints and pre-emption.

akoumpa: After syncing with @HuiyingLi , I'm proposing to guard the symlink creation in a try/except and have additional text-base pointers (e.g., `LATEST.txt` and `LOWEST_VAL.txt`) that contain one line with the relative filepath.

jrbourbeau: > I assume DBRX does not support symlinks?

Yeah, that's correct. It's backed by cloud blob storage (instead of a POSIX filesystem) where symlinks don't exist. 

> If that's the case is there a case to detect the environment and guard the return based on it?

Databricks sets various environment variables we could use to detect "is this running on Databricks". But IIUC we'd still want to support loading from previously saved `LATEST` checkpoints, right?

akoumpa: s3 also has the same restrictions (@HuiyingLi ), so I was thinking instead to guard os.symlink on a try/catch with an OSError and also write the txt file.

jrbourbeau: Sounds good -- I'll update this PR accordingly

PR #1053, opened 2026-01-14
fix: fix custom llama and qwen state dict adapter concat and split

HuiyingLi:
The expected layout is [Q,K,V]. If load/save with _safe_concat/split on local tensor only while when multigpu FSDP shard on dim0 (default), the shards have incorrect q/k/v content (interleaved), causing multigpu fsdp (only) to have high loss, while the single gpu remains correct.

### DIFF
--- a/nemo_automodel/components/models/common/combined_projection/state_dict_adapter.py
+++ b/nemo_automodel/components/models/common/combined_projection/state_dict_adapter.py
@@ -242,9 +242,7 @@ class CombinedProjectionStateDictAdapter:
                 gate_up_actual_size = gate_up_weight.shape[0]
                 local_intermediate_size = gate_up_actual_size // 2
 
-                gate_weight, up_weight = gate_up_weight.split(
-                    [local_intermediate_size, local_intermediate_size], dim=0
-                )
+                gate_weight, up_weight = gate_up_weight.split([local_intermediate_size, local_intermediate_size], dim=0)
 
                 hf_state_dict[f"{prefix}.mlp.gate_proj.weight"] = gate_weight
                 hf_state_dict[f"{prefix}.mlp.up_proj.weight"] = up_weight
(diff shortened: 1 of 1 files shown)


nemo_automodel/components/models/common/combined_projection/state_dict_adapter.py
```diff
@@ -37,55 +38,58 @@ def _is_dtensor(tensor: torch.Tensor) -> bool:
...
-        if sum(local_split_sizes) != local_size:
-            raise RuntimeError(
-                f"Split size mismatch: local_split_sizes={local_split_sizes} sum to {sum(local_split_sizes)}, "
-                f"but local tensor size is {local_size}. This indicates an incorrect split calculation."
-            )
-
-        local_splits = local_tensor.split(local_split_sizes, dim=dim)
+        try:
+            return list(tensor.split(split_sizes, dim=dim))
+        except Exception:
```
ZhiyuLi-Nvidia: May I know the reason to handle exception here?

nemo_automodel/components/models/common/combined_projection/state_dict_adapter.py
```diff
@@ -37,55 +38,58 @@ def _is_dtensor(tensor: torch.Tensor) -> bool:
...
+            pass
+
+        # Fallback for small 1D Shard(0) tensors (e.g., fused QKV bias)
+        if (
+            tensor.ndim == 1
+            and dim == 0
+            and len(tensor.placements) == 1
+            and isinstance(tensor.placements[0], Shard)
+            and tensor.placements[0].dim == 0
```
ZhiyuLi-Nvidia: Emm there are 5 conditional checks. So there would be all together 2^5 cases, would `else` handle the other cases?

tests/unit_tests/distributed/test_parallelizer_utils.py
```diff
@@ -227,6 +227,163 @@ def fake__fully_shard(mod, *, mesh, mp_policy, offload_policy):
...
+        num_attention_heads=2,
+        num_key_value_heads=2,
+        hidden_size=4,
+        head_dim=2,
+        tie_word_embeddings=False,
+    )
+    adapter = mod.CombinedProjectionStateDictAdapter(cfg)
+
+    hf_sd = {
+        "model.layers.0.mlp.gate_proj.weight": mod.torch.ones((3, 4)),
```
ZhiyuLi-Nvidia: can we use random values so that make it a more challenging test cases?

We can set random seed for repro as well.

HuiyingLi: Updated the test to random val, thanks!

akoumpa (approved):

PR #1066, opened 2026-01-15
feat: add consolidation for Databricks

adil-a:
Fixes issue https://github.com/NVIDIA-NeMo/Automodel/issues/995


nemo_automodel/components/checkpoint/_backports/consolidate_hf_safetensors.py
```diff
@@ -558,13 +562,54 @@ def _consolidate_safetensors_files(
...
-    # Step 3: Write actual tensor data from input files to output files
-    _write_data(input_files_data, output_files_data, num_threads)
+    if use_staging:
+        # Use staging directory for writing files first, then copy to final destination.
+        # This works around remote storage systems that don't support direct-append or non-sequential writes.
+        if staging_dir is not None:
+            os.makedirs(staging_dir, exist_ok=True)
+            temp_dir = tempfile.mkdtemp(prefix="safetensors_consolidate_", dir=staging_dir)
+        else:
+            temp_dir = tempfile.mkdtemp(prefix="safetensors_consolidate_")
```
jrbourbeau, line 573: More of a style comment then anything else, but would something like 

```python
with tempfile.TemporaryDirectory(prefix="safetensors_consolidate_", dir=staging_dir) as temp_dir:
    # Existing code here
    ...
```

work here? Handles cleanup a little nicer

nemo_automodel/components/checkpoint/_backports/consolidate_hf_safetensors.py
```diff
@@ -558,13 +562,54 @@ def _consolidate_safetensors_files(
...
             metadata, size = _get_safetensors_file_metadata(f)
             input_files_data[safetensor_file] = _InputFileData(metadata_size=size, metadata=metadata)
-    # Step 1: Parse metadata to determine tensor shapes and types
-    _parse_input_metadata(input_files_data, output_files_data)
 
-    # Step 2: Write metadata headers to output files
-    _write_metadata(output_files_data)
-    # Step 3: Write actual tensor data from input files to output files
-    _write_data(input_files_data, output_files_data, num_threads)
+    if use_staging:
```
jrbourbeau, line 566: I think it's safe to assume when `staging_dir` is specified that they also want to use `use_staging`

```suggestion
    if use_staging or staging_dir:
```

Also makes it one less option the user needs to specify (when using a non-default staging directory) 

EDIT: I may be confused about how CLI options are getting mapped to local variables here. The thing I want to point out is that if a user does ``--checkpoint.staging_dir <my-dir>`` from the CLI, that they don't also have to specify another CLI option turning staging on overall (maybe that's `--checkpoint.single_rank_consolidation`)

PR #1107, opened 2026-01-25
fix: leave num_epochs unset if max_steps is specified

### LINKED ISSUE #1095: num_epochs should not default to 10 if max_steps is specified in step scheduler
**Describe the bug**

If the step scheduler section of config file specifies `max_steps` but not `num_epochs`, the value of `num_epochs` defaults to 10 and this can effectively overwrite the value of `max_steps`, depending on the size of the dataset.

Run a recipe with `num_epochs` unset and `max_steps` set to a value more than 10x the number of steps needed per epoch.

**Expected behavior**

`num_epochs` remains unset if `max_steps` is specified instead.

Sparlitu:
# What does this PR do ?

This PR aims to fix the case where max_steps is specified in the scheduler and num_epochs is not, effectively defaulting to 10 and affecting the number of steps in training.

# Changelog

- Changed StepScheduler logic to allow `num_epochs` to be None if `max_steps` is provided.
- Added backward compatibility check to default `num_epochs` to 10 if nothing is set.
- Added a unit test `test_scheduler_max_steps_priority` in `tests/unit_tests/components/training/test_step_scheduler.py`.
- Applied `ruff` formatting to modified files.

Fixes #1095

### DIFF
--- a/nemo_automodel/components/training/step_scheduler.py
+++ b/nemo_automodel/components/training/step_scheduler.py
@@ -36,11 +36,11 @@ def _calculate_max_steps(
     return num_epochs * epoch_len
 
 
-def _calculate_num_epochs(max_steps: int, epoch_len: Optional[int], default_num_epochs: int = 10) -> int:
+def _calculate_num_epochs(max_steps: Optional[int], epoch_len: Optional[int], default_num_epochs: int = 10) -> int:
     """
     Calculate the number of epochs out of maximum number of steps.
     """
-    if epoch_len is None:
+    if epoch_len is None or max_steps is None:
         return default_num_epochs
     return ceil(max_steps / epoch_len)
 
@@ -101,13 +101,10 @@ class StepScheduler(Stateful):
 
         # This is for backward compatibility in the sense that num_epochs's default value was 10
         if num_epochs is None:
-            if max_steps is None:
-                num_epochs = 10
-            else:
-                num_epochs = _calculate_num_epochs(
-                    max_steps,
-                    self.epoch_len,
-                )
+            num_epochs = _calculate_num_epochs(
+                max_steps,
+                self.epoch_len,
+            )
 
         self.num_epochs = num_epochs
         assert num_epochs is None or num_epochs > 0, (


nemo_automodel/components/training/step_scheduler.py
```diff
@@ -81,8 +83,15 @@ def __init__(
         assert start_step >= 0, "start_step must be greater than or equal to 0"
         self.epoch = start_epoch
         assert start_epoch >= 0, "start_epoch must be greater than or equal to 0"
+
+        # This is for backward compatibility in the sense that num_epochs's default value was 10
+        if num_epochs is None and max_steps is None:
+            num_epochs = 10
+
         self.num_epochs = num_epochs
```
akoumpa, line 109: Thanks for making this.

From a quick read, I think it's missing a function that derives the `num_epochs` from `max_steps` when `num_epochs` is None? what do you think?

Sparlitu: Great catch. Would you like something similar to _calculate_max_steps? What should the default_num_epochs be then? 10? Here's what I'm thinking.

```
def _calculate_num_epochs(
    max_steps: int, epoch_len: Optional[int], default_num_epochs: int = 10
) -> int:
    """
    Calculate the number of epochs out of maximum number of steps.
    """
    if epoch_len is None:
        return default_num_epochs
    return ceil(max_steps / epoch_len)
```

What do you say?

akoumpa: yeah I think that would a good idea :) Thanks a lot @Sparlitu

nemo_automodel/components/training/step_scheduler.py
```diff
@@ -101,13 +101,10 @@ def __init__(
...
-                    max_steps,
-                    self.epoch_len,
-                )
+            num_epochs = _calculate_num_epochs(
+                max_steps,
+                self.epoch_len,
+            )
 
         self.num_epochs = num_epochs
         assert num_epochs is None or num_epochs > 0, (
```
Sparlitu, line 110: I think it's safe to also remove the redundant check in the assertion, number of epochs cannot be None anymore. Maybe also move the comment to _calculate_num_epochs because where it is right now, I feel it's of no use. What do you think?

akoumpa: agree, but not worth rerunning ci for this one :)

akoumpa (approved):
Thanks a lot for contributing @Sparlitu!

PR #1116, opened 2026-01-26
feat: transformers v4 API

adil-a:
Refactor that moves model specific distributed logic into `auto_model.py` to enhance the "drop-in replacement" experience of NeMo Automodel. Also introduces a new .save_pretrained API similar to HF.


nemo_automodel/_transformers/auto_model.py
```diff
@@ -493,7 +546,8 @@ def _retry(**override):
 
         _verify_sdpa_support(model, is_hf_model, cp_size)
 
-        model.config.update({"nemo_version": __version__})
```
akoumpa, line 594: is `nemo_version` preserved elsewhere?

PR #1130, opened 2026-01-29
feat: disable validation completely when val_every_steps is -1

hemildesai:
When pp is enabled, val seq length needs to match train seq length. For some quick tests, it is useful to disable validation in this case


hemildesai:
/ok to test [f9698e2](https://github.com/NVIDIA-NeMo/Automodel/pull/1130/commits/f9698e21101e76a0d83037fa4973d34ffb68ee38)

akoumpa (changes requested):
can we keep this to a dev branch instead?

PR #1147, opened 2026-02-01
feat: remove checkpointer from Automodel class

adil-a:
QOL improvement; checkpointer creation and work is done inside the class now.


nemo_automodel/_transformers/auto_model.py
```diff
@@ -539,6 +539,7 @@ def apply_model_infrastructure(
     model_name_or_path=None,
     load_base_model=False,
     cache_dir=None,
+    pretrained_model_name_or_path="",
```
akoumpa, line 541: do we need both `model_name_or_path` and `pretrained_model_name_or_path`?

PR #1166, opened 2026-02-04
feat: refactor retriever code

adil-a:
image

Convergence runs done over the course of refactoring. The ones with `*main*` are from the main branch, and the ones with `*refactor*` are the final code ones.

### DIFF
--- a/tests/functional_tests/llm_pretrain_and_kd/customizer_retrieval/test_customizer_retrieval.py
+++ b/tests/functional_tests/llm_pretrain_and_kd/customizer_retrieval/test_customizer_retrieval.py
@@ -98,6 +98,7 @@ def _build_eval_model(device: torch.device):
         use_liger_kernel=False,
         use_sdpa_patching=False,
         torch_dtype=torch.bfloat16,
+        trust_remote_code=True,
     ).to(device).eval()
 
 


nemo_automodel/components/models/llama_bidirectional/model.py
```diff
@@ -0,0 +1,283 @@
...
+            )
+
+        hidden_states = self.norm(hidden_states)
+        return BaseModelOutputWithPast(
+            last_hidden_state=hidden_states,
+            past_key_values=past_key_values,
+        )
+
+
+def _pool(last_hidden_states: torch.Tensor, attention_mask: torch.Tensor, pool_type: str) -> torch.Tensor:
```
rnyak: if def pool already defined in the biencoder.py do we need to define it here again? can we import it?

nemo_automodel/components/models/llama_bidirectional/model.py
```diff
@@ -0,0 +1,283 @@
...
+            loss=loss,
+            logits=pooled_logits,
+            past_key_values=transformer_outputs.past_key_values,
+            hidden_states=transformer_outputs.hidden_states,
+            attentions=transformer_outputs.attentions,
+        )
+
+
+# Export for ModelRegistry auto-discovery
+ModelClass = LlamaBidirectionalModel
```
rnyak: would we be able to do ` ModelClass = LlamaBidirectionalForSequenceClassification` under this line when we want to train a cross-encdoer, aka reranker?
basically can we define two `ModelClass` here?

nemo_automodel/components/models/llama_bidirectional/model.py
```diff
@@ -0,0 +1,283 @@
...
+    config_class = LlamaBidirectionalConfig
+
+    def __init__(self, config: LlamaConfig):
+        """
+        Initialize LlamaBidirectionalModel.
+
+        Args:
+            config: Model configuration
+        """
+        super().__init__(config)
```
rnyak, line 96: may I ask how we are removing the casual masking here? 

i think these line are missing

```
        # Disable causal attention for all layers
        for layer in self.layers:
            layer.self_attn.is_causal = False

```

adil-a: Hey, yeah it is in the forward pass in the same file. Llama passes `**kwargs` into the attention kernel call, and we pass `is_causal=False` there. We also explicitly pass in an unmasked attention mask. Setting it to `False` as a class attribute is a no-op for Llama. See here: https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L236 notice that it is unused even though it is set.

nemo_automodel/components/models/llama_bidirectional/model.py
adil-a: Yes, we can pass a list. I'll make the change, thank you.

examples/biencoder/llama_embed_nemotron_8b/llama_embed_nemotron_8b.yaml
```diff
@@ -19,23 +19,20 @@
...
   val_every_steps: 3000
   num_epochs: 1
 
-dist_env:
-  backend: nccl
-  timeout_minutes: 1
-
 model:
-  _target_: nemo_automodel.components.models.biencoder.NeMoAutoModelBiencoder.from_pretrained
+  _target_: nemo_automodel._transformers.auto_model.NeMoAutoModelForBiencoder.from_pretrained
```
rnyak: for consistency shall it be `nemo_automodel.NeMoAutoModelForBiencoder.from_pretrained`?

adil-a: Yes you are correct that should be changed

examples/biencoder/llama3_2_1b_biencoder.yaml
```diff
@@ -19,45 +19,47 @@
...
+eval_negative_size: 4
+temperature: 0.02
+
 step_scheduler:
   global_batch_size: 128
   local_batch_size: 4
   ckpt_every_steps: 500
   val_every_steps: 500
   num_epochs: 1
+  max_steps: 50
```
rnyak: better to remove this or comment it out. the original yaml does not set max_steps.

examples/biencoder/llama3_2_1b_biencoder.yaml
```diff
@@ -122,11 +124,11 @@ optimizer:
   weight_decay: 0.01
   adam_w_mode: true
   bias_correction: true
-  master_weights: true
+  master_weights: false
 
 # Learning rate scheduler
 lr_scheduler:
-  lr_warmup_steps: 100
+  lr_warmup_steps: 5
```
rnyak: can we set this to `100` as in the original yaml file.

nemo_automodel/components/models/llama_bidirectional/model.py
oliverholworthy: > Setting it to False as a class attribute is a no-op for Llama


@adil-a I wanted to flag one thing regarding the is_causal change.

I don't believe setting is_causal = False on the attention modules was a no-op. While it's true that LlamaAttention never references self.is_causal directly within its own class, the attribute is read externally by the attention backend functions. 

LlamaAttention.forward passes it's own instance `self` to the attention backend:

```
attn_output, attn_weights = attention_interface(
    self,  # <-- module passed as first arg
    query_states, key_states, value_states, attention_mask, ...
    **kwargs,
)
```

The SDPA and Flash Attention backends then read the attribute from the module:

```
is_causal = is_causal if is_causal is not None else getattr(module, "is_causal", True)
```

So the original `__init__` approach was functional.

That said, the kwargs approach is functionally equivalent since the backends check the explicit is_causal kwarg first before falling back to the module attribute. Both work. My slight preference for the `__init__` approach is that bidirectionality is signaled as a property of the model itself, not of a particular forward call, and setting it as an attribute makes that intent clearer.

examples/biencoder/llama3_2_1b_biencoder.yaml
```diff
@@ -31,33 +35,30 @@ dist_env:
   timeout_minutes: 1
 
 model:
-  _target_: nemo_automodel.components.models.biencoder.NeMoAutoModelBiencoder.from_pretrained
```
oliverholworthy, line 34: Question about the `NeMoAutoModelForBiencoder` naming: I noticed the class was renamed from `NeMoAutoModelBiencoder` to `NeMoAutoModelForBiencoder`. The `For` prefix follows the HuggingFace `AutoModelFor*` convention, but in HF that pattern maps to task types (`ForCausalLM`, `ForSequenceClassification`, etc.) where the class inherits from the corresponding HF AutoModel. "Biencoder" is an architecture, not a task, and this class is standalone, it doesn't subclass _BaseNeMoAutoModelClass or any HF `AutoModelFor*`.

So I feel like the name containing `For` is a bit misleading and doesn't follow the same convention as the other classes we have along those lines. Do you think it's worth further consideration to this name?

nemo_automodel/components/models/llama_bidirectional/model.py
adil-a: @ybabakhin mentioned that there are cases where setting it as the class attribute is not sufficient. Let's keep it as is for now and we can revisit this in the PR that deals with adding further kinds of custom models (including just HF ones without any modifications)

examples/biencoder/llama3_2_1b_biencoder.yaml
adil-a: That's a fair point. Happy to rename here

nemo_automodel/components/models/llama_bidirectional/model.py
oliverholworthy: > setting it as the class attribute is not sufficient

Setting as class attribute is indeed not sufficient, in the same way passing is_causal=False as a kwarg is also not sufficient either. We also need to construct the bidirectional mask correctly.

Which is particularly relevant where attention implementation is sdpa/eager rather than flash attention. (related PR: #1349)

adil-a: Ok, I will include this back in the MR for supporting the other models. At least for Llama this is not an issue currently

PR #1181, opened 2026-02-06
feat: v5 refactor for device mesh only model init

adil-a:
for ease of review: https://app.devin.ai/review/NVIDIA-NeMo/Automodel/pull/1181

qwen3: https://wandb.ai/Nemo-automodel/adil-workspace?nw=dr1sucu78rt
llama 3.2: https://wandb.ai/Nemo-automodel/adil-workspace?nw=84mxts1rcfy

### DIFF
--- a/nemo_automodel/recipes/llm/train_ft.py
+++ b/nemo_automodel/recipes/llm/train_ft.py
@@ -66,6 +66,7 @@ from nemo_automodel.components.loss.masked_ce import MaskedCrossEntropy
 from nemo_automodel.components.optim.scheduler import OptimizerParamScheduler
 from nemo_automodel.components.optim.utils import build_dion_optimizer, is_dion_optimizer
 from nemo_automodel.components.quantization.fp8 import build_fp8_config
+from nemo_automodel.components.training.model_output_utils import get_final_hidden_states
 from nemo_automodel.components.training.rng import ScopedRNG, StatefulRNG
 from nemo_automodel.components.training.step_scheduler import StepScheduler
 from nemo_automodel.components.training.utils import (
@@ -1198,7 +1199,7 @@ class TrainFinetuneRecipeForNextTokenPrediction(BaseRecipe):
                     logits=getattr(out, "logits", out),
                     labels=labels,
                     model=model,
-                    hidden_states=out.hidden_states[-1] if getattr(out, "hidden_states", None) is not None else None,
+                    hidden_states=get_final_hidden_states(out),
                     num_label_tokens=num_label_tokens,
                 )
                 loss_buffer.append(local_loss.clone().detach())
--- a/docker/common/uv-pytorch.lock
+++ b/docker/common/uv-pytorch.lock
@@ -2752,16 +2752,16 @@ wheels = [
 
 [[package]]
 name = "megatron-fsdp"
-version = "0.1.0"
+version = "0.2.3"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "einops" },
     { name = "packaging" },
     { name = "torch", marker = "sys_platform == 'never'" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/95/8d/07223a667484e03311c8a0db6f2a19f69d0845c1661195ebb7ed86bf39ec/megatron_fsdp-0.1.0.tar.gz", hash = "sha256:3c0714e6f9fa484af72d2693f1633711094473ce42967cfa81c74a630c841556", size = 82754, upload-time = "2025-10-08T15:04:31.998Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/10/58/1dc5f0a79634cb6dd11715b62c994e6e1d43de952a2ea914c07b78bc3c48/megatron_fsdp-0.2.3.tar.gz", hash = "sha256:d12d93b20e3e5d214ae44d115976b209ef7352221f5ec07767bcbcb978ec5098", size = 86472, upload-time = "2026-02-06T16:30:27.88Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/6f/44/213003e3e9cd4d2014d59ffcb699ceb2b1b0ae6db6015d024718c63bc89e/megatron_fsdp-0.1.0-py3-none-any.whl", hash = "sha256:c4a4190f8fdbd0e89f7f47a6a9b218f709498ef0a676729449dad7e55688a10c", size = 83287, upload-time = "2025-10-08T15:04:30.676Z" },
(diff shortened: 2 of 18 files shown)


nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -130,60 +129,86 @@ def build_model_and_optimizer(
     cfg_model,
     cfg_opt,
     cfg_peft,
-    model_wrapper,
     seed,
     has_packed_sequence=False,
+    dp_size=None,
```
akoumpa: Is it possible to simplify here a bit? I was thinking if we could replace all the *_size variables by querying the meshes instead.

nemo_automodel/_transformers/auto_model.py
```diff
@@ -802,20 +1058,31 @@ def from_pretrained(
...
         sdpa_method: Optional[List[SDPBackend]] = None,
         torch_dtype="auto",
         attn_implementation: str = DEFAULT_ATTN_IMPLEMENTATION,
         quantization_config=None,
         force_hf: bool = False,
-        model_wrapper=None,
-        autopipeline: AutoPipeline | None = None,
-        parallelize_fn: Callable | None = None,
+        # Parallelism sizes (None = infer from device_mesh if provided)
+        dp_size: Optional[int] = None,
```
akoumpa: I'm somewhat at cross-road here. 

The HF API defines the tp_size, therefore if we are to extend that convention we need to use all the ep_size/cp_size etc, which on a second thought may not be the cleanest solution?

Given a device-mesh with standardized submesh names, the tp_size/ep_size variable become redundant and in fact a possible source of error. For example, if the submesh for TP has size 8 but the user passes tp_size of 2 then this would cause troubles. From this perspective, the UI is asking for trouble. I prefer to pass-in the device-mesh rather than create it inside the from_pretrained (which i think is the reason tp_size was introduced). Having the ability to pass-in the device-mesh is useful in cases of two models where you want to allocate GPUs [a-b] to model a and [c-g] to model b.

So I'm thinking if we want to keep tp_size (since it's in the HF API) and derive the rest from the device_mesh (-- as I mentioned, at cross-road).

nemo_automodel/_transformers/auto_model.py
```diff
@@ -802,20 +1058,31 @@ def from_pretrained(
...
+        dp_replicate_size: Optional[int] = None,
+        tp_size: Optional[int] = None,
+        pp_size: Optional[int] = None,
+        cp_size: Optional[int] = None,
+        ep_size: Optional[int] = None,
+        # Device mesh
+        device_mesh: Optional["DeviceMesh"] = None,
+        moe_mesh: Optional["DeviceMesh"] = None,
```
akoumpa, line 324: I don't like this, but all things considered I think it might be acceptable (CF is on monday anyway).

My alternative would be to subclass `DeviceMesh` into a new class called `MultiViewDeviceMesh` and be able to provide both the moe and non-moe meshes with a single object.

nemo_automodel/_transformers/auto_model.py
```diff
@@ -1119,17 +1453,76 @@ def _retry(**override):
...
+        # Initialize infrastructure objects (will be populated from configs if provided)
+        model_wrapper = None
+        autopipeline = None
+        parallelize_fn = None
+        qat_quantizer = None
+        loss_fn = None
+
+        # Instantiate from configs if provided
+        if distributed_config is not None:
+            model_wrapper, autopipeline, parallelize_fn, qat_quantizer = instantiate_infrastructure(
```
akoumpa, line 488: can we break this function into smaller ones? just by the name i feel it's doing too many things?

nemo_automodel/_transformers/auto_model.py
adil-a: Firstly, I agree it's redundant the way we are passing it right now. You should need one or the other, not both. I don't really like the idea of creating the device mesh inside from_pretrained because the program will need it at a certain point, and it's awkward if from_pretrained returns a device mesh alongside the model.

IMO it's best to redact all the *_size parameters here in favor of just keeping the device mesh. I think this is an ok design decision (and that it feels somewhat like tech debt on HF side that they have both tp_size and device mesh).

nemo_automodel/_transformers/auto_model.py
adil-a: @hemildesai mentioned it might be possible to combine both the device meshes as part of the new PyT release. We can look at this improvement for next release.

nemo_automodel/_transformers/auto_model.py
adil-a: The function itself does break down to many other functions, but I want to keep the core logic inside the from_config and from_pretrained short and quick to read. The other option is to bring all the functions `instantiate_infrastructure` calls but put them inside the from_config and from_pretrained methods instead, but it might pollute it. I guess it's just a design decision, lmk your thoughts

nemo_automodel/recipes/llm/train_ft.py
adil-a: Yep, should be possible. from_pretrained and from_config call a function to extract the relevant sizes so I'll do the same here (we should still fix the redundancy in parameters in from_pretrained)

nemo_automodel/_transformers/auto_model.py
akoumpa: cool let's drop the new `*_size`s and follow the device mesh.

nemo_automodel/recipes/llm/train_ft.py
akoumpa: cool, let's do it then!

nemo_automodel/_transformers/auto_model.py
akoumpa: ok, can you add a comment that `instantiate_infrastructure`/`_instantiate_distributed` etc are temporary and no other change? I think we'd be better off refactoring this, even if we have to refactor the fsdp2manager too.

nemo_automodel/components/distributed/fsdp2.py
```diff
@@ -30,261 +27,58 @@
...
+        device_mesh: DeviceMesh,
+        moe_mesh: Optional[DeviceMesh] = None,
+    ):
+        self.config = config
+        self.device_mesh = device_mesh
+        self.moe_mesh = moe_mesh
+
+        # Extract config fields for easy access
+        self.sequence_parallel = config.sequence_parallel
+        self.use_hf_tp_plan = config.use_hf_tp_plan
```
akoumpa: let's deprecate `use_hf_tp_plan` and `custom_tp_plan` in favor of `tp_plan`.

We can use the `tp_plan` argument from `from_pretrained`/`from_config`, if it's None then it can look up the internal registry.

nemo_automodel/_transformers/auto_model.py
adil-a: https://github.com/pytorch/pytorch/blob/main/torch/distributed/device_mesh.py#L1321

PR #1206, opened 2026-02-10
fix: enable NemotronHConfig FLOPs calculation in flops_utils.py

ShiftyBlock:
# What does this PR do ?

Enables NemotronHConfig FLOPs calculation by adding it to the config lookup dictionary and fixing attribute name mismatches with the published HuggingFace config.
# Changelog

- Added NemotronHConfig to the mapping in get_flops_formula_for_hf_config()
- Changed config.mamba_state_dim to config.ssm_state_size in _mamba_layer_flops()
- Changed config.mamba_num_groups to config.n_groups in _mamba_layer_flops()
- Replaced config.is_hybrid_model check with hasattr(config, "hybrid_override_pattern") in _hybrid_model_flops()

**Pre checks**:

# Additional Information
Attribute names were verified against the NemotronHConfig published in nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16.

Co-authored with @rishipython: he originally discovered that nemotronh_flops was unused in the config-to-formula mapping, which led to the investigation of the downstream attribute mismatches.

### DIFF
--- a/nemo_automodel/components/utils/flops_utils.py
+++ b/nemo_automodel/components/utils/flops_utils.py
@@ -532,9 +532,19 @@ def _mamba_layer_flops(config, gbs, seq_len):
     """Model FLOPs for Mamba layer. We ignore part of the flops of scan because the
     chunk size is not known from model config."""
     hs = config.hidden_size
-    mamba_state_dim = config.mamba_state_dim
+    if hasattr(config, "mamba_state_dim"):
+        mamba_state_dim = config.mamba_state_dim
+    elif hasattr(config, "ssm_state_size"):
+        mamba_state_dim = config.ssm_state_size
+    else:
+        raise ValueError("Expected config to have 'mamba_state_dim' or 'ssm_state_size'")
     mamba_head_dim = config.mamba_head_dim
-    mamba_num_groups = config.mamba_num_groups
+    if hasattr(config, "mamba_num_groups"):
+        mamba_num_groups = config.mamba_num_groups
+    elif hasattr(config, "n_groups"):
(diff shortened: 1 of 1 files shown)


nemo_automodel/components/utils/flops_utils.py
```diff
@@ -532,9 +532,9 @@ def _mamba_layer_flops(config, gbs, seq_len):
     """Model FLOPs for Mamba layer. We ignore part of the flops of scan because the
     chunk size is not known from model config."""
     hs = config.hidden_size
-    mamba_state_dim = config.mamba_state_dim
+    mamba_state_dim = config.ssm_state_size
```
akoumpa:
```suggestion
    if hasattr(config, 'mamba_state_dim'):
       mamba_state_dim = config.mamba_state_dim
    elif hasattr(config, 'ssm_state_size'):
       mamba_state_dim = config.ssm_state_size
    else:
       raise ValueError("Expected config to have `mamba_state_dim` or `ssm_state_size`)
```

Hi @ShiftyBlock , thanks a lot for the fix!

I would propose a more defensive approach, for example, to keep the other attribute name in case some other model uses it? And similarly for the other cases in the file. Please let me know what you think :)

davidyang05: Thank you @akoumpa for the quick feedback. A defensive approach makes sense - definitely possible other model configs might use it. I've gone ahead and made those changes you've requested, please let me know if you have anything else you'd like to add!

akoumpa: looking great, thanks a lot @ShiftyBlock!

PR #1207, opened 2026-02-10
fix: config fixes

adil-a:
GPT-OSS, qlora, model init fixes

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -37,6 +37,12 @@ FROM ${BASE_IMAGE} AS update_base_container
 ENV PIP_NO_CACHE_DIR=1
 WORKDIR /opt
 
+ENV DEBIAN_FRONTEND=noninteractive
+# Address CVE-2025-68973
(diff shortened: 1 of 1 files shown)


examples/llm_finetune/qwen/qwen1_5_moe_a2_7b_qlora.yaml
```diff
@@ -42,7 +42,7 @@ model:
 
 peft:
   _target_: nemo_automodel.components._peft.lora.PeftConfig
-  target_modules: ["*.experts.*"]
+  match_all_linear: True
```
akoumpa, line 45: 🤔

adil-a: experts lora is unsupported currently due to the change from HF and MoEs. They're only parameter objects (instead of nn.Linear modules you can easily swap out).

This'll have to be a feature at a later point.

PR #1300, opened 2026-02-17
feat: add MoE LoRA rank scaling and torch_mm to MoE LoRA

hemildesai:
wandb - https://wandb.ai/Nemo-automodel/automodel-moe-lora

Implement per-expert LoRA rank scaling based on [LoRA Without Regret](https://thinkingmachines.ai/blog/lora/): when `moe_rank_scaling=True`, MoE expert modules receive a LoRA rank of `dim // n_activated_experts` while standard Linear modules keep the full rank. This allows training a separate LoRA per expert with appropriately reduced rank, matching the paper's recommendation. Default is `False`, preserving existing behavior.

Also renames `lora_moe.py` to `lora_experts.py` for clarity, updates Qwen3 MoE configs for torch_mm experts backend, and adds a new TE packed-sequence LoRA config.

- **nemo_automodel/components/_peft/lora.py**
  - Add `moe_rank_scaling: bool = False` to `PeftConfig`
  - Wire through `from_dict` deserialization
  - In `apply_lora_to_linear_modules`, compute scaled rank for MoE modules when enabled, with validation (ValueError if rank < 1) and a warning on non-exact division

- **nemo_automodel/components/_peft/lora_moe.py to lora_experts.py**
  - Rename for clarity; update all imports

- **tests/unit_tests/_peft/test_lora_moe.py**
  - Add 5 new tests for moe_rank_scaling:
    - Basic scaling (dim=16, n_activated=2 gives MoE rank=8, Linear rank=16)
    - Default off (both keep full rank)
    - Floor-division with warning (dim=16, n_activated=3 gives rank=5)
    - dim too small raises ValueError
    - Output equivalence with zero-init B matrices

- **examples/llm_finetune/qwen/**
  - Update qwen3_moe_30b_lora.yaml (torch_mm experts, disable actckpt)
  - Update qwen3_moe_30b_te_packed_sequence.yaml (safetensors format)
  - Add qwen3_moe_30b_te_packed_sequence_lora.yaml

### DIFF
--- a/tests/unit_tests/_peft/test_lora_experts.py
+++ b/tests/unit_tests/_peft/test_lora_experts.py
@@ -760,6 +760,396 @@ def test_deepep_lora_relu2_non_gated_shapes(device):
     assert lora_module.lora_down_A.shape == (4, 32, 4)
 
 
+# ---------- torch_mm forward path tests ----------
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA required")
+def test_grouped_experts_lora_forward_torch_mm(moe_config, device):
+    """Test forward pass of GroupedExpertsLoRA using _forward_grouped_mm (use_torch_mm=True)."""
+    orig_experts = GroupedExperts(moe_config).to(device)
+    with torch.no_grad():
+        orig_experts.init_weights(buffer_device=device)
+
+    # Force use_torch_mm so forward dispatches to _forward_grouped_mm
+    orig_experts.use_torch_mm = True
+    lora_experts = GroupedExpertsLoRA(orig_experts, lora_dim=4, alpha=8).to(device)
+    assert lora_experts.use_torch_mm is True
+
+    num_tokens = 10
+    x = torch.randn(num_tokens, 16, device=device)
+    token_mask = torch.ones(num_tokens, dtype=torch.bool, device=device)
+    weights = torch.rand(num_tokens, 2, device=device)
+    indices = torch.randint(0, 4, (num_tokens, 2), device=device)
+
+    out = lora_experts(x, token_mask, weights, indices)
+    assert out.shape == (num_tokens, 16)
+    assert not torch.isnan(out).any()
+    assert not torch.isinf(out).any()
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA required")
+def test_grouped_experts_lora_forward_torch_mm_equivalence(moe_config, device):
+    """With zero-init B, torch_mm path should match the loop path output."""
+    orig_experts = GroupedExperts(moe_config).to(device)
+    with torch.no_grad():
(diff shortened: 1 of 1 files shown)


examples/llm_finetune/qwen/qwen3_moe_30b_lora.yaml
```diff
@@ -34,7 +34,7 @@ model:
     attn: te
     linear: torch
     rms_norm: te
-    experts: te
+    experts: torch_mm
```
adil-a, line 37: Why do we need to change this to torch_mm? What about gmm?

Secondly, do we need to change the rest of the custom moe configs that are applying lora?

examples/llm_finetune/qwen/qwen3_moe_30b_lora.yaml
adil-a, line 1: There is no moe rank scaling config included here.

Also a wider comment: do we need to include the moe rank scaling parameter? Can we make it the default if it gives good downstream perf, and include a comment on this in our docs?

nemo_automodel/components/_peft/lora_experts.py
```diff
@@ -148,6 +62,9 @@ def __init__(self, orig_module: GroupedExperts, lora_dim=8, alpha=32, lora_A_ini
             self.gate_up_proj_bias.data.copy_(orig_module.gate_up_proj_bias.data)
             self.down_proj_bias.data.copy_(orig_module.down_proj_bias.data)
 
+        # Copy backend setting from original (super().__init__ defaults to False without backend)
+        self.use_torch_mm = orig_module.use_torch_mm
```
adil-a, line 66: same q as before; what about gmm?

examples/llm_finetune/qwen/qwen3_moe_30b_lora.yaml
adil-a: Maybe you're using torch_mm because load balancing is a bigger issue with LoRA?

hemildesai: torch_mm generally seems to give better performance when not perfectly balanced, so changed the default here.

examples/llm_finetune/qwen/qwen3_moe_30b_lora.yaml
hemildesai: I am not sure if moe rank scaling is canonicalized yet so left the default as `False` for now.

nemo_automodel/components/_peft/lora_experts.py
hemildesai: `gmm` will still be available for LoRA.

hemildesai:
> Another experiment we should do if not already done is to compare the convergence and throughput of this PR with the way we apply LoRA to MoEs on main

Yes I can do that.

PR #1301, opened 2026-02-17
fix: skip load-before-shard when expert parallelism is active

hemildesai:
When EP > 1, loading the checkpoint before FSDP/EP sharding can cause issues.

This adds `ep_size` to the `load_before_shard` condition so that EP configurations use the post-shard checkpoint loading path.

- Extract `_should_load_before_shard()` helper into `_transformers/utils.py` for testability (avoids heavy transformer_engine import chain)
- Call the helper from `apply_model_infrastructure()` in `infrastructure.py`
- Add 13 unit tests covering every condition of the load-before-shard decision (EP, TP, PP, PEFT, pretrained path, load_base_model, and parametrized combinations)

### DIFF
--- a/nemo_automodel/_transformers/infrastructure.py
+++ b/nemo_automodel/_transformers/infrastructure.py
@@ -30,6 +30,7 @@ from typing import TYPE_CHECKING, Optional, Union
 
 import torch
 
+from nemo_automodel._transformers.utils import _should_load_before_shard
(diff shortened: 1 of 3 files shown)


hemildesai:
/ok to test [8c7043b](https://github.com/NVIDIA-NeMo/Automodel/pull/1301/commits/8c7043b034fe0e3e9807f5ea6b51033139b315e4)

adil-a (approved):

PR #1303, opened 2026-02-17
ci: Update transformers to latest version 5.2.0

thomasdhc:
beep boop 🤖: Updating transformers to latest version on pypi

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,7 +84,7 @@ dependencies = [
     "pyyaml",
     "torch<=2.10.0",
     "torchdata",
-    "transformers>=5.0.0",
+    "transformers<=5.2.0",
     "wandb",
     "torchao",
     "mlflow",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -84,7 +84,7 @@ dependencies = [
     "pyyaml",
     "torch<=2.10.0",
     "torchdata",
-    "transformers>=5.0.0",
```
oliverholworthy, line 87: Does dropping the lower bound mean we'd want to support transformers 4.x? The current spec has >=5.0.0, and we may want to keep that if the codebase is built around 5.x APIs.

On the upper bound: `<=5.2.0` feels a bit restrictive to me, since it would mean we'd need a dependency bump PR for every new minor release even if nothing breaks. And it would actively prevent users from using newer versions of transformers with Automodel even if nothing actually breaks. In my experience, unnecessarily tight upper bounds tend to be a pain point for downstream users of libraries. 

Would something like `>=5.0.0<6` work instead? That way we're protected against a major version change but still can enable use with 5.x releases without friction.

PR #1324, opened 2026-02-18
fix: combined projection

ZhiyuLi-Nvidia:
# What does this PR do ?

## Problem
The custom model's combined projections (qkv_proj, gate_up_proj) use ColwiseParallel for tensor parallelism, but the weight layout was a naive concatenation:
* QKV: [all Q rows | all K rows | all V rows]
* gate_up: [all gate rows | all up rows]
With ColwiseParallel (which shards dim 0 evenly), this gives each TP rank the wrong mix of Q/K/V heads — e.g., rank 0 gets all Q and some K, rank 1 gets remaining K and all V. This produces silently incorrect results, especially under GQA where Q and KV head counts differ.

# Changelog

Interleaved weight layout so that ColwiseParallel sharding naturally gives each rank complete, matched groups:
* QKV: KV-head-grouped layout [Q_group_0 | K_0 | V_0 | Q_group_1 | K_1 | V_1 | ...] — each TP rank gets whole KV-head groups with their corresponding Q heads.
* gate_up: Row-interleaved layout [gate_0, up_0, gate_1, up_1, ...] — each TP rank gets matched gate/up pairs.

Also included: DCP-based base model loading option

# Tests
Match hf and custom implementation loss curve
* llama3-8b TP4FSDP2: https://wandb.ai/nvidia/automodel-dev-zhiyul/workspace?nw=62e6rt6s4kv
* llama3-70b: https://wandb.ai/nvidia/automodel-dev-zhiyul?nw=43ugbs5opby

# Additional Information

- Related to # (issue)

### DIFF
--- a/tests/functional_tests/llm_pretrain_and_kd/L2_TP_Output_Parity_Minified.sh
+++ b/tests/functional_tests/llm_pretrain_and_kd/L2_TP_Output_Parity_Minified.sh
@@ -23,7 +23,7 @@ KL_THRESHOLD="${KL_THRESHOLD:-1e-6}"
 
 torchrun --nproc_per_node=2 --nnodes=1 \
     tests/functional_tests/llm_pretrain_and_kd/run_tp_output_parity_minified.py \
-    --models qwen3 qwen3_seq_cls ministral3 \
+    --models qwen3 qwen3_seq_cls ministral3 llama qwen2 \
     --sequence_parallel both \
     --kl_threshold "${KL_THRESHOLD}"
 
--- a/tests/functional_tests/llm_pretrain_and_kd/run_tp_output_parity_minified.py
+++ b/tests/functional_tests/llm_pretrain_and_kd/run_tp_output_parity_minified.py
@@ -20,6 +20,8 @@ variants of:
 - Qwen3ForCausalLM (HF)
 - Qwen3ForSequenceClassification (HF)
 - Ministral3ForCausalLM (NeMo Automodel custom)
+- LlamaForCausalLM (NeMo Automodel custom, combined QKV + gate_up projections)
+- Qwen2ForCausalLM (NeMo Automodel custom, combined QKV + gate_up projections)
 
 It also validates both tensor-parallel plans:
 - sequence_parallel=False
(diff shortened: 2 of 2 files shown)


nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -122,6 +122,9 @@ class CheckpointingConfig:
     # This should be used for remote storage systems that don't support direct-append or non-sequential writes.
     staging_dir: str | None = None  # Optional directory for staging files during consolidation.
     # If provided, temp files will be created here instead of system temp. Useful when system temp has limited space.
+    # If True, load base model via DCP (distributed read) instead of "every rank loads full state dict".
+    # Can avoid slow sequential CPU load + distribute when many ranks each do a full ckpt read and CPU→GPU copy.
+    use_dcp_for_base_model_load: bool = False
```
akoumpa: let's wait for this, I'm not sure we want to include it. 

I expect the safetensors load bottleneck to be solved in https://github.com/NVIDIA-NeMo/Automodel/pull/1163

nemo_automodel/components/models/common/combined_projection/state_dict_adapter.py
```diff
@@ -73,6 +73,41 @@ def __init__(self, config):
...
+        self.group_size = self.num_attention_heads // self.num_key_value_heads
+
+    def _interleave_qkv(self, q: torch.Tensor, k: torch.Tensor, v: torch.Tensor) -> torch.Tensor:
+        """Interleave Q, K, V by KV-head groups for TP-correct ColwiseParallel sharding.
+
+        Layout: [Q_group_0 | K_0 | V_0 | Q_group_1 | K_1 | V_1 | ...]
+        where each group has (group_size * head_dim) Q rows, head_dim K rows, head_dim V rows.
+        """
+        rest = q.shape[1:]
+        q = q.view(self.num_key_value_heads, self.group_size * self.head_dim, *rest)
```
akoumpa: Just double checking, is view correct or should it use reshape?

nemo_automodel/components/checkpoint/checkpointing.py
ZhiyuLi-Nvidia: OK. Let me revert this part.

nemo_automodel/components/models/common/combined_projection/state_dict_adapter.py
ZhiyuLi-Nvidia: Yeap. Reshape should be more robust. 

In all of my experiments, view is working well, I think q should be contiguous. Anyway, let me change it to reshape.

ZhiyuLi-Nvidia: Updated.

nemo_automodel/components/checkpoint/checkpointing.py
ZhiyuLi-Nvidia: Reverted.

PR #1335, opened 2026-02-19
fix: qwen3_8b_hellaswag_pp_peft recipe

ZhiyuLi-Nvidia:
# What does this PR do ?

As titled, remove explicitly set cache_dir and add pad sequence length.

# Changelog

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/examples/llm_finetune/llama3_2/llama_3_2_3b_instruct_squad.yaml
+++ b/examples/llm_finetune/llama3_2/llama_3_2_3b_instruct_squad.yaml
@@ -34,6 +34,10 @@ model:
   pretrained_model_name_or_path: meta-llama/Llama-3.2-3B-Instruct
   torch_dtype: bf16
 
+  backend:
+    _target_: nemo_automodel.components.models.common.BackendConfig
+    rms_norm: "torch"
+
 checkpoint:
   enabled: true
   checkpoint_dir: checkpoints/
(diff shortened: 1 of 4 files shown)


examples/llm_finetune/qwen/qwen3_8b_hellaswag_pp_peft.yaml
```diff
@@ -82,7 +82,9 @@ packed_sequence:
 
 dataloader:
   _target_: torchdata.stateful_dataloader.StatefulDataLoader
-  collate_fn: nemo_automodel.components.datasets.utils.default_collater
+  collate_fn:
+    _target_: nemo_automodel.components.datasets.utils.default_collater
+    pad_seq_len_divisible: 320
```
akoumpa: Hi @ZhiyuLi-Nvidia. for this one can you share a little more context why it is needed? I thought for fp8 we needed the padding, but I'm not sure what I'm missing here.

ZhiyuLi-Nvidia: The root cause is torch pp doesn't support variant seq len. This is to do the padding to avoid the above issue.

Could you take a look at the logs: https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/0t1oh7bz/logs

```
torchrun --nproc-per-node=8 nemo_automodel/recipes/llm/train_ft.py --config examples/llm_finetune/qwen/qwen3_8b_hellaswag_pp_peft.yaml 
```

Without the change, The pipeline compiled based on the constant cached the shape [2, 168, 4096] from training step.
When validation started, the first validation batch happened to be shorter (seq_len=72), resulting in a shape of [2, 72, 4096]. The pipeline validation logic detected that 72 != 168 and raised the PipeliningShapeError.

Issue #1336, opened 2026-02-19
Potential loss masking conflict when pad_token_id == eos_token_id (Case study on Qwen-like settings)


wswaq:
### **Description**

I have identified a critical issue in the current training logic that causes significant performance degradation (specifically in Perplexity/PPL) when training models where the **Pad Token** and **EOS Token** share the same ID.

This is the default configuration for the **Qwen** series (e.g., Qwen1.5/2/2.5), where `<|endoftext|>` serves as both `pad_token` and `eos_token`.

### **The Bug**

In the current implementation, the loss mask is generated by checking if a token matches `self._pad_token_id`.

```python
# Current problematic implementation
loss_mask[labels == self._pad_token_id] = 0.0

# For padded sequences, ensure the embedding layer can map the token ID
tokens[tokens == self._pad_token_id] = 0
labels[labels == self._pad_token_id] = 0

```

### **Technical Analysis**

When `pad_token_id == eos_token_id`, this logic makes no distinction between **actual sequence termination** and **batch padding**.

1. **Loss Signal Erasure**: The model never receives gradients for the EOS token. It effectively learns that "predicting the end of a sentence doesn't matter," which breaks the model's ability to learn sequence boundaries.
2. **Infinite Generation**: During inference, models trained with this bug often suffer from "repetitive loops" or "run-on sentences" because the probability of the EOS token was never optimized.
3. **PPL Regression**: In my experiments with a **500M parameter model** trained on **40B tokens**, this bug caused the **WikiText-2 PPL** to inflate to **50+**, whereas it should realistically be around **20-25**.

### **Evidence**

* **Model:** Qwen-based (500M)
* **Dataset:** 40B Tokens
* **Observed Behavior:** PPL is nearly double compared to smaller models (300M) trained on a different repo with correct masking logic. The only significant delta identified is this specific masking block.

akoumpa:
hi @wswaq , thanks for raising this, I'm looking into this issue and should have a fix soon.

akoumpa:
Thanks a lot @wswaq , very much appreciate this, I will follow up with a second PR.

qiaochuz-nv:
## QA RCCA Analysis

**Fix PR:** #1338 (+1154/-27)
**Root Cause:** When `pad_token_id == eos_token_id` (Qwen-like models), loss masking incorrectly masks EOS tokens, causing PPL degradation.

[...]

PR #1338, opened 2026-02-19
fix: loss masking with pad eos collision

akoumpa:
# What does this PR do ?

Please see https://github.com/NVIDIA-NeMo/Automodel/issues/1336 for the problem description. 

# Changelog

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/tests/unit_tests/_transformers/test_auto_tokenizer.py
+++ b/tests/unit_tests/_transformers/test_auto_tokenizer.py
@@ -18,7 +18,7 @@ from unittest.mock import patch
 from transformers.tokenization_utils_base import BatchEncoding
 
 from nemo_automodel._transformers.auto_tokenizer import NeMoAutoTokenizer
-from nemo_automodel._transformers.tokenization.nemo_auto_tokenizer import _add_token, _remap_system_role
+from nemo_automodel._transformers.tokenization.nemo_auto_tokenizer import _add_token
 
 
 class _StubHFTokenizer:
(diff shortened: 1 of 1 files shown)


nemo_automodel/components/datasets/llm/formatting_utils.py
```diff
@@ -24,6 +26,47 @@
...
+    """
+    if pad_token_id != eos_token_id:
+        return sequence == pad_token_id
+
+    mask = torch.zeros(sequence.shape, dtype=torch.bool, device=sequence.device)
+    non_pad_positions = (sequence != pad_token_id).nonzero(as_tuple=True)[0]
+    if non_pad_positions.numel() > 0:
+        last_content_idx = non_pad_positions[-1].item()
+        # last_content_idx + 1 → real EOS (keep), last_content_idx + 2 → padding
+        mask[last_content_idx + 2 :] = True
```
ZhiyuLi-Nvidia, line 63: Does it hold true when pad_token_id and eos_token_id are different?

ZhiyuLi-Nvidia (approved):

nemo_automodel/components/datasets/llm/formatting_utils.py
```diff
@@ -32,15 +75,37 @@ def _pad_to_seq_length(sample, pad_token_id, seq_length):
...
 
 
+_warned_add_pad_token = set()
+
+
 def _add_pad_token(tokenizer):
     """Add pad token to tokenizer if not present."""
     pad_token_id = None
-    if getattr(tokenizer, "pad_token_id", None) is None:
+    if getattr(tokenizer, "pad_token_id", None) is None and not "no_pad_id" in _warned_add_pad_token:
```
oliverholworthy: This `_warned_add_pad_token` set is module-level, but it also gates the tokenizer.pad_token_id = tokenizer.eos_token_id assignment. On a second call with a different tokenizer that also has pad_token_id=None, the fallback won't execute, leaving pad_token_id as None. Should we only guard the `logger.warning()`, not the assignment itself, it the intent here is to only log the warning once.

nemo_automodel/_transformers/tokenization/nemo_auto_tokenizer.py
```diff
@@ -64,6 +98,24 @@ def from_pretrained(cls, pretrained_model_name_or_path, *args, add_bos_token=Tru
         tokenizer.__class__ = type(cls.__name__, (cls, base_tokenizer_cls), {})
         return tokenizer
 
+    def apply_chat_template(self, conversation, *args, **kwargs):
```
oliverholworthy: What's the connection between this apply_chat_template method and the pad/eos fix? If unrelated, would it make sense in a separate PR?

nemo_automodel/components/datasets/llm/formatting_utils.py
akoumpa: Yeah, I think so.

nemo_automodel/_transformers/tokenization/nemo_auto_tokenizer.py
akoumpa: Thanks @oliverholworthy , I've reverted this.

nemo_automodel/components/datasets/llm/formatting_utils.py
akoumpa: Thanks @oliverholworthy , I've fixed the issue.

PR #1349, opened 2026-02-20
fix: biencoder bidirectional masking and consistent attn_implementation default

oliverholworthy:
- Replace `_update_causal_mask` with `create_bidirectional_mask` in `LlamaBidirectionalModel` to fix bidirectional attention with newer transformers versions and non-flash attention backends
- Add `attn_implementation` option to the mining recipe so the attention backend can be configured at inference time
- Default `attn_implementation` to None (sdpa) instead of flash_attention_2 in NeMoAutoModelBiencoder

## Motivation

The old `_update_causal_mask` override had three problems:

1. Dead method on transformers >= 4.53: upstream removed `_update_causal_mask` from LlamaModel, so our override was never called and the model silently fell back to causal masking with sdpa/eager backends.
2. Wrong mask format for SDPA/eager: it returned None or passed through the raw 2D mask without expanding to 4D float, causing shape/dtype errors in scaled_dot_product_attention.
3. No backend awareness: Flash Attention 2 expects a 2D (batch, seq_len) mask, while SDPA/eager need a 4D (batch, 1, seq_len, seq_len) float mask. The old code didn't distinguish between them.

The new _create_bidirectional_mask:
- Uses native `create_bidirectional_mask` from transformers.masking_utils when available (transformers >= 5.0)
- Falls back to backend-aware logic for older transformers: 2D passthrough for flash_attention_2, _prepare_4d_attention_mask for SDPA/eager

**Pre checks**:

### DIFF
--- a/tests/unit_tests/models/biencoder/test_llama_bidirectional_model.py
+++ b/tests/unit_tests/models/biencoder/test_llama_bidirectional_model.py
@@ -96,67 +96,47 @@ def test_llama_bidirectional_model_init_and_mask():
         vocab_size=128, hidden_size=32, num_hidden_layers=1, num_attention_heads=1, intermediate_size=64, pad_token_id=0
     )
     model = LlamaBidirectionalModel(cfg)
-#    assert all(getattr(layer.self_attn, "is_causal", True) is False for layer in model.layers)
-    assert model._update_causal_mask(None) is None
+    model.eval()
+
+    # All attention layers should be non-causal
+    assert all(getattr(layer.self_attn, "is_causal", True) is False for layer in model.layers)
+
+    # Forward with padding mask produces valid output
+    input_ids = torch.randint(0, cfg.vocab_size, (1, 3))
     mask = torch.tensor([[1, 1, 0]])
-    input_tensor = torch.randn(1, 3, 32)
-    out_mask = model._update_causal_mask(mask, input_tensor=input_tensor)
-    assert out_mask is not None and out_mask.dim() == 4
-    all_ones_out = model._update_causal_mask(torch.ones_like(mask), input_tensor=input_tensor)
-    assert all_ones_out is not None and (all_ones_out == 0.0).all()
+    out = model(input_ids=input_ids, attention_mask=mask)
+    assert out.last_hidden_state is not None and out.last_hidden_state.shape == (1, 3, 32)
+
+    # Forward without attention mask also works
+    out_no_mask = model(input_ids=input_ids)
+    assert out_no_mask.last_hidden_state is not None and out_no_mask.last_hidden_state.shape == (1, 3, 32)
 
 
-def test_update_causal_mask_equivalence_with_old_impl():
-    """The old _update_causal_mask used a data-dependent `(mask == 0).any()` branch
-    that is incompatible with torch.export / ONNX.  The new implementation delegates
-    to _prepare_4d_attention_mask.  This test checks semantic equivalence: the same
-    positions are masked/unmasked when the 4D output is applied as an additive bias
-    to attention scores."""
-
-    def old_update_causal_mask(attention_mask):
-        """Original implementation (pre-ONNX fix)."""
-        if attention_mask is not None and (attention_mask == 0.0).any():
-            return attention_mask
-        return None
-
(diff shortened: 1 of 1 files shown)


nemo_automodel/_transformers/auto_model.py
```diff
@@ -734,7 +734,7 @@ def from_pretrained(
         share_encoder: bool = True,
         pooling: str = "avg",
         l2_normalize: bool = True,
-        attn_implementation: str = "flash_attention_2",
+        attn_implementation: Optional[str] = None,
```
adil-a: why shouldn't we use FA2 by default? we leave mem savings and speed on the table otherwise

nemo_automodel/_transformers/auto_model.py
adil-a, line 861: we have retry methods if fa2 doesn't work here btw

nemo_automodel/components/models/llama_bidirectional/model.py
```diff
@@ -33,6 +33,16 @@
 from transformers.processing_utils import Unpack
 from transformers.utils import TransformersKwargs
 
+# Check if native create_bidirectional_mask exists (transformers >= 5.0)
+try:
```
adil-a: can we enforce v5 now? automodel requires it so I think this code block is unnecessary

nemo_automodel/components/models/llama_bidirectional/model.py
```diff
@@ -94,15 +104,45 @@ def __init__(self, config: LlamaConfig):
...
+                tokens and 0 for padding, or None.
+
+        Returns:
+            A 4D float mask for sdpa/eager, a 2D mask for flash_attention_2,
+            or None when no masking is needed.
+        """
+        if attention_mask is None:
+            return None
+
+        if _HAS_NATIVE_BIDIRECTIONAL_MASK:
```
adil-a: same as above, let's just enforce v5 and use this

nemo_automodel/recipes/biencoder/mine_hard_negatives.py
```diff
@@ -69,6 +69,8 @@
     # Model loading parameters (loaded directly, not from config)
     "model_name_or_path": None,  # Required: path to model checkpoint
     "tokenizer_name_or_path": None,  # Optional: defaults to model_name_or_path
+    # Attention implementation for model loading
+    "attn_implementation": None,  # None = use model default; "sdpa", "flash_attention_2", "eager"
```
adil-a, line 73: same comment as above as auto_model.py

nemo_automodel/components/models/llama_bidirectional/model.py
oliverholworthy: this code I believe ends up in the checkpoint we save to disk, so while the automodel library my dependon on >=5, an open question may be whether the checkpoint we save we want to support earlier versions

nemo_automodel/_transformers/auto_model.py
oliverholworthy: For consistency i've changed to `DEFAULT_ATTN_IMPLEMENTATION` to match the other classes in this file, which uses `flash_attention_2` if the flash_attn package is installed (falling back to sdpa otherwise).

nemo_automodel/components/models/llama_bidirectional/model.py
oliverholworthy: We can revisit the checkpoint portability question in a follow-up, For now I've changed to a direct import as suggested.

nemo_automodel/components/models/llama_bidirectional/model.py
oliverholworthy: updated

nemo_automodel/_transformers/auto_model.py
oliverholworthy: In what scenario would this Exception handler be reached?

nemo_automodel/recipes/biencoder/mine_hard_negatives.py
oliverholworthy: This is intentionally None, since the Model implementation determines it's own default. We only pass a value through to the model class if this is not None

PR #1356, opened 2026-02-23
refactor: extract initialize_model_weights from load_base_model

hemildesai:
## Summary
- Extract weight initialization logic from `Checkpointer.load_base_model` into a new `Checkpointer.initialize_model_weights` static method, so `from_config` (which skips checkpoint loading) can still materialize meta-device parameters and call `initialize_weights`
- Move `_init_peft_adapters` into `initialize_model_weights` via a new `peft_init_method` parameter, keeping PEFT adapter init alongside weight init regardless of checkpoint loading path
- Add `need_post_shard_init` block in `apply_model_infrastructure` that calls `initialize_model_weights` after parallelisms are applied, decoupled from the checkpoint loading condition
- Make `from_config` accept `load_base_model` as a kwarg (defaults to `False`) for flexibility

## Test plan
- [x] Added 11 new unit tests for `initialize_model_weights` (materialization, HF init reset, Nemotron skip, peft_init_method forwarding, signature check)
- [x] Added 3 tests for the `load_before_shard` path (call ordering, single init call, no peft_init_method in load_base_model)
- [x] Added 1 test for peft_init_method forwarding through `apply_model_infrastructure`
- [x] Added 5 tests for post-shard init (from_config, from_pretrained, non-meta, MegatronFSDP skip)
- [x] Added 2 tests for `from_config` load_base_model kwarg forwarding
- [x] All 62 tests pass (`pytest tests/unit_tests/checkpoint/test_checkpointing.py tests/unit_tests/_transformers/test_infrastructure.py`)

### DIFF
--- a/nemo_automodel/_transformers/auto_model.py
+++ b/nemo_automodel/_transformers/auto_model.py
@@ -573,7 +573,7 @@ class _BaseNeMoAutoModelClass(_BaseAutoModelClass):
             peft_config=peft_config,
             fp8_config=fp8_config,
             compile_config=compile_config,
-            load_base_model=False,
+            load_base_model=kwargs.pop("load_base_model", False),
             **kwargs,
         )
 
--- a/nemo_automodel/_transformers/infrastructure.py
+++ b/nemo_automodel/_transformers/infrastructure.py
@@ -414,12 +414,12 @@ def apply_model_infrastructure(
     checkpoint_already_loaded = False
     if load_before_shard:
         lora_a_init = getattr(peft_config, "lora_A_init", None)
+        checkpointer.initialize_model_weights(model, device, peft_init_method=lora_a_init)
         checkpointer.load_base_model(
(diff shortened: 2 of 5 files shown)


nemo_automodel/_transformers/auto_model.py
```diff
@@ -573,7 +573,7 @@ def from_config(
             peft_config=peft_config,
             fp8_config=fp8_config,
             compile_config=compile_config,
-            load_base_model=False,
+            load_base_model=kwargs.pop("load_base_model", False),
```
adil-a, line 576: why do we need this? if we want to load a base model then what stops us from going the .from_pretrained route?

hemildesai: For models like Deepseek v3.2 the config class isn't available in transformers so it errors out if you use .from_pretrained

Issue #1364, opened 2026-02-24
NemotronV3 EP training deadlocks with variable-length token counts across ranks


ShiftyBlock:
**Describe the bug**

Expert Parallelism (EP) training deadlocks on the first forward pass when ranks receive different token counts (due to variable length sequence packing). The deadlock is at this line: DTensor.from_local(x, placements=[Shard(0)]). The root cause is DTensor.from_local with Shard(0) computes logical_dim_size = local_size × world_size independently per rank without cross-rank communication, so ranks disagree on buffer sizes for the subsequent all-gather.
The deadlock results in a timeout error like this: [Rank 0] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=2, OpType=COALESCED, NumelIn=18720, NumelOut=149760, Timeout(ms)=600000) ran for 600007 milliseconds before timing out.

Full log: https://pastebin.com/Y6re6g3S

**Steps/Code to reproduce bug**

Please list *minimal* steps or code snippet for us to be able to reproduce the bug.

Run EP training with variable-length input data (e.g. SQuAD QA fine-tuning with sequence packing):
torchrun --nproc-per-node=8 --nnodes=1 \
  nemo_automodel/recipes/llm/train_ft.py \
  --config nemotron_ep8_train.yaml

Link to nemotron_ep8_train.yaml: https://pastebin.com/DWDtgm4Z
This config has been tested and verified on an 8xH100 node.

To reproduce faster (fail in ~30s instead of 10min), add debug flags:
NCCL_DEBUG=INFO TORCH_DISTRIBUTED_DEBUG=DETAIL TORCH_NCCL_ASYNC_ERROR_HANDLING=1 \
TORCH_NCCL_BLOCKING_WAIT=1 torchrun --nproc-per-node=8 --nnodes=1 \
  nemo_automodel/recipes/llm/train_ft.py --config nemotron_ep8_train.yaml

**Expected behavior**

EP training should handle variable-length token counts across ranks without deadlocking. 

**Additional context**

qiaochuz-nv:
## QA RCCA Analysis

**Fix PR:** #1365
**Root Cause:** Expert Parallelism (EP) training deadlocked when ranks received different token counts from variable-length sequence packing — all-to-all collective hung waiting for mismatched buffer sizes.

**Trigger Config:** EP + variable-length sequence packing

**Coverage Assessment:**

| Dimension | Bug Config | Nearest Test | Match |
[...]

PR #1365, opened 2026-02-24
fix: EP collective deadlock with variable-length token counts

ShiftyBlock:
# What does this PR do ?

Fixes NCCL communication deadlocks when using Expert Parallelism. 

# Changelog

- Replace DTensor.from_local(x, [Shard(0)]).full_tensor() with all_gather for the max length, then padding to the max length, gather, trim padding.
- Replace DTensor Partial→Shard(0) reduce-scatter with all_reduce + narrow to support original per-rank token boundaries and resolve "RuntimeError: Backend nccl does not support reduce_scatter_tensor_coalesced" error
- Add gradient anchor (y + x * 0.0) so all ranks participate in backward collectives regardless of routing sparsity

**Pre checks**:

# Additional Information

## Three independent failure points
Each fix is independently necessary — removing any one causes a deadlock or crash:

- Input all-gather: DTensor.full_tensor() deadlocks because each rank expects different chunk sizes from peers.
- Output redistribution: Consider the case of 4 ranks that own [18, 24, 30, 12] tokens respectively. DTensor Partial→Shard(0) even-splits tokens (ex: 84/4=21 each) instead of respecting original ownership, causing data corruption or the reduce_scatter_tensor_coalesced unsupported error.
- Backward participation: Ranks with no routed tokens skip the backward all-reduce, which causes deadlocks because all ranks must participate in the all-reduce 

## Validation
Baseline (deadlocks on first forward): https://pastebin.com/UBgHF8mf
Fixed (loss 4.69→0.06 over 100 steps): https://pastebin.com/f2AMEpqd
Config used: https://pastebin.com/DWDtgm4Z (8×H100, EP=8, SQuAD fine-tuning)

## Communication Overhead
The fix uses all_reduce + narrow instead of reduce_scatter because reduce_scatter requires even splits. Measured overhead on 8×GPU NVLink with torch.cuda.Event timing:

Op | Time/layer
-- | --
all_reduce (fix) | 0.121 ms
reduce_scatter (theoretical) | 0.071 ms
Overhead | 0.050 ms/layer
Across 20 MoE layers | ~1 ms/step
% of step time | ~0.2%

- Related to #1364 

Thanks to @AlexLuu1 for reporting the post-adapted NanoV3 training issue during validation that motivated this fix.

### DIFF
--- a/nemo_automodel/components/moe/experts.py
+++ b/nemo_automodel/components/moe/experts.py
@@ -38,22 +38,6 @@ from nemo_automodel.components.moe.megatron.moe_utils import (
 from nemo_automodel.components.moe.megatron.token_dispatcher import MoEFlexTokenDispatcher, TokenDispatcherConfig
 
 # ── EP variable-length collective helpers ──
-_ep_collective_group_cache: dict[tuple[int, ...], dist.ProcessGroup] = {}
-
-
-def _get_or_create_ep_collective_group(ep_mesh: DeviceMesh) -> dist.ProcessGroup:
-    """Return a dedicated ProcessGroup for EP collectives.
-
-    Using a separate group from FSDP avoids cross-feature collective ordering
-    interference when EP and FSDP both communicate over world-size groups.
-    """
-    mesh_ranks = tuple(int(r) for r in ep_mesh.mesh.flatten().tolist())
-    cached = _ep_collective_group_cache.get(mesh_ranks)
-    if cached is not None:
-        return cached
-    pg = dist.new_group(ranks=list(mesh_ranks))
-    _ep_collective_group_cache[mesh_ranks] = pg
-    return pg
 
 
 class _AllGatherConcatVarlenFn(Function):
@@ -314,7 +298,7 @@ class GroupedExperts(nn.Module):
 
         # EP variable-length all-gather
         if ep_size > 1:
-            ep_group = _get_or_create_ep_collective_group(ep_mesh)
+            ep_group = ep_mesh.get_group()
             local_num_tokens = x.size(0)
 
             # Exchange per-rank token counts


nemo_automodel/components/moe/experts.py
```diff
@@ -34,6 +37,62 @@
...
+_ep_collective_group_cache: dict[tuple[int, ...], dist.ProcessGroup] = {}
+
+
+def _get_or_create_ep_collective_group(ep_mesh: DeviceMesh) -> dist.ProcessGroup:
+    """Return a dedicated ProcessGroup for EP collectives.
+
+    Using a separate group from FSDP avoids cross-feature collective ordering
+    interference when EP and FSDP both communicate over world-size groups.
+    """
+    mesh_ranks = tuple(int(r) for r in ep_mesh.mesh.flatten().tolist())
```
akoumpa: Thanks @ShiftyBlock , but I wonder if it would make sense to use the ep_mesh's group instead of creating a new one?

davidyang05: Hi @akoumpa, good call, updated to use ep_mesh.get_group(). Many thanks for your suggestions!
Just ran a sanity run to make sure training still worked: https://pastebin.com/k630jqSM

hemildesai:
Hi @ShiftyBlock, I tested this fix on fixed seq length Moonlight 16b and the convergence looks good but there's a ~12.5% regression in TPS. See image below:

Screenshot 2026-02-24 at 1 30 51 PM

Blue is main, Red is this branch.

I think this is fine for now as the `torch` dispatcher backend is more for correctness and maximum compatibility. I will create an issue to add a torch native all to all dispatcher backend for better native ep comms.

hemildesai:
Thanks for the suggestion @ShiftyBlock. I think its fine to just have a single path for torch native all gather dispatcher for now as it is mostly meant for correctness, comparison and for hardware architectures that don't support DeepEP, otherwise it is recommended to use the DeepEP dispatcher for best perf.

Moreover, #1375 should support both fixed and varlen sequences and also use less memory than all gather, so we can provide a more performant torch native option there. Happy to accept a PR for #1375 if you are interested, https://github.com/pytorch/torchtitan/blob/40327250468861f43b8d1437ecf68112709125b4/torchtitan/distributed/expert_parallel.py#L102-L181 should serve as a good reference point.

PR #1380, opened 2026-02-25
feat: support loading biencoder datasets directly from HuggingFace Hub

oliverholworthy:
- Updates `data_dir_list` parameter to `make_retrieval_dataset` that accepts `hf://` URIs, enabling biencoder training directly from HuggingFace 
- Supports subset(s) (`hf://nvidia/embed-nemotron-dataset-v1/FEVER`)
- Mixing HF and local sources in the same training run is supported
- Fully backwards compatible: existing `data_dir_list` configs work unchanged

## Motivation

Training a biencoder on `nvidia/embed-nemotron-dataset-v1` currently requires a manual data preparation step: run `data_preparation.py` to download from HF, resolve `ids_only` text, and write local JSON+Parquet files, then point the YAML config at those local files.

With this change, users can write:

```yaml
dataloader
  dataset:
    data_dir_list:
      - hf://nvidia/embed-nemotron-dataset-v1/FEVER
```
 
 and have the loader fetch, resolve, and load everything on the fly.

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo_automodel/components/datasets/llm/retrieval_dataset.py
+++ b/nemo_automodel/components/datasets/llm/retrieval_dataset.py
@@ -283,7 +283,19 @@ def _list_hf_subsets(repo_id: str) -> List[str]:
 
 
 def _load_hf_subset(repo_id: str, subset: str):
-    """Load a single HF subset and return ``(normalized_data_list, CorpusInfo)``."""
+    """Load a single HF subset and return ``(normalized_data_list, CorpusInfo)``.
+
+    Note:
+        The direct ``hf://`` path currently expects the Automodel retrieval schema:
+        - ``{subset}/dataset_metadata.json`` with ``corpus_id`` metadata
+        - ``{subset}_corpus`` split with corpus columns like ``id`` and ``text``
+        - ``{subset}`` split with query columns like ``question`` and ``pos_doc``
+
+        FEVER and SyntheticClassificationData from
+        ``nvidia/embed-nemotron-dataset-v1`` are examples that follow this layout.
+        Datasets with different structures should use a custom adapter/preprocessor
+        before calling this loader.
+    """
 
     # 1. Download dataset_metadata.json
     meta_path = hf_hub_download(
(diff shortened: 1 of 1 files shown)


examples/biencoder/llama3_2_1b_biencoder.yaml
```diff
@@ -53,9 +53,7 @@ dataloader:
   dataset:
     _target_: nemo_automodel.components.datasets.llm.make_retrieval_dataset
     data_dir_list:
-      - /adasif/retriever_models_research/training_datasets/nqsh_shuffled_50k.json
-      - /adasif/retriever_models_research/training_datasets/mldr_en_perc95_small.json
-      - /adasif/retriever_models_research/training_datasets/miracl_train_es_llama3_1b_4m_512len.json
+      - hf://nvidia/embed-nemotron-dataset-v1/FEVER
```
adil-a, line 56: IIRC @ybabakhin mentioned FEVER and one other dataset during the call (forgetting the name). Can we include that one as well?

nemo_automodel/components/datasets/llm/retrieval_dataset.py
adil-a, line 1: Can we include some comments here that will tell the user that the preprocessing is only for FEVER + whichever other dataset would work? Just so any dev that might want to make amends can get a quicker understanding that they'd need to implement a custom preprocessor for their dataset.

examples/biencoder/llama3_2_1b_biencoder.yaml
oliverholworthy: The other two that will work are:

- [SyntheticClassificationData](https://huggingface.co/datasets/nvidia/embed-nemotron-dataset-v1/tree/main/SyntheticClassificationData)
- [MIRACL](https://huggingface.co/datasets/nvidia/embed-nemotron-dataset-v1/tree/main/MIRACL)

I've added SyntheticClassificationData here too, however MIRACL is much larger (1.34 GB) vs FEVER which is 90 MB, which might be a bit too large for a quick-start example?

nemo_automodel/components/datasets/llm/retrieval_dataset.py
oliverholworthy: Great suggestion @adil-a . I've extended the docstrings and error messages to make it clearer what format is currently supported

examples/biencoder/llama3_2_1b_biencoder.yaml
adil-a: Good point. Yes, let's stick to SyntheticClassificationData then :) Thank you!

adil-a (approved):
Thanks for addressing the comments. Approved :)

PR #1410, opened 2026-02-27
fix: combined projection bias loading and rms_norm numerical instability

ZhiyuLi-Nvidia:
# What does this PR do ?

* combine projection bias is not loaded correctly 
  - failed model: affect models with qkv projection bias including qwen. And llama is not affected 
  - Root cause: When loading checkpoints via DCP for models with biases (e.g. Qwen2 with TP), the state dict adapter all-gathers 1D bias DTensors from Shard(0) to Replicate so reshape/split/cat can operate on the full data — but never restores * the original placement. load_state_dict then copies a Replicate tensor into a Shard(0) parameter, silently truncating the bias values. This only affects 1D bias tensors (e.g. qkv_proj.bias); 2D weight tensors use Shard(1) and are never gathered.
  - Fix: Introduce a _gather_1d_bias / _restore_1d_bias helper pair that explicitly bracket bias-only operations in to_hf() and from_hf(). The interleave/deinterleave methods remain pure tensor math with no DTensor awareness. Weight paths are completely untouched.

* instability with te norm:
  - failed model: both 2 models are very sensitive to te norm
  - root cause: te norm seems to be implemented with bf16 without fp32 upcast. We observed ~5% max numerical differences for a single norm op; these differences accumulate and lead to a completely different initial loss.
  - Fix: use torch norm for Qwen2.5-7B Llama3.2-3B 

## Convergence Test

* Qwen2.5-7B: https://wandb.ai/nvidia/automodel-dev-zhiyul?nw=ouoluvgy6a
* llama3.2-3B: https://wandb.ai/nvidia/automodel-dev-zhiyul/workspace?nw=7nq9jmggxxm

# Changelog

- fix qkv projection bias loading (by assigning the right placement)
- use torch te norm for Qwen2.5-7B Llama3.2-3B 

**Pre checks**:

- [] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md)
- [] Did you write any new necessary tests?

# Additional Information

- Related to # (issue)


examples/llm_finetune/llama3_2/llama_3_2_3b_instruct_squad_peft.yaml
```diff
@@ -33,6 +33,9 @@ model:
   _target_: nemo_automodel.NeMoAutoModelForCausalLM.from_pretrained
   pretrained_model_name_or_path: meta-llama/Llama-3.2-3B-Instruct
   torch_dtype: bf16
+  backend:
+    _target_: nemo_automodel.components.models.common.BackendConfig
+    rms_norm: "torch"
```
akoumpa, line 38: Hi @ZhiyuLi-Nvidia , thanks for making this, would it be possible to root-cause the instability? For example, is it some version issue in TE or something else? Thank you.

ZhiyuLi-Nvidia: Created the issue https://github.com/NVIDIA-NeMo/Automodel/issues/1432 for future investigation and tracking. Will address is in future PRs.

akoumpa (approved):

PR #1411, opened 2026-02-27
feat: Add native Comet ML experiment tracking

LoganVegnaSHOP:
## Summary

- Add `CometLogger` class in `nemo_automodel/components/loggers/comet_utils.py` — mirrors the existing `MLflowLogger` pattern
- Wire Comet logging into `TrainFinetuneRecipeForNextTokenPrediction`:
  - Training metrics (loss, TPS, grad_norm, lr, mem) via `log_train_metrics()`
  - Validation metrics (val_loss) via `log_val_metrics()`
  - MoE load balance metrics via `_log_moe_metrics()`
- Enable by adding a `comet:` block to the training YAML config:

```yaml
comet:
  project_name: "my-project"
  experiment_name: "my-run"      # optional, auto-generated from model name
  workspace: "my-workspace"      # optional, uses COMET_WORKSPACE env var
  api_key: null                  # optional, uses COMET_API_KEY env var
  tags: ["finetune", "llama"]    # optional
```

## Motivation

Currently, Comet ML users must rely on Comet's wandb auto-patcher, which does not
reliably intercept `wandb.log()` calls because NeMo gates all wandb logging behind
`if wandb.run is not None` — requiring a valid wandb API key and active run even when
wandb is only used as a bridge to Comet. Native Comet support (matching the existing
MLflow integration pattern) eliminates this fragile dependency.

## Test plan

- [ ] Run a fine-tuning job with `comet:` config and verify metrics appear in Comet dashboard
- [ ] Verify training metrics (loss, TPS, grad_norm, lr, mem) are logged at `log_remote_every_steps` frequency
- [ ] Verify validation metrics (val_loss) are logged on each validation step
- [ ] Verify MoE load balance metrics are logged (on MoE models)
- [ ] Verify that omitting `comet:` config has no effect (backward compatible)
- [ ] Verify `comet_ml` not installed raises clear ImportError

### DIFF
--- a/tests/unit_tests/training/test_train_ft_mlflow_logging.py
+++ b/tests/unit_tests/training/test_train_ft_mlflow_logging.py
@@ -67,13 +67,26 @@ def test_log_train_metrics_calls_mlflow(monkeypatch):
     recipe.metric_logger_train = types.SimpleNamespace(log=lambda x: None)
     mlflow_mock = Mock()
     recipe.mlflow_logger = types.SimpleNamespace(log_metrics=mlflow_mock)
+    recipe.comet_logger = None
 
     # Avoid cuda calls on environments without GPUs
     import torch.cuda
 
     monkeypatch.setattr(torch.cuda, "reset_peak_memory_stats", lambda: None, raising=False)
 
-    log_data = MetricsSample(step=7, epoch=1, metrics={"loss": 1.23, "grad_norm": 0.5, "lr": 1e-3, "mem": 0.1, "tps": 10.0, "tps_per_gpu": 5.0, "num_label_tokens": 42})
+    log_data = MetricsSample(
+        step=7,
+        epoch=1,
+        metrics={
+            "loss": 1.23,
+            "grad_norm": 0.5,
+            "lr": 1e-3,
+            "mem": 0.1,
+            "tps": 10.0,
+            "tps_per_gpu": 5.0,
+            "num_label_tokens": 42,
+        },
+    )
     recipe.log_train_metrics(log_data)
 
     mlflow_mock.assert_called_once()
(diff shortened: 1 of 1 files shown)


nemo_automodel/components/loggers/comet_utils.py
```diff
@@ -0,0 +1,156 @@
...
+    comet_config = cfg.get("comet", {})
+    if not comet_config:
+        raise ValueError("Comet configuration not found in config")
+
+    project_name = comet_config.get("project_name", "automodel-experiment")
+    workspace = comet_config.get("workspace", None)
+    api_key = comet_config.get("api_key", None)
+    experiment_name = comet_config.get("experiment_name", "")
+    tags = list(comet_config.get("tags", []))
+    auto_metric_logging = comet_config.get("auto_metric_logging", False)
```
hemildesai: I am wondering if its better to raise an error if the key is not present instead of using defaults for some of these like project name, api keys, etc. Wdyt?

LoganVegnaSHOP: Yeah that makes sense, certainly for project name

akoumpa:
Hi @LoganVegnaSHOP , thanks a lot for the feature.

Unfortunately, the linter [fails](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/22617479876/job/65533876982?pr=1411) and i cannot push to your branch, would you mind taking a look (uv run ruff format/check --fix <file.py>) ?

thank you.

hemildesai:
/ok to test [95d8842](https://github.com/NVIDIA-NeMo/Automodel/pull/1411/commits/95d88426a4cc89451cd2067014bff9012e55e2fb)

PR #1419, opened 2026-03-01
cp: torch mm expert backend

HuiyingLi:
cp torch_mm backend from main.


nemo_automodel/components/moe/layers.py
hemildesai: You might want to remove this class method and use the function at the top throughout.

HuiyingLi: @hemildesai  thank you! Fixed!

PR #1441, opened 2026-03-03
feat: context-parallel with nemotron v3

adil-a:
## Summary

Adds context parallelism (CP) support for NemotronV3 hybrid models (interleaved attention + Mamba2 + MoE). NemotronV3 requires a dual-path CP strategy because attention layers and Mamba/SSM layers have fundamentally different parallelism requirements:

- **Attention layers**: Use TE's sequence-parallel CP (p2p or allgather), splitting the sequence across ranks
- **Mamba/SSM layers**: Use a hidden-parallel strategy (all-to-all), where each rank processes the full sequence but only a subset of heads

### Supported configurations

| Format | Backend | Sequence Packing | CP Support |
|--------|---------|-------------------|------------|
| THD | TE | Yes | ✅ DualChunkSwap via TE p2p |
| THD | TE | No | ✅ DualChunkSwap via TE p2p |
| BSHD | SDPA | No | ✅ via PyTorch `context_parallel()` |

### Key changes

**Mamba hidden-parallel CP** (`mamba_cp.py` — new file):
- `MambaContextParallel` class orchestrating CP for Mamba2 mixer layers
- Autograd-aware all-to-all (`_all_to_all_cp2hp` / `_all_to_all_hp2cp`) to redistribute between sequence-sharded and hidden-sharded layouts
- DualChunkSwap undo/redo (`_undo_attention_load_balancing` / `_redo_attention_load_balancing`) so SSM kernels see tokens in sequential order
- THD packed sequence support via `_deinterleave_packed_seqs` / `_reinterleave_packed_seqs` for correct token ordering after all-to-all with multiple packed sequences
- Parameter slicing (conv1d, dt_bias, A_log, D) returning views so gradients flow to full parameters

**NemotronV3 model changes** (`layers.py`, `model.py`):
- Attention layers wired to TE CP via `DotProductAttention.set_context_parallel_group()`
- Mamba layers wired via `MambaContextParallel` attached as `mixer.cp`
- `seq_idx` computation from `cu_seqlens` for correct SSM state resets at sequence boundaries
- THD format support in attention (cu_seqlens, max_seqlen, qkv_format passthrough)

**Bug fixes**:
- Guard `padding_mask` for `pad_token_id=None` (NemotronV3 tokenizer returns `None`) in `thd_utils.py`, `cp_utils.py`, `model_utils.py` — `tensor == None` returns Python `False` (a scalar bool), not a tensor
- Fix `cu_seqlens` global/local mismatch in Mamba CP — batch cu_seqlens are GLOBAL (pre-TE-partitioning), so derive local as `cu_seqlens // cp_size` instead of erroneously computing `cu_seqlens * cp_size`
- Fix `seq_idx` in mixer layer — remove erroneous `* cp_size` since cu_seqlens is already global
- Fix `padding_idx=None` crash in `pack_dataset` when packing with CP > 1 (`getattr(tokenizer, "pad_token_id", 0)` returns `None` when the attribute exists but is `None`)

**Test infrastructure**:
- Consolidated `run_nemotron_v3_attention_cp.py` into `run_attention_cp.py` as `--model_type nemotron_v3` with `--configs` arg for multi-format testing (bshd_te, thd_te, bshd_sdpa)
- Fixed pre-existing test bugs in Qwen3MoE and DeepSeekV3 CP tests: `rope_theta` access via `get_rope_config()`, missing `cp_size`/`cp_rank` params, missing param grad all-reduce, DeepSeek `freqs_cis` format
- New `run_hybrid_nemotron_v3_cp.py` — end-to-end hybrid model test (4 configs)
- New `run_mamba_cp.py` — Mamba2Mixer layer test (4 configs including THD packed)
- Added `L2_Context_Parallel` to CI workflow matrix (`cicd-main.yml`)
- Fixed Qwen3.5MoE shell script (was incorrectly using pytest on a non-pytest script)

### THD format and sequence packing with CP

In THD format, the batch is flattened to a 1D `[T, H]` tensor with `cu_seqlens` tracking sequence boundaries. For CP with DualChunkSwap:

1. TE partitions tokens across CP ranks using `thd_get_partitioned_indices`
2. After all-to-all in Mamba CP, tokens are in rank-major order — `_deinterleave_packed_seqs` rearranges to sequence-major so each sequence's tokens are contiguous
3. `_undo_attention_load_balancing` permutes from DualChunkSwap order `[c0, c3, c1, c2]` back to sequential `[c0, c1, c2, c3]` for the SSM kernel
4. After SSM, the reverse transforms (`_redo` + `_reinterleave`) restore the layout for attention

Note: With CP > 1 and sequence packing, each sequence is padded to be divisible by `2 * cp_size` before packing. This changes pack composition compared to CP=1 (fewer sequences per pack due to padding overhead), resulting in different `num_label_tokens` per step — this is expected behavior, not a correctness issue.

### Convergence results

https://api.wandb.ai/links/nvidia-nemo-fw-public/f7qpoo9i


nemo_automodel/components/distributed/thd_utils.py
```diff
@@ -127,7 +127,9 @@ def process_input_for_thd(
         # Pass cu_seqlens_padded here since CP doesn't support padding between sequences correctly, the labels or loss mask will ensure that loss is computed correctly.
         "cu_seqlens": cu_seqlens_padded,
         "labels": labels_thd,
-        "padding_mask": (input_ids_thd == padding_token_id),
+        "padding_mask": (input_ids_thd == padding_token_id)
+        if padding_token_id is not None
+        else torch.zeros_like(input_ids_thd, dtype=torch.bool),
```
hemildesai: Why this else? We should just error out here maybe?

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -509,7 +509,7 @@ def build_dataloader(
                     split=cfg_ds.split,
                     pack_size=packed_sequence_size,
                     max_packs=getattr(cfg_ps, "max_packs", None),
-                    padding_idx=getattr(tokenizer, "pad_token_id", 0),
+                    padding_idx=getattr(tokenizer, "pad_token_id", None) or 0,
```
hemildesai: Why this change? `getattr(tokenizer, "pad_token_id", 0)` should be the same as `getattr(tokenizer, "pad_token_id", None) or 0`

nemo_automodel/components/distributed/cp_utils.py
```diff
@@ -379,7 +379,9 @@ def _shard_thd_chunk_for_te(
         "cu_seqlens": cu_seqlens_padded.to(torch.int32).contiguous(),
         "max_seqlen": torch.tensor(max_seqlen).to(torch.int32).to(device=cu_seqlens_padded.device),
         "qkv_format": qkv_format,
-        "padding_mask": (batch["input_ids"] == padding_token_id).bool().contiguous(),
+        "padding_mask": (batch["input_ids"] == padding_token_id).bool().contiguous()
```
hemildesai: Same as the comment in thd_utils, maybe we should error out here?

nemo_automodel/components/distributed/thd_utils.py
adil-a: it's an issue for nemotron v3. the tokenizer doesn't have a padding token

nemo_automodel/recipes/llm/train_ft.py
adil-a: for nano v3 the tokenizer doesn't have a pad token (it's set to None).

getattr(tokenizer, "pad_token_id", 0) -> None

getattr(tokenizer, "pad_token_id", None) or 0 -> 0

nemo_automodel/components/distributed/thd_utils.py
adil-a: reverted.

nemo_automodel/recipes/llm/train_ft.py
adil-a: reverted

nemo_automodel/components/distributed/cp_utils.py
adil-a: reverted

hemildesai (approved):

PR #1449, opened 2026-03-04
feat: add reranker training

adil-a:
## Summary

Adds cross-encoder (reranker) training support alongside the existing bi-encoder pipeline. Cross-encoders jointly attend to query-document pairs, producing relevance scores that are significantly more accurate than bi-encoder similarity — a critical capability for retrieval reranking.

### Cross-Encoder Model & Training Recipe
- **`NeMoAutoModelCrossEncoder`** — New auto-model class with full infrastructure support (FSDP2, PEFT, kernel patching)
- **`CrossEncoderModel`** — Wraps any `ForSequenceClassification` backbone. Routes through `LlamaBidirectionalForSequenceClassification` for Llama models, falls back to HF `AutoModelForSequenceClassification` for all others
- **`TrainCrossEncoderRecipe`** — Extends `TrainBiEncoderRecipe` with cross-entropy loss on reshaped logits, training accuracy tracking, and validation with accuracy@1 and MRR metrics
- **`accuracy()` / `batch_mrr()`** — Pure-function metric utilities for training and validation

### Data Pipeline
- **`CrossEncoderCollator`** — Concatenates query+passage via configurable prompt template, tokenizes, pads. Compatible with `NeMoAutoTokenizer`
- **`flatten_bi_encoder_to_cross_encoder()`** — Transforms grouped bi-encoder data to flattened cross-encoder format. Same data files work for both model types via `model_type` config switch
- **`make_retrieval_dataset(model_type="cross_encoder")`** — Unified dataset factory supporting both bi-encoder and cross-encoder formats

### Architecture Simplification
- Removed dual-encoder (`lm_q`/`lm_p`) design with `share_encoder` — single backbone is simpler and aligns with modern embedding models
- Simplified `EncoderStateDictAdapter` (was `BiencoderStateDictAdapter`) — symmetric `model.` prefix strip/add
- **Any HF model as encoder backbone** — `AutoModel`/`AutoModelForSequenceClassification` fallback means users are not limited to custom bidirectional models
- Unified naming: `Biencoder` -> `BiEncoder`, `train_n_passages` -> `n_passages`

### CI & Testing
- **Functional test**: 2-GPU FSDP2 training (32 steps) + quality evaluation asserting finetuned model improves positive-pair scoring and achieves >=75% ranking accuracy
- **Unit tests**: `accuracy()`, `batch_mrr()`, `CrossEncoderCollator` (output shapes, labels, padding), `flatten_bi_encoder_to_cross_encoder()` (value validation), liger/SDPA retry for cross-encoder
- **`L2_Retrieval` CI job** added to `cicd-main.yml`

### Example
```yaml
model:
  _target_: nemo_automodel.NeMoAutoModelCrossEncoder.from_pretrained
  pretrained_model_name_or_path: meta-llama/Llama-3.2-1B
  num_labels: 1
  pooling: avg
dataloader:
  dataset:
    _target_: nemo_automodel.components.datasets.llm.make_retrieval_dataset
    model_type: cross_encoder
    n_passages: 5
  collate_fn:
    _target_: nemo_automodel.components.datasets.llm.CrossEncoderCollator
    rerank_max_length: 512
    prompt_template: "question:{query} \n \n passage:{passage}"
```

## Test plan
- [x] 2-GPU cross-encoder training completes without errors (32 steps, FSDP2)
- [x] Finetuned model pos-score improvement > 0 (observed: +9.35)
- [x] Finetuned ranking accuracy >= 75% (observed: 92.0%)
- [x] 22 unit tests pass (accuracy, batch_mrr, collator, flatten, retry)
- [x] `ruff check` clean on `nemo_automodel/`
- [x] CI `L2_Retrieval` job passes

### DIFF
--- a/tests/functional_tests/retrieval/L2_Cross_Encoder_Finetune.sh
+++ b/tests/functional_tests/retrieval/L2_Cross_Encoder_Finetune.sh
@@ -15,11 +15,9 @@
 
 set -exo pipefail
 
-COVERAGE_ARGS="--data-file=/workspace/.coverage --source=/workspace/ --parallel-mode"
-
 # Run the cross-encoder recipe with 2 GPUs (FSDP2).
 python -m torch.distributed.run --nproc_per_node=2 --nnodes=1 \
-    -m coverage run ${COVERAGE_ARGS} \
+    -m coverage run \
     -m nemo_automodel.recipes.retrieval.train_cross_encoder \
     --config \
     tests/functional_tests/retrieval/recipe_cross_encoder.yaml \
@@ -28,7 +26,7 @@ python -m torch.distributed.run --nproc_per_node=2 --nnodes=1 \
     --dataloader.dataset.data_dir_list $TEST_DATA_DIR/embedding_testdata/training.jsonl
 
 # Compare baseline vs finetuned cross-encoder checkpoint.
-python3 -m coverage run ${COVERAGE_ARGS} \
+python3 -m coverage run \
     tests/functional_tests/retrieval/compare_cross_encoder_models.py \
     $TEST_DATA_DIR/llama-nemotron-embed-1b-v2 \
     /workspace/output/cross_encoder_inline/checkpoints/epoch_0_step_31/ \


nemo_automodel/_transformers/encoder.py
```diff
@@ -69,71 +68,67 @@ def pool(last_hidden_states: torch.Tensor, attention_mask: torch.Tensor, pool_ty
...
 
-    def forward(self, input_dict: dict, encoder: str = "query") -> Optional[torch.Tensor]:
-        """Forward pass — delegates to encode().
-
-        Going through forward() ensures FSDP2 unshard hooks fire via __call__.
-        """
-        return self.encode(input_dict, encoder)
+    def forward(self, input_dict: dict = None, **kwargs) -> Optional[torch.Tensor]:
+        """Forward pass -- going through __call__ ensures FSDP2 unshard hooks fire."""
+        if self.task == "classification":
```
adil-a: rename to score

nemo_automodel/_transformers/retrieval.py
adil-a, line 1: rename lm_q -> model
ABC for the encoder

nemo_automodel/recipes/encoder/train_cross_encoder.py
```diff
@@ -0,0 +1,187 @@
...
+                    }
+                    labels = batch.pop("labels")
+
+                    model = self.model_parts[0]
+                    train_ctx = (
+                        torch.amp.autocast("cuda", dtype=torch.bfloat16) if torch.cuda.is_available() else nullcontext()
+                    )
+                    with train_ctx:
+                        outputs = model(**batch, return_dict=True)
+                        logits = outputs.logits.view(-1, self.train_n_passages)
```
rnyak: we basically want to use same value both for train_n_passage both for training and validation? in the [yaml file ](https://github.com/NVIDIA-NeMo/Automodel/blob/main/examples/biencoder/llama3_2_1b_biencoder.yaml#L59)we have `train_n_passage` defined twice under `dataloader` and `validation_dataloader`. what if user put different value for each.  according to line above we'd be still using train_n_passages defined under `dataloader` field in the yaml right?

examples/encoder/cross_encoder/llama3_2_1b.yaml
```diff
@@ -0,0 +1,142 @@
...
+
+# Optional: Uncomment to enable validation
+# validation_dataloader:
+#   _target_: torchdata.stateful_dataloader.StatefulDataLoader
+#   dataset:
+#     _target_: nemo_automodel.components.datasets.llm.make_retrieval_dataset
+#     model_type: crossencoder
+#     data_dir_list: training_datasets/validation.json
+#     data_type: eval
+#     train_n_passages: 5
```
rnyak: I think we better name this to val_n_passages? bcs in the current case it should be identical to train_n_passages, I wonder what happens if we set it to a different number than the training. I left a related comment about this below.

nemo_automodel/_transformers/encoder.py
rnyak: `cls_last` and `cls` they are identical, I think we can get rid of `cls_last` ?

examples/encoder/cross_encoder/llama3_2_1b.yaml
```diff
@@ -0,0 +1,142 @@
...
+# limitations under the License.
+
+# To run this recipe, please use the following command:
+# python examples/encoder/cross_encoder/finetune.py --config examples/encoder/cross_encoder/llama3_2_1b.yaml
+# Or with torchrun for multi-GPU:
+# torchrun --nproc-per-node=8 examples/encoder/cross_encoder/finetune.py --config examples/encoder/cross_encoder/llama3_2_1b.yaml
+
+seed: 42
+
+train_n_passages: 5
```
rnyak: why we duplicate this in three different places? isnt that risky? why not define only once.

examples/encoder/cross_encoder/llama3_2_1b.yaml
```diff
@@ -0,0 +1,142 @@
...
+eval_negative_size: 4
```
rnyak: I think we are overcomplicating the yaml. why we need the eval_neg_size? 

option 1:
- set train_n_passages only once in the yaml, use it everywhere, both for training and evaluation
option 2:
- set train_n_passages, and val_n_passages, and use train_n_passages for training process, the second for val process.

nemo_automodel/recipes/encoder/train_retriever_encoder.py
```diff
@@ -425,8 +427,8 @@ def _run_validation_epoch(self, val_dataloader):
                     query, passage = _unpack_qp(batch)
 
                     model = self.model_parts[0]
-                    q_reps = model(query, encoder="query")
-                    p_reps = model(passage, encoder="passage")
+                    q_reps = model(query)
+                    p_reps = model(passage)
 
                     n_passages = self.eval_negative_size + 1
```
rnyak: a bit unnecessary, related to my comment above. if we go with option 1 then  `n_passages` is basically same as  ` self.train_n_passages`.

examples/encoder/cross_encoder/llama3_2_1b.yaml
adil-a: I will make the change to support the second option

docs/guides/llm/retrieval-dataset.md
```diff
@@ -103,4 +103,4 @@ dataloader:
 ## Requirements
 
 - `pos_doc` must be **non-empty**.
-- If training requests negatives (e.g., `train_n_passages > 1`), `neg_doc` must contain **at least one** document (the loader will cycle negatives if you provide fewer than needed).
+- If training requests negatives (e.g., `n_passages > 1`), `neg_doc` must contain **at least one** document (the loader will cycle negatives if you provide fewer than needed).
```
rnyak: can you elaborate the "cycle negatives" part if I set n_passages=1? I think the code should give an error in that case unless we do in batch negative, i.e. using other samples in the same mini-batch as negatives.

nemo_automodel/components/datasets/llm/retrieval_dataset.py
```diff
@@ -452,7 +452,7 @@ def _transform_func(examples, num_neg_docs, corpus_dict, use_dataset_instruction
         if num_neg_docs > 0 and len(negatives) == 0:
             raise ValueError(
                 f"neg_doc is empty for example {i_example} but {num_neg_docs} negative(s) requested "
-                f"(train_n_passages > 1). Provide negatives or set train_n_passages=1."
+                f"(n_passages > 1). Provide negatives or set n_passages=1."
```
rnyak: same comment as above. if we set `n_passages=1` what happens? in-batch neg sampling?

examples/retrieval/cross_encoder/llama3_2_1b.yaml
```diff
@@ -76,16 +70,14 @@ dataloader:
 #     model_type: crossencoder
 #     data_dir_list: training_datasets/validation.json
 #     data_type: eval
-#     train_n_passages: 5
-#     eval_negative_size: 4
+#     n_passages: 5
```
rnyak, line 76: I saw that you now do `self.val_n_passages = self.cfg.get("validation_dataloader.dataset.n_passages", self.train_n_passages)`, I assume it should be fine a user can set train and val n_passages to different numbers.

nemo_automodel/components/datasets/llm/retrieval_dataset.py
```diff
@@ -666,7 +699,7 @@ def make_retrieval_dataset(
     dataset = make_retrieval_dataset(
         data_dir_list=args.data_dir_list,
         data_type=args.data_type,
-        train_n_passages=args.train_n_passages,
+        n_passages=args.n_passages,
         eval_negative_size=args.eval_negative_size,
```
rnyak, line 704: where it'd read `args.eval_negative_size`? do we still have this `eval_negative_size` in  the yaml file? I cannot find it.

currently validation crushes. bcs we dont have `eval_negative_size` in the yaml file and the default `eval_negative_size` is `10`.  I'd recommend not do to `eval_negative_size=args.eval_negative_size` instead we can do 
eval_negative_size=args.n_passages-1`, but this `n_passages` should come from validation_dataloader in the yaml file.

examples/retrieval/cross_encoder/llama3_2_1b.yaml
```diff
@@ -0,0 +1,134 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# To run this recipe, please use the following command:
+# python examples/encoder/cross_encoder/finetune.py --config examples/encoder/cross_encoder/llama3_2_1b.yaml
+# Or with torchrun for multi-GPU:
+# torchrun --nproc-per-node=8 examples/encoder/cross_encoder/finetune.py --config examples/encoder/cross_encoder/llama3_2_1b.yaml
+
+seed: 42
+
```
rnyak, line 21: we miss temperature arg here. what level we should add it? under `model:` below? otherwise it wont pickup and propagated [here](https://github.com/NVIDIA-NeMo/Automodel/blob/adil/retriever-feat/nemo_automodel/components/models/llama_bidirectional/model.py#L232), it will use the default.

nemo_automodel/components/models/llama_bidirectional/model.py
rnyak, line 240: same comment above. how is this temperature is populated/propagated here? I think it is not read from the yaml file if we dont add it under `model:`, is it?

examples/retrieval/cross_encoder/llama3_2_1b.yaml
```diff
@@ -0,0 +1,134 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# To run this recipe, please use the following command:
+# python examples/encoder/cross_encoder/finetune.py --config examples/encoder/cross_encoder/llama3_2_1b.yaml
+# Or with torchrun for multi-GPU:
+# torchrun --nproc-per-node=8 examples/encoder/cross_encoder/finetune.py --config examples/encoder/cross_encoder/llama3_2_1b.yaml
+
+seed: 42
```
rnyak, line 20: when I se the seed to `0` I am getting an assertion error.. we should be able to set the seed to `0`, right?

nemo_automodel/_transformers/retrieval.py
```diff
@@ -69,162 +66,120 @@ def pool(last_hidden_states: torch.Tensor, attention_mask: torch.Tensor, pool_ty
...
-        if self.l2_normalize:
-            embeds = F.normalize(embeds, dim=-1)
+        # HuggingFace consolidated checkpoint compatibility
+        self.name_or_path = os.path.dirname(inspect.getfile(type(self.model)))
+        self.state_dict_adapter = EncoderStateDictAdapter()
+        encoder_class_name = self.model.__class__.__name__
+        self.config.architectures = [encoder_class_name]
+        self.config.auto_map = {
+            "AutoModel": f"model.{encoder_class_name}",
+            "AutoConfig": f"model.{encoder_class_name.replace('Model', 'Config')}",
```
ybabakhin: This doesn't work for `LlamaBidirectionalForSequenceClassification` now. `config.json` ends up being:
```
{
  "architectures": [
    "LlamaBidirectionalForSequenceClassification"
  ],
  "attention_bias": false,
  "attention_dropout": 0.0,
  "auto_map": {
    "AutoConfig": "model.LlamaBidirectionalForSequenceClassification",
    "AutoModel": "model.LlamaBidirectionalForSequenceClassification"
  },
```

nemo_automodel/components/models/llama_bidirectional/model.py
ybabakhin, line 236: Is `pooling` defined in the cross-encoder config, and propagated here properly?

nemo_automodel/_transformers/retrieval.py
```diff
@@ -69,162 +66,120 @@ def pool(last_hidden_states: torch.Tensor, attention_mask: torch.Tensor, pool_ty
...
 
     @classmethod
     def build(
         cls,
         model_name_or_path: str,
-        share_encoder: bool = True,
+        task: str = None,
         pooling: str = "avg",
         l2_normalize: bool = True,
         trust_remote_code: bool = False,
```
rnyak, line 229: why trust_remote_code is False? do we need to set "trust_remote_code: true" in the yaml file?

nemo_automodel/components/datasets/llm/retrieval_collator.py
```diff
@@ -233,3 +236,44 @@ def _convert_dict_to_list(self, input_dict: Dict[str, List]) -> List[Dict[str, A
...
+            self.args = kwargs.pop("args")
+        super().__init__(*args, **kwargs)
+
+    def __call__(self, features: List[Dict[str, Any]]) -> "BatchEncoding":
+        query_examples = [x["question"] for x in features]
+        doc_examples = [x["doc_text"] for x in features]
+        num_labels = features[0].get("num_labels") if features else None
+
+        def format_text(q, p):
+            return f"question:{q} \n \n passage:{p}"
```
gabrielspmoreira: Should the cross-encoder prompt be configurable (e.g. in the YAML) ?

nemo_automodel/components/models/llama_bidirectional/model.py
gabrielspmoreira: The `LlamaBidirectionalForSequenceClassification` extends `LlamaForSequenceClassification` from transformers, adding a classification MLP in the output.
Many Auto regressive models in transformers don't add the corresponding "ForSequenceClassification" implementation. 
I am thinking that we could make it possible to turn any model into a CrossEncoder if we add the MLP layer on [CrossEncoderModel](https://github.com/NVIDIA-NeMo/Automodel/blob/5ac16e964a4c9981e6f886b5b2945d9375afb5f9/nemo_automodel/_transformers/encoder.py#L257 constructor and forward method). What do you think?

Note: MLP weights should be saved in the same checkpoint of the other model weights

examples/retrieval/bi_encoder/llama_embed_nemotron_8b/llama_embed_nemotron_8b.yaml
```diff
@@ -13,13 +13,12 @@
...
-# python examples/biencoder/finetune.py --config examples/biencoder/llama3_2_1b_biencoder.yaml
+# python examples/encoder/bi_encoder/finetune.py --config examples/encoder/bi_encoder/llama_embed_nemotron_8b/llama_embed_nemotron_8b.yaml
 # Or with torchrun for multi-GPU:
-# torchrun --nproc-per-node=8 examples/biencoder/finetune.py --config examples/biencoder/llama3_2_1b_biencoder.yaml
+# torchrun --nproc-per-node=8 examples/encoder/bi_encoder/finetune.py --config examples/encoder/bi_encoder/llama_embed_nemotron_8b/llama_embed_nemotron_8b.yaml
 
 seed: 125
 
-train_n_passages: 5
 temperature: 0.02
```
rnyak, line 22: is it picked up from this level? or it should be under `model:`?

nemo_automodel/_transformers/retrieval.py
adil-a: as discussed, we need to pass this in the yaml file explicitly if needed for the model

nemo_automodel/components/datasets/llm/retrieval_dataset.py
adil-a: fixed so that eval_negative_size defaults to None, and is then computed as n_passages - 1. n_passages are configurable from the yaml for train/val sets

nemo_automodel/components/models/llama_bidirectional/model.py
adil-a: fixed

nemo_automodel/_transformers/retrieval.py
adil-a: fixed, thank you

nemo_automodel/components/models/llama_bidirectional/model.py
adil-a: you can pass it in through the yaml under `model`

nemo_automodel/components/datasets/llm/retrieval_collator.py
adil-a: added

nemo_automodel/components/models/llama_bidirectional/model.py
adil-a: this is a good idea but I feel like we need to think more deeply about this and where the edge cases are. I'd suggest tackling this in a separate PR. I do like this idea nonetheless

examples/retrieval/bi_encoder/llama_embed_nemotron_8b/llama_embed_nemotron_8b.yaml
adil-a: please check the most recent yaml

PR #1461, opened 2026-03-05
fix: handle missing reset_parameters in Qwen3_5MoeBlock.init_weights()

### LINKED ISSUE #1430: [Qwen3.5 MoE] `Qwen3_5MoeBlock.init_weights()` crashes on HF `Qwen3_5MoeRMSNormGated` missing `reset_parameters()`
This issue was discovered while adding Qwen3.5-MoE support in NeMo-RL. https://github.com/NVIDIA-NeMo/RL/issues/2025

## Describe the bug

When initializing weights for the Qwen 3.5 MoE model, `Qwen3_5MoeBlock.init_weights()` in `nemo_automodel/components/models/qwen3_5_moe/model.py` raises an `AttributeError` because it unconditionally calls `self.linear_attn.norm.reset_parameters()`, but the HuggingFace `Qwen3_5MoeRMSNormGated` class does not implement this method.

`Qwen3_5MoeBlock` inherits from `Block` (qwen3_next) and in its `__init__` replaces `self.linear_attn` with the HuggingFace native `Qwen3_5MoeGatedDeltaNet`. That module's `.norm` is a `Qwen3_5MoeRMSNormGated` from HuggingFace transformers, which is a simple `nn.Module` that initializes `self.weight = nn.Parameter(torch.ones(hidden_size))` but does **not** define a `reset_parameters()` method.

Run any Qwen3.5 MoE configuration that triggers weight initialization for the linear attention layers:

## Expected behavior

`init_weights()` should handle the case where the norm module does not have a `reset_parameters()` method, and fall back to manually resetting the weight parameter.

## Additional context

**File**: `nemo_automodel/components/models/qwen3_5_moe/model.py` — `Qwen3_5MoeBlock.init_weights()`

**Error**:
```
File ".../nemo_automodel/components/models/qwen3_5_moe/model.py", line 99, in init_weights
    self.linear_attn.norm.reset_parameters()
AttributeError: 'Qwen3_5MoeRMSNormGated' object has no attribute 'reset_parameters'
```

**Proposed fix**:

- Add a `hasattr` check before calling `reset_parameters()`. If the method does not exist (HF norm), fall back to `self.linear_attn.norm.weight.data.fill_(1.0)`, which achieves the same reset effect as the HF constructor's `torch.ones(hidden_size)` initialization.

zpqiu:
## Summary

- Fixes #1430
- `Qwen3_5MoeBlock.init_weights()` crashes with `AttributeError` because HuggingFace's `Qwen3_5MoeRMSNormGated` does not implement `reset_parameters()`
- Added a `hasattr` guard and fallback to `weight.data.fill_(1.0)`, which matches HF's constructor initialization (`torch.ones(hidden_size)`)

## Test plan

- [x] Verify Qwen3.5 MoE weight initialization no longer raises `AttributeError`
- [x] Confirm norm weights are correctly initialized to ones in both code paths

### DIFF
--- a/tests/unit_tests/models/qwen3_5_moe/test_qwen3_5_moe_model.py
+++ b/tests/unit_tests/models/qwen3_5_moe/test_qwen3_5_moe_model.py
@@ -227,6 +227,25 @@ class TestQwen3_5MoeBlock:
         assert mock_trunc.call_count == 5
         mock_norm_reset.assert_called_once()
 
+    def test_init_weights_linear_attention_norm_without_reset_parameters(self, text_config_with_linear, moe_config, backend_config):
+        """Fallback path when norm (e.g. HF Qwen3_5MoeRMSNormGated) lacks reset_parameters."""
+        block = Qwen3_5MoeBlock(1, text_config_with_linear, moe_config, backend_config)
+
+        # Replace norm with a simple module that has weight but NO reset_parameters
+        fake_norm = torch.nn.Module()
+        fake_norm.weight = torch.nn.Parameter(torch.zeros(block.linear_attn.norm.weight.shape))
+        block.linear_attn.norm = fake_norm
+
+        with (
+            patch.object(block.input_layernorm, "reset_parameters"),
+            patch.object(block.post_attention_layernorm, "reset_parameters"),
+            patch.object(block.mlp, "init_weights"),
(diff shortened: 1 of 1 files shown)


HuiyingLi:
Hi @zpqiu , thank you so much for the PR! Looks great! Could you please add test also? Thank you!

PR #1489, opened 2026-03-08
fix: skip model.to(device) after checkpoint loading (tied params + FSDP)

terrykong:
## Summary

After FSDP sharding + checkpoint loading, calling `model.to(device)` triggers FSDP's `reset_sharded_param()` which fails on tied parameters. For tied `lm_head`/`embed_tokens` with TP=2, it tries to copy the full unsharded size (128256) into the local shard (64128):
```
RuntimeError: start (0) + length (128256) exceeds dimension size (64128)
```
### Upstream references
- https://github.com/pytorch/pytorch/issues/151085 (FSDP2 tied params + to(device))
- https://github.com/pytorch/torchtitan/issues/940 (torchtitan workaround for same issue)

## Changes
- In `infrastructure.py`, skip `model.to(device)` when `should_load_checkpoint` is True (checkpoint loading already places parameters on the correct device)
- Added two unit tests:
  - `test_skips_model_to_device_when_checkpoint_loaded` — verifies `model.to()` is NOT called when checkpoint was loaded
  - `test_calls_model_to_device_when_no_checkpoint` — verifies `model.to()` IS called in the from_config path
  
## Test plan
- [x] New unit tests for both code paths (checkpoint loaded vs not)  
- [ ] Run existing infrastructure unit tests to verify no regressions"

### DIFF
--- a/nemo_automodel/_transformers/infrastructure.py
+++ b/nemo_automodel/_transformers/infrastructure.py
@@ -520,8 +520,8 @@ def apply_model_infrastructure(
     if autopipeline is None:
         print_trainable_parameters(model)  # Once model's been sharded
         # Ensure model is on the correct device.
-        # Skip if checkpoint was loaded (params are already on device) to avoid triggering
-        # FSDP's reset_sharded_param which fails on tied parameters after parallelization.
+        # Skip when post-shard init was performed (params are already on the target device)
+        # to avoid triggering FSDP's reset_sharded_param which fails on tied parameters.
         if not need_post_shard_init:
             try:
                 model.to(device, non_blocking=True)


nemo_automodel/_transformers/infrastructure.py
```diff
@@ -498,15 +498,20 @@ def apply_model_infrastructure(
...
-        try:
-            model.to(device, non_blocking=True)
-        except NotImplementedError as e:
-            if "Cannot copy out of meta tensor" in str(e):
-                logger.warning("model.to(device) failed (meta tensors); using model.to_empty(device=device) instead.")
-                model.to_empty(device=device)
-            else:
-                raise
         print_trainable_parameters(model)  # Once model's been sharded
+        # Ensure model is on the correct device; AutoPipeline takes care of it internally.
```
hemildesai: Minor nit: Can we remove `AutoPipeline takes care of it internally.` from the comment? `autopipeline` should be None inside the branch.

nemo_automodel/_transformers/infrastructure.py
```diff
@@ -498,15 +498,20 @@ def apply_model_infrastructure(
...
+        # Skip if checkpoint was loaded (params are already on device) to avoid triggering
+        # FSDP's reset_sharded_param which fails on tied parameters after parallelization.
+        if not should_load_checkpoint:
```
hemildesai: Model is initialized here - https://github.com/NVIDIA-NeMo/Automodel/blob/50c488ceee923a57d760ba688143d254451dc09e/nemo_automodel/_transformers/infrastructure.py#L476 so can we change the condition to `if not need_post_shard_init`?

terrykong: https://github.com/NVIDIA-NeMo/Automodel/pull/1489/changes/949f6acf449cd0801cb0a6ad73550e970c684d63

nemo_automodel/_transformers/infrastructure.py
terrykong: https://github.com/NVIDIA-NeMo/Automodel/pull/1489/changes/949f6acf449cd0801cb0a6ad73550e970c684d63

PR #1491, opened 2026-03-09
feat: VDR feedback: Common inference utility

pthombre:
- Unified diffusion inference utility: Replaced separate model-specific generation scripts (flux_generate.py, wan_generate.py, wan_validate.py) with a single unified generate.py entry point supporting Flux, Wan, and HunyuanVideo models, driven by per-model YAML configs (generate_flux.yaml, generate_wan.yaml, generate_hunyuan.yaml)          
  - T5 bf16 compatibility patch: Added nemo_automodel/shared/t5_patches.py — a monkey-patch that replaces apex's FusedRMSNorm (which doesn't support bfloat16) with a native T5LayerNorm. Applied in the Flux processor before loading T5 models, and exported from nemo_automodel.shared                                                               
  - Flux processor fixes: Cast CLIP/T5 embeddings to bfloat16 in tools/diffusion/processors/flux.py to reduce memory and ensure dtype consistency                                                                                                                                                                                                       
  - Flux finetuning config updates: Adjusted flux_t2i_flow.yaml defaults (learning rate, guidance scale, flow shift, checkpoint frequency, removed pipeline_spec block)                                                                                                                                                                                 
  - Test fix: Fixed test_dataloader.py — tests were passing batch_size= to SequentialBucketSampler which expects base_batch_size=

### DIFF
--- a/tests/unit_tests/_transformers/test_auto_model.py
+++ b/tests/unit_tests/_transformers/test_auto_model.py
@@ -937,6 +937,10 @@ class TestBuildModelRetryDepth:
             patch("nemo_automodel._transformers.auto_model._init_model") as mock_init,
             patch("nemo_automodel._transformers.auto_model.get_world_size_safe", return_value=2),
             patch("nemo_automodel._transformers.auto_model._verify_sdpa_support"),
+            patch(
+                "nemo_automodel._transformers.capabilities.attach_capabilities_and_validate",
+                return_value=sentinel_model,
+            ),
             patch("nemo_automodel._transformers.auto_model.apply_model_infrastructure", return_value=sentinel_model),
             patch("nemo_automodel._transformers.auto_model.get_hf_config", return_value=mock_config),
             patch("nemo_automodel._transformers.auto_model._maybe_dequantize_fp8_for_peft", return_value=False),


nemo_automodel/shared/t5_patches.py
```diff
@@ -0,0 +1,73 @@
...
+    This must be called before loading any T5 models.
+
+    This function is idempotent and safe to call multiple times.
+    """
+    global _T5_PATCH_APPLIED
+    if _T5_PATCH_APPLIED:
+        return
+
+    try:
+        from apex.normalization import FusedRMSNorm
```
akoumpa: is there a way to avoid importing apex? I want to avoid the dependency if possible 🙏

pthombre: Removed the apex dependency

examples/diffusion/generate/generate.py
```diff
@@ -0,0 +1,443 @@
...
+import torch.distributed as dist
+
+from nemo_automodel.components.config._arg_parser import parse_args_and_load_config
+from nemo_automodel.components.loggers.log_utils import setup_logging
+from nemo_automodel.shared.t5_patches import patch_t5_layer_norm
+
+logger = logging.getLogger(__name__)
+
+# Pipeline class name -> output type mapping
+_PIPELINE_OUTPUT_TYPES = {
```
adil-a, line 54: ooc since only one kind of pipeline has the image output type, do we need this map? can we just do an if statement to check if it's a `FluxPipeline` and then just set it to image for that?

examples/diffusion/generate/generate.py
```diff
@@ -0,0 +1,443 @@
...
+
+    dtype_str = getattr(cfg.inference, "dtype", "bfloat16")
+    torch_dtype = _resolve_dtype(dtype_str)
+
+    checkpoint_dir = Path(checkpoint)
+    if not checkpoint_dir.exists():
+        raise FileNotFoundError(f"Checkpoint directory not found: {checkpoint_dir}")
+
+    ema_path = checkpoint_dir / "ema_shadow.pt"
+    consolidated_path = checkpoint_dir / "consolidated_model.bin"
```
adil-a, line 210: separate comment but how come the consolidated models are bin files? this doesn't seem like the HF format right?

examples/diffusion/generate/generate.py
```diff
@@ -0,0 +1,443 @@
...
+    elif sharded_dir.is_dir() and any(name.endswith(".distcp") for name in os.listdir(sharded_dir)):
+        logger.info("Loading sharded FSDP checkpoint from %s", sharded_dir)
+        pipe.transformer = _load_sharded_fsdp_checkpoint(pipe.transformer, str(sharded_dir), torch_dtype)
+        pipe.transformer.to("cuda", dtype=torch_dtype)
+        logger.info("Loaded sharded FSDP checkpoint")
+    else:
+        logger.warning("No recognized checkpoint format found in %s, using base model weights", checkpoint_dir)
+
+
+def _load_sharded_fsdp_checkpoint(transformer, sharded_dir, torch_dtype=torch.bfloat16):
```
adil-a, line 234: in addition to the comment above, would it make sense to use Automodel's checkpointer to do the checkpoint loading work? Does it work with diffusion models?

nemo_automodel/shared/transformers_patches.py
adil-a, line 1: in line with the other patch files in this module, can we call it transformers_patches instead?

nemo_automodel/shared/t5_patches.py
```diff
@@ -0,0 +1,72 @@
...
+
+from __future__ import annotations
+
+import logging
+
+import torch
+
+_logger = logging.getLogger(__name__)
+
+_T5_PATCH_APPLIED = False
```
adil-a: I think we should remove this. It's only here for the test suite and the test suites should depend on directly checking if the fix was applied rather than depending on a global flag.

examples/diffusion/generate/generate.py
pthombre: As diffusion support grows it might be useful to have this mapping saved. I think in current form we can scrap it, but this was added keeping future support in mind.

examples/diffusion/generate/generate.py
pthombre: This utility is used to load finetuned/pretrained automodel checkpoints for inference? Ideally most of the models we run can be accommodated on a single GPU for inference. Does the checkpointer implementation in AM support loading just the model to run inference?

nemo_automodel/shared/transformers_patches.py
pthombre: Added

nemo_automodel/shared/t5_patches.py
pthombre: Agreed. This is fixed

chtruong814 (approved):

PR #1527, opened 2026-03-11
fix: NaN loss in NemotronH from-scratch pretraining with FSDP2

chloechiaw:
## What does this PR do?

Fixes two issues that cause `loss=NaN` and `grad_norm=0.0` during from-scratch pretraining of NemotronHForCausalLM with FSDP2 and expert parallelism.

### 1. Remove `output_dtype` from FSDP2 MixedPrecisionPolicy (`config.py`)

`output_dtype=torch.bfloat16` causes FSDP2 to call `.to(bfloat16)` on the module forward output. For custom models returning a raw `torch.Tensor` (e.g. NemotronHForCausalLM), this detaches the autograd graph, silently zeroing all gradients and producing NaN loss.

HF models returning `CausalLMOutput` dataclasses are unaffected because the cast propagates through tensor fields without breaking gradients.

### 2. Add explicit weight init for custom models built from config (`model_init.py`, `auto_model.py`)

Under deferred-init (`no_init_weights` + `init_empty_weights` + FSDP2), `reset_parameters()` is suppressed and `to_empty()` leaves parameters as uninitialized garbage memory. The existing `initialize_weights()` only covers a subset of parameters. Anything it misses (e.g. Mamba `norm.weight`) remains as garbage, causing NaN output from the first Mamba CUDA kernel.

Adds `initialize_custom_model_from_config()` which runs OUTSIDE the `init_empty_weights` context and provides two-phase init:

1. **Global fallback**: unconditionally inits ALL params to safe defaults (normal for weights, zeros for biases, ones for norms)
2. **Architecture-specific init**: calls `initialize_weights()` on top (dt_bias, A_log, residual rescaling, etc.)

Only runs when `is_custom_model and not load_base_model` (from_config path). `from_pretrained` and checkpoint resume are unaffected.

### DIFF
--- a/nemo_automodel/_transformers/auto_model.py
+++ b/nemo_automodel/_transformers/auto_model.py
@@ -93,6 +93,7 @@ from nemo_automodel._transformers.model_init import (
     _init_model,
     get_hf_config,
     get_is_hf_model,
+    initialize_custom_model_from_config,
     no_hf_meta_device,
     resolve_sdpa_method,
 )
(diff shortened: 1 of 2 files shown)


chloechiaw:
@ZhiyuLi-Nvidia Hi Zhiyu, yep sorry we are waiting for the node to finish training for the class assignment, will get one PR with all of the fixes to you after this weekend. We ran into other surprising results and might need more code edits.

PR #1532, opened 2026-03-12
feat: add v4_compatible ckpt

akoumpa:
# What does this PR do ?

Model checkpoint assets such as `config.json` are not backwards compatible with transformers v4. 

This PR introduces a workaround to support v4-compatible `config.json` on transformers v5. 
Runtime-model config modifications (apart from quantization) will not be honored when using the `v4_compatible` flag.

Use this with caution.

# Changelog

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo_automodel/components/checkpoint/addons.py
+++ b/nemo_automodel/components/checkpoint/addons.py
@@ -69,21 +69,21 @@ class ConsolidatedHFAddon:
             # save the config.json file
             if hasattr(model_part, "config"):
                 v4_compatible = kwargs.get("v4_compatible", False)
-                if v4_compatible and _config_exists(original_model_path, "config.json"):
-                    _save_original_config_json(original_model_path, hf_metadata_dir, "config.json")
+                config_name = "config.json"
+                if v4_compatible and _config_exists(original_model_path, config_name):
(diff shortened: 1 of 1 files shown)


nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -123,6 +123,8 @@ class CheckpointingConfig:
     # This should be used for remote storage systems that don't support direct-append or non-sequential writes.
     staging_dir: str | None = None  # Optional directory for staging files during consolidation.
     # If provided, temp files will be created here instead of system temp. Useful when system temp has limited space.
+    v4_compatible: bool = False  # If True, save the original pretrained config.json (with quantization_config removed)
```
hemildesai, line 126: Should we default to True?

akoumpa: I think not, I think the opt-in approach is safer. This is a best-effort feature, to unblock customizer who is still using transformers v4 for deployment.

akoumpa: but please let me know if you see a strong case for setting this to True by default.

PR #1576, opened 2026-03-18
fix: checkpointing for PEFT.

akoumpa:
Previously, the state_dict in the modelstate class had an if/elseif/else statement where peft was handled in two caces and non-peft on the third one.

The first case of peft, was handling correctly, while the second was including buffers causing issues in downstream consumers.

This fix simplifies the logic (simple if/else) and bypassed the issues with the buffer.

I believe the issue was introduced in https://github.com/NVIDIA-NeMo/Automodel/commit/5b4996826938300ed36c6e4e2bc409c403f6dd88

# What does this PR do ?

Add a one line overview of what this PR aims to accomplish.

# Changelog

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo_automodel/components/checkpoint/checkpointing.py
+++ b/nemo_automodel/components/checkpoint/checkpointing.py
@@ -1285,19 +1285,20 @@ def _convert_checkpoint_with_transformers(
 
         # Now apply all the conversions
         for first_param_name, mapping in param_name_to_mapping.items():
-            # convert() returns (collected_tensors_dict, conversion_errors)
+            # convert() returns dict or (dict, errors) depending on transformers version
             result = mapping.convert(first_param_name, model=model, config=model.config)
-            assert isinstance(result, tuple), (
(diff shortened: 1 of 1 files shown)


nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -1258,41 +1258,55 @@ def _convert_checkpoint_with_transformers(
...
             try:
-                realized_value = mapping.convert(first_param_name, model=model, config=model.config)
+                # convert() returns (collected_tensors_dict, conversion_errors)
+                result = mapping.convert(first_param_name, model=model, config=model.config)
+                assert isinstance(result, tuple), (
+                    f"Expected convert() to return (dict, errors) tuple, got {type(result)}"
+                )
+                realized_value, _conversion_errors = result
+                assert isinstance(realized_value, dict), (
+                    f"Expected convert() first element to be dict, got {type(realized_value)}"
```
adil-a: why do we need these asserts now?

PR #1622, opened 2026-03-30
fix: cast input dtype in LinearLoRA.forward() to prevent dtype mismatch

### LINKED ISSUE #1540: gpt-oss single-gpu PEFT fails with type-mismatch error
**Describe the bug**

LoRA on GPT-OSS fails due to BF16/FP32 type mismatch. Behavior occurs on main, not in v0.3.0

```
[rank0]:   File "/opt/Automodel/nemo_automodel/components/models/gpt_oss/model.py", line 256, in forward
[rank0]:     hidden = self.model(
[rank0]:              ^^^^^^^^^^^
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1778, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1789, in _call_impl
[rank0]:     return forward_call(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/Automodel/nemo_automodel/components/models/gpt_oss/model.py", line 167, in forward
[rank0]:     h = layer(
[rank0]:         ^^^^^^
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1778, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1789, in _call_impl
[rank0]:     return forward_call(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/Automodel/nemo_automodel/components/models/gpt_oss/model.py", line 66, in forward
[rank0]:     attn_out = self.self_attn(
[rank0]:                ^^^^^^^^^^^^^^^
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1778, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1789, in _call_impl
[rank0]:     return forward_call(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/Automodel/nemo_automodel/components/models/gpt_oss/layers.py", line 104, in forward
[rank0]:     q = self.q_proj(x)
[rank0]:         ^^^^^^^^^^^^^^
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1778, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1789, in _call_impl
[rank0]:     return forward_call(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/Automodel/nemo_automodel/components/_peft/lora.py", line 239, in forward
[rank0]:     res = F.linear(x, self.weight, bias)
[rank0]:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: RuntimeError: mat1 and mat2 must have the same dtype, but got Float and BFloat16
```

```
step_scheduler:
  global_batch_size: 1
  local_batch_size: 1
  ckpt_every_steps: 500
  max_steps: 200
  num_epochs: 200
dist_env:
  backend: nccl
  timeout_minutes: 10
rng:
  _target_: nemo_automodel.components.training.rng.StatefulRNG
  seed: 1111
  ranked: true
model:
  pretrained_model_name_or_path: openai/gpt-oss-20b
  trust_remote_code: false
  backend:
    _target_: nemo_automodel.components.models.common.BackendConfig
    attn: te
    linear: te
    rms_norm: te
    rope_fusion: true
    experts: gmm
    dispatcher: deepep
    fake_balanced_gate: false
    enable_hf_state_dict_adapter: true
    enable_deepep: true
peft:
  match_all_linear: true
  dim: 8
  alpha: 32
  use_triton: true
checkpoint:
  enabled: false
  checkpoint_dir: checkpoints/
  model_save_format: safetensors
distributed:
  strategy: fsdp2
  tp_size: 1
  pp_size: 1
  ep_size: 1
  cp_size: 1
  sequence_parallel: false
  activation_checkpointing: false
loss_fn:
  _target_: nemo_automodel.components.loss.te_parallel_ce.TEParallelCrossEntropy
dataset:
  _target_: nemo_automodel.components.datasets.llm.squad.make_squad_dataset
  dataset_name: rajpurkar/squad
  split: train
packed_sequence:
  packed_sequence_size: 4096
  max_packs: 200
dataloader:
  _target_: torchdata.stateful_dataloader.StatefulDataLoader
  collate_fn: nemo_automodel.components.datasets.utils.default_collater
  shuffle: true
optimizer:
  _target_: torch.optim.AdamW
  betas:
  - 0.9
  - 0.95
  eps: 1.0e-05
  lr: 0.0001
  weight_decay: 0.0
  fused: true
```

**Expected behavior**

Same behavior as in v0.3.0

**Additional context**

[Full internal log](https://gitlab-master.nvidia.com/dl/jet/ci/-/jobs/280919218)

stanley1208:
# What does this PR do?

Fix a dtype mismatch error in `LinearLoRA.forward()` that causes LoRA training to fail when the input tensor dtype differs from the model weight dtype (e.g., float32 input with bfloat16 weights).

Fixes #1540.

## Root Cause

When model weights are loaded in bfloat16 (standard for LLMs), both the main linear layer (`self.weight`) and LoRA adapter layers (`lora_A`, `lora_B`) are in bfloat16. If the input `x` arrives as float32 (e.g., without autocast), `F.linear` raises:

RuntimeError: expected mat1 and mat2 to have the same dtype, but got: float != c10::BFloat16

## Fix

Cast input `x` to match `self.weight.dtype` at the top of `forward()`:

```python
x = x.to(self.weight.dtype)

This single line handles all downstream operations — the main linear, lora_A, lora_B, and DoRA paths — without needing per-call dtype casts.

Verification
Reproduced and verified on A100 GPU:

original_linear = nn.Linear(768, 768, bias=False).to(dtype=torch.bfloat16, device='cuda')
lora_linear = LinearLoRA(original_linear, dim=8, alpha=32).to('cuda')

# Before fix: RuntimeError (dtype mismatch)
# After fix: works correctly
x = torch.randn(2, 128, 768, dtype=torch.float32, device='cuda')
output = lora_linear(x)  # SUCCESS: shape=[2, 128, 768], dtype=bfloat16

Both float32 and bfloat16 inputs produce correct outputs after the fix.

### DIFF
--- a/nemo_automodel/components/_peft/lora.py
+++ b/nemo_automodel/components/_peft/lora.py
@@ -225,6 +225,8 @@ class LinearLoRA(nn.Linear):
             Tensor: Output tensor of shape (batch_size, out_features).
         """
         # pylint: disable=C0115,C0116
+        # Cast input to match weight dtype to avoid dtype mismatch (e.g. float32 input with bfloat16 weights)
+        x = x.to(self.weight.dtype)
         # If LinearLoRA is used to monkey-patch a nn.Linear module, we want to use nn.Linear's
         # forward in the case where it uses quantized weights. We store a reference to nn.Linear's
         # forward in `super_fwd` attribute. If the attribute does not exist we do the usual linear.

### CODE REVIEW

File: nemo_automodel/components/_peft/lora.py [hunk 1]
```diff
@@ -225,6 +225,8 @@ def forward(self, x):
             Tensor: Output tensor of shape (batch_size, out_features).
         """
         # pylint: disable=C0115,C0116
+        # Cast input to match weight dtype to avoid dtype mismatch (e.g. float32 input with bfloat16 weights)
+        x = x.to(self.weight.dtype)
```
akoumpa, line 229: Hi @stanley1208 , thanks for the PR, have you tried this with QLoRA by any chance?

stanley1208: @akoumpa Thanks for the quick review! I haven't tested with QLoRA yet — I verified with standard LoRA on A100 with bfloat16 weights and float32 input. 

For QLoRA, the `lora_dtype` parameter already handles setting the LoRA adapter dtype explicitly (lora.py line 108-109), and the `super_fwd` path (line 231-233) is used for quantized weights, which bypasses this cast. So the fix should be safe for QLoRA as well — but I'm happy to verify on Colab if you'd like. Let me know!

akoumpa: Thanks @stanley1208 for the reply!

To be honest I'm not sure this is the root cause for 1540, I feel there might be some issue related to initialization, where the whole model should be initialized with bfloat16, thus when it gets the token embeddings those should be in bfloat16 not float32.

akoumpa: In any case, I've started CI let's see how that goes 🙏

stanley1208: @akoumpa That makes sense — my fix is a defensive guard, but the root cause may be in model initialization. Happy to investigate the init/checkpoint loading path if that would be helpful!

akoumpa: @stanley1208 , definitely, if you have the BW I would greatly appreciate looking into the init/ckpt loading path 🙇

stanley1208: @akoumpa I traced the init/checkpoint loading path. Here's what I found:

The dtype inconsistency likely originates in `checkpointing.py` line 396. For custom models (like GPT-OSS), `_is_custom_model()` returns `True`, which **skips** the `_load_hf_checkpoint_preserving_dtype()` path — the function that explicitly loads weights while preserving their checkpoint dtype (e.g., bfloat16).

The flow:
1. `_init_model()` creates the model with `local_torch_dtype()` context (should be bfloat16)
2. But during checkpoint loading, custom models don't go through `_load_hf_checkpoint_preserving_dtype()`
3. The alternate loading path may not preserve dtype consistently
4. After loading, `embed_tokens` may produce float32 hidden states
5. These float32 hidden states flow into bfloat16 LoRA layers → dtype mismatch

The condition at `checkpointing.py:396`:
```python
_is_bin_checkpoint(model_path) or (is_safetensors and not _is_custom_model(...))

Comment (akoumpa) [reply nemo_automodel/components/_peft/lora.py:229]: This is awesome @stanley1208 , thanks for figuring this out so fast!

I wonder in this case what's the best way to proceed. From the one side we could add a similar `_load_hf_checkpoint_preserving_dtype` function, but that would need to:
(a) avoid allocating more memory than necessary, that is a problem for devices with unified memory (e.g., DGX Spark)
(b) handle things like dequantization, where the base checkpoint has quantized weights (e.g., mxfp4) but training happens in bf16
(c) weight remapping, where the base checkpoint has weights which are then translated to a different shape (as now happens in transformers v5).

Do you feel comfortable with the code to write a fix for this case? We are still in the process of adding the checkpoint test-suite, so please let me know. 

CC @adil-a

Comment (stanley1208) [reply nemo_automodel/components/_peft/lora.py:229]: @akoumpa Thanks! I'd love to take this on. The constraints you listed (memory, dequantization, weight remapping) make this more nuanced than a simple dtype cast — I want to make sure I get it right.

Could you point me to any existing examples of how custom models handle checkpoint loading currently? That would help me understand the expected flow before I propose a solution. Happy to start with a design sketch before writing code.

Comment (akoumpa) [reply nemo_automodel/components/_peft/lora.py:229]: This is great @stanley1208 , 

For the custom models, I would look at https://github.com/NVIDIA-NeMo/Automodel/tree/main/nemo_automodel/components/models and pick gpt_oss/llama/mistral4 as example s, you can run something like
```
from nemo_automodel import NeMoAutoModelForCausalLM
model = NeMoAutoModelForCausalLM.from_pretrained("/path/to/gpt_oss_20b") # you can also use the hf id
```
This should internally trigger the custom model path (if you want to use the default you can do so by passing `force_hf=True`. 

From there I would look into the [`_build_model`](https://github.com/NVIDIA-NeMo/Automodel/blob/177b4cf6b1640dad97f3f19d6ca5e18c0c6346a5/nemo_automodel/_transformers/auto_model.py#L177) method, [`_init_model`](https://github.com/NVIDIA-NeMo/Automodel/blob/177b4cf6b1640dad97f3f19d6ca5e18c0c6346a5/nemo_automodel/_transformers/model_init.py#L239) and the [`apply_model_infrastructure`](https://github.com/NVIDIA-NeMo/Automodel/blob/177b4cf6b1640dad97f3f19d6ca5e18c0c6346a5/nemo_automodel/_transformers/infrastructure.py#L321) (probably this one is the most important).  The code got a bit complicated I must admit, so please let me know if you have any questions :)

stanley1208: @akoumpa Thanks for the pointers! I'll dig into _build_model, _init_model, and apply_model_infrastructure and come back with a design sketch. Will reach out if I have questions.

PR #1650, opened 2026-04-01
fix: Float32RMSNorm torch.compile crash on PyTorch 2.11+

hemildesai:
## Summary
- Extract `Float32RMSNorm.forward` into a standalone `_float32_rms_norm_fwd()` compiled function to reduce dynamo guard-state combinations (eliminates guards on `self` and module state)
- Bump default dynamo `cache_size_limit` to 64 unconditionally in `compile_utils.py` so per-method `@torch.compile` decorators don't hit `FailOnRecompileLimitHit` from varying ndim/autocast/grad_mode combinations
- Root cause: PyTorch 2.11 is stricter about recompilation limits (default=8), and MoE training with variable-length sequences triggers 8+ guard-state combinations

## Test plan
- [x] Verified fix runs 50 steps of Qwen3 MoE 30B SFT with `rms_norm: torch_fp32` on PyTorch 2.11 (torch 2.11.0a0+eb65b36914.nv26.02) without crash
- [x] TPS/gpu: 3138 (fixed torch_fp32) vs 3218 (TE RMSNorm) — comparable performance
- [x] Pre-commit hooks pass (ruff, ruff-format)
- [ ] CI unit tests

### DIFF
--- a/examples/llm_finetune/nemotron/nemotron_nano_v3_hellaswag_peft_te.yaml
+++ b/examples/llm_finetune/nemotron/nemotron_nano_v3_hellaswag_peft_te.yaml
@@ -16,6 +16,7 @@
 # To run this recipe, please use the following command:
 # python examples/llm_finetune/finetune.py --config examples/llm_finetune/nemotron/nemotron_nano_v3_hellaswag_peft_te.yaml
 
+recipe: TrainFinetuneRecipeForNextTokenPrediction
 
 step_scheduler:
   global_batch_size: 8


nemo_automodel/components/utils/compile_utils.py
```diff
@@ -248,3 +248,11 @@ def build_compile_config(cfg: Optional[Dict[str, Any]]) -> CompileConfig:
 
 # Apply Flash Attention fix when module is imported (dynamo config happens per-compilation)
 _FLASH_ATTENTION_FIX_APPLIED = apply_flash_attention_compile_fix()
+
+# Set a reasonable dynamo cache limit at import time so that per-method @torch.compile
```
akoumpa, line 252: @hemildesai do you recall which config triggers this the gpt-oss + var seq len?

hemildesai: This is triggered when using variable sequence lengths with torch_fp32 rms norm. Specifically, the tulu3 qwen3 moe 30b config triggers this.

akoumpa: Thanks @hemildesai

Issue #1639, opened 2026-03-31
workaround required for padding_mask


akoumpa:
**Describe the bug**

In particular for https://github.com/NVIDIA-NeMo/Automodel/pull/1631, after fixing the tied weights issue, I run into another issue where the default collator was creating a "padding_mask" although the model is unable to consume it, thus crashing the run.

1631 included a filtering step to remove the padding_mask, but it would be better not to have to do any filtering, as it pollutes/complicates the code.

**Steps/Code to reproduce bug**
Run the yaml that I have attached to 1631

**Expected behavior**
Run without crashing / having to filter the batch.

**Additional context**

PR #1652, opened 2026-04-02
fix: stop default collater from adding padding_mask

zeel2104:
# What does this PR do ?

Stop `default_collater` from creating a `padding_mask` for generic padded batches, so models that do not accept that argument no longer need downstream filtering and do not crash.

# Changelog

- Remove synthesized `padding_mask` creation from `nemo_automodel.components.datasets.utils.default_collater`
- Update the default collator unit test to validate the new batch contract
- Keep the fix scoped to the generic collator path without changing THD/context-parallel handling

**Pre checks**:

# Additional Information

- Related to #1639


hemildesai (changes requested):
Hi, thanks a lot for your contribution. 

We need padding mask in the batch for correctly calculating aux_loss for MoEs. Rn, padding mask needs to be passed explicitly to handle some corner cases. As a result, I think an appropriate fix would be to add padding_mask to the batch if the model supports it, skip it otherwise.

akoumpa:
@hemildesai + @adil-a ping

hemildesai:
/ok to test [d8799e7](https://github.com/NVIDIA-NeMo/Automodel/pull/1652/commits/d8799e751b7f4d3a95c3a094872e54e5886eb2ae)

PR #1672, opened 2026-04-03
feat: Add Llada SFT support

pthombre:
● What does this PR do?

  Add supervised fine-tuning (SFT) support for the LLaDA diffusion language model using the existing MDLM strategy.

  Changelog

  - Add _patch_remote_code_compat() to handle LLaDA's older trust_remote_code model code against newer transformers (missing all_tied_weights_keys, old tie_weights signature, missing use_cache config default)
  - Add retry path in _from_pretrained_parent_class that applies compat patches on AttributeError/TypeError from remote code models
  - Add supervise_padding option to DLLMCollator to control whether padding tokens contribute to loss (needed for LLaDA SFT)
  - Add LengthGroupedSampler for text-only datasets that groups similar-length sequences into batches to reduce padding waste, with state_dict/load_state_dict for checkpoint resume
  - Add group_by_length dataloader config option that switches from StatefulDistributedSampler to LengthGroupedSampler
  - Add normalization_mode property to DLLMStrategy base class ("supervised" or "noise") to control loss denominator; MDLM defaults to "supervised" (total loss_mask==1 count)
  - Update MDLMCrossEntropyLoss to match the reference dllm framework: loss = sum(CE * mask * 1/p_mask) / sum(maskable)
  - Update MDLMStrategy.prepare_batch to remove attention_mask (MDLM models are bidirectional)
  - Add filter_forward_kwargs(model, batch) call before model(**batch) in dLLM recipe to handle models without **kwargs in their forward signature (e.g., LLaDA)
  - Track and log Train/supervised_tokens metric
  - Add llada_sft.yaml example config for LLaDA-8B-Base SFT on tulu-3-sft-mixture
  - Add unit tests for dLLM strategy, batch filtering, collator supervise_padding, sampler statefulness, loss numerical correctness, and remote code compat patch

  Pre checks:

  - Make sure you read and followed https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md
  - Did you write any new necessary tests?
  - Did you add or update any necessary documentation?

  Additional Information

  First dLLM model to use the MDLM strategy, so the loss formula change has no backwards compatibility impact.


nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -550,17 +550,34 @@ def build_dataloader(
...
-                "rank": dp_rank,
-                "shuffle": shuffle,
-            }
-            sampler = StatefulDistributedSampler(
-                ds,
-                seed=seed,
-                drop_last=True,
-                **dist_sampler_kwargs,
-            )
+            group_by_length = cfg_dl.get("group_by_length", False)
```
akoumpa, line 563: do we need this change?

pthombre: Llada training also includes padding tokens in the loss mask. Generally grouping samples by length allows us to reduce the effective number of padding tokens we need to add

PR #1837, opened 2026-04-14
refactor(retrieval): configure text attention mode

oliverholworthy:
# What does this PR do?

Make retrieval text self-attention configurable through `is_causal`, without changing vision attention.

An explicit YAML/API value overrides the checkpoint's text-config value. Without either setting, bi-encoders default to bidirectional attention (`false`), while cross-encoders preserve the constructed scoring backbone's native mode.

# Changelog

- Apply and persist the policy across registered, Hugging Face fallback, and extracted text backbones.
- Synchronize existing boolean `is_decoder` settings so encoder-only backbones construct the requested mask.
- Preserve native encoder-decoder scoring (including T5/BART); reject unsupported uniform overrides without changing encoder or cross-attention.
- Restrict composite-model changes to the text tower. Llama Nemotron VL uses `llm_config.is_causal`; its legacy outer `bidirectional_attention` input remains accepted but is ignored and no longer serialized.
- Register the local Nemotron VL config before loading, so cold-start behavior does not depend on earlier model imports.
- Reuse native Llama forward behavior to preserve hidden states and attention outputs in both modes.
- Update retrieval recipes, documentation, and tensor contracts, with masking, checkpoint-restoration, text/vision separation, and forward/backward parity tests.

# Before your PR is "Ready for review"

**Pre checks**:

- [x] Read and followed the [contributor guidelines](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md).
- [x] Added necessary tests.
- [x] Updated necessary documentation.

# Validation

- Local affected suite: 319 passed on both Transformers 5.15.1 (pinned) and 5.16.1, with CI runtime-budget checks enabled and no tests excluded.
- Local executable patch coverage: 100% (139/139 lines; 62/62 branches originating on changed lines).
- Replaced six subprocess cases with two offline, in-process config-loading checks. Verified they catch the missing registration even with a warm Hugging Face registry.
- Native T5/BART reload and direct/extracted scorer parity use tiny real models, FP32 score tolerances, and exact weight checks; no mocks added.
- Ruff and `git diff --check` passed.
- Both-mode Llama output and gradient parity passes on CPU/fp32 with eager and SDPA; GPU/distributed validation remains CI-owned.

PR #1838, opened 2026-04-14
feat: add extract_submodel parameter to build_encoder_backbone

oliverholworthy:
# What does this PR do ?

Add an `extract_submodel` parameter to `build_encoder_backbone` for generic VLM text-backbone extraction via dotted attribute path.

When `extract_submodel` is set, the parent model is loaded with HuggingFace `AutoModel`, the requested submodel is extracted, and the extracted text config is used to build the appropriate retrieval backbone. Supported extracted text backbones, such as Llama and Ministral3, still use the registered retrieval classes where available; unsupported embedding backbones are returned as extracted HF models; unsupported reranking backbones use the HF sequence-classification wrapper.

This avoids adding new model-specific VLM extraction paths for each architecture.

# Changelog

- Add `extract_submodel: str | None` parameter to `build_encoder_backbone()`
- Support dotted attribute extraction, e.g. `"language_model"` or `"model.language_model"`
- Convert supported extracted text backbones to registered retrieval backbones for embedding and reranking
- Support HF sequence-classification fallback for extracted `task="score"` backbones without a registered reranker
- Add functional tests with tiny local VLM checkpoints for embedding and reranking extraction paths

**Pre checks**:

# Additional Information

The parameter flows through the existing YAML config system via `**kwargs` passthrough:

```yaml
model:
  _target_: nemo_automodel.NeMoAutoModelBiEncoder.from_pretrained
  pretrained_model_name_or_path: mistralai/Ministral-3-3B-Base-2512
  extract_submodel: language_model
  pooling: avg
  l2_normalize: true
```

`extract_submodel` is a named parameter on `build_encoder_backbone`, so it is consumed there and not forwarded to HF's `from_pretrained`.

The same extraction path is also used by cross-encoder/reranking builds via `task="score"`.

Companion to #1837 (is_causal refactor). These two PRs are independent and can be reviewed/merged in any order.

### DIFF
--- a/nemo_automodel/_transformers/retrieval.py
+++ b/nemo_automodel/_transformers/retrieval.py
@@ -16,6 +16,7 @@
 
 import inspect
 import os
+from copy import deepcopy
 from typing import Optional
 
 import torch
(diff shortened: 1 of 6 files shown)


nemo_automodel/_transformers/retrieval.py
```diff
@@ -152,10 +155,23 @@ def build_encoder_backbone(
...
     if task == "score":
-        return AutoModelForSequenceClassification.from_pretrained(
+        model = AutoModelForSequenceClassification.from_pretrained(
             model_name_or_path, trust_remote_code=trust_remote_code, **hf_kwargs
         )
-    return AutoModel.from_pretrained(model_name_or_path, trust_remote_code=trust_remote_code, **hf_kwargs)
+    else:
+        model = AutoModel.from_pretrained(model_name_or_path, trust_remote_code=trust_remote_code, **hf_kwargs)
+
+    if extract_submodel is not None:
```
rnyak: @oliverholworthy this functionality is in the fallback section only meaning it wont cover those added in the supported_backbones dict, such as llama or ministral3 etc, if I am not mistaken. is there a specific reason this was added here? thanks.

oliverholworthy:
Addressed the `SUPPORTED_BACKBONES` review comment from https://github.com/NVIDIA-NeMo/Automodel/pull/1838#discussion_r3113081128.

`extract_submodel` is now handled before `SUPPORTED_BACKBONES` dispatch: when it is set, `build_encoder_backbone` loads the parent model with HuggingFace Auto classes, extracts the requested dotted path, and returns that submodel. This covers supported parent model types like `llama` / `mistral3` instead of only the fallback path.

Added `test_extract_submodel_supported_backbone_uses_hf_parent` to cover this explicitly.

oliverholworthy:
Updated the `extract_submodel`: it now extracts from the HF parent first, then checks the extracted submodel config against `SUPPORTED_BACKBONES`. If the extracted text backbone is supported, it instantiates the registered bidirectional backbone and loads the extracted clean state dict into it; otherwise it returns the extracted HF submodel unchanged.

This means supported text backbones such as Llama / Ministral still end as `LlamaBidirectionalModel` / `Ministral3BidirectionalModel`, while unsupported extracted text backbones keep the generic fallback behavior.

thomasdhc:
@oliverholworthy As a apart of this PR you could remove this omit for codecov: https://github.com/NVIDIA-NeMo/Automodel/blob/main/pyproject.toml#L324

Looks like it was causing issues before but since the file is in place, we should add it back. Thanks!

rnyak (approved):

adil-a (approved):

akoumpa (approved):

PR #1841, opened 2026-04-14
ci: add sync-skills workflow

ko3n1g:
## Summary

- Adds `sync-skills.yml` workflow to keep `skills/` in sync with agent directories via the `_sync_skills.yml@v0.91.0` reusable workflow
- Removes `skills/.claude` directory
- Adds **Git & PR conventions** section to `AGENTS.md` so agents know to use `type(scope)?: description` format for commit messages and PR titles (prevents bracket-prefix anti-patterns like `[ci] fix: …`)

## Test plan

- [ ] CI passes
- [ ] Skills sync correctly on next push to `main`


thomasdhc (dismissed):

thomasdhc:
@akoumpa Please have a look as well

akoumpa:
Hi @ko3n1g , thanks for making this.

Initially, i used the .claude/skills so that claude would find them out of the box, however, this created problem, because people mount their skills on .claude/skills and then git was causing issues, thus reverted back to using skills/.

Do you know if this would introduce any similar conflict?

ko3n1g:
> Hi @ko3n1g , thanks for making this.
> 
> Initially, i used the .claude/skills so that claude would find them out of the box, however, this created problem, because people mount their skills on .claude/skills and then git was causing issues, thus reverted back to using skills/.
> 
> Do you know if this would introduce any similar conflict?

Yeah I think that would still lead to issues. Perhaps we want to add these folders to .gitignore as a follow-up this PR. We need a single source of truth so that we can symlink it for other agents into read-only dirs

thomasdhc (dismissed):

akoumpa (dismissed):

akoumpa:
🤔 would a symlink not work tho?

thomasdhc (approved):

PR #1862, opened 2026-04-15
fix(collate): auto-derive assistant turn markers for non-Qwen models

khazic:
## Problem

`build_labels_from_template` has a fast marker-based path for Qwen-family
processors, but silently falls back to the fragile `build_labels` for any
other processor type (Gemma4, Llama, Mistral, ...).

`build_labels` re-tokenizes the assistant text standalone and searches for
it in `input_ids`. Because BPE is context-sensitive, the standalone
tokenization often differs from the in-sequence one → `_find_pattern_indices`
returns -1 → all labels = -100 → `num_label_tokens=0` → **nan loss**.

Observed on Gemma4-31B-it: intermittent and systematic `num_label_tokens=0`
steps causing nan loss and zero grad_norm during SFT training.

## Root cause

`Gemma4Processor` is not in `_IMSTART_TEMPLATE_PROCESSORS`, and
`convert_tokens_to_ids("<start_of_turn>")` returns the wrong id (3 instead
of 105) on the Gemma4 tokenizer, so a simple Gemma4-specific entry would
not be sufficient.

## Fix

Add two new helpers:

- **`_derive_turn_markers(tokenizer)`** — applies a minimal dummy conversation
  with a sentinel string and extracts the assistant-turn start marker and
  end-of-turn id directly from the resulting token ids. Works for *any* model
  without requiring per-processor registration. Also handles `BatchEncoding`
  (a `UserDict` subclass, not `dict`) correctly.

- **`_build_labels_from_markers(...)`** — shared scanning loop, eliminates
  duplication between the Qwen fast-path and the new general path.

`build_labels_from_template` now tries in order:
1. Fast path — Qwen via `convert_tokens_to_ids` (zero overhead)
2. General — `_derive_turn_markers` for all other processors
3. Last-resort — `build_labels` BPE fallback with an explicit warning

## Verification

Gemma4-31B-it SFT on 8×H100: `num_label_tokens` went from 0 (nan loss)
to the expected assistant response length after this fix.

### DIFF
--- a/nemo_automodel/components/datasets/vlm/collate_fns.py
+++ b/nemo_automodel/components/datasets/vlm/collate_fns.py
@@ -247,8 +247,8 @@ def _get_stop_token_id(tokenizer) -> Optional[int]:
 
 
 # Processor types whose chat template uses ``<|im_start|>``/``<|im_end|>``
-# markers.  For these we can reliably locate assistant turns by scanning the
-# token ids instead of re-tokenizing (which is sensitive to BPE context).
+# markers.  For these we use the fast ``_get_assistant_marker`` /
+# ``_get_stop_token_id`` helpers (no dummy-conversation overhead).
 _IMSTART_TEMPLATE_PROCESSORS = frozenset(
     {
         "Qwen2VLProcessor",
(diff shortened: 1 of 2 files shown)


HuiyingLi:
Thank you so much! ❤️   @khazic

HuiyingLi (approved):
Thank you very much!

PR #1870, opened 2026-04-16
fix(metric_logger): handle non-scalar tensor metrics without crashing

khazic:
## Summary

- `stack_and_move_tensor_metrics_to_cpu` called `.item()` unconditionally on `stacked_by_name[name][pos]`
- `.item()` raises `RuntimeError` if the tensor has more than one element (i.e., any non-scalar metric)
- Fix: use `.item()` for scalars (`numel() == 1`) and `.tolist()` for multi-element tensors, both of which produce JSON-serialisable Python types

## Test plan

- [ ] Log a scalar tensor metric → still works, `.item()` path taken
- [ ] Log a 1-D tensor metric (e.g. per-token loss array) → now works, `.tolist()` path taken, no `RuntimeError`


khazic:
Closing in favour of a clean PR with only the single fix commit.

PR #1895, opened 2026-04-17
fix: pass unnormalized residual to MoE gate in Gemma4 decoder layer

### LINKED ISSUE #1852: Gemma4 MoE: gate receives double-normalized input, causing high gen_kl_error in RL
**Describe the bug**

Discovered this by observing high gen_kl_error when working on https://github.com/NVIDIA-NeMo/RL/issues/2212, the 26B MoE model.

The Gemma4 MoE decoder layer (`Gemma4MoEDecoderLayer`) sends the same pre-normalized tensor to both the gate and the experts. The gate applies its own internal `RMSNorm`, resulting in double normalization of the gate input. This produces incorrect routing decisions in every MoE layer, causing a large logit divergence vs the HF reference implementation.

In HF transformers, the router receives the **raw residual** (unnormalized) and applies a single `RMSNorm` internally, while the experts receive `pre_feedforward_layernorm_2(residual)` (normalized). In Automodel, both receive `pre_feedforward_layernorm_2(residual)`, so the gate sees `RMSNorm_no_scale(RMSNorm_with_scale(x))` instead of `RMSNorm_no_scale(x)`.

Load Gemma4-26B-A4B-it in both HF transformers and Automodel forward the same input, and compare logits.

**Expected behavior**

The gate should receive the unnormalized residual (same as HF), so it applies exactly one `RMSNorm`. The experts should receive the output of `pre_feedforward_layernorm_2`. Logits should match HF within bf16 tolerance.

**Additional context**
FYI. I drafted a fix https://github.com/sharonyu-115/Automodel/commit/7804b7032b18220364f02fbdf0a0daaaa52e8b9c, verified with the NeMo RL training, `gen_kl_error` (KL divergence between vLLM inference and Automodel training logprobs) was 0.116 before fix, dropped to 0.0011 after fix.

sharonyu-115:
# What does this PR do ?

This PR is to address issue: https://github.com/NVIDIA-NeMo/Automodel/issues/1852?reload=1%3Freload%3D1
The Gemma4 MoE decoder layer (Gemma4MoEDecoderLayer) sends the same pre-normalized tensor to both the gate and the experts. The gate applies its own internal RMSNorm, resulting in double normalization of the gate input. This produces incorrect routing decisions in every MoE layer, causing a large logit divergence vs the HF reference implementation.

# Changelog

The Gemma4 MoE decoder layer was sending `pre_feedforward_layernorm_2(x)` to **both** `moe.gate` and `moe.experts`. `Gemma4Gate` applies its own RMSNorm internally (`Gemma4Gate.forward` in `nemo_automodel/components/models/gemma4_moe/model.py`), so the gate saw a double-normalized input and learned degenerate routing.

HuggingFace's reference `Gemma4TextMoeDecoderLayer` passes the **raw post-attention residual** to the router (single RMSNorm inside the router) and only the normalized input to the experts. This PR restores that split:

- `Gemma4MoE.forward` now accepts an optional `gate_input` kwarg.
- When provided, the gate uses `gate_input` for routing; experts still consume the normalized `x`.
- `Gemma4MoEDecoderLayer` calls `self.moe(moe_input, padding_mask, gate_input=x)` so routing sees the raw residual.

Backward-compatible: callers that don't pass `gate_input` get the previous single-input behavior.

**Pre checks**:

# Additional Information

- Related to https://github.com/NVIDIA-NeMo/Automodel/issues/1852?reload=1%3Freload%3D1


athitten:
Thank you so much @sharonyu-115 for identifying the bug and fixing it. I verified the convergence just to be safe with your branch and it looks great, the fix helps with it (better than what we had before). Appreciate your contribution here! Approved the PR

Screenshot 2026-04-17 at 5 12 22 PM

athitten (approved):
LGTM, TYSM @sharonyu-115

akoumpa (approved):

PR #1904, opened 2026-04-18
feat: TP+PP support for Gemma4 VLM (with tied lm_head fix and 31B recipes)

khazic:
## Summary

- Add TP+PP support for Gemma4 ConditionalGeneration VLM:
  register the model in the parallelizer and add Gemma4-specific
  pipeline forwards (text backbone + VLM outer), including
  `layer_types`-based RoPE dispatch, safe `tie_weights` skipping for
  PP stages without `embed_tokens`/`lm_head`, and PP chunking that
  handles `image_position_ids`.
- Fix tied `lm_head` loading on the PP last stage. PP last stages own
  `lm_head` but not `embed_tokens`, so they cannot share storage
  locally. HF tied-embedding checkpoints (and DCP checkpoints saved
  before this fix) omit `lm_head.weight` entirely, which made the DCP
  planner fail with `Missing key in checkpoint state_dict:
  lm_head.weight`. The fix:
  - Splits `uses_tied_lm_head` (config flag) from
    `has_local_tied_lm_head` (true only when local `lm_head` and
    `embed_tokens` actually share storage). Saved state dicts now keep
    `lm_head.weight` on PP last stages.
  - Adds `materialize_missing_tied_lm_head` /
    `get_tied_lm_head_source_names` helpers using
    `_tied_weights_keys` plus HF fallbacks.
  - In the safetensors fast load path, materializes the missing
    `lm_head.weight` from the embedding tensor.
  - In the standard DCP load path, when `lm_head.weight` is absent
    from checkpoint metadata, routes the embedding source key into
    the lm_head tensor and renames back after load. This now also
    applies on init load so a fresh HF tied-embedding base loads
    cleanly into a PP last stage.
- Add two desensitized example recipes for Gemma4 31B VLM:
  - `examples/vlm_finetune/gemma4/gemma4_31b_tp4_pp2.yaml` — single
    node, 8 GPUs (TP4 x PP2 x DP1).
  - `examples/vlm_finetune/gemma4/gemma4_31b_tp4_pp4.yaml` — 4 nodes,
    32 GPUs (TP4 x PP4 x DP2), with a multi-node `torchrun` snippet.
  Both use `google/gemma-4-31B-it` and the public MedPix-VQA dataset.

## Test plan

- [x] Added/updated unit tests in
  `tests/unit_tests/checkpoint/test_checkpointing.py`
  (`has_local_tied_lm_head`, `materialize_missing_tied_lm_head`,
  PP-last-stage save behavior).
- [x] Manual run of `gemma4_31b_tp4_pp2.yaml` (single node, 8 GPU).
- [x] Manual run of `gemma4_31b_tp4_pp4.yaml` (4 nodes, 32 GPU).
- [ ] CI: ruff, unit tests, recipe smoke tests.

### DIFF
--- a/examples/vlm_finetune/gemma4/gemma4_31b_tp4_pp2.yaml
+++ b/examples/vlm_finetune/gemma4/gemma4_31b_tp4_pp2.yaml
@@ -1,0 +1,110 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Single-node fine-tune of Gemma4 31B with TP=4 and PP=2.
+# Topology: 1 node x 8 GPUs = 8 GPUs total (TP4 x PP2, DP1).
+# Gemma4 31B has 46 decoder layers, split evenly: 23 layers on each PP rank.
+#
+# torchrun --nproc-per-node=8 examples/vlm_finetune/finetune.py \
+#          -c examples/vlm_finetune/gemma4/gemma4_31b_tp4_pp2.yaml
+
+recipe: FinetuneRecipeForVLM
+
+step_scheduler:
+  global_batch_size: 8
+  local_batch_size: 1
+  ckpt_every_steps: 500
+  val_every_steps: 500
+  num_epochs: 2
+
+dist_env:
+  backend: nccl
+  timeout_minutes: 60
+
+rng:
+  _target_: nemo_automodel.components.training.rng.StatefulRNG
+  seed: 42
+  ranked: true
+
+model:
+  _target_: nemo_automodel.NeMoAutoModelForImageTextToText.from_pretrained
+  pretrained_model_name_or_path: google/gemma-4-31B-it
+  torch_dtype: torch.bfloat16
+  use_liger_kernel: true
+  use_sdpa_patching: false
+  attn_implementation: sdpa
+  text_config:
+    use_cache: false
+
+checkpoint:
+  enabled: true
+  checkpoint_dir: vlm_checkpoints/gemma4_31b_it_tp4_pp2/
+  model_save_format: torch_save
+  save_consolidated: false
+
+distributed:
+  strategy: fsdp2
+  dp_size: none
+  tp_size: 4
+  cp_size: 1
+  pp_size: 2
+  ep_size: 1
+
+  sequence_parallel: false
+  activation_checkpointing: true
+
+  pipeline:
+    pp_schedule: 1f1b
+    pp_microbatch_size: 1
+    scale_grads_in_schedule: false
+
+loss_fn:
+  _target_: nemo_automodel.components.loss.masked_ce.MaskedCrossEntropy
+
+dataset:
+  _target_: nemo_automodel.components.datasets.vlm.datasets.make_medpix_dataset
+  path_or_dataset: mmoukouba/MedPix-VQA
+  split: train[:1000]
+
+dataloader:
+  _target_: torchdata.stateful_dataloader.StatefulDataLoader
+  num_workers: 0
+  pin_memory: true
+  collate_fn:
+    _target_: nemo_automodel.components.datasets.vlm.collate_fns.gemma4_prefix_collate_fn
+
+validation_dataset:
+  _target_: nemo_automodel.components.datasets.vlm.datasets.make_medpix_dataset
+  path_or_dataset: mmoukouba/MedPix-VQA
+  split: validation[:500]
+
+validation_dataloader:
+  _target_: torchdata.stateful_dataloader.StatefulDataLoader
+  collate_fn:
+    _target_: nemo_automodel.components.datasets.vlm.collate_fns.gemma4_prefix_collate_fn
+
+optimizer:
+  _target_: torch.optim.AdamW
+  lr: 1e-5
+  weight_decay: 0.01
+  betas: [0.9, 0.95]
+
+lr_scheduler:
+  lr_decay_style: cosine
+
+freeze_config:
+  freeze_embeddings: true
+  freeze_vision_tower: true
+  freeze_audio_tower: true
+  freeze_language_model: false
--- a/nemo_automodel/components/distributed/pipelining/hf_utils.py
+++ b/nemo_automodel/components/distributed/pipelining/hf_utils.py
@@ -239,16 +239,240 @@ def create_pipeline_forward_causal_lm() -> Callable:
     return pipeline_forward_causal_lm
 
 
+def create_pipeline_forward_gemma4_text() -> Callable:
+    """Pipeline-compatible forward for the Gemma4 text decoder backbone.
+
+    Works for both HF Gemma4TextModel (dense path) and Gemma4MoETextModelBackend (MoE path).
+    Handles:
+    - Optional embed_tokens (None on non-first PP stages; hidden states arrive in input_ids slot)
+    - Both full_attention and sliding_attention causal masks (Gemma4 uses mixed layer types)
+    - Per-layer-type position embeddings: Gemma4RotaryEmbedding.forward(x, pos_ids, layer_type)
+    """
+
+    def pipeline_forward_gemma4_text(
+        self,
+        input_ids: Optional[torch.LongTensor] = None,
+        attention_mask: Optional[torch.Tensor] = None,
+        position_ids: Optional[torch.LongTensor] = None,
+        inputs_embeds: Optional[torch.FloatTensor] = None,
+        cache_position: Optional[torch.LongTensor] = None,
+        padding_mask: Optional[torch.Tensor] = None,
+        **kwargs,
+    ) -> torch.Tensor:
+        if inputs_embeds is None:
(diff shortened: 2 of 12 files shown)


khazic:
Training log from a live run on the 4-node TP4 x PP4 x DP2 recipe (Gemma4 31B VLM) on our side — loss is converging and grad norm / tps look stable:

```
2026-04-18 06:57:50 | INFO | root | step 247 | epoch 0 | loss 0.2537 | grad_norm 12.3819 | lr 1.99e-05 | mem 29.18 GiB | tps 4006.84(2003.42/gpu) | num_label_tokens 138
2026-04-18 06:57:52 | INFO | root | step 248 | epoch 0 | loss 0.1538 | grad_norm 12.6614 | lr 1.99e-05 | mem 29.18 GiB | tps 3973.81(1986.90/gpu) | num_label_tokens 220
2026-04-18 06:57:54 | INFO | root | step 249 | epoch 0 | loss 0.2785 | grad_norm 9.9974  | lr 1.99e-05 | mem 29.18 GiB | tps 4010.05(2005.02/gpu) | num_label_tokens 216
2026-04-18 06:57:56 | INFO | root | step 250 | epoch 0 | loss 0.2776 | grad_norm 9.7785  | lr 1.99e-05 | mem 29.18 GiB | tps 3983.61(1991.81/gpu) | num_label_tokens 273
2026-04-18 06:57:59 | INFO | root | step 251 | epoch 0 | loss 0.3938 | grad_norm 16.7286 | lr 1.99e-05 | mem 29.18 GiB | tps 3296.05(1648.03/gpu) | num_label_tokens 270
2026-04-18 06:58:01 | INFO | root | step 252 | epoch 0 | loss 0.4191 | grad_norm 13.9407 | lr 1.99e-05 | mem 29.18 GiB | tps 3912.06(1956.03/gpu) | num_label_tokens 287
2026-04-18 06:58:03 | INFO | root | step 253 | epoch 0 | loss 0.0856 | grad_norm 18.3720 | lr 1.99e-05 | mem 29.18 GiB | tps 4002.60(2001.30/gpu) | num_label_tokens 284
2026-04-18 06:58:05 | INFO | root | step 254 | epoch 0 | loss 0.1639 | grad_norm 6.9901  | lr 1.99e-05 | mem 29.18 GiB | tps 3984.99(1992.49/gpu) | num_label_tokens 439
```

Config: 4 nodes x 8 GPUs (32 GPUs total), TP=4, PP=4, DP=2, `pp_microbatch_size=1`, 1f1b schedule, `activation_checkpointing=true`, bf16, Gemma4 31B VLM. Memory is steady at ~29 GiB/GPU and per-GPU tps is around 2k. Training started cleanly after the tied `lm_head` init fix in this PR — without it, the PP last stage failed with `Missing key in checkpoint state_dict: lm_head.weight` during base model load.

khazic:
@HuiyingLi Pushed `2a29a88` to fix the checks that were failing on the previous run:

- **Linting (I001)**: sorted imports in `checkpointing.py` and `parallelizer.py`.
- **Import contract**: `parallelizer.py` no longer imports `components.models.gemma4_moe.model.Gemma4ForConditionalGeneration` (which transitively pulled `components.checkpoint` and broke the independence contract). It now imports the HF base class from `transformers.models.gemma4.modeling_gemma4`, which is what dense Gemma4 31B VLM actually instantiates anyway.
- **L0 unit tests**:
  - `test_model_state_disables_tied_embeddings_for_non_tied_models` updated for the new `uses_tied_lm_head` / `has_local_tied_lm_head` split on `ModelState`.
  - `test_validate_model_with_tied_embeddings` and `test_validate_multiple_issues`: MockModel now provides `lm_head` + `embed_tokens` sharing the same weight tensor, so the validator's stricter "truly tied" check actually triggers.
  - Added a no-op `update_seq_len` to `_MockAutoPipeline` to match the new `self.pp.update_seq_len(input_ids.shape[1])` call added in the VLM PP path.

Locally `ruff check . && ruff format --check . && lint-imports` all pass (ruff 0.9.10 matching CI).

Could you kick off a fresh test run? `/ok to test 2a29a882cd...` when you have a moment — thanks!

khazic:
Also pushed `cf772a4d` addressing the Claude review:

1. **`patch_hf_model_for_pp` scope (bug)** — the Gemma4 branch was gated on structural `model.model.language_model is not None`, which also matches KimiVL / Mistral4 / Qwen3 VL MoE / LlavaOneVision / Kimi K25 VL. Gate is now `config.model_type == "gemma4"` (plus `config.text_config.model_type` for nested VLM configs), so unrelated VLMs fall back to the generic CausalLM forward.
2. **`_tied_weights_keys` shape** — added a comment explaining the dict-only branch in `get_tied_lm_head_source_names`: HF upstream uses a list (targets only, source resolved via `get_input_embeddings()` + the existing fallbacks), NeMo custom models use a target→source dict. The list case is already handled implicitly; the comment makes that explicit instead of looking like an oversight.
3. **Test coverage for the new VLM patch branch** — added three cases to `TestPatchHfModelForPp`:
   - Gemma4 via `config.model_type` → gets `pipeline_forward_gemma4_text` / `pipeline_forward_gemma4_vlm`.
   - Gemma4 via nested `config.text_config.model_type` → same.
   - Non-Gemma4 VLM with `model.language_model` → falls back to the generic `pipeline_forward` / `pipeline_forward_causal_lm` (regression guard for the bug #1 above).

Locally clean again: `ruff check . && ruff format --check . && lint-imports` all pass.

Could you re-trigger? `/ok to test cf772a4d79...` when you have a moment — thanks!

khazic:
Pushed `7c34c9fb` to address the L2 failures. Two issues rooted in the tied-lm_head rename:

1. **`checkpointing.py:_maybe_build_consolidated_index`** still referenced `model_state.is_tied_lm_head`, which I renamed to `uses_tied_lm_head` / split into `has_local_tied_lm_head`. That raised `AttributeError` during save on every tied-embedding model — the root cause for L2_Retrieval / L2_HF_Transformer_Finetune / L2_Pretrain_and_KD, and for most of the L2_HF_DCP failures. Switched it to `has_local_tied_lm_head` (only drop `lm_head` from the save map when it is actually an alias of the embedding; PP last stages keep their own `lm_head`).

2. **`has_local_tied_lm_head` was too strict**: it required `lm_head.weight is embed_tokens.weight`, which is broken by FSDP2 / TP sharding (DTensor wrapping breaks tensor identity). For single-rank Gemma3 VL 4B this made `has_local_tied_lm_head=False` after sharding, so the save path stopped dropping `lm_head`, and `test_dcp_vlm::test_vlm_dcp_checkpoint` saw a stray `lm_head.weight` key on disk. Relaxed to "config says tied AND both local weights exist"; that matches the old behavior for single-rank and still returns `False` for PP partitions missing one side.

Verified locally: `ruff check .`, `ruff format --check .`, `lint-imports` all clean; existing `test_has_local_tied_lm_head_is_false_for_pp_last_stage_like_partition` / `test_model_state_keeps_pp_last_stage_lm_head_in_saved_state_dict` still pass under the new definition.

`/ok to test 7c34c9fb...` when you have a moment — thanks!

nemo_automodel/components/distributed/pipelining/hf_utils.py
```diff
@@ -239,16 +239,240 @@ def pipeline_forward_causal_lm(
...
+    )
+
+    if inner_model is not None and text_backbone is not None:
+        # VLM with nested text backbone (e.g. Gemma4): patch text backbone and VLM outer
+        if patch_inner_model:
+            text_backbone.forward = types.MethodType(
+                create_pipeline_forward_gemma4_text(), text_backbone
+            )
+        if patch_causal_lm_model:
+            model.forward = types.MethodType(create_pipeline_forward_gemma4_vlm(), model)
```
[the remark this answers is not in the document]
akoumpa: Hi @khazic what do you think about this one?

akoumpa (approved):
Thanks a lot @khazic !

nemo_automodel/components/distributed/pipelining/hf_utils.py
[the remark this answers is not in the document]
khazic: Hi @akoumpa — good catch, this was already addressed in a follow-up commit `cf772a4d`

examples/vlm_finetune/gemma4/gemma4_31b_tp4_pp2.yaml
```diff
@@ -0,0 +1,110 @@
...
+# Gemma4 31B has 46 decoder layers, split evenly: 23 layers on each PP rank.
+#
+# torchrun --nproc-per-node=8 examples/vlm_finetune/finetune.py \
+#          -c examples/vlm_finetune/gemma4/gemma4_31b_tp4_pp2.yaml
+
+recipe: FinetuneRecipeForVLM
+
+step_scheduler:
+  global_batch_size: 8
+  local_batch_size: 1
```
HuiyingLi, line 26: @khazic  with pp2 the lbs shouldn't be 1?

nemo_automodel/components/distributed/pipelining/hf_utils.py
HuiyingLi, line 41: @khazic missing embed_vision for gemma4 for pp?

khazic: Good catch! When testing I was only running pure-text inputs, so the multimodal path wasn't exercised and `embed_vision` slipped through. Fixed by adding it to `MULTIMODAL_SUFFIXES` in `hf_utils.py`.

examples/vlm_finetune/gemma4/gemma4_31b_tp4_pp2.yaml
khazic: Good catch! Technically `lbs=1` still runs without errors — the code checks `n_microbatches < len(stages)` where `len(stages)` is the per-rank stage count (= 1 for 1F1B), so no warning is triggered. But with only 1 microbatch for 2 PP stages there is zero pipeline overlap (100% bubble). Fixed to `local_batch_size: 2` so we get 2 microbatches and proper 1F1B overlap.

PR #1734, opened 2026-04-08
ci: Update to transformers v5.5

athitten:
# What does this PR do ?

Upgrade to transformers v5.5 and updates mistral_common. Both of them are required for gemma4.

# Changelog

- `tests/unit_tests/models/deepseek_v3/test_dsv3_layers.py` — Updated mock configs to set rope_parameters = rope_scaling to match v5.5's DeepseekV3Config behavior.
- `nemo_automodel/components/models/gemma4_moe/model.py` — Added bidirectional compat between expert_intermediate_size (pre-v5.5) and moe_intermediate_size (v5.5 rename) in Gemma4ForConditionalGeneration.__init__ and Gemma4MoETextModelBackend.__init__.
- `tests/unit_tests/models/gemma4/test_gemma4_model.py` — Updated test configs to use moe_intermediate_size instead of expert_intermediate_size.
- `tests/unit_tests/models/gemma4/test_gemma4_state_dict_adapter.py` — Updated mock state dict to use v5.5 key format (experts.gate_up_proj, router.per_expert_scale).
- `nemo_automodel/components/models/nemotron_parse/model.py` — Handle MBartDecoderLayer.forward returning a single tensor in v5.5 (was tuple in v5.3).
- `nemo_automodel/components/models/nemotron_v3/model.py` — Replace pre-created DynamicCache from GenerationMixin with NemotronHybridCache in prepare_inputs_for_generation.
- `_reinit_non_persistent_buffers` in `nemo_automodel/components/checkpoint/checkpointing.py` since in v5.5 Gemma3RotaryEmbedding's per-layer-type inv_freq buffers, embed_scale, and SigLIP position_ids buffers are non-persistent (not saved in checkpoint) and so they have to be initialized explicitly. Creates a generalized method instead of the previous once which was specific to rope. Also adds appropriate unit test.
- Adds new message pattern added in transformers v5.4 "attempted to run this operator with Meta tensors" to `_build_model` method in `nemo_automodel/_transformers/auto_model.py` to retry building without meta device. Transformers v5.5 added a `torch.equal()` call inside `tie_weights()` (line 2539 of `modeling_utils.py`) to compare tied parameter values. When _build_model wraps from_pretrained inside init_empty_weights() (meta device context), the model parameters are meta tensors, and torch.equal doesn't support them — producing a NotImplementedError. Retry logic is already there in _build_model method, the PR adds this error message so that retry is done even for this case. (this is done in #1765 )
[...]

PR #1906, opened 2026-04-18
fix: restore Qwen3.5 + Phi-4-MM nightly CI after transformers v5.5 update

HuiyingLi:
## Summary

Two nightly VLM finetune CI jobs broke after the transformers v5.5 bump (#1734). This PR fixes both.

## Changes

**`nemo_automodel/components/models/qwen3_5_moe/cp_linear_attn.py`** — port `CPAwareGatedDeltaNet._forward_no_cp` to the transformers v5.5 per-layer cache API:
- `cache_params.has_previous_state` → `cache_params.has_previous_state(self.layer_idx)` (now a method)
- Read states only when `use_precomputed_states` is true
- Read via `cache_params.layers[layer_idx].{conv,recurrent}_states` instead of the removed top-level dicts
- Write via `update_conv_state` / `update_recurrent_state` methods instead of `conv_states[idx] = ...`

Without this, every forward pass with a fresh `DynamicCache` raised `AttributeError: 'DynamicCache' object has no attribute 'conv_states'`.

**`nemo_automodel/_transformers/kernel_patches.py`** (+ wire-up from `utils.py`) — bridge the legacy `_supports_flash_attn_2` class flag to v5.5's `_supports_flash_attn`. transformers v5.5 renamed the attribute and switched the `_flash_attn_can_dispatch` check to the new name only (defaulting to `False` on `PreTrainedModel`). Remote-code models pinned against ≤v5.3 (e.g. `microsoft/Phi-4-multimodal-instruct` sets `_supports_flash_attn_2 = True`) are unaware of the rename, so their FA2 support becomes invisible to v5.5 and `attn_implementation="flash_attention_2"` raises `ValueError: Phi4MMForCausalLM does not support Flash Attention 2`.

Fix: install a property on `PreTrainedModel._supports_flash_attn` that falls back to the legacy flag when a subclass hasn't set the new one. Subclasses that set `_supports_flash_attn` directly still shadow the property via normal MRO lookup, so native v5.5 models are unaffected. Called from `apply_cache_compatibility_patches()` so it runs at the same setup point as the other v5 compat shims.

After the bridge, Phi-4-MM dispatches to FA2 on v5.5 (confirmed by `is_flash_attn_greater_or_equal_2_10` being called during forward; memory also drops from ~11.37 GiB SDPA → ~10.97 GiB FA2).

## Tests

- `TestPatchLegacyFlashAttnFlag` (in `tests/unit_tests/_transformers/test_auto_model.py`): property installed, idempotent, legacy-True bridges, explicit new-flag True/False both shadow, base default preserved, legacy-False does not bridge, nearest-in-MRO wins.
- `TestForwardNoCpV55CacheAPI` (in `tests/unit_tests/models/qwen3_5_moe/test_cp_linear_attn.py`): training-style `DynamicCache` runs without error, no-cache path still works, `update_conv_state` / `update_recurrent_state` invoked with the layer's `layer_idx`, `has_previous_state(layer_idx)` called as a method.

## Linked CI jobs

- Qwen3.5: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/300041612
- Phi-4-MM: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/300041609

## Test plan

- [x] Reproduced both failures locally against the same container image, then confirmed the fixes pass
- [x] `torchrun --nproc-per-node=8 examples/vlm_finetune/finetune.py -c examples/vlm_finetune/qwen3_5/qwen3_5_4b.yaml` (max_steps=2) — steps 0/1 + validation + checkpoint, exit 0
- [x] `torchrun --nproc-per-node=8 examples/vlm_finetune/finetune.py -c examples/vlm_finetune/phi4/phi4_mm_cv17.yaml` (max_steps=1) — step 0 (loss 2.8924, FA2 confirmed) + validation + checkpoint, exit 0
- [x] New unit tests: `pytest tests/unit_tests/_transformers/test_auto_model.py::TestPatchLegacyFlashAttnFlag tests/unit_tests/models/qwen3_5_moe/test_cp_linear_attn.py::TestForwardNoCpV55CacheAPI` — 12 passed
- [ ] Nightly VLM CI green

## Not in this PR

`nemotron_parse_v1_1` nightly (job 300041608) fails offline because the CI's HF cache is missing `nvidia/C-RADIOv2-H`. That's a cache-seeding fix on the CI side (add `huggingface-cli download nvidia/C-RADIOv2-H` to the pre-cache step), not a library bug — tracked separately.

### DIFF
--- a/nemo_automodel/_transformers/kernel_patches.py
+++ b/nemo_automodel/_transformers/kernel_patches.py
@@ -136,6 +136,51 @@ def _patch_liger_kernel(model):
         raise RuntimeError("Failed to patch model")
 
 
+def _patch_legacy_flash_attn_flag():
+    """Bridge the legacy ``_supports_flash_attn_2`` class flag to v5.5's
+    ``_supports_flash_attn``.
+
+    transformers v5.5 renamed the FA2-support attribute from
+    ``_supports_flash_attn_2`` to ``_supports_flash_attn`` and switched the
+    dispatch check at ``_flash_attn_can_dispatch`` to the new name only.
+    Remote-code models pinned against <=v5.3 (e.g. microsoft/Phi-4-multimodal-instruct
+    sets ``_supports_flash_attn_2 = True`` in its modeling file) are not aware
+    of the rename, so their FA2 support is invisible to v5.5 and
+    ``attn_implementation="flash_attention_2"`` raises ``ValueError``.
+
+    Install a property on ``PreTrainedModel._supports_flash_attn`` that falls
+    back to the legacy flag when a subclass has not set the new one. Subclasses
+    that set ``_supports_flash_attn = True`` directly still shadow the property
+    via normal MRO lookup, so native models are unaffected.
+    """
+    import transformers.modeling_utils as mu
+
+    base = mu.PreTrainedModel
+    if getattr(base, "_nemo_fa2_flag_bridged", False):
+        return
+
+    # Capture the base-class default (``False`` on v5.5) so the fallback
+    # preserves original behavior when no flag is set anywhere.
+    _base_default = base.__dict__.get("_supports_flash_attn", False)
+
+    def _supports_flash_attn_fget(self):
+        cls = type(self)
+        for klass in cls.__mro__:
+            # Stop at the base — the property lives here; anything below is
+            # just the captured default.
+            if klass is base:
+                break
+            d = klass.__dict__
+            if "_supports_flash_attn" in d:
+                return d["_supports_flash_attn"]
+            if d.get("_supports_flash_attn_2") is True:
+                return True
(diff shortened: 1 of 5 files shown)


HuiyingLi:
Updated the PR description to match the current diff — earlier version referred to a prior `model_init.py` approach that was dropped in favor of the `_patch_legacy_flash_attn_flag` bridge in `kernel_patches.py`.

/claude review

HuiyingLi:
PR description now updated correctly (previous edit was swallowed by a GraphQL warning on the Projects classic API). Could you re-review?

/claude review

akoumpa (approved):

PR #1914, opened 2026-04-20
feat: add gemma4 moe ring CP

khazic:
# What does this PR do?

Adds a Gemma4-specific Context Parallelism (CP) path for **dense and MoE** VLM
fine-tuning, without changing the generic CP behavior of any other model.

Gemma4 cannot use the existing causal ring-attention CP path for this workload:
its decoder mask combines **image-token bidirectional attention**, **sliding-window
attention**, and **packed-sequence document boundaries**, and its **large head
dimension (256)** rules out FlashAttention-2 and the SDPA flash/mem-efficient
kernels. This PR routes only Gemma4 decoder self-attention through a **model-owned
p2p ring FlexAttention** path, plugged into a small model-agnostic seam in the CP
infrastructure.

At a high level, each CP rank keeps its local **query** shard. K/V (and the token
metadata needed to build masks) are **rotated around the CP ranks via point-to-point
ring exchange** — they are never fully all-gathered. For each ring step the active
Gemma4 layer builds a `torch.compile(flex_attention)` block mask (full / sliding,
packed-document boundaries, image bidirectional groups, padding rows), computes
attention against the current K/V chunk, and the per-chunk outputs are combined with
an **online-softmax (logsumexp) merge**. Only the local query output shard is returned.

# Design

CP attention routing is model-agnostic (`apply_cp` / `attach_cp_attention_hooks`):

- **TE `DotProductAttention`** → TE's own context-parallel group (unchanged).
- **Non-TE attention exposing `run_cp_manual_attention`** → the model owns its CP
  transport + compute through that seam (Gemma4's flex ring).
- **Any other non-TE attention** → the classic DTensor SDPA CP path.

`cp_utils` contains **no model names** — only the TE-vs-not split and the seam; the
Gemma4 ring lives entirely in `components/models/gemma4_moe/cp_attention.py` and is
attached via `attach_gemma4_cp_ring_attention`. This keeps the path easy to extend to
other models that need their own CP attention.

# Changelog

- **`components/models/gemma4_moe/cp_attention.py`** (new)
  - Gemma4 p2p ring FlexAttention CP: ring K/V collection + p2p exchange, per-chunk
    FlexAttention with layer-aware masks (causal, sliding-window, packed-document
    boundaries, image-token bidirectional groups, padding-only query rows), online
    softmax merge, a custom autograd `Function` with ring grad exchange, the
    `run_cp_manual_attention` entry, and `attach_gemma4_cp_ring_attention`.

- **`components/models/gemma4_moe/model.py`**
  - CP-aware forward for dense and MoE paths.
  - `prepare_model_inputs_for_cp` / `prepare_inputs_embeds_for_cp` to build text+image
    `inputs_embeds` on the unsharded sequence before CP sharding.
  - Packed / vision-bidirectional FlexAttention block-mask mapping; pad-token resolution.

- **`components/distributed/cp_utils.py`**
  - Model-agnostic CP attention seam + SDPA-swap hooks; manual contiguous sequence
    sharding for the model-owned path; single-document `_packed_seq_ids` synthesis;
    padding / position-id helpers. Generic models keep the DTensor `context_parallel`
    path.

- **`components/moe/parallelizer.py`**
  - Capability-based `apply_cp` routing (TE vs hooks; model-owned vs generic DTensor).

- **`_transformers/infrastructure.py`**
  - Attach the dense (non-TE) CP hooks when `cp_size > 1`.

- **`recipes/vlm/finetune.py`**
  - Pre-compute `inputs_embeds` before CP sharding for Gemma4 + image inputs.
  - Sum validation label-token counts over CP (the VLM recipe shards labels before
    counting), so `val_loss` divides like-for-like.

- **`examples/vlm_finetune/gemma4/gemma4_26b_a4b_moe_medpix_ep8cp2_4k.yaml`**
  - Runnable Gemma4-26B-A4B MoE MedPix EP8×CP2 config at 4k sequence length.

- **Tests**
  - Comprehensive unit tests for the ring attention (mask builders, ring/grad p2p
    exchange, online-softmax merge, autograd, manual-attention entry), the CP wiring
    seam and hooks, the packed/vision mask mapping, parallelizer routing, the
    infrastructure CP-hook attach, and the validation token-counting fix —
    bringing the PR's changed lines to ~100% unit coverage (CPU-only, no GPU
    required; FlexAttention exercised in CPU-eager mode).

**Pre checks**:

# Test Results

## Targeted behavior

- Gemma4 CP2 routes through the compiled FlexAttention p2p ring. Observed log line
  (EP8×CP2, 2k seq, 26B-A4B MoE):
  - `Gemma4 CP using compiled flex_attention p2p ring. Q=(1, 16, 1024, 256) K_local=(1, 8, 1024, 256) head_dim=256->256 cp_rank=0 cp_size=2`
- Packed-sequence mask correctness target:
  - `CP2 forward(A alone) == CP2 forward(packed [A, B])[:, A slice]`
  - `CP2 forward(B alone) == CP2 forward(packed [A, B])[:, B slice]`
- Text-only, image, packed, and no-pack inputs all route through the same Gemma4 CP
  ring implementation, with layer-specific mask construction.
- Unit tests verify the ring forward and gradients match a full-attention SDPA
  reference at `cp_size=1`, and the online-softmax merge matches a single full
  softmax.
- ep8cp1 ep8cp2 parity runs:
https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/plvv3r6b
https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/itjikprm

image

- single node 32k seqlen on mock https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/bm3lch7d
image


khazic:
Heads-up on validation status: I'm currently running end-to-end tests on 4 nodes (32x A100-80GB) for both the MoE path (`gemma4_26b_a4b_moe` / `gemma-4-E4B`) and the 31B dense path with `cp_size=2`. Will post the loss-parity numbers (CP=1 vs CP=2) and step-time figures once the runs finish. The mock-dataset configs in this PR are already validated locally; please hold off on merging until the multi-node runs come back green.

khazic:
## 4-node smoke-test results (32x H100-80GB)

End-to-end CP verification on 4 nodes (32 x H100-80GB) with Gemma4-E4B MoE (~7B total, 4B activated).

### Loss parity: CP=1 vs CP=2 (matched per-GPU shard size)

To verify the ring-attention + flex_attention path is mathematically equivalent to the non-CP baseline, I ran two matched configurations (same per-GPU `S_local`, different `cp_size`):

| step | CP=1, seq=2048 | CP=2, seq=4096 | Δ |
|------|----------------|----------------|----|
| 0    | 16.9419        | 17.1810        | 0.24 |
| 1    | 10.1633        | 10.4407        | 0.28 |
| 2    | 8.4085         | 8.7233         | 0.31 |
| 3    | 7.1558         | 7.4028         | 0.24 |

Trajectories overlap within noise. Residual Δ is from different global batch sizes (32 vs 16) sampling different mock examples per step plus reduction-order FP differences. `grad_norm` tracks similarly: `1256 -> 68` (CP=1) vs `1136 -> 77` (CP=2).

### Topology / infra notes

- `cp_size=2, dp_size=16` (CP within node over NVLink, DP across nodes over TCP)
- `activation_checkpointing: true`, `PYTORCH_ALLOC_CONF=expandable_segments:True`
- `NCCL_IB_HCA=^=all` + `NCCL_NET_PLUGIN=none` were needed to work around a NCCL 2.27.5 bug in the K8s container where the proxy thread kept attempting IB-type handshakes even with `NCCL_IB_DISABLE=1`. Not related to this PR.

### 31B dense: not yet validated

I have not yet run the 31B dense config end-to-end. Earlier attempts OOM'd during backward on a 32x 80GB cluster with pure FSDP+CP (weights + Adam state + flex_attention backward workspace together exceed per-GPU budget). 31B long-context training almost certainly needs TP alongside CP to shard the hidden dimension as well; that combination is not covered by this PR and will be a follow-up. The 31B yamls in this PR are provided as configuration templates; they are not expected to run as-is without either TP or a larger cluster.

### Why all-gather + flex_attention instead of ring-attention

The all-gather variant here is deliberately simple: each rank gathers full K/V along the sequence dimension then runs a compiled `flex_attention` with a per-rank causal block mask. We chose this over a ring-attention implementation for two reasons:

1. **Correctness under mixed DTensor/plain-tensor paths.** PyTorch's native `context_parallel` DTensor dispatch into SDPA kept hitting a "mixed torch.Tensor and DTensor" crash in the attention inner_fn (the compiled decoder propagates a local-tensor Q while K/V return as DTensors after all-gather). All-gather-then-local-SDPA avoids the mixed path entirely and works on every PyTorch version we tested.
2. **Reused infra.** flex_attention already gives us Flash-style O(N) memory via its Triton tiled kernel and accepts arbitrary per-rank block masks; we just had to wrap it in `torch.compile` and pass `kernel_options` to shrink block sizes when `head_dim >= 256` (the Gemma4 31B global-attention-layer variant has `head_dim=512`).

A proper ring-attention implementation would save the `S_full` K/V memory per rank (O(N) in seq instead of O(N/CP)), and could let us use Flash Attention directly (because each ring chunk is a standard causal attention). That is the right long-term direction but is a larger change; keeping it out of this PR so the diff stays reviewable.

### Known limitations (follow-up PR candidates)

- `create_block_mask` materializes a dense `[S_local, S_full]` bool mask, which allocates ~1 GiB at `seq=16K` and OOMs before flex_attention even starts. Needs a sparse or streaming construction.
- CE loss materializes a fp32 `[B*S_local, 262144]` logit tensor (~4 GiB at `seq=8K`). Chunked CE would unblock longer contexts.
- Ring-attention path (see above) to eliminate the `S_full` K/V memory footprint.

khazic:
A few more notes to round out the review context.

## Ring attention is the right long-term direction

The all-gather variant in this PR is a pragmatic starting point, not the ideal design. A proper ring attention implementation (each rank keeps only its local K/V shard and rotates them in a ring) would be strictly better along every axis:

- **Memory**: avoids the per-rank `S_full` K/V materialization, so attention memory goes from `O(S_full/CP)` (current) back to `O(S_local)` per rank — the actual reason people reach for CP in the first place.
- **Compute path compatibility**: each ring chunk becomes a standard causal (or fully-open) attention on contiguous Q, K, V of the same length, so Flash Attention accepts it directly. We then do not need flex_attention at all and recover its best-case throughput.
- **Communication overlap**: ring P2P sends can be overlapped with per-chunk Flash attention, which the all-gather pattern cannot.

I chose the all-gather + flex_attention path in this PR because it (a) avoids the "mixed torch.Tensor / DTensor" crash in PyTorch's native `context_parallel` dispatch on current PyTorch versions, and (b) reuses flex_attention's already-tested block-sparse kernel so the CP-specific code stays small and reviewable. The follow-up should be a real ring attention wrapper; I'm happy to take that as a separate PR.

## VLM path: image and no-image both validated

The recipe-level changes in `finetune.py` handle both paths:

- **Text-only** (no `pixel_values` in the batch): `make_cp_batch_and_ctx` shards `input_ids`, `labels`, and `position_ids` directly. The `gemma4_4b_cp2_mock.yaml` smoke test runs this path (mock dataset with `num_images_per_sample: 0`); loss-parity numbers are in the previous comment.
- **With images** (`pixel_values` present): we cannot CP-shard `input_ids` before image features are merged, because image tokens are concentrated at the start of the sequence and the `masked_scatter` inside HF Gemma4 cannot redistribute image features across CP ranks correctly (each rank would see a different number of image-token slots). Solution: at the recipe layer (`finetune.py:920-954`), when CP is active and `pixel_values` is present, we pre-compute `inputs_embeds` on the full unsharded sequence via a `_pre_embed_only=True` pass through `model.__call__`, then hand `inputs_embeds` (not `input_ids`) to `make_cp_batch_and_ctx`. Also validated on a mock VLM dataset.
- We route the pre-embed through `model.__call__` (not direct method call) specifically so the FSDP2 pre-forward hook fires and all-gathers the sharded `embed_tokens` weight into a Replicate placement — calling the method directly bypasses the hook and we hit a mixed DTensor / plain-tensor error. This is why `prepare_inputs_embeds_for_cp` is exposed via `_pre_embed_only` instead of called directly.

## Debug decisions worth flagging for the review

These are the non-obvious ones we landed on during a multi-day debugging session; a reviewer may want to push back or suggest a better shape.

1. **flex_attention must be called through `torch.compile`.** Without `torch.compile` the eager implementation materializes the full `[B, nH, S_local, S_full]` scores matrix and OOMs immediately on any real sequence length — the exact opposite of flex_attention's purpose. We lazily compile once per process and cache (`cp_utils.py:160-171`). A fresh warning appeared in PyTorch 2.10 that surfaces this; without the compile wrapper I spent an embarrassing amount of time misdiagnosing the OOM as a memory-pressure problem.

2. **`enable_gqa=True` must be passed explicitly to flex_attention.** Gemma 4 uses GQA (E4B has 16 Q / 2 KV heads; 31B has 32 Q / 16 KV). Our fallback SDPA path already auto-enabled this flag, but the initial flex_attention call did not, which caused it to always raise `ValueError: Expect query and key/value to have the same number of heads`, silently falling back to SDPA MATH and OOMing. One-line fix in `cp_utils.py:260-266`.

3. **Triton kernel_options must be shrunk for large `head_dim`.** Gemma 4 31B's "global attention" layer variant uses `head_dim=512`. With default flex_attention `BLOCK_M/BLOCK_N=128` and `num_stages=3`, the required Triton shared memory is ~200 KB, which exceeds A100/H100's ~163 KB limit; the kernel errors out with `No valid triton configs. OutOfMemoryError: out of resource: triton_tem_fused_flex_attention_0 Required: 200704 Hardware limit: 166912`. We now pass `kernel_options={"BLOCK_M": 32, "BLOCK_N": 32, "num_stages": 1, ...}` whenever `head_dim >= 256` (`cp_utils.py:265-283`). Throughput trade-off: 32x32 blocks are ~2-3x slower than the default for the same kernel, but they fit. Could be made adaptive in a future PR.

4. **Silent except-catch on the flex_attention call is intentional — with a logger.** The primary path is wrapped in `try: ... except Exception as err:` so that any Triton compile failure (e.g. on an unsupported shape) falls back to SDPA rather than crashing the run, but we log the exception type, shapes, and traceback once per process (`cp_utils.py:266-284`) so the silent fallback does not hide an actual kernel bug. This was added after a multi-hour chase where flex_attention was failing silently and the only visible symptom was an unrelated-looking 1 GiB OOM deep inside SDPA.

5. **Manual sequence slicing, not `context_parallel` DTensor dispatch.** PyTorch's `torch.distributed.tensor.experimental.context_parallel` wraps Q/K/V as DTensors and relies on the DTensor SDPA dispatch to trigger ring communication. On the PyTorch versions we tested this breaks because the compiled decoder propagates a plain-tensor Q through the graph while K/V all-gather comes back as DTensors, and the inner_fn trips a "mixed torch.Tensor and DTensor" assert. We moved to manual sequence slicing in `make_cp_batch_and_ctx` and put the CP communication inside an SDPA hook (`attach_cp_sdpa_hooks`), which keeps activations outside of attention as plain tensors end-to-end.

6. **Gemma4 per-layer-input 640 GiB broadcast.** Orthogonal to CP but hit during this work. Older transformers versions of `Gemma4TextModel.get_per_layer_inputs` call the fallback path `[B, S, 1, H] == [1, 1, vocab, H]` → `[B, S, vocab, H]` bool when `input_ids is None`, which is ~640 GiB on Gemma4-E4B. `_patch_gemma4_per_layer_inputs` in `infrastructure.py:109-144` replaces it with a `(inputs_embeds @ embed_tokens.T).argmax(-1)` reverse lookup (~1 GiB, and correct as long as Gemma's `inputs_embeds = embed_tokens(ids) * sqrt(H)` invariant holds, which it does on all Gemma 4 variants).

Happy to split any of the above into their own PRs if the reviewer prefers smaller diffs.

HuiyingLi:
> ## 4-node smoke-test results (32x H100-80GB)
> End-to-end CP verification on 4 nodes (32 x H100-80GB) with Gemma4-E4B MoE (~7B total, 4B activated).
> 
> ### Loss parity: CP=1 vs CP=2 (matched per-GPU shard size)
> To verify the ring-attention + flex_attention path is mathematically equivalent to the non-CP baseline, I ran two matched configurations (same per-GPU `S_local`, different `cp_size`):
> 
> step	CP=1, seq=2048	CP=2, seq=4096	Δ
> 0	16.9419	17.1810	0.24
> 1	10.1633	10.4407	0.28
> 2	8.4085	8.7233	0.31
> 3	7.1558	7.4028	0.24
> Trajectories overlap within noise. Residual Δ is from different global batch sizes (32 vs 16) sampling different mock examples per step plus reduction-order FP differences. `grad_norm` tracks similarly: `1256 -> 68` (CP=1) vs `1136 -> 77` (CP=2).
> 
> ### Topology / infra notes
> * `cp_size=2, dp_size=16` (CP within node over NVLink, DP across nodes over TCP)
> * `activation_checkpointing: true`, `PYTORCH_ALLOC_CONF=expandable_segments:True`
> * `NCCL_IB_HCA=^=all` + `NCCL_NET_PLUGIN=none` were needed to work around a NCCL 2.27.5 bug in the K8s container where the proxy thread kept attempting IB-type handshakes even with `NCCL_IB_DISABLE=1`. Not related to this PR.
> 
> ### 31B dense: not yet validated
> I have not yet run the 31B dense config end-to-end. Earlier attempts OOM'd during backward on a 32x 80GB cluster with pure FSDP+CP (weights + Adam state + flex_attention backward workspace together exceed per-GPU budget). 31B long-context training almost certainly needs TP alongside CP to shard the hidden dimension as well; that combination is not covered by this PR and will be a follow-up. The 31B yamls in this PR are provided as configuration templates; they are not expected to run as-is without either TP or a larger cluster.
> 
> ### Why all-gather + flex_attention instead of ring-attention
> The all-gather variant here is deliberately simple: each rank gathers full K/V along the sequence dimension then runs a compiled `flex_attention` with a per-rank causal block mask. We chose this over a ring-attention implementation for two reasons:
> 
> 1. **Correctness under mixed DTensor/plain-tensor paths.** PyTorch's native `context_parallel` DTensor dispatch into SDPA kept hitting a "mixed torch.Tensor and DTensor" crash in the attention inner_fn (the compiled decoder propagates a local-tensor Q while K/V return as DTensors after all-gather). All-gather-then-local-SDPA avoids the mixed path entirely and works on every PyTorch version we tested.
> 2. **Reused infra.** flex_attention already gives us Flash-style O(N) memory via its Triton tiled kernel and accepts arbitrary per-rank block masks; we just had to wrap it in `torch.compile` and pass `kernel_options` to shrink block sizes when `head_dim >= 256` (the Gemma4 31B global-attention-layer variant has `head_dim=512`).
> 
> A proper ring-attention implementation would save the `S_full` K/V memory per rank (O(N) in seq instead of O(N/CP)), and could let us use Flash Attention directly (because each ring chunk is a standard causal attention). That is the right long-term direction but is a larger change; keeping it out of this PR so the diff stays reviewable.
> 
> ### Known limitations (follow-up PR candidates)
> * `create_block_mask` materializes a dense `[S_local, S_full]` bool mask, which allocates ~1 GiB at `seq=16K` and OOMs before flex_attention even starts. Needs a sparse or streaming construction.
> * CE loss materializes a fp32 `[B*S_local, 262144]` logit tensor (~4 GiB at `seq=8K`). Chunked CE would unblock longer contexts.
> * Ring-attention path (see above) to eliminate the `S_full` K/V memory footprint.

- We can do 31B in a separate PR. Do you mind removing the 31B template config

HuiyingLi:
> ## 4-node smoke-test results (32x H100-80GB)
> End-to-end CP verification on 4 nodes (32 x H100-80GB) with Gemma4-E4B MoE (~7B total, 4B activated).
> 
> ### Loss parity: CP=1 vs CP=2 (matched per-GPU shard size)
> To verify the ring-attention + flex_attention path is mathematically equivalent to the non-CP baseline, I ran two matched configurations (same per-GPU `S_local`, different `cp_size`):
> 
> step	CP=1, seq=2048	CP=2, seq=4096	Δ
> 0	16.9419	17.1810	0.24
> 1	10.1633	10.4407	0.28
> 2	8.4085	8.7233	0.31
> 3	7.1558	7.4028	0.24
> Trajectories overlap within noise. Residual Δ is from different global batch sizes (32 vs 16) sampling different mock examples per step plus reduction-order FP differences. `grad_norm` tracks similarly: `1256 -> 68` (CP=1) vs `1136 -> 77` (CP=2).
> 
> ### Topology / infra notes
> * `cp_size=2, dp_size=16` (CP within node over NVLink, DP across nodes over TCP)
> * `activation_checkpointing: true`, `PYTORCH_ALLOC_CONF=expandable_segments:True`
> * `NCCL_IB_HCA=^=all` + `NCCL_NET_PLUGIN=none` were needed to work around a NCCL 2.27.5 bug in the K8s container where the proxy thread kept attempting IB-type handshakes even with `NCCL_IB_DISABLE=1`. Not related to this PR.
> 
> ### 31B dense: not yet validated
> I have not yet run the 31B dense config end-to-end. Earlier attempts OOM'd during backward on a 32x 80GB cluster with pure FSDP+CP (weights + Adam state + flex_attention backward workspace together exceed per-GPU budget). 31B long-context training almost certainly needs TP alongside CP to shard the hidden dimension as well; that combination is not covered by this PR and will be a follow-up. The 31B yamls in this PR are provided as configuration templates; they are not expected to run as-is without either TP or a larger cluster.
> 
> ### Why all-gather + flex_attention instead of ring-attention
> The all-gather variant here is deliberately simple: each rank gathers full K/V along the sequence dimension then runs a compiled `flex_attention` with a per-rank causal block mask. We chose this over a ring-attention implementation for two reasons:
> 
> 1. **Correctness under mixed DTensor/plain-tensor paths.** PyTorch's native `context_parallel` DTensor dispatch into SDPA kept hitting a "mixed torch.Tensor and DTensor" crash in the attention inner_fn (the compiled decoder propagates a local-tensor Q while K/V return as DTensors after all-gather). All-gather-then-local-SDPA avoids the mixed path entirely and works on every PyTorch version we tested.
> 2. **Reused infra.** flex_attention already gives us Flash-style O(N) memory via its Triton tiled kernel and accepts arbitrary per-rank block masks; we just had to wrap it in `torch.compile` and pass `kernel_options` to shrink block sizes when `head_dim >= 256` (the Gemma4 31B global-attention-layer variant has `head_dim=512`).
> 
> A proper ring-attention implementation would save the `S_full` K/V memory per rank (O(N) in seq instead of O(N/CP)), and could let us use Flash Attention directly (because each ring chunk is a standard causal attention). That is the right long-term direction but is a larger change; keeping it out of this PR so the diff stays reviewable.
> 
> ### Known limitations (follow-up PR candidates)
> * `create_block_mask` materializes a dense `[S_local, S_full]` bool mask, which allocates ~1 GiB at `seq=16K` and OOMs before flex_attention even starts. Needs a sparse or streaming construction.
> * CE loss materializes a fp32 `[B*S_local, 262144]` logit tensor (~4 GiB at `seq=8K`). Chunked CE would unblock longer contexts.
> * Ring-attention path (see above) to eliminate the `S_full` K/V memory footprint.

Thank you! @khazic I'm curious how would you do proper ringattn if flashattn wont work? 
What about TE attn? But that prob needs rewriting the model as custom model.

khazic:
Thanks @HuiyingLi and the review bot — addressed the feedback in `92e88df8`:

- **Removed the 31B-dense example configs** (`gemma4_31b_cp2*.yaml`) as requested. 31B-long-context needs CP + TP and will be a separate PR.
- **Updated the stale docstring** in `make_cp_batch_and_ctx` that still described the old DTensor-wrap path — CP communication now lives entirely inside `_cp_sdpa` via explicit K/V all-gather + compiled flex_attention.
- **Added a guarding comment** on the `torch.no_grad()` pre-embed wrapper in `finetune.py` spelling out that it assumes `freeze_embeddings: true` and will silently drop embedding gradients if someone unfreezes them. Happy to replace the `no_grad` with a conditional (`freeze_embeddings: true` → `no_grad`, otherwise regular) in a follow-up if you'd prefer.
- **Added unit tests for `prepare_inputs_embeds_for_cp`** covering both the text-only path and the image-feature `masked_scatter` path.
- **Fixed the ruff errors** that broke the linting job (unused local in `infrastructure.py`, three I001 import-order issues in `cp_utils.py`).
- **Tightened the two DTensor tests** so they only run when a process group is actually initialized, and dropped a dead redundant `with patch(...)` wrapper plus an unused `Replicate` import that the bot flagged.

`ruff check` now passes locally; CI should be green aside from any pre-existing flakes.

khazic:
### Correction on wording — this PR does **not** implement ring attention

I realized the phrasing in my earlier comment (["To verify the ring-attention + flex_attention path..."](https://github.com/NVIDIA-NeMo/Automodel/pull/1914#issuecomment-4281939401)) is misleading. Apologies for the confusion. To be precise:

**What this PR actually implements is an _all-gather_ + flex_attention variant, not ring attention.**

`_cp_sdpa` in `cp_utils.py`:

```python
key_full = _all_gather_seq(key)    # single all-gather -> every rank holds full K
val_full = _all_gather_seq(value)  # single all-gather -> every rank holds full V
# Each rank then has: local Q (S_local)  +  full K, V (S_full)
out = _get_compiled_flex_attn()(query, key_full, val_full, block_mask=per_rank_causal_mask)
```

Differences from a real ring-attention implementation:

| | this PR (all-gather) | proper ring attention |
|---|---|---|
| per-rank K/V memory | **O(S_full)** | O(S_local) |
| communication | single all-gather | N−1 P2P send/recv (can overlap with compute) |
| mask | per-rank causal **float** mask (non-standard, Q/K not same length) | standard `is_causal=True` / no-mask per chunk |
| attention kernel | flex_attention (Flash rejects our mask) | Flash Attention works directly |

### Why Flash Attention is not reachable from this PR

Not because of Gemma 4, but because of the all-gather design I chose:

- Flash accepts only `is_causal=True` (Q and K same length) or no mask — **it rejects any custom float mask**.
- After all-gather, Q has length `S_local` while K/V have length `S_full`, so Q and K are **not the same length**.
- The mask each rank needs is "rectangular causal with a rank offset": `Q_local[i]` may attend `K[j]` iff `r * S_local + i >= j`. That is not expressible in Flash's API.

So I had to fall back to `flex_attention`, which accepts arbitrary block masks. **The Flash incompatibility is a consequence of the all-gather architecture, not of Gemma 4.**

### Why all-gather and not ring in this PR

Ring attention would have let us use Flash directly (each ring chunk is a standard causal or fully-open attention on equal-length Q/K/V), and would have reduced per-rank K/V memory from O(S_full) to O(S_local) — which is the whole point of CP. I chose all-gather anyway because:

1. Implementing ring attention requires an N−1-step P2P pipeline with online-softmax `(O, L)` state merge, a custom autograd `Function` for the ring backward pass, and (for good utilization) zig-zag load balancing. That is a large, from-scratch component.
2. PyTorch's existing `torch.distributed.tensor.experimental._attention.context_parallel_attention` implements ring internally but requires Q/K/V to all be DTensors. In the versions we tested, that path trips a "mixed `torch.Tensor` and DTensor" assert in the compiled decoder's inner_fn (Q is produced as a plain local tensor, K/V come back from DTensor all-gather, and SDPA dispatch cannot reconcile the two). The cleanest way to sidestep that was to leave DTensor out of the picture entirely — which forced manual all-gather, which forced a custom mask, which forced flex_attention.
3. Scope — keeping the diff reviewable.

### What should be different / next PR

A proper ring-attention implementation is the right follow-up. Concretely:
- Per-step P2P ring (can use `torch.distributed.nn.functional.send` to get autograd) with Flash Attention as the per-chunk kernel.
- Online-softmax `(O, L)` merge across ring steps.
- Zig-zag Q/K split for load balance under causal masks.

Once that lands we get: O(S_local) K/V memory, Flash kernel throughput, and we can drop the `create_block_mask` dense-materialization dependency that currently OOMs at seq ≥ 16K.

Happy to take that on as a separate PR. Sorry again for the sloppy wording — the actual code in this PR is the simpler all-gather variant.

khazic:
@huiyinbg Could you help run a CI test on this PR?\n\nConfig: `examples/vlm_finetune/gemma4/gemma4_E4B_cp4_mock_image.yaml`\nModel: `google/gemma-4-E4B-it`, CP=4, 4 nodes × 8 H100-80GB, multimodal mock dataset (1 image/sample, max_length=8192).\n\nLocal validation (4 nodes, 32×H100-80GB) ran 21 steps, loss 17.29→3.98, no OOM or NaN. Thanks!

khazic:
@HuiyingLi FYI — the follow-up PR for Gemma4 31B CP+TP support will be submitted in a few days. I'm currently a bit occupied with a part-time food delivery job. Thanks for your patience!

examples/vlm_finetune/gemma4/k8s_gemma4_cp4_test.yaml
HuiyingLi, line 1: Could we delete this one, we only include automodel configs in this dir. Thank you!

khazic: done it

nemo_automodel/components/distributed/cp_utils.py
```diff
@@ -125,52 +125,283 @@ def _self_attn_pre_forward_hook(_module, module_args, module_kwargs):
...
-      - Everything after (O proj + residual + MLP) is compiled and fused.
+    """Inject CP-aware SDPA into self_attn modules for CP>1 correctness.
+
+    Replaces F.scaled_dot_product_attention during each self_attn forward with a
+    manual all-gather ring-attention implementation that avoids PyTorch's internal
+    DTensor-based ring attention (which has a mixed torch.Tensor/DTensor bug in some
+    PyTorch versions).
+
+    Algorithm (all-gather variant):
+      1. All-gather K and V from all CP ranks → full-sequence K_full, V_full.
```
akoumpa: Hi @khazic , thank you for the contribution, i wonder if this will replicate compute / mem for the layers? Since this is used in multiple models, I feel it's a bit risky change.

akoumpa: my recommendation would be to switch to the well-established CP from transformerengine, that would help us derisk and also bring the great TE perf to gemma4. Please let me know what you think @khazic

khazic: Hi @akoumpa, I understand your point. I am fully capable of bringing TE CP support into automodel. I just need a few days to handle some personal stuff (I deliver food for a living between PRs), and I will send that PR in a few days. Thanks for the direction!

examples/vlm_finetune/gemma4/gemma4_E4B_cp4_mock_image.yaml
```diff
@@ -0,0 +1,110 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
athitten: @khazic mind renaming this to `gemma4_4b_cp4_mock_image.yaml` ? Just to be consistent with other yamls and not confuse users

examples/vlm_finetune/gemma4/gemma4_E4B_cp4_mock_image.yaml
```diff
@@ -0,0 +1,110 @@
...
+  timeout_minutes: 30
+
+rng:
+  _target_: nemo_automodel.components.training.rng.StatefulRNG
+  seed: 42
+  ranked: true
+
+model:
+  _target_: nemo_automodel.NeMoAutoModelForImageTextToText.from_pretrained
+  pretrained_model_name_or_path: /llm-align/open_models/gemma4/gemma-4-E4B-it
```
athitten: Should add the HF model-id

nemo_automodel/_transformers/infrastructure.py
```diff
@@ -630,17 +630,18 @@ def apply_model_infrastructure(
...
+            attach_linear_attn_position_hooks,
         )
 
-        is_compile_enabled = isinstance(model_wrapper, FSDP2Manager) and model_wrapper.enable_compile
-        cp_mesh = mesh.device_mesh["cp"] if is_compile_enabled else None
+        cp_mesh = mesh.device_mesh["cp"]
 
         model_parts = model.parts if hasattr(model, "parts") else [model]
         for mp in model_parts:
+            mp._cp_enabled = True
```
hemildesai: This will set it for all models right?

HuiyingLi:
Good catch — `_force_repeat_kv_for_sdpa` was dead (the ring CP path owns GQA via `enable_gqa` in its flex/SDPA call, never HF's `sdpa_attention`). Removed it, its two tests, and the now-unused imports in dadd6df.

/claude review

nemo_automodel/components/moe/parallelizer.py
```diff
@@ -478,29 +478,41 @@ def apply_cp(model: torch.nn.Module, cp_mesh: DeviceMesh, cp_comm_type: str = "p
     # Prefer nested text modules when present (VLM models)
     _model = get_text_module(_model)
 
-    # Set model-level flag so the forward pass can null out attention_mask.
+    # Set CP flags so wrapper and text-module forward paths can prepare
+    # full-sequence embeddings and null out local attention masks.
     # With CP the mask is not sharded along the sequence dim and TE asserts
     # "Padding mask not supported with context parallelism!".
+    model._cp_enabled = True
```
hemildesai, line 485: Where is this used?

HuiyingLi: this one is used by gemma4. the one below by qwen35. I have another PR on qwen35, will unify the two flags here. They are checks before vlm CP, if the placeholder img tokens are already pre-embd.

nemo_automodel/_transformers/infrastructure.py
HuiyingLi: @hemildesai  deleted, thank you!

hemildesai (approved):

PR #1928, opened 2026-04-21
fix: break FA2 retry loop for packed sequences on FA2-unsupported models

adil-a:
## Summary

- Fixes the 5-retry FA2 loop hit by `nemotron_nano_4b_squad` CI (e.g. [job 301287481](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/301287481)), where `NemotronHForCausalLM` + `packed_sequence_size=1024` produced 5 identical `does not support Flash Attention 2` failures before terminating.
- `_apply_preload_overrides` forces `attn_implementation="flash_attention_2"` when `has_packed_sequence=True` and `cp_size==1`. It runs at the start of every `_build_model` call, including retries, so the fallback cascade in the `except ValueError` handler (FA2 → SDPA → eager) never sticks — each retry resets to FA2 and hits the same error.
- New guard: if `has_packed_sequence=True` and `attn_implementation` has already fallen back off FA2/FA3 (i.e. we're on a retry whose previous iteration failed FA2), raise a clear `ValueError` naming the model and pointing at the two fixes (disable packed sequences, or use an FA2/FA3-supporting model) instead of re-forcing FA2.
- Scope check: only `examples/llm_finetune/nemotron/nemotron_nano_4b_squad.yaml` combines a NemotronH model with `packed_sequence_size > 0` in the repo, so no other CI configs change behavior.

## Test plan

- [x] New unit tests in `tests/unit_tests/_transformers/test_kernel_patches.py` covering the guard (10 cases: first-call forcing, None attn, retry-with-sdpa, retry-with-eager, error message contents, FA3 pass-through, no-packed passthrough, CP forces SDPA, TP disables Liger, packed+CP errors).
- [x] All 78 existing `tests/unit_tests/_transformers/test_auto_model.py` tests still pass.
- [x] Local 8×H100 repro of the failing CI job: pre-fix log showed 104 `NemotronH ... FA2` errors (retry loop across 8 ranks); post-fix shows 16 (one per rank, chained) and the new clear error surfaces as the final exception.
- [x] `ruff check`, `ruff format --check`, and `lint-imports --no-cache` all clean.

### DIFF
--- a/nemo_automodel/_transformers/auto_model.py
+++ b/nemo_automodel/_transformers/auto_model.py
@@ -321,6 +321,11 @@ class _BaseNeMoAutoModelClass(_BaseAutoModelClass):
                 has_packed_sequence,
                 attn_implementation,
                 use_liger_kernel,
+                model_name_or_path=(
+                    pretrained_model_name_or_path_or_config
+                    if isinstance(pretrained_model_name_or_path_or_config, str)
+                    else None
+                ),
             )
         device = torch.cuda.current_device()
 
--- a/nemo_automodel/_transformers/kernel_patches.py
+++ b/nemo_automodel/_transformers/kernel_patches.py
@@ -209,7 +209,15 @@ def _get_next_fallback_attn(attn_implementation: str) -> str:
         return priorities[0]
 
 
-def _apply_preload_overrides(tp_size, cp_size, has_packed_sequence, attn_implementation, use_liger_kernel):
+def _apply_preload_overrides(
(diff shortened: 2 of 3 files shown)


adil-a:
Replacing with a simpler fix: disable packed sequences in the yaml (NemotronH doesn't support FA2, and FA2 is required for packed sequences). The 'clearer error' approach in this PR doesn't actually let the CI test pass.

Issue #1883, opened 2026-04-16
NaN loss with SDPA attention when local_batch_size > 1 on Gemma 4 4B


HuiyingLi:
## Summary

Training `gemma4_4b.yaml` with `attn_implementation=sdpa` and `local_batch_size=2` produces NaN loss early in training.

This is a known limitation in transformers, opening this issue to help with tracking.
- https://github.com/huggingface/transformers/issues/45201

## Environment

- Model: `google/gemma-4-E4B-it`
- Config: `examples/vlm_finetune/gemma4/gemma4_4b.yaml`
- `transformers`: 5.5.0.dev0
- `torch`: 2.11.0a0+eb65b36914.nv26.02 (CUDA 13.1)
- Hardware: 8× H100 80GB, single node
- Precision: bf16, FSDP2

## Repro

```bash
torchrun --nproc-per-node=8 nemo_automodel/recipes/vlm/finetune.py \
  -c examples/vlm_finetune/gemma4/gemma4_4b.yaml \
  --step_scheduler.max_steps=30 \
  --step_scheduler.local_batch_size=2 \
  --step_scheduler.global_batch_size=16 \
  --model.attn_implementation=sdpa
```

## Observed

`grad_norm` becomes NaN at step 11, loss NaN from step 12 onward:

```
step 10 | loss 2.6030 | grad_norm  9.9375
step 11 | loss 2.3224 | grad_norm     nan
step 12 | loss    nan | grad_norm     nan
...
[val] step 29 | loss nan
```

wandb: https://wandb.ai/Nemo-automodel/gemma4-4b-attn-lbs/runs/4yb1b74k

<img width="934" height="240" alt="Image" src="https://github.com/user-attachments/assets/08fa8b83-1fe7-468c-a49b-601f13f69b72" />

khazic:
I previously ran into the same issue when using your scripts, and I already fixed that bug. However, my environment uses PyTorch 2.10 and Transformers 5.5.3, so I’ll test it in your environment to see exactly where the problem is.

athitten:
Note to avoid confusion: This issue is orthogonal to the [lack of FA support for gemma4](https://github.com/huggingface/transformers/issues/45201). They are two independent issues.
cc: @akoumpa @HuiyingLi

athitten:
Can't be reproduced with transformers v5.5 version. Closing for now, to open a new issue if we encounter NaN again.

PR #1985, opened 2026-04-22
fix: guard against zero label tokens causing NaN loss in VLM training

khazic:
# What does this PR do ?

Add a defensive guard against division by zero in \`MaskedCrossEntropy\` when \`num_label_tokens=0\`.

# Changelog

- \`masked_ce.py\`: return \`loss * 0.0\` instead of dividing by zero when \`num_label_tokens=0\`
- \`finetune.py\`: guard PP reporting loss normalization against \`num_label_tokens=0\`
- \`test_masked_ce.py\`: add regression test for the empty-supervision case

**Pre checks**:

# Additional Information

Related to #1883.

I attempted to reproduce the SDPA NaN described in #1883 using the exact repro command on transformers 5.5.0 / PyTorch 2.11.0 (CUDA 13.1) / 8xH100, but could not reproduce it. The issue was filed against \`transformers==5.5.0.dev0\`, and I believe the underlying SDPA masking bug has since been fixed in the stable 5.5.0 release.

While investigating, I noticed that \`MaskedCrossEntropy\` has no guard for \`num_label_tokens=0\`, which produces \`NaN\` via division by zero. In the **multi-GPU training path**, \`num_label_tokens\` is all-reduced across DP ranks, so hitting zero in practice would require every sample across every rank to have no valid labels simultaneously -- extremely unlikely. However, two paths are genuinely exposed:

1. **Validation loop**: \`num_label_tokens\` is computed per-batch with no all-reduce. If a single validation batch has all labels set to \`-100\` (e.g. due to label-building failure, as already noted in the \`collate_fns.py\` comment: *"may produce nan loss"*), \`MaskedCrossEntropy\` produces NaN, which propagates into the reported validation loss.
2. **Single-GPU training**: the all-reduce is a no-op (world size = 1), so a single bad sample is enough to trigger it.

This PR adds a minimal defensive guard so that an empty-supervision batch contributes zero loss instead of NaN, keeping training and validation metrics clean.

### DIFF
--- a/nemo_automodel/components/loss/masked_ce.py
+++ b/nemo_automodel/components/loss/masked_ce.py
@@ -82,5 +82,7 @@ class MaskedCrossEntropy(nn.Module):
         loss = F.cross_entropy(logits, labels, reduction=self.reduction)
         if num_label_tokens is not None:
             assert self.reduction == "sum", "num_label_tokens is only supported when reduction is 'sum'"
+            if num_label_tokens == 0:
+                return loss * 0.0
             loss = loss / num_label_tokens
         return loss
--- a/nemo_automodel/recipes/vlm/finetune.py
+++ b/nemo_automodel/recipes/vlm/finetune.py
@@ -1133,7 +1133,7 @@ class FinetuneRecipeForVLM(BaseRecipe):
         if self.pp_enabled:
             # PP uses sum reduction per microbatch (no internal normalization).
             # Divide by num_label_tokens to get the mean loss, same as non-PP.
-            reporting_loss = reporting_loss / num_label_tokens
+            reporting_loss = reporting_loss / num_label_tokens if num_label_tokens > 0 else reporting_loss * 0.0
             reporting_loss = reporting_loss.float().to(self.dist_env.device)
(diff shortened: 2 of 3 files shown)


khazic:
**Additional reproduction note**

While the trigger condition is unlikely with the default `model_max_length=131072`, we confirmed the bug fires in practice under high-resolution multi-image inputs combined with a shorter `max_length` setting.

In that scenario, the image tokens occupy most of the sequence budget and right-truncation completely removes the assistant response. The resulting sample has all labels set to `-100`, so `num_label_tokens=0` in the validation loop (which has no all-reduce guard). `MaskedCrossEntropy` then divides by zero, producing `loss: tensor(nan)`, and the NaN propagates into the reported validation loss via `total_loss += local_loss.item() * num_label_tokens` (`nan * 0 = nan` in Python float arithmetic).

Multi-GPU training is not affected in practice because `num_label_tokens` is all-reduced across DP ranks there — a single bad sample cannot zero out the global count. The exposed paths are the **validation loop** (per-batch count, no all-reduce) and **single-GPU training** (all-reduce is a no-op).

athitten (approved):
LGTM, TYSM @khazic !

Issue #1917, opened 2026-04-20
use lazy map for dataset preprocessing


akoumpa:
**Is your feature request related to a problem? Please describe.**
AM uses .map(fn) to transform datasets (e.g., apply template, tokenize etc). Unfortunately, this happens during startup and runs over the whole dataset. In addition, the current caching functionality often fails, incurring additional delay on repeated runs.

```
2026-04-20 12:09:22 | INFO | httpx | HTTP Request: HEAD https://huggingface.co/akoumpa/Devstral-Small-2-24B-Instruct-2512-BF16/resolve/main/config.json "HTTP/1.1 307 Temporary Redirect"
2026-04-20 12:09:22 | INFO | httpx | HTTP Request: HEAD https://huggingface.co/api/resolve-cache/models/akoumpa/Devstral-Small-2-24B-Instruct-2512-BF16/6d6f531753869fcd8cbbaefa7ee94b93966d2d04/config.json "HTTP/
1.1 200 OK"
2026-04-20 12:09:23 | INFO | nemo_automodel._transformers.auto_tokenizer | Using custom tokenizer MistralCommonBackend for model type 'mistral3'
2026-04-20 12:09:23 | INFO | httpx | HTTP Request: GET https://huggingface.co/api/models/akoumpa/Devstral-Small-2-24B-Instruct-2512-BF16/tree/main?recursive=true&expand=false "HTTP/1.1 200 OK"
2026-04-20 12:09:24 | INFO | httpx | HTTP Request: HEAD https://huggingface.co/akoumpa/Devstral-Small-2-24B-Instruct-2512-BF16/resolve/main/tekken.json "HTTP/1.1 302 Found"
2026-04-20 12:09:24 | INFO | mistral_common.tokens.tokenizers.tekken | Non special vocabulary size is 130072 with 1000 special tokens.
2026-04-20 12:09:24 | INFO | httpx | HTTP Request: HEAD https://huggingface.co/datasets/rajpurkar/squad/resolve/main/README.md "HTTP/1.1 307 Temporary Redirect"
2026-04-20 12:09:24 | INFO | httpx | HTTP Request: HEAD https://huggingface.co/api/resolve-cache/datasets/rajpurkar/squad/7b6d24c440a36b6815f21b70d25016731768db1f/README.md "HTTP/1.1 200 OK"
[...]

PR #2007, opened 2026-04-23
feat: lazy preprocessing for text datasets

edjson:
# What does this PR do ?

Replace `.map(fn)` dataset processing with `LazyMappedDataset`. This processes items on the fly, and caches results.

# Changelog

- Added `nemo_automodel/components/datasets/lazy_mapped_dataset.py` with `LazyMappedDataset` class.

- Updated `nemo_automodel/components/datasets/llm/xlam.py` and `nemo_automodel/components/datasets/llm/squad.py` to          use `LazyMappedDataset`.

- Updated `tests/unit_tests/datasets/llm/test_xlam.py` to match new `LazyMappedDataset` behavior.

**Pre checks**:

# Additional Information

- Related to #1917


nemo_automodel/components/datasets/lazy_mapped_dataset.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+
+    def __init__(self, dataset, map_fn, cache_size=None):
+        self._dataset = dataset
+        self._map_fn = map_fn
+
+        if cache_size is None:
+            cache_size = len(dataset)
+
+        self._cache_size = cache_size
```
akoumpa, line 50: Hi @edjson , do you know if datasets provides any facility for caching on disk that we could leverage here? I realized today, that inevitably, this will cache in memory, which for large datasets may be trouble. Please let me know what you think. I know the LRU is already in place, and should help mitigate the problem, but I'm wondering whether there's additional options we could explore. Thanks.

edjson: Hello,
Thank you for raising this. HuggingFace does have some disk caching methods we can leverage including: 

- **Enable streaming.** If datasets are too large we can use streaming to download and process data on the fly, which is good for large data sets.

- **Memory Mapping.** `load_dataset` downloads and caches data to the disk, however we can set the location for the arrow cache file in `map()` that way we have it written to a SSD instead of a HDD. We can set `HF_DATASETS_CACHE` as an environment variable to direct all caching to a specific path for a simpler system-wide approach. Also the `num_proc`  parameter in `.map()` when using multiprocessing, each worker loads its own shard into memory simultaneously multiplying RAM usage so specifying the cache location is worth exploring.

- **Disable in-memory caching.** Making sure that `keep_in_memory` is set to false, that way datasets will not load data into memory. There is a global toggle in `datasets.disable_caching()`/`datasets.enable_caching()` if we want full control over when caching happens

- **Intermediately save to the disk.** We can use `.save_to_disk` after processing to free RAM space on frequent runs. This preserves the Arrow format so re-loading is fast and does not require reprocessing.

Given the concern with large datasets, streaming and Arrow cache redirection may be the best starting points. I would be happy to open a separate PR to explore adding an optional `cache_dir` parameter to the dataset functions. That way users would pass a `cache_file_name` to `.map()` to specify a cache location in the YAML config.

akoumpa: hmm, I'm thinking maybe we go with the LRU and bound it to a number of samples e.g., 10k samples or 64MB of RAM, since this is only used for text and yes consider saving the processed text on a disk cache file.

I'm looking for the generic solution/abstraction that can be applied to both text datasets but also text + multimedia datasets as necessary in VLM finetuning, that's my main motivation here, but I agree, we don't have to fix everything in one go. for this one, my only ask is the bounded LRU + the optional on-disk cache for processed examples, and I'd prefer to keep that logic inside `LazyMappedDataset` rather than delegating to HF's `.map(cache_file_name=...)`, so the same wrapper works for non-HF / VLM datasets down the line.

streaming and `HF_DATASETS_CACHE` I think we can leave out of scope here. and yes, happy for the broader `cache_dir` exploration to land in a follow-up PR as you suggested, thanks!

nemo_automodel/components/datasets/lazy_mapped_dataset.py
```diff
@@ -0,0 +1,102 @@
...
+        map_fn: A callable that accepts a single example and returns the
+            transformed example.
+        cache_size: Number of processed items to cache. Defaults to the full
+            dataset size. Set to 0 to disable caching.
+
+    Returns:
+        A map-style dataset that applies map_fn lazily on each item access.
+    """
+
+    def __init__(self, dataset, map_fn, cache_size=None):
```
akoumpa:
```suggestion
    def __init__(self, dataset, map_fn, cache_size=10000):
```

nemo_automodel/components/datasets/lazy_mapped_dataset.py
```diff
@@ -0,0 +1,102 @@
...
+    Dataset wrapper that applies a transform function on-the-fly instead of
+    preprocessing the whole dataset upfront with .map(fn).
+
+    Args:
+        dataset: Any object that supports ``__len__`` and ``__getitem__``
+            (e.g. a Hugging Face ``datasets.Dataset``).
+        map_fn: A callable that accepts a single example and returns the
+            transformed example.
+        cache_size: Number of processed items to cache. Defaults to the full
+            dataset size. Set to 0 to disable caching.
```
akoumpa:
```suggestion
        cache_size: Number of processed items to cache. Defaults to the 10k
            dataset samples. Set to 0 to disable caching or None to cache all.
```

akoumpa: Hi @edjson , I'm going to apply this + the default value in the constructro, if that's ok with you, to unblock CI and merge this

akoumpa (approved):
Thanks a lot @edjson

PR #2045, opened 2026-04-24
fix: Unsafe deserialization via `torch.load` on dataset cache files

tomaioo:
## Summary

Security: Unsafe deserialization via `torch.load` on dataset cache files

## Problem

**Severity**: `High` | **File**: `nemo_automodel/components/datasets/diffusion/text_to_image_dataset.py:L45`

The dataset loader deserializes `.pt`/cache files using `torch.load(cache_file, map_location="cpu")` without `weights_only=True` or an equivalent safe format. If an attacker can modify cache files or metadata paths, this can lead to arbitrary code execution during loading because `torch.load` uses pickle by default.

## Solution

Use `torch.load(..., weights_only=True)` where possible, or migrate cache serialization to a non-executable format (e.g., safetensors). Validate that `cache_file` resolves under an expected trusted directory before loading.

## Changes

- `nemo_automodel/components/datasets/diffusion/text_to_image_dataset.py` (modified)


akoumpa (approved):
Thank you @tomaioo !

PR #2058, opened 2026-04-27
fix(vlm): broaden fake-image vision-token detection beyond Qwen

khazic:
## Summary

The fake-image hack injects a 56×56 white image into pure-text samples so
the visual encoder stays active under FSDP / DeepSpeed Zero3 (every parameter
must participate in the all-gather collective, or training hangs).  Its
correctness depends on masking the resulting vision tokens out of attention
via `mask_fake_vision_tokens_batch` / `mask_fake_vision_tokens_single`.

The previous `_get_vision_token_ids` was copy-pasted from Qwen and only
recognised:
- `image_token_id` / `video_token_id` processor attributes
- four `<|vision_start|>`-family token strings

For non-Qwen models this silently returned an empty set, so the masking
step was a no-op and fake-image vision tokens remained visible to attention.

### Confirmed impact per model family

| Model family | Collate path | Status before this fix |
|---|---|---|
| Qwen2.5-VL / Qwen3-VL | `qwen2_5_collate_fn` | ✅ Working — `image_token_id` found |
| **Kimi K2.5 / KimiVL** | `kimi_k25_vl_collate_fn` / `kimi_vl_collate_fn` | ❌ **Full bug** — `media_placeholder_token_id` not scanned, `<\|media_pad\|>` not in string list |
| **Gemma3 / Gemma4** | `default_collate_fn` (via `gemma4_prefix_collate_fn`) | ❌ **Full bug** — `image_token_id` only accessible via `config.attribute_map`, not on processor directly; tokenizer uses `<image>` not `<\|image_pad\|>` |
| **Pixtral / Mistral3** | `default_collate_fn` | ⚠️ **Partial bug** — `image_token_id` found so `[IMG]` is masked, but `image_break_token_id` (`[IMG_BREAK]`) and `image_end_token_id` (`[IMG_END]`) were missed |
| LLaVA-OneVision | `llava_onevision_collate_fn` | ❓ Uncertain — Qwen2-based tokenizer likely has `<\|image_pad\|>` at id 151655, but actual image token is id 151646; needs real processor to verify |

Both trigger paths are affected:
- **`__getitem__` path** (`datasets.py:963–1063`): unconditionally active for any pure-text sample — no `preload_media` flag required
- **`collate_fn` path**: activated when `_injected_fake=True` (requires `preload_media=True`)

## Changes

**`nemo_automodel/components/datasets/vlm/fake_image.py`**
- Extend `_get_vision_token_ids` to scan three sources:
  1. 14 known processor / `processor.config` attribute names, covering
     Qwen (`image_token_id`, `vision_start/end_token_id`, …),
     Kimi (`media_placeholder_token_id`),
     LLaVA / Mistral3 / Gemma3 (`image_token_index`, `boi/eoi_token_index`, …),
     Pixtral (`image_break_token_id`, `image_end_token_id`)
  2. 13 curated vision token strings (`<\|media_pad\|>`, `<image>`, `[IMG]`, …)
  3. Keyword fuzzy scan over `tokenizer.added_tokens_decoder` for automatic
     coverage of future or custom VLMs
- Add `_warn_no_vision_tokens`: emit a **one-time `logger.warning`** per
  processor class when detection finds nothing, turning a silent no-op into
  a visible signal

**`tests/unit_tests/datasets/vlm/test_fake_image.py`** (new file)
- 22 unit tests covering Qwen / Kimi / LLaVA / Mistral / Gemma3 detection
  paths, `processor.config` attribute path, `added_tokens_decoder` fuzzy
  match, robustness edge cases (raising tokenizer, non-int attribute),
  warning emission, and one-shot dedup — all mocked, no GPU or model weights
  required

## Test plan
- [x] `ruff format` / `ruff check` clean
- [x] 22/22 unit tests pass on GPU server (Python 3.12, pytest 9.0.2)
- [ ] CI green on `NVIDIA-NeMo/Automodel`

### DIFF
--- a/nemo_automodel/components/datasets/vlm/fake_image.py
+++ b/nemo_automodel/components/datasets/vlm/fake_image.py
@@ -28,8 +28,11 @@ keeping the visual encoder active.
 """
 
 import copy
+import logging
 
 from PIL import Image as PILImage
+
+logger = logging.getLogger(__name__)
 
 # Constant 56x56 white PIL image used as the fake placeholder.
 _FAKE_IMAGE = PILImage.new("RGB", (56, 56), (255, 255, 255))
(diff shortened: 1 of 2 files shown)


nemo_automodel/components/datasets/vlm/fake_image.py
```diff
@@ -90,35 +93,140 @@ def inject_fake_image_into_conversation(conversation):
     return conversation
 
 
+# Attribute names that VLM processors / configs use to expose vision token IDs.
+# Different model families use different names; we scan all of them.
+_VISION_TOKEN_ID_ATTRS = (
```
HuiyingLi, line 98: Would this registry be useful and could be reused elsewhere? Instead of inside fake_image.py?

khazic: Currently `_VISION_TOKEN_ID_ATTRS` and `_VISION_TOKEN_STRINGS` are only used inside `fake_image.py` for the "scan all possible names" approach needed for robust fake-image token masking. Other components access model-specific attributes directly. Happy to extract them to a shared `vision_token_utils.py` if you see a concrete consumer that would benefit — otherwise I'd prefer to keep them co-located with their only user for now.

HuiyingLi (approved):
LGTM thank you!

PR #2059, opened 2026-04-27
fix(vlm): support multi-image token expansion in _expand_image_tokens

khazic:
## Problem

`_expand_image_tokens` in `collate_fns.py` assumed **exactly one image per sample** by hard-coding `grid_thws[0]` and `placeholder_positions[0]`. For multi-image conversations (e.g. Kimi K2.5 VL with several images per turn), only the first placeholder was expanded correctly; the remaining placeholders were left as single dummy tokens. This causes a sequence-length / shape mismatch at the PP (Pipeline Parallelism) boundary, leading to incorrect outputs or runtime errors.

## Fix

- Replace the single-image path with a loop that pairs each placeholder position with its corresponding `grid_thws` row, building the expanded sequence piece-by-piece.
- Add a `ValueError` guard when the placeholder count does not match `grid_thws.shape[0]`, making mismatches immediately visible instead of silently corrupting the batch.
- Update the docstring to reflect `(N, 3)` grid shape and the new `Raises` section.

## Tests

Added 4 new unit tests to `TestExpandImageTokens`:

| Test | What it covers |
|---|---|
| `test_expand_image_tokens_multi_image_different_sizes` | 2 images with different grid sizes |
| `test_expand_image_tokens_multi_image_same_size` | 2 images with identical grids |
| `test_expand_image_tokens_three_images` | 3 images |
| `test_expand_image_tokens_mismatch_raises` | `ValueError` on placeholder/grid count mismatch |

All 10 tests in `TestExpandImageTokens` pass (6 pre-existing + 4 new).

## Checklist

- [x] No GPU / real weights required — pure-torch mock tests
- [x] Existing single-image tests unchanged and still pass
- [x] DCO sign-off included

### DIFF
--- a/nemo_automodel/components/datasets/vlm/collate_fns.py
+++ b/nemo_automodel/components/datasets/vlm/collate_fns.py
@@ -776,54 +776,62 @@ def _expand_image_tokens(
     media_token_id: int,
     merge_kernel_size: Tuple[int, int] = _DEFAULT_MERGE_KERNEL,
 ) -> Tuple[torch.Tensor, torch.Tensor]:
-    """Expand single image placeholder tokens to the correct number based on grid_thws.
+    """Expand image placeholder tokens to the correct patch counts based on grid_thws.
 
     For PP, this ensures the sequence length is fixed BEFORE the model forward pass,
     eliminating dynamic sequence expansion inside the model.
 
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
LGTM thank you!

PR #2060, opened 2026-04-27
fix(vlm): use build_labels_from_template in kimi_vl_collate_fn

khazic:
## Problem

`kimi_vl_collate_fn` was calling `build_labels` to construct training labels. `build_labels` re-tokenizes the assistant text **standalone** and then searches for that token sequence inside the full `input_ids` via `_find_pattern_indices`.

Due to BPE context sensitivity, standalone tokenization can produce different token IDs than in-context tokenization:

```
# standalone
tokenizer("sure") → [1234]

# in-context (same word, different boundary due to surrounding tokens)
tokenizer("...Assistant: sure<|im_end|>") → [..., 5678]
```

When `_find_pattern_indices` fails to find a match, the sample's labels stay all `-100`. That sample contributes **zero loss silently** — the forward pass still runs (consuming GPU compute), but no gradient flows and the training data is effectively wasted.

## Fix

Replace `build_labels` with `build_labels_from_template` in `kimi_vl_collate_fn`. This function locates assistant turns by scanning `input_ids` for the structural markers the chat template inserts (special-token IDs that are fixed regardless of BPE context), so matching cannot fail. It falls back to `build_labels` only if no markers can be derived, making the change safe for all tokenizer types.

This brings `kimi_vl_collate_fn` in line with `kimi_k25_vl_collate_fn`, which already used `build_labels_from_template`.

## Tests

Updated the five `kimi_vl_collate_fn` unit tests to monkeypatch `build_labels_from_template` instead of `build_labels`. All 6 tests pass.

## Checklist

- [x] One-line production change, zero behaviour change for correct samples
- [x] DCO sign-off included

### DIFF
--- a/nemo_automodel/components/datasets/vlm/collate_fns.py
+++ b/nemo_automodel/components/datasets/vlm/collate_fns.py
@@ -742,7 +742,7 @@ def kimi_vl_collate_fn(
 
     batch = processor(**processor_kwargs)
 
-    labels = build_labels(
+    labels = build_labels_from_template(
         batch["input_ids"],
         conversations,
         processor,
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
LGTM, thank you so much!

PR #2061, opened 2026-04-27
fix(vlm): use build_labels_from_template in phi4_mm_collate_fn

khazic:
## Problem

`phi4_mm_collate_fn` was calling `build_labels` to construct training labels. `build_labels` re-tokenizes the assistant text **standalone** and searches for that token sequence inside the full `input_ids` via `_find_pattern_indices`.

Due to BPE context sensitivity, standalone tokenization can produce different token IDs than in-context tokenization:

```
# standalone
tokenizer("sure") → [1234]

# in-context (same word, different boundary due to surrounding tokens)
tokenizer("...Assistant: sure<|im_end|>") → [..., 5678]
```

When `_find_pattern_indices` fails to find a match, the sample's labels stay all `-100`. That sample contributes **zero loss silently** — the forward pass still runs (consuming GPU compute), but no gradient flows and the training data is effectively wasted.

## Fix

- Replace `build_labels` with `build_labels_from_template` in `phi4_mm_collate_fn`. This function locates assistant turns via structural chat-template markers and falls back to `build_labels` only if no markers can be derived, making the change safe for all tokenizer types.
- Fix a stale docstring in `gemma4_prefix_collate_fn` that incorrectly referenced `build_labels` when the function actually uses `build_labels_from_template` via `default_collate_fn`.

This is the same fix applied to `kimi_vl_collate_fn` in #2060.

## Tests

Updated the four affected `phi4_mm_collate_fn` unit tests to monkeypatch `build_labels_from_template` instead of `build_labels`. All 4 tests pass.

## Checklist

- [x] One-line production change, zero behaviour change for correct samples
- [x] DCO sign-off included

### DIFF
--- a/nemo_automodel/components/datasets/vlm/collate_fns.py
+++ b/nemo_automodel/components/datasets/vlm/collate_fns.py
@@ -496,7 +496,7 @@ def phi4_mm_collate_fn(examples, processor):
         else:
             batch["input_mode"] = 0
 
-    labels = build_labels(
+    labels = build_labels_from_template(
         batch["input_ids"],
         conversations,
         processor,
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
LGTM, thank you so much!

PR #2062, opened 2026-04-27
fix(vlm): mask fake vision tokens in kimi_k25_vl_collate_fn

khazic:
## Problem

Under FSDP / DeepSpeed Zero3, pure-text samples have a 56×56 white fake image injected so the visual encoder stays active during all-gather / reduce-scatter (every parameter must participate in the collective). The injected vision tokens should be invisible to attention — their `attention_mask` must be set to 0 — so they do not contaminate the model's representations.

All other VLM collate functions call `mask_fake_vision_tokens_batch` for this purpose:

| collate function | masks fake tokens |
|---|---|
| `default_collate_fn` | ✅ |
| `qwen2_5_collate_fn` | ✅ |
| `qwen3_omni_collate_fn` | ✅ |
| `kimi_vl_collate_fn` | ✅ |
| `llava_onevision_collate_fn` | ✅ |
| **`kimi_k25_vl_collate_fn`** | ❌ **missing** |

Without the masking, real text tokens attend to fake image tokens (attention_mask=1), slightly polluting the hidden states of pure-text samples during FSDP/Zero3 training. Labels are unaffected (fake tokens are in a user turn → automatically -100), so the loss is correct, but the attention computation is not clean.

## Fix

- Track `kept_examples` alongside `kept_conversations` through the per-sample processing loop (including the early overlong-drop path that already filtered `conversations` but not `examples`).
- After stacking the result tensors, call `mask_fake_vision_tokens_batch(result, processor, fake_indices)` — identical to every other collate function.

## Tests

Added two unit tests:

- `test_kimi_k25_vl_collate_fn_fake_image_mask` — verifies `mask_fake_vision_tokens_batch` is called with batch index 0 when `_injected_fake=True`.
- `test_kimi_k25_vl_collate_fn_non_fake_not_masked` — verifies `mask_fake_vision_tokens_batch` is **not** called when no sample is injected.

All 14 `kimi_k25` tests pass.

## Checklist

- [x] DCO sign-off included

### DIFF
--- a/nemo_automodel/components/datasets/vlm/collate_fns.py
+++ b/nemo_automodel/components/datasets/vlm/collate_fns.py
@@ -845,6 +845,7 @@ def kimi_k25_vl_collate_fn(
     # Optionally drop overlong samples before processing
     if max_length is not None and drop_overlong:
         conversations, _kept = _drop_overlong_samples(conversations, processor, max_length)
+        examples = [examples[i] for i in _kept]
 
     # Get media token ID
     media_token_id = getattr(processor, "media_placeholder_token_id", None)
@@ -858,6 +859,7 @@ def kimi_k25_vl_collate_fn(
     # Process each sample individually, dropping any that exceed max_length
     # after token expansion.
     kept_conversations = []
+    kept_examples = []
     all_expanded = []
     all_pixel_values = []
     all_grid_thws = []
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
LGTM, thank you so much!

PR #2071, opened 2026-04-28
fix(checkpoint): allow partial optimizer load for params with no gradient state

khazic:
## Problem

MoE models with non-differentiable routing (sigmoid, top-k hard selection)
have gate weights that sit in the optimizer `param_groups` but **never
receive gradients**.  AdamW lazily initialises per-parameter state only
on the first backward pass, so after training these weights have an empty
`{}` state entry.

On resume, `get_optimizer_state_dict(flatten_optimizer_state_dict=True)`
produces **placeholder tensors** for every `param_group` entry — including
the no-gradient gate weights.  DCP then raises:

```
RuntimeError: Missing key in checkpoint state_dict:
    optim.state.model.layers.1.mlp.gate.weight.step
```

This affects any MoE model whose routing gate is not differentiable
(e.g. HYV3, and potentially others using sigmoid or hard top-k routing).

## Fix

Pass `DefaultLoadPlanner(allow_partial_load=True)` when loading optimizer
state so that checkpoint keys absent from the state dict — and state dict
keys absent from the checkpoint — are silently skipped.

For params that never received a gradient this is the correct semantic:
they resume with fresh (zero) optimizer state, identical to a cold start.

## Test plan

- [x] Ran HYV3 4-layer P1 resume on 8×A100 — no longer crashes on `gate.weight`
- [x] Resume loss at step 50 matches original run within expected bf16 tolerance
- [ ] Existing checkpoint CI tests

### DIFF
--- a/nemo_automodel/components/checkpoint/checkpointing.py
+++ b/nemo_automodel/components/checkpoint/checkpointing.py
@@ -375,7 +375,7 @@ class Checkpointer:
         """
         optimizer_state = OptimizerState(model, optimizer, scheduler, is_peft=self.config.is_peft)
         state_dict = optimizer_state.state_dict()
-        self._do_load(state_dict, os.path.join(weights_path, "optim"))
+        self._do_load(state_dict, os.path.join(weights_path, "optim"), allow_partial_load=True)
         optimizer_state.load_state_dict(state_dict)
 
     @torch.no_grad()
@@ -790,6 +790,7 @@ class Checkpointer:
         path: str,
         storage_reader: Optional[_HuggingFaceStorageReader] = None,
         is_init_step: bool = False,
+        allow_partial_load: bool = False,
     ) -> dict[str, torch.Tensor]:
         """
         Load a state dictionary from `path` using DCP or PEFT special-case logic.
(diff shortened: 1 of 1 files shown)


khazic:
Closing this PR — the root cause has been identified and fixed elsewhere.

**Root cause**

The `Missing key` error during optimizer checkpoint loading was traced back to `gate_bias_update_factor=1e-3` being incorrectly set in `HYV3ForCausalLM`. With a non-zero factor, `update_moe_gate_bias()` called `Gate.update_bias()` on every step, which exercises optimizer-related state paths tied to `e_score_correction_bias`. This diverged from the official Tencent Hy3-preview SFT procedure, which treats `e_score_correction_bias` as a static pre-trained buffer and never updates it during fine-tuning.

**Fix**

`gate_bias_update_factor` is set to `0.0` for HYV3 in PR #2072, and `update_moe_gate_bias()` is guarded to skip the call when the factor is zero. After applying these changes, the full P1 checkpoint save/resume test (100 steps, FSDP2 + EP4 + PP2) passes cleanly:
- No `Missing key` errors on save or resume
- `val_loss` at step 99 identical across the original run and the resumed run (7.4278)

The partial-load planner change introduced here is therefore not needed for this use case. Closing in favour of #2072.

PR #2085, opened 2026-04-29
feat: Enable AC and dynamic-resolution vision path for Nemotron-Omni

yuekaizhang:
## Motivation

When using Nemotron-Omni / Nano v3 VL in RL-style training where the rollout is served by vLLM, two gaps prevented training from running correctly and efficiently:

1. Related to https://huggingface.co/nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-BF16/commit/8e00a7f6c072777ccb412f356024d9c92f602c80. **Train/rollout vision divergence.** vLLM uses a [`DynamicResolutionImageTiler`](https://github.com/vllm-project/vllm/blob/4b95e9cec4e9a1a90d3f4b2afa62e88459b2b90e/vllm/transformers_utils/processors/nano_nemotron_vl.py#L257) (one variable-resolution tensor per image, no tile splitting), but Automodel only had the static InternVL-style tile path. Same image → different ViT embeddings on the two sides → broken rollout/train logprob agreement.
2. **No activation checkpointing for the VLM.** `apply_ac` looked up `hidden_size` / `num_experts` only under `config.text_config` and wrapped layers on the outer module — both wrong for Nemotron-Omni, which nests the LM under `config.llm_config` and `model.language_model`. 

## What this PR adds

- **Dynamic-resolution vision path** (`nemo_automodel/components/models/nemotron_omni/model.py`)
  - New `extract_feature_dynamic` that crops each padded image back to its real `(h, w)` and runs RADIO per-image, matching vLLM's `NanoNemotronVLMultimodal` numerically (boundary-cast to ViT dtype, eval-mode spectral reparam).
  - New `_pixel_shuffle_dynamic_res` ported from vLLM — splits features by per-image patch counts, shuffles each, concatenates.
  - `forward()` now accepts `imgs_sizes` and dispatches to the dynamic-res branch when present; falls back to the existing tile-based `image_flags` path otherwise (fully backward compatible).

- **AC works for the VLM** (`nemo_automodel/components/moe/parallelizer.py`)
  - `apply_ac` config lookup now also consults `config.llm_config` (in addition to `text_config`).
  - Routes through `get_text_module(_model)` so blocks are wrapped under `language_model.layers`, not the outer wrapper or the vision tower.

### DIFF
--- a/nemo_automodel/components/models/nemotron_omni/model.py
+++ b/nemo_automodel/components/models/nemotron_omni/model.py
@@ -575,6 +575,112 @@ class NemotronOmniForConditionalGeneration(HFCheckpointingMixin, nn.Module, MoEF
 
         return vit_embeds
 
+    def extract_feature_dynamic(
+        self,
+        pixel_values: torch.Tensor,
+        imgs_sizes: "torch.Tensor | list[tuple[int, int]]",
+    ) -> torch.Tensor:
+        """Dynamic-resolution feature extraction (no tile splitting).
+
+        Matches vLLM's dynamic-resolution vision path for Nano v3 VL /
+        Nemotron-Omni (see 3rdparty/vllm/vllm/model_executor/models/
+        nano_nemotron_vl.py). Required when the rollout uses
+        DynamicResolutionImageTiler — tile-based `extract_feature` would
+        produce different embeddings and break rollout/train logprob
+        agreement.
+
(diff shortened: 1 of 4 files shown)


hijkzzz:
awesome MR

akoumpa (approved):
Thanks a lot @yuekaizhang

PR #1824, opened 2026-04-14
refactor: Remove separate moe_mesh references

edjson:
# What does this PR do ?

Removes separate moe_mesh parameter form NeMoAutoModel so users only need to pass in device_mesh to the NeMoAutoModel class.

# Changelog

- Add specific line by line info of high level changes in this PR.

# Before your PR is "Ready for review"

**Pre checks**:

- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?

If you haven't finished some of the above items you can still open "Draft" PR.

# Additional Information

- Follows up on #1684
- closes #1699

PR #2123, opened 2026-05-04
refactor: Remove separate moe_mesh references

edjson:
# What does this PR do ?

Removes `moe_mesh` from the user-facing API

# Changelog

- `auto_model.py`: removed `moe_mesh` from signature, docstrings, `_retry` forwarding, and `MeshContext.from_meshes` calls
- `extract_nemo_activations.py`: removed stale `moe_mesh=` 
- `test_auto_model.py`: added `TestMoeMeshRemovedFromAPI` to guard against regressions

**Pre checks**:

# Additional Information

- Related to #1824


nemo_automodel/_transformers/auto_model.py
```diff
@@ -578,7 +576,7 @@ def from_pretrained(
         if tp_plan is not None and distributed_config is not None:
             distributed_config.tp_plan = tp_plan
 
-        mesh = MeshContext.from_meshes(device_mesh, moe_mesh)
+        mesh = MeshContext.from_meshes(device_mesh)
```
akoumpa, line 656: 🤔 do we need MeshContext then?

edjson: Hello, 
Yes I believe `instantiate_infrastructure` still expects `MeshContext` so we would need to keep it to satisfy `mesh: Optional[MeshContext] = None`.

adil-a:
I think we still have `moe_mesh` references inside the recipes folder here `https://github.com/NVIDIA-NeMo/Automodel/tree/main/nemo_automodel/recipes`. Could you please go over them and double check where moe_mesh should be removed? Since the NeMoAutoModelFor... classes are being changed, we will need to change call sites as well

nemo_automodel/_transformers/auto_model.py
adil-a, line 656: since we are removing moe_mesh from .from_meshes call, we need to somehow get the moe_mesh inside the MeshContext. Likely you can do this extraction in the post init method https://github.com/NVIDIA-NeMo/Automodel/blob/8cb1f5ce7701a5d92f7478d9d4ea100263478ed1/nemo_automodel/components/distributed/mesh.py#L121

adil-a: I think we should be able to get the moe_mesh from the device mesh directly

adil-a: @hemildesai is the change in already where we flatten the ep mesh into the device mesh or not yet? @edjson is helping us removing the moe mesh calls from the user-facing API but we can't do this until the moe mesh is extractable from the device mesh

edjson:
@akoumpa could you test this when you get the chance please?

edjson:
Closing in favor of #2266

PR #2152, opened 2026-05-06
fix(deepseek_v4): use scatter_add_ for indexer_topk mask

### LINKED ISSUE #2144: DeepSeek V4 sparse-attention mask drops pool position 0 due to `scatter_` overwrite on clamped `-1` indices
**Describe the bug**

In `DeepseekV4Attention.forward` (`nemo_automodel/components/models/deepseek_v4/layers.py`, around L905–L913), `indexer_topk` (shape `[B, S, K]`, with `-1` marking causally-masked / invalid slots) is converted into a boolean `indicator` mask of shape `[B, S, P]` via:

```python
valid = indexer_topk != -1                              # [B, S, K]
safe_idx = indexer_topk.clamp(min=0)                    # -1 -> 0 to avoid OOB
indicator = torch.zeros((B, S, P), dtype=torch.bool, device=...)
indicator.scatter_(-1, safe_idx, valid)
```

This is incorrect. `clamp(min=0)` collapses **all `-1` entries onto the same destination index `0`**, and `scatter_` uses overwrite semantics when there are duplicate indices along the scattered dim. As a result, whenever a query's `indexer_topk[b, s, :]` contains both:

1. a valid entry pointing to pool position `0` (`valid=True`), **and**
2. one or more `-1` entries (`valid=False`, but `safe_idx=0` after clamp),

the `False` writes can overwrite the `True` write at `indicator[b, s, 0]`, depending on the order along K. The mask then incorrectly forces `-inf` at pool position `0` for that query, so the query silently fails to attend to a pool slot it was supposed to attend to.

Standalone reproducer of just the mask-construction snippet:

```python
import torch

B, S, K, P = 1, 1, 4, 5

# Query wants to attend to pool positions 2 and 0;
# the trailing two slots are causally masked (-1).
indexer_topk = torch.tensor([[[2, 0, -1, -1]]])   # [B, S, K]

valid = indexer_topk != -1                        # [[[T, T, F, F]]]
safe_idx = indexer_topk.clamp(min=0)              # [[[2, 0, 0, 0]]]

# --- current (buggy) implementation ---
indicator = torch.zeros((B, S, P), dtype=torch.bool)
indicator.scatter_(-1, safe_idx, valid)
print("buggy   :", indicator)

# --- expected ---
print("expected:", torch.tensor([[[True, False, True, False, False]]]))

# --- proposed fix ---
counts = torch.zeros((B, S, P), dtype=torch.int32)
counts.scatter_add_(-1, safe_idx, valid.to(torch.int32))
fixed = counts > 0
print("fixed   :", fixed)
```

Output:

```
buggy   : tensor([[[False, False,  True, False, False]]])
expected: tensor([[[ True, False,  True, False, False]]])
fixed   : tensor([[[ True, False,  True, False, False]]])
```

**Additional context**

Suggested fix — replace overwrite with count-then-threshold, so invalid slots contribute `0` and cannot stomp on valid writes:

```python
valid = indexer_topk != -1
safe_idx = indexer_topk.clamp(min=0)
indicator_counts = torch.zeros(
    (B, S, P), dtype=torch.int32, device=full_kv.device,
)
indicator_counts.scatter_add_(-1, safe_idx, valid.to(torch.int32))
indicator = indicator_counts > 0
```

Downstream impact: in NeMo-RL GRPO training on DSV4, after applying this fix the observed `gen_kl_error` drops from **0.02591** to **0.02494**.

jQizhang:
# What does this PR do ?

fixed #2144 

Fix the DSV4 sparse-attention compressed mask so a query can still attend to pool position 0 when its indexer_topk row contains both a valid 0 and at least one -1 (clamped-to-0) slot.

# Changelog

- nemo_automodel/components/models/deepseek_v4/layers.py
    - Add _build_indexer_topk_compressed_mask(attention_mask, indexer_topk, n_pooled) helper. Builds the [B, S, P] additive compressed-position mask from indexer_topk using a count-then-threshold pattern.
  - tests/unit_tests/models/deepseek_v4/test_dsv4_layers.py
    - Add TestDeepseekV4AttentionMask::test_indexer_topk_mask_preserves_pool_zero_with_clamped_invalid_entries.

**Pre checks**:

# Additional Information

- Related to # (issue)
#2144

### DIFF
--- a/nemo_automodel/components/models/deepseek_v4/layers.py
+++ b/nemo_automodel/components/models/deepseek_v4/layers.py
@@ -478,6 +478,31 @@ def eager_attention_with_sink(
     return torch.matmul(probs, value_states).transpose(1, 2).contiguous(), probs
 
 
+def _build_indexer_topk_compressed_mask(
+    attention_mask: torch.Tensor,
+    indexer_topk: torch.Tensor,
+    n_pooled: int,
+) -> torch.Tensor:
+    """Build the additive compressed-position mask for Indexer-selected pool IDs."""
+    min_val = torch.finfo(attention_mask.dtype).min
+    indexer_topk = indexer_topk.to(device=attention_mask.device)
+    batch, seq_len = indexer_topk.shape[:2]
+    valid = indexer_topk != -1  # [B, S, K]
+    safe_idx = indexer_topk.clamp(min=0)
+    indicator_counts = torch.zeros(
+        (batch, seq_len, n_pooled),
+        dtype=torch.int32,
+        device=attention_mask.device,
+    )
+    indicator_counts.scatter_add_(-1, safe_idx, valid.to(torch.int32))
+    indicator = indicator_counts > 0
+    return torch.where(
+        indicator,
+        torch.zeros((), dtype=attention_mask.dtype, device=attention_mask.device),
+        torch.full((), min_val, dtype=attention_mask.dtype, device=attention_mask.device),
+    )  # [B, S, P]
+
+
 class DeepseekV4Indexer(nn.Module):
     """HF PR 45616 port.  Picks the top-k compressed positions per query when
     ``compress_ratio == 4``.  Owns its own pool at ``index_head_dim`` plus a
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
LGTM thank you so much!

PR #2161, opened 2026-05-06
feat(nemotron-v3): add Multi-Token Prediction (MTP) support

adil-a:
## Summary

Adds Multi-Token Prediction (MTP) support for Nemotron Super V3, including the model-side scaffolding, state-dict adapter, and recipe integration. The MTP auxiliary head produces per-depth hidden states during training that are converted to per-depth losses by the recipe-side `calculate_mtp_loss`, which inherits whatever efficient loss class the recipe configures (FusedLinearCrossEntropy, MaskedCrossEntropy, etc.). Auto-detected at runtime from the HF config (`num_nextn_predict_layers`, `mtp_hybrid_override_pattern`); the only training-time knob is `model.mtp_loss_scaling_factor` (default `0.1`).

### MTP head produces meaningful predictions on real Super V3 weights

PEFT smoke test (1×8 H100) on the published Super V3 checkpoint, scoring three numbers on the model's own greedy chat-template generation:

| Metric | Value | Chance baseline |
|---|---|---|
| Main-head top-1 (`argmax(logits[t]) == input_ids[t+1]`) | **85.15%** | 0.000763% |
| MTP-depth-1 top-1 (`argmax(mtp_logits[t]) == input_ids[t+2]`) | **72.28%** | 0.000763% |
| Speculative accept (`argmax(mtp_logits[t]) == argmax(logits[t+1])`) | **74.26%** | 0.000763% |

MTP-depth-1 predicts the **+2** token (one past what the main head predicts), so it's a strictly harder task. 72% accuracy at this is ~95,000× chance and is in the published DeepSeek-V3 range for the analogous setup.

### Saved checkpoint state-dict parity (multi-node full SFT)

99-step Hellaswag SFT on Super V3 across 4 nodes. Comparing the saved consolidated checkpoint at `epoch_0_step_99` vs the base HF checkpoint: **42,683 keys total / 1,040 `mtp.*` keys on both sides — 0 missing, 0 extra**. The state-dict adapter round-trips the MTP namespace cleanly to the HF flat layout.

### Base vs finetuned on the same Hellaswag input

Same 64-token Hellaswag sample evaluated on both base and SFT'd model:

| Metric | Base | FT (step 99) | Δ |
|---|---|---|---|
| Main top-1 | 48.39% | 51.61% | +3.22pp |
| MTP-depth-1 top-1 | 35.48% | 38.71% | +3.22pp |
| Speculative accept | 69.35% | 69.35% | 0 |

MTP head improved alongside the main head (same magnitude), and the speculative accept rate is preserved — both heads moved in lockstep, consistent with end-to-end joint training.

### Resume-from-checkpoint parity

10-step PEFT run, save at step 4, resume in a second run, compare losses with MTP loss actively contributing:

| Step | Run A (clean) | Run B (resume) | abs Δ | rel % |
|---|---|---|---|---|
| 5 (immediate post-resume) | 3.302634 | 3.3026 | 3.4e-05 | 0.001% |
| 6 | 2.488104 | 2.4820 | 6.1e-03 | 0.245% |
| 7 | 3.121054 | 3.1079 | 1.3e-02 | 0.422% |

Step-5 (the critical first-post-resume step) is bit-identical at logged precision; subsequent steps within bf16 noise.

### Unit tests

20 unit tests covering: pattern parsing, cumulative `roll_tensor` semantics, `MTPConfig` enable/disable logic, model construction with/without MTP, MTP forward emission of `mtp_per_depth_h`, eval-mode gating, gradient backprop into fusion modules, state-dict round-trip exactness for D=1 and D=2, and forward determinism under the new `NemotronHCausalLMOutputWithPast` dataclass output.

## Test plan

- [x] Unit tests pass (`tests/unit_tests/models/nemotron_v3/test_nemotron_v3_mtp{,_parity}.py`)
- [x] PEFT smoke test on Super V3 (1×8 H100) — MTP loss non-zero, training converges
- [x] Multi-node full SFT (4×8 H100, 99 steps) — saved consolidated HF checkpoint, all 1,040 `mtp.*` keys present
- [x] Resume-from-checkpoint parity — step-5 loss bit-identical
- [x] MTP head correctness vs chance — 72%/74% top-1 / accept rate vs 8e-6 chance
- [x] Base-vs-FT comparison on same input — MTP and main move in lockstep


adil-a:
Addressed both review comments in `aa98993e`:

1. **E402 — dataclass between import blocks**: moved `NemotronHCausalLMOutputWithPast` below all top-level imports.
2. **Dead `position_ids` in `MTPModule.forward`**: removed it from the signature and the caller. You're right that it was rolled but never consumed; rolling positions per depth would actually be wrong for MTP attention RoPE (we want main-position RoPE preserved), and `NemotronV3Block` doesn't accept `position_ids` directly anyway.

Also ran `ruff format` + `ruff check --fix` over `nemo_automodel/`.

adil-a:
Correction on the `position_ids` removal — I was wrong, restored in `ef55a638`.

Cross-checked against Megatron-LM's reference (`megatron/core/transformer/multi_token_prediction.py:890-905`): Megatron rolls **both** `input_ids` and `position_ids` with `shifts=-1` and forwards rolled `position_ids` through the per-depth transformer layers, so RoPE-using attention sees the original token position even though slot indices are shifted.

NemotronV3's attention happens to be positionless (`layers.py:35` — "GQA attention for NemotronV3 (no RoPE)"), so for *this* model `position_ids` is a no-op. But `MTPModule` lives in `components/models/common/mtp/` and is intended as model-agnostic scaffolding; stripping the rolling would silently break any future RoPE-using sublayer that reuses it. Now the rolled `position_ids` is forwarded through `block_kwargs` so RoPE-aware sublayers can consume it; positionless sublayers swallow it via `**attn_kwargs`.

nemo_automodel/components/moe/parallelizer.py
```diff
@@ -73,6 +73,21 @@ def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
...
+def _iter_moe_blocks(model: nn.Module, _model: nn.Module):
+    """Yield decoder blocks that may contain MoE sublayers.
+
+    Covers the main backbone (``_model.layers``) plus an optional MTP
+    auxiliary head (``model.mtp.layers``) when present. MTP sublayers are
+    not registered under ``_model.layers`` but carry the same MoE structure
+    and must receive the same EP / FSDP treatment so their state-dict
+    round-trips cleanly.
+    """
+    yield from (block for _, block in _model.layers.named_children())
```
akoumpa:
```suggestion
    yield from _model.layers.children()
```

nemo_automodel/components/moe/parallelizer.py
```diff
@@ -73,6 +73,21 @@ def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
...
+    mtp_module = getattr(model, "mtp", None)
+    if mtp_module is not None and hasattr(mtp_module, "layers"):
+        yield from (block for _, block in mtp_module.layers.named_children())
```
akoumpa:
```suggestion
        yield from mtp_module.layers.children()
```

nemo_automodel/components/moe/parallelizer.py
```diff
@@ -73,6 +73,21 @@ def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
         )
 
 
+def _iter_moe_blocks(model: nn.Module, _model: nn.Module):
+    """Yield decoder blocks that may contain MoE sublayers.
+
+    Covers the main backbone (``_model.layers``) plus an optional MTP
```
akoumpa: I feel the names are not very infromative,

nemo_automodel/components/models/common/mtp/mtp.py
```diff
@@ -0,0 +1,231 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Callable
+
+import torch
+import torch.nn as nn
+
+# Hybrid layer symbols matching HF Nemotron-H's ``hybrid_override_pattern``
```
akoumpa: perhaps this would fit better in the nemotron_v3 directory? I'm not sure if the symbol to block naming convention carries over to other models.

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -851,6 +851,71 @@ def calculate_loss(loss_fn, **kwargs) -> torch.Tensor:
...
+    D = len(mtp_per_depth_h)
+    cur_labels = labels
+    total = mtp_per_depth_h[0].new_zeros(())
+    for k, h_k in enumerate(mtp_per_depth_h):
+        cur_labels = roll_tensor(cur_labels, shifts=-1, dim=-1)
+        masked = cur_labels.clone()
+        n_invalid = min(k + 1, masked.shape[-1])
+        masked[..., -n_invalid:] = ignore_index
+
+        if isinstance(loss_fn, FusedLinearCrossEntropy):
```
akoumpa, line 898: I wonder if both paths are exercised?

nemo_automodel/components/models/common/mtp/mtp.py
adil-a: moved

nemo_automodel/recipes/llm/train_ft.py
adil-a: I just added unit tests for both paths, should be good I believe

nemo_automodel/components/moe/parallelizer.py
```diff
@@ -84,20 +99,21 @@ def apply_ep(model: nn.Module, ep_mesh: DeviceMesh, moe_mesh: DeviceMesh | None
...
-                    module=moe_module.experts,
-                    device_mesh=ep_mesh,
-                    parallelize_plan=ExpertParallel(),
-                )
+        if not isinstance(moe_module, MoE):
+            continue
+        # GroupedExpertsTEGroupedLinear uses TE's GroupedLinear which creates
+        # local experts directly. It doesn't support DTensor wrapping, so we
+        # skip distribute_module entirely and just initialize token dispatcher.
+        if isinstance(moe_module.experts, GroupedExpertsTE):
```
akoumpa, line 116: Hi @adil-a,  I see that it gates the for loop on being an `MoE` instance, but [`GroupedExpertsTE`](https://github.com/NVIDIA-NeMo/Automodel/blob/3c9bc91de93503794cfe2041353e893395edf2ed/nemo_automodel/components/moe/experts.py#L796) does not subclass `MoE` , so I'm wondering how this will work.

adil-a: The outer gate is on the **layer class** `MoE` (the module that owns the router + experts + shared expert), not on the experts attribute. `GroupedExpertsTE` is what `MoE.experts` gets set to when the backend is configured for TE + DeepEP (`layers.py:622`), but the `MoE` wrapper itself is unchanged — so `isinstance(moe_module, MoE)` is True regardless of the experts implementation in use.

The TE path is then picked up by the inner `isinstance(moe_module.experts, GroupedExpertsTE)` check at line 109, which routes to `init_token_dispatcher` instead of `parallelize_module`. This is the same flow the existing `examples/llm_benchmark/nemotron/nemotron_nano_v3_te_deepep.yaml` config exercises.

nemo_automodel/components/moe/state_dict_mixin.py
```diff
@@ -525,13 +525,16 @@ def _convert_single_merged_expert_to_hf_split_experts(
...
+                in emitted HF keys (used for namespaces outside the main backbone,
+                e.g. ``"mtp."``).
 
         Returns:
             List of (fqn, tensor) tuples in HuggingFace format, or None if not an expert tensor
         """
         n_experts = self.moe_config.n_routed_experts
         inter_dim = self.moe_config.moe_inter_dim
-        prefix = self._hf_prefix
+        prefix = kwargs.get("prefix_override") or self._hf_prefix
```
akoumpa: @adil-a I feel it will improve code clarity if `prefix_override` was moved to a keyword argument (with a default value) instead of going via kwargs. Would that be a big cut to implement?

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -1426,6 +1491,16 @@ def _forward_backward_step(
...
                     num_label_tokens=num_label_tokens,
                 )
+                mtp_per_depth_h = getattr(out, "mtp_per_depth_h", None)
+                if mtp_per_depth_h is not None:
+                    local_loss = local_loss + calculate_mtp_loss(
+                        self.loss_fn,
+                        mtp_per_depth_h=mtp_per_depth_h,
+                        labels=labels,
+                        model=model,
+                        scaling_factor=getattr(out, "mtp_loss_scaling_factor", 0.1),
```
akoumpa: @adil-a what's the case for needing getattr here? I'm a bit hesitant to inject a default value here, so I'm wondering if there's a better way to handle this.

nemo_automodel/components/moe/parallelizer.py
adil-a: Done in 97ff4135 (now part of 44e3ef21 after rebase). Same simplification at line 88 below.

nemo_automodel/components/moe/parallelizer.py
adil-a: Done in 97ff4135 (now in 44e3ef21).

nemo_automodel/components/moe/parallelizer.py
adil-a: Renamed in 44e3ef21: `(model: nn.Module, _model: nn.Module)` → `(model_wrapper: nn.Module, backbone: nn.Module)`, with docstring spelling out which is the outer wrapper vs. the inner decoder stack. Callers unchanged.

nemo_automodel/components/moe/state_dict_mixin.py
adil-a: Done in 44e3ef21. `_convert_single_merged_expert_to_hf_split_experts` now exposes `prefix_override: str | None = None` as a keyword-only argument with a default. `**kwargs` is kept on the signature so the base callers that forward arbitrary state-dict kwargs (e.g. `exclude_key_regex`) continue to compile without churning every adapter; only the `prefix_override` access changes (`kwargs.get(...)` → named arg). The lone caller that explicitly sets it (`nemotron_v3/state_dict_adapter.py:198`, the `mtp.` path) was already passing it as a named kwarg.

nemo_automodel/recipes/llm/train_ft.py
adil-a: Good catch — the defensive default was unnecessary. Dropped in 44e3ef21: it now reads `out.mtp_loss_scaling_factor` directly with a short comment explaining that `mtp_per_depth_h is not None` already guarantees the source dataclass declares the field (`NemotronHCausalLMOutputWithPast` declares both as paired fields). A future model that emits `mtp_per_depth_h` without also setting `mtp_loss_scaling_factor` should now surface as a loud `AttributeError` rather than silently using `0.1`.

akoumpa (approved):
Thanks a ton @adil-a this looks great

PR #2172, opened 2026-05-07
fix(vlm): forward get_rope_index to neat packing for mRoPE models

khazic:
## What

When VLM neat-packing is enabled for an mRoPE-aware model (Qwen2.5-VL, Qwen3-VL, Qwen3-VL-MoE, Qwen3-Omni), the recipe never passes `model.get_rope_index` to `neat_pack_dataset_vlm`. With it absent, packed `position_ids` are emitted as 1D `range(seq_len)` per sample and the model's internal mRoPE auto-fill is skipped.

This PR plumbs `get_rope_index` from the model through `build_dataloader` to `neat_pack_dataset_vlm`. Models without the method (Mistral3, LLaVA-OV, KimiVL, Gemma4-VLM) keep prior behavior since `getattr(model, "get_rope_index", None)` returns `None`.

## Exact trigger conditions

All of the following must hold simultaneously:

1. **VLM recipe** (`recipes/vlm/finetune.py`) — not LLM/diffusion/retrieval.
2. **Packing enabled** — `packed_sequence.pack_size > 0` (or legacy `dataset.packing.enabled`).
3. **Pretokenize on** — `packed_sequence.pretokenize: true` (so the path goes through `PreTokenizedDatasetWrapper` + `neat_pack_dataset_vlm`).
4. **Model implements `get_rope_index`** — i.e. mRoPE-aware: `Qwen2_5_VLForConditionalGeneration`, `Qwen3VLForConditionalGeneration`, `Qwen3VLMoeForConditionalGeneration`, `Qwen3OmniMoeForConditionalGeneration`.
5. **No user-supplied `post_tokenize_hook_fn`** that pre-computes 3D `position_ids`.

The path is **independent of CP**. CP-specific mRoPE handling (PR #1482, `cp_utils.py:294`) is correct *given* 3D `position_ids` enter CP — that fix shards correctly on `dim=2` for `ndim==3`. This PR is about how 3D `position_ids` originate in the first place under packing.

Unaffected combinations (no behavior change):
- Non-packing VLM training — `default_collate_fn` deliberately omits `position_ids` so the model auto-calls `get_rope_index` (see comment at `collate_fns.py:1277-1281`).
- Non-mRoPE VLMs — `getattr(model, "get_rope_index", None)` returns `None`, packing falls through the existing 1D path, identical to current behavior.
- LLM neat packing — uses `neat_pack_dataset` (not `_vlm`), unrelated.

## Code path (failure case before this PR)

```
[...]

PR #2180, opened 2026-05-07
fix(vlm): ceil-divide PP chunker so trailing samples are not dropped

khazic:
## Summary

\`_chunk_vlm_media\` partitioned samples per microbatch with \`batch_size // n_microbatches\`. When that division is uneven the trailing \`batch_size % n_microbatches\` samples never get assigned to a chunk — their image tensors are silently dropped. Meanwhile \`schedule.step\` slices \`input_ids\`/\`labels\` via \`torch.tensor.chunk(n_microbatches)\` which uses ceil-sized chunks and covers every sample, so the affected microbatches end up with media tokens in text but no pixel data, breaking vision scatter or producing wrong outputs.

This patch switches all three internal branches to ceil division so the chunker mirrors \`tensor.chunk\` semantics.

## Changelog

- **fix(vlm)**: replace \`batch_size // n_microbatches\` with \`-(-batch_size // n_microbatches)\` (ceil) in \`_chunk_vlm_media\`'s Gemma4 multi-image, general flat-patches, and legacy 1-image-per-sample branches.
- **test(vlm)**: add three regression tests covering uneven \`batch_size\` across each branch (\`batch_size=7, n_microbatches=3\` for general/Gemma4; \`batch_size=5, n_microbatches=3\` for legacy). Each asserts no sample is dropped and chunk sizes match \`tensor.chunk\`.

## Test plan

- [x] \`uv run pytest tests/unit_tests/recipes/test_finetune_vlm_helpers.py\` — 68 passed, 3 skipped (skips are pre-existing \`fused_linear_ce\` GPU cases)
- [x] \`uv run ruff format --check\` and \`ruff check\` on \`recipes/vlm/finetune.py\` — clean
- [ ] PP=3 functional verification on a VLM recipe with \`local_batch_size: 7\` (requires multi-GPU host)

### DIFF
--- a/nemo_automodel/recipes/vlm/finetune.py
+++ b/nemo_automodel/recipes/vlm/finetune.py
@@ -282,7 +282,11 @@ def _chunk_vlm_media(
         # Gemma4 multi-image: pixel_values is [N_total_images, max_patches, patch_dim].
         # All images are padded to the same max_patches, so split by image count directly.
         cumsum_images = torch.cumsum(n_images_per_sample, dim=0)
-        samples_per_mb = batch_size // n_microbatches
+        # Ceil division so trailing samples are not dropped when batch_size is not
+        # divisible by n_microbatches.  Mirrors torch.tensor.chunk(n) semantics used
+        # by the PP schedule on input_ids/labels — keeps image chunks aligned with
+        # text chunks even on uneven splits.
+        samples_per_mb = -(-batch_size // n_microbatches)
         for mb_idx in range(n_microbatches):
             s_start = mb_idx * samples_per_mb
             s_end = min(s_start + samples_per_mb, batch_size)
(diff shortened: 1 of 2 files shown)


khazic:
Rebased onto latest `main` (`174ba8d2`) to resolve the conflict with #2172. New head: `31024b63`.

- `nemo_automodel/recipes/vlm/finetune.py`: auto-merged (this PR touches the `samples_per_mb` lines, #2172 added `get_rope_index` plumbing in different functions of the same file).
- `tests/unit_tests/recipes/test_finetune_vlm_helpers.py`: manually resolved — the three new `test_uneven_batch_size_*` methods are kept inside `TestChunkVlmMedia` (right after `test_fallback_mismatched_images`), and #2172's module-level `get_rope_index forwarding tests for build_dataloader` section is preserved verbatim after the class.

Verified locally: `uv run pytest tests/unit_tests/recipes/test_finetune_vlm_helpers.py` → 70 passed, 3 skipped (67 from `main` + this PR's 3 new tests; nothing from #2172 was lost).

HuiyingLi (dismissed):

HuiyingLi (approved):

PR #2181, opened 2026-05-07
fix(vlm): fail loudly in PP chunker when pixel_values cannot be aligned

khazic:
## Summary

\`_chunk_vlm_media\`'s legacy fallback (\`n_images != batch_size\` AND \`n_images_per_sample is None\` AND pixel_values is not shaped \`[batch_size, ...]\`) used to dump every image into mb0, emit empty pixel_values for mb1..N, and log a warning. Subsequent microbatches whose text still carried media tokens would then scatter into empty tensors — silent corruption rather than a crash.

This patch replaces the warn-and-continue with \`raise ValueError\`, forcing the calling collate to align with the chunker's contract instead of training on broken data.

## Changelog

- **fix(vlm)**: \`_chunk_vlm_media\` now raises \`ValueError\` when no layout matches, with a message listing the actual shapes / counts and pointing at the two valid options (\`pixel_values [batch_size, ...]\` or \`n_images_per_sample\` set).
- **test(vlm)**: rewrite \`test_fallback_mismatched_images\` and \`test_pp_vlm_chunking_mismatched_images_and_batch\` to assert the new \`pytest.raises\` contract instead of the old "warn and continue" behavior.

## Behavioral note

This is a breaking change for any caller relying on the previous warn-and-continue behavior. Such callers were already producing incorrect training data — empty pixel_values scattered into media tokens cannot yield meaningful gradients — so the breakage is intentional. All in-tree collates already set \`n_images_per_sample\`, so no in-tree recipe is affected.

## Test plan

- [x] \`uv run pytest tests/unit_tests/recipes/test_finetune_vlm_helpers.py\` — 65 passed, 3 skipped (skips are pre-existing \`fused_linear_ce\` GPU cases)
- [x] \`uv run ruff check\` on \`recipes/vlm/finetune.py\` — clean

### DIFF
--- a/nemo_automodel/recipes/vlm/finetune.py
+++ b/nemo_automodel/recipes/vlm/finetune.py
@@ -326,13 +326,20 @@ def _chunk_vlm_media(
             patch_end = cumsum[img_end - 1].item() if img_end > 0 else 0
             pixel_values_chunks.append(pixel_values[int(patch_start) : int(patch_end)])
     else:
-        pixel_values_chunks.append(pixel_values)
-        image_grid_chunks.append(image_grid)
-        for _ in range(n_microbatches - 1):
-            pixel_values_chunks.append(pixel_values[:0])
-            image_grid_chunks.append(image_grid[:0])
-        logging.warning(
-            f"VLM chunking: n_images={n_images} != batch_size={batch_size}, giving all images to first microbatch"
+        # No layout matched. Previously this branch logged a warning and gave all
+        # images to mb0 with empty tensors for mb1..N, but trailing microbatches
+        # whose text still contains media tokens would then scatter into empty
+        # pixel_values — silent corruption. Fail loudly so the calling collate is
+        # forced to provide n_images_per_sample (or align n_images with batch_size)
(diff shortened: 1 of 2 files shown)


HuiyingLi (dismissed):

khazic:
Rebased onto latest `main` (`174ba8d2`) to resolve the conflict with #2172. New head: `4af1b609`.

- `nemo_automodel/recipes/vlm/finetune.py`: auto-merged (the two PRs touch different functions in the same file).
- `tests/unit_tests/recipes/test_finetune_vlm_helpers.py`: manually resolved — kept this PR's `pytest.raises(ValueError, ...)` body for `test_fallback_mismatched_images_raises`, and preserved the `get_rope_index forwarding tests for build_dataloader` section added by #2172 verbatim.

Verified locally: `uv run pytest tests/unit_tests/recipes/test_finetune_vlm_helpers.py` → 67 passed, 3 skipped (the +2 vs. the original PR description are #2172's two new `test_build_dataloader_*` tests, confirming nothing was lost).

HuiyingLi (approved):

PR #2191, opened 2026-05-08
feat(deepseek-v4): add Multi-Token Prediction (MTP) training support

khazic:
## Summary

Adds Multi-Token Prediction (MTP) training support for DeepSeek V4 (Flash). MTP layers run as standard pre-norm attention + MoE blocks (no HC machinery), with rotary embeddings shared from the main backbone. The auxiliary loss is computed via the recipe-side `calculate_mtp_loss` and added to the main CE loss.

### Model side
- `components/models/common/mtp/`: model-agnostic scaffold (`MTPConfig`, `MTPModule`, `roll_tensor`).
- `components/models/deepseek_v4/mtp.py`: V4-specific `DeepseekV4MTPSublayer` and `build_deepseek_v4_mtp` factory. `compress_ratios` is forced to `None` for MTP attention to avoid `IndexError` past the backbone layer count; rotary refs are stored via `object.__setattr__` so they don't pollute `state_dict`.
- `components/models/deepseek_v4/model.py`: `DeepseekV4ForCausalLM` now constructs `self.mtp` when `num_nextn_predict_layers > 0` and returns a `DeepseekV4CausalLMOutput` dataclass (`logits` + optional `mtp_per_depth_h`).

### State-dict adapter
- `from_hf` runs MTP layers (`layers.{N+k}.*`) through the same dequantize / aggregate-experts / rename pipeline as the backbone (renumber to `layers.{k}.*`, run pipeline, re-prefix to `mtp.layers.{k}.*`). Previously MTP keys bypassed dequantization and FP8/FP4 buffers were left raw.
- `to_hf` rewrites `mtp.layers.{k}.*` into `model.layers.{N+k}.*` and runs the unified split / rename / quantize path; an explicit fallback strips the leftover `model.` prefix for fusion-only modules (`eh_proj` / `enorm` / `hnorm` / `final_layernorm`) that have no entry in the rename table.

### Recipe (`recipes/llm/train_ft.py`)
- `calculate_mtp_loss`: per-depth CE through the configured loss class (FusedLinearCE / MaskedCE), summed with `loss_scaling_factor / D` weighting.
- `_forward_backward_step` (non-PP branch) reads `out.mtp_per_depth_h` and adds the MTP loss to the main loss.
- `_mtp_is_enabled(cfg, model_parts)` + setup-time guard: raises `NotImplementedError` if pipeline parallelism is enabled together with MTP, since the PP schedule does not currently aggregate the MTP auxiliary loss. **PP + MTP is intentionally deferred to a follow-up PR.**

### Tests
- `test_deepseek_v4_mtp.py`: config / construction / forward / backward / state-dict coverage.
- `test_dsv4_state_dict_adapter.py`: MTP round-trip for layer rename, FP8 dequantize, expert aggregation, and the fusion-only fallback in both directions.
- `test_dsv4_model_smoke.py`: updated to read `.logits` from the new dataclass output.

## Overlap with #2161

PR #2161 (Nemotron V3 MTP) introduces the same `calculate_mtp_loss` helper and the same non-PP integration in `_forward_backward_step`. Those two regions are byte-identical between the branches.

This is intentional — both PRs need the same recipe-side scaffolding, and the model-agnostic MTP base (`components/models/common/mtp/`) is shared. **When #2161 lands first, those duplicated lines will be auto-resolved on rebase**, and this PR will reduce to the V4-specific changes (model, MTP sublayer, adapter, PP guard, V4 tests).

## Test plan

wandb: https://wandb.ai/Nemo-automodel/huiyingl_workspace?nw=nwuserhuiyingl
image


nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -916,6 +908,44 @@ def calculate_mtp_loss(
...
+            loss = loss + calculate_mtp_loss(
+                self.loss_fn,
+                mtp_per_depth_h=mtp_per_depth_h,
+                labels=labels,
+                model=self.model,
+                scaling_factor=scaling_factor,
+            )
+        return loss
+
+
```
adil-a: this shouldn't be inside train_ft imo

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -1290,6 +1323,36 @@ def _log_moe_metrics(self, step: int, wandb_log_fn) -> None:
...
+            f"rank={self.dist_env.rank}, "
+            f"pp_has_last_stage={self.pp.info.has_last_stage}, "
+            f"owns_mtp_module={getattr(last_stage_model, 'mtp', None) is not None}, "
+            f"mtp_enabled={mtp_enabled}, "
+            f"mtp_depth={mtp_depth}, "
+            f"mtp_loss_scaling_factor={get_mtp_loss_scaling_factor(last_stage_model)}, "
+            f"schedule_loss={type(self.pp.info.schedule._loss_fn).__name__}"
+        )
+        logger.info(mtp_msg)
+        if self.cfg.get("debug_mtp_pp", False):
```
adil-a: i assume this function was here for debugging purposes?

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -1500,7 +1563,7 @@ def _forward_backward_step(
                         mtp_per_depth_h=mtp_per_depth_h,
                         labels=labels,
                         model=model,
-                        scaling_factor=out.mtp_loss_scaling_factor,
+                        scaling_factor=getattr(out, "mtp_loss_scaling_factor", 0.1),
```
adil-a: @akoumpa left a comment in my old PR that he doesn't want a default value being used. let's please revert this change

nemo_automodel/components/moe/parallelizer.py
```diff
@@ -57,6 +57,32 @@ def _get_cp_stream() -> torch.cuda.Stream:
...
+
+
+def _get_moe_module(block: nn.Module) -> MoE | None:
+    moe_module = getattr(block, "moe", None)
+    if isinstance(moe_module, MoE):
+        return moe_module
+    moe_module = getattr(block, "mlp", None)
+    if isinstance(moe_module, MoE):
+        return moe_module
+    return None
```
adil-a:
```suggestion
def _get_moe_module(block: nn.Module) -> MoE | None:
    for name in ("moe", "mlp"):
        module = getattr(block, name, None)
        if isinstance(module, MoE):
            return module
```

yiakwy-xpu-ml-framework-team:
@HuiyingLi  @khazic Do we have accuracy and accept length report ?

HuiyingLi:
@yiakwy-xpu-ml-framework-team Thank you for the comment. I will work on an accuracy report.

adil-a (approved):
LGTM, awesome work everyone!

PR #2192, opened 2026-05-08
fix(pp): preserve VLM forward when class opts in via _pp_keep_self_forward

khazic:
## What

`patch_hf_model_for_pp` unconditionally replaced `model.forward` with the generic `pipeline_forward_causal_lm` for any VLM that is not Gemma4 or Mistral3. Chunk-aware VLMs (Qwen3-VL-MoE, Qwen3.5-MoE, KimiVL, Kimi-K2.5-VL) keep the per-microbatch `pixel_values` in `self._vlm_pixel_values_chunks` and fetch them inside their **own** forward; the generic forward never reads that attribute, so under PP the vision tower silently never ran and image tokens were trained against placeholder text embeddings.

The existing recipes worked around this by setting `patch_inner_model: false` and `patch_causal_lm_model: false` in their YAMLs, but the framework had no way to enforce it — any new PP+VLM recipe that forgot the two flags would silently corrupt training.

## Repro (pre-fix)

```python
from nemo_automodel.components.distributed.pipelining.hf_utils import patch_hf_model_for_pp
# build a chunk-aware VLM, attach _vlm_pixel_values_chunks, call patch with defaults
patch_hf_model_for_pp(model, patch_inner_model=True, patch_causal_lm_model=True)
# model.forward is now generic causal_lm; pixel_values never reaches inner model
```

`grep _vlm_pixel_values_chunks pipelining/hf_utils.py` confirms only the dedicated `gemma4_vlm` and `mistral3_vlm` forwards read the chunks.

## Changelog

- Add opt-in class attribute `_pp_keep_self_forward`. When `True`, `patch_hf_model_for_pp` is a no-op for both inner and outer modules — preserves the model's chunk-aware forward.
- `validate_hf_model_for_pipeline_support` now refuses to start a PP job for a VLM that has neither a dedicated forward (gemma4 / mistral3) nor `_pp_keep_self_forward`, so the same misconfiguration cannot reach training silently again.
- Mark `Qwen3VLMoeForConditionalGeneration`, `Qwen3_5MoeForConditionalGeneration`, `KimiVLForConditionalGeneration`, `KimiK25VLForConditionalGeneration` with `_pp_keep_self_forward = True`.
- Unit tests for: the patch escape hatch, validation passing for chunk-aware VLMs, validation passing for dedicated-forward VLMs, validation raising for unsupported VLM-PP combos.

## Test plan

- [x] `pytest tests/unit_tests/distributed/pipelining/test_hf_utils.py` — 58 passed
- [x] `pytest tests/unit_tests/models/{qwen3_vl_moe,qwen3_5_moe,kimivl,kimi_k25_vl}/` — 489 passed (3 pre-existing flash-attn-env failures unrelated to this change)
- [x] End-to-end repro: confirmed `pixel_values` now reaches inner model under default `patch_*=True` for chunk-aware VLMs
- [ ] CI: L0 + L2 VLM PP suites

### DIFF
--- a/nemo_automodel/components/distributed/pipelining/functional.py
+++ b/nemo_automodel/components/distributed/pipelining/functional.py
@@ -36,6 +36,7 @@ from nemo_automodel.components.distributed.pipelining.hf_utils import (
     MULTIMODAL_SUFFIXES,
     TEXT_MODULE_ATTRS,
     get_text_module,
+    model_keeps_self_forward,
     patch_hf_model_for_pp,
 )
 
@@ -509,9 +510,18 @@ def split_model_into_stages(
         """Build a pipeline stage from specified module names."""
         # Deep copy the model
         stage_model = copy.deepcopy(model)
-        patch_hf_model_for_pp(
-            stage_model, patch_inner_model=patch_inner_model, patch_causal_lm_model=patch_causal_lm_model
-        )
+        # Two model implementation paths:
+        #   - HF / dedicated-patch path (LLMs, Gemma4 VLM, Mistral3 VLM): the
+        #     PP-aware forward lives in ``patch_hf_model_for_pp``.
+        #   - Custom-impl path that handles PP itself (Qwen3-VL-MoE, KimiVL,
+        #     Kimi-K2.5-VL, Qwen3.5-MoE): the model class declares
+        #     ``_pp_keep_self_forward = True`` so its own ``forward`` (which
+        #     pulls per-microbatch pixel_values from ``_vlm_pixel_values_chunks``)
+        #     stays intact.
+        if not model_keeps_self_forward(stage_model):
+            patch_hf_model_for_pp(
+                stage_model, patch_inner_model=patch_inner_model, patch_causal_lm_model=patch_causal_lm_model
+            )
         # Create a set of modules to keep
         modules_to_keep = set(module_names)
         modules_sorted = sorted(modules_to_keep, key=lambda x: x.split(".")[-1])
--- a/nemo_automodel/components/distributed/pipelining/hf_utils.py
+++ b/nemo_automodel/components/distributed/pipelining/hf_utils.py
@@ -586,14 +586,25 @@ def _is_gemma4_vlm(model: torch.nn.Module) -> bool:
     return getattr(text_config, "model_type", None) == "gemma4"
 
 
+def model_keeps_self_forward(model: torch.nn.Module) -> bool:
+    """Return True when *model* opts out of pipeline-aware forward patching.
+
+    Used by the pipeline split call site to skip ``patch_hf_model_for_pp``
+    entirely for models whose own ``forward`` is already PP-aware (typically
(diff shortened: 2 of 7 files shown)


HuiyingLi:
Hi @khazic , we have two paths for model implementation, HF implementation v.s. custom implementation (mostly moe). The flag `patch_hf_model_for_pp` is for HF implementation models. For custom moe models I don't think we need the patch since we have the model forward.

khazic:
Thanks @HuiyingLi — that framing is right and I've restructured the PR accordingly in `1a09f17`.

The decision of whether to patch now lives at the call site (`functional._build_stage_from_modules`), and `patch_hf_model_for_pp` is back to a single concern (pick the right patch flavour for HF / Gemma4 / Mistral3). The custom-impl MoE classes that own their PP-aware forward — Qwen3-VL-MoE, Qwen3.5-MoE, KimiVL, Kimi-K2.5-VL — now skip the patcher entirely instead of entering it and bailing.

One detail worth flagging: I kept the per-class `_pp_keep_self_forward = True` marker rather than deciding "skip patch iff custom-impl" purely from `MODEL_ARCH_MAPPING` membership. Mistral3 is also a custom-impl class registered in NeMo, but it actually does need `patch_hf_model_for_pp` for its dedicated `pipeline_forward_mistral3_vlm`. A blanket "custom = skip" rule would silently misroute Mistral3, so the flag stays as the precise signal. The call site reads it via the new `model_keeps_self_forward(model)` helper.

The validation guard added earlier in the PR (catches a future custom-impl VLM that forgets the marker) is unchanged.

New SHA for `/ok to test`: `1a09f173296d5635b9d95365a790bed2d69cf461`.

HuiyingLi:
We guard the PP patch by checking _pp_keep_self_forwad, but this marker is only added to a subset of models? What about the other models?

khazic:
Good question! The other VLMs fall into one of two categories:

1. **Gemma4 / Mistral3** — already have dedicated `pipeline_forward_*_vlm` functions inside `patch_hf_model_for_pp`, so they don't need the flag.
2. **Any VLM without (1) or (2)** — `validate_hf_model_for_pipeline_support` will now raise a `ValueError` at job start, preventing silent training corruption. See the new validation block at `hf_utils.py:707-735` and the test `test_validate_unsupported_vlm_pp_combination_raises`.

So no VLM can silently fall through to the generic CausalLM forward anymore — it either has explicit support or the job refuses to start.

khazic:
Friendly ping @HuiyingLi — CI is fully green on the latest SHA (`3b1a513`) after your re-trigger, including `L2_HF_Transformer_VLM`, `L2_Context_Parallel`, and `L2_Pretrain_and_KD`. The earlier scope question is addressed by the explicit `validate_hf_model_for_pipeline_support` raise plus the new test, so any VLM that doesn't opt in will refuse to start rather than silently fall through.

Happy to address any further comments — just flagging it for another look whenever you have a moment. Thanks!

HuiyingLi (approved):
Thanks for the explanation. We should gate the silent drop of vision for HF vlm PP. Thanks for the PR.

PR #2202, opened 2026-05-10
fix: add combined projection single-tensor conversion

taivu1998:
## Problem

NeMo-RL issue https://github.com/NVIDIA-NeMo/RL/issues/2072 reports that LoRA weight syncing fails for Qwen2/Llama Automodel policies because `CombinedProjectionStateDictAdapter` does not implement `convert_single_tensor_to_hf`.

The full `to_hf()` path already knows how to split combined `qkv_proj` and `gate_up_proj` tensors, but NeMo-RL's DTensor weight sync streams one tensor at a time and calls the per-tensor adapter hook. Without that hook, the sync path either errors or has to fall back to `force_hf=true`.

## Root Cause

`CombinedProjectionStateDictAdapter` supported full state-dict conversion only. The single-tensor conversion API was missing for the same combined projection layouts:

- `qkv_proj` -> `q_proj`, `k_proj`, `v_proj`
- `gate_up_proj` -> `gate_proj`, `up_proj`
- LoRA-A duplication for input-side adapter weights
- output-side/base/LoRA-B splitting
- 1-D bias gathering/restoration semantics

## Changes

- Add `convert_single_tensor_to_hf()` to `CombinedProjectionStateDictAdapter`.
- Reuse the existing grouped QKV and gate/up deinterleave helpers so single-tensor conversion matches `to_hf()`.
- Preserve `exclude_key_regex` behavior for skipped tensors such as `_extra_state`.
- Add regression tests for base weights, biases, LoRA-A duplication, LoRA-B splitting, pass-through tensors, and excluded tensors.
- Move heavy Llama/PEFT functional-test imports into the functional helper so lightweight adapter tests can collect independently.

## Validation

- `uvx ruff check ...`
- `uvx ruff format --check ...`
- `git diff --check`
- `/usr/local/bin/python3.13 -m py_compile ...`
- `pytest tests/unit_tests/models/common/test_combined_projection_state_dict_adapter.py`

The full combined-projection adapter test file passed locally: 22 passed.

## Related

- Fixes the Automodel side of NVIDIA-NeMo/RL#2072.

### DIFF
--- a/nemo_automodel/components/models/common/combined_projection/state_dict_adapter.py
+++ b/nemo_automodel/components/models/common/combined_projection/state_dict_adapter.py
@@ -366,6 +366,97 @@ class CombinedProjectionStateDictAdapter:
 
         return hf_state_dict
 
+    @staticmethod
+    def _filter_excluded_keys(
+        fqn_tensors: list[tuple[str, Any]],
+        exclude_key_regex: Optional[str] = None,
+    ) -> list[tuple[str, Any]]:
+        """Apply the same post-conversion exclusion semantics as to_hf()."""
+        if not exclude_key_regex:
+            return fqn_tensors
+        return [(key, value) for key, value in fqn_tensors if not re.match(exclude_key_regex, key)]
+
+    def convert_single_tensor_to_hf(self, fqn: str, tensor: Any, **kwargs) -> list[tuple[str, Any]]:
+        """Convert a single tensor from native format to HuggingFace format.
+
+        Args:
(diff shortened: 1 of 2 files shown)


akoumpa:
Hi @taivu1998 , thanks a lot for the contribution 🙇 , main must have moved and now there's a few small conflict, would you mind taking a look? Once the conflicts have been resolved, I can help trigger the CI pipeline. Thank you again for your contribution.

PR #2204, opened 2026-05-11
fix(recipes): correct validation loss averaging in LLM KD recipe

khazic:
## Summary
- `_run_validation_epoch` in the LLM KD recipe accumulated per-token-averaged losses from `_forward_backward_step` without un-averaging first
- This caused `val_loss` to be divided twice (artificially small) and `ce_loss`/`kd_loss` to be reported as raw sums (artificially large)
- Fix: multiply each batch's per-token loss by its `num_label_tokens` before accumulating, then divide by `total_num_label_tokens` at the end for a proper weighted average — matching the parent `FinetuneRecipe` pattern

## Test plan
- [ ] Run LLM KD training with validation enabled and verify val metrics are in the same range as training metrics
- [ ] Confirm `ce_loss` and `kd_loss` in validation logs are per-token averages, not raw sums

### DIFF
--- a/nemo_automodel/recipes/llm/kd.py
+++ b/nemo_automodel/recipes/llm/kd.py
@@ -797,9 +797,9 @@ class KnowledgeDistillationRecipeForNextTokenPrediction(TrainFinetuneRecipeForNe
             for mp in self.model_parts:
                 mp.eval()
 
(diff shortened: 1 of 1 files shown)


HuiyingLi (approved):
Thank you for fixing!

PR #2231, opened 2026-05-14
feat(loggers): MLflow run resumption, accurate run status, and VLM/MoE coverage

rob-luke:
# What does this PR do ?

Make MLflow logging restart-resilient, status-accurate, and consistent
across LLM / VLM / dLLM recipes — by switching from a custom wrapper
class to MLflow's native fluent API.

## Benefits for end users (operators running training)

- **Restart resilience.** When a job is preempted (Slurm) or manually
  restarted from a checkpoint, the same MLflow run is resumed — loss,
  LR, validation, and grad-norm curves stay contiguous in the UI
  instead of fracturing into new runs.
- **Opt-out resume.** If you don't want this restart-resilient
  behaviour (e.g. you reuse a `checkpoint_dir` across unrelated
  experiments and don't want their telemetry stitched together), set
  `mlflow.resume: false` and every launch will start a fresh run.
- **Run status now reflects reality.** Crashed runs report `FAILED`,
  cancelled runs report `KILLED`, only clean exits report `FINISHED`.
  Previously every termination — crash, SIGTERM, success — showed as
  `FINISHED`, so the UI couldn't distinguish them.
- **Hparam values are stable across reruns.** Previously, any callable
  in the config (e.g. `collate_fn: torch.utils.data.default_collate`)
  was logged with Python's default `repr`, which embeds the in-process
  memory address — `<function default_collate at 0x7f8b1234>`. The
  address is different on every launch, so identical configs logged
  different hparam values every time. Now those callables render as
  their dotted import path, so `collate_fn` is the literal string
  `torch.utils.data.default_collate` in every run and MLflow's compare
  view shows real diffs.
- **Full config preserved as an artifact.** `config.yaml` is uploaded
  on first launch; each resume adds a timestamped
  `config.resumed-<UTC-ts>.yaml` snapshot, so mid-flight config changes
  (extended `max_steps`, swapped dataset, etc.) leave an audit trail.
- **VLM finetuning and MoE metrics are logged to MLflow alongside
  wandb.** Previously VLM was wandb-only, and MoE expert-utilization /
  auxiliary-loss metrics were sent to wandb/comet but not MLflow.
- **Run notes from YAML.** Set `mlflow.description: "..."` and it
  renders as the Description panel in the MLflow UI.
- **Configurable hparam granularity.** New `mlflow.flatten_depth`
  controls how deeply nested config is split into individual hparams
  (default 1, matching prior behaviour).

## Benefits for developers and maintainers

- **Built on MLflow's supported upstream fluent API.** We no longer
  maintain a custom wrapper around an upstream API, and future MLflow
  features (auto-logging, system metrics, etc.) become available
  without reimplementing them in our wrapper.
- **Symmetry with wandb.** Each recipe now has parallel
  `if wandb.run is not None: wandb.log(...)` and
  `if mlflow.active_run() is not None: mlflow.log_metrics(...)`
  blocks; the two integrations look architecturally identical, so
  anyone who can read one can read the other.
- **No rank bookkeeping.** Non-rank-0 ranks naturally have no active
  MLflow run, so logging calls no-op; the explicit
  `dist.get_rank() == 0` checks scattered across the old wrapper are
  gone.
- **Pure-function helpers are unit-testable.** Test surface grows from
  ~460 to ~665 lines; 44 unit tests covering resume paths, status
  paths, hook chaining, metric coercion, and config flattening.

## What is intentionally NOT changed

- **YAML schema is additive only.** The five existing keys
  (`experiment_name`, `run_name`, `tracking_uri`, `artifact_location`,
  `tags`) behave identically; three new optional keys (`resume`,
  `description`, `flatten_depth`) are added with safe defaults.
- **Other loggers (wandb, comet) untouched** — same call sites, same
  config keys, same behaviour.

## One behavioural removal worth flagging

The previous `build_mlflow` silently injected three tags from `cfg`:

```python
tags["model"] = cfg.model.pretrained_model_name_or_path
tags["global_batch_size"] = cfg.step_scheduler.global_batch_size
tags["local_batch_size"] = cfg.step_scheduler.local_batch_size
```

These clashed with our internal systems. And these values are logged as parameters in mlflow. So duplicating them in tags enforced a preference on all users. Users can still manually add these tags if they wish.

# Changelog

- Remove `MLflowLogger` class and `build_mlflow()` builder.
- Add `configure_mlflow()` over MLflow's fluent API; recipes now log
  via module-level `mlflow.log_params` / `mlflow.log_metrics` gated by
  `mlflow.active_run()`.
- Add `to_float_metrics()` helper that drops non-numeric values (e.g.
  the `timestamp` string from `MetricsSample.to_dict()`) and coerces
  tensors to floats.
- Add `end_active_run_as_killed()` and wire it into the SIGTERM path
  in `DistributedSignalHandler`, so cancelled runs end with status
  `KILLED`.
- Add `_install_mlflow_failure_hook()` that chains a `sys.excepthook`
  to mark the active run `FAILED` on uncaught exceptions; idempotent
  and chains to the previous excepthook.
- Add run resumption: `MLFLOW_RUN_ID` env var (explicit override) or
  a `mlflow_run_id` sidecar in `checkpoint_dir` (implicit, gated by
  `mlflow.resume`, default `true`). New runs always write the sidecar.
- Add `mlflow.resume`, `mlflow.description`, `mlflow.flatten_depth`
  config keys; existing keys unchanged.
- Switch config logging from `cfg.to_dict()` to
  `cfg.to_yaml_dict(use_orig_values=True)` so callables render as
  dotted paths instead of memory-address `repr`s.
- Log full config as a `config.yaml` artifact; resume launches write a
  timestamped `config.resumed-<UTC-ts>.yaml` snapshot and skip param
  re-logging (which would raise `MlflowException` for any changed
  value).
- Add MLflow logging to the VLM finetune recipe (previously
  wandb-only).
- Add MoE metrics path to MLflow in the LLM recipe.
- Auto-injected `tags["model" / "global_batch_size" / "local_batch_size"]`
  removed; tags are now exactly what the user puts in YAML.
- Tests: 44 unit tests in `tests/unit_tests/loggers/test_mlflow_utils.py`
  + updates in `tests/unit_tests/training/test_train_ft_mlflow_logging.py`
  covering all of the above.

**Pre checks**:

- [x] Did you write any new necessary tests? (44 unit tests added / updated)

# Additional Information

- Related to # (issue) — none filed; open to filing a tracking issue if maintainers prefer.


rob-luke:
Sorry there were linting errors @akoumpa, I submitted a fix and hope they are fixed now.

pyproject.toml
```diff
@@ -430,6 +430,7 @@ ignore_imports = [
     "nemo_automodel.components.distributed.optimized_tp_plans -> nemo_automodel.components.models.llama.model",
     "nemo_automodel.components.distributed.optimized_tp_plans -> nemo_automodel.components.models.gemma4_moe.model",
     "nemo_automodel.components.distributed.optimized_tp_plans -> nemo_automodel.components.models.mistral3_vlm.model",
+    "nemo_automodel.components.training.signal_handler -> nemo_automodel.components.loggers.mlflow_utils",
```
akoumpa: Hi @rob-luke , thanks for the contribution, I'm wondering if it is possible to avoid this? 🤔

nemo_automodel/components/loggers/mlflow_utils.py
```diff
@@ -12,217 +12,238 @@
...
+    Also installs a `sys.excepthook` so crashed jobs report as FAILED rather
+    than FINISHED. After this call the recipe logs via module-level
+    `mlflow.log_params` and `mlflow.log_metrics` directly; on non-rank-0
+    processes `mlflow.active_run()` is None so those calls become no-ops
+    naturally.
+
+    Returns the active run on rank 0, or None when MLflow is not configured
+    or on non-rank-0 processes.
     """
+    if not (dist.is_initialized() and dist.get_rank() == 0):
```
akoumpa, line 40: I see here, that if it's not rank-0 then it won't configure it, so I'm thinking that it may not need the signal handler, but please let me know what you think @rob-luke

pyproject.toml
rob-luke: Great suggestion, I have tried another approach, please let me know if the new approach is better or if you have an alternative suggestion.

nemo_automodel/components/loggers/mlflow_utils.py
rob-luke: Great insight, @akoumpa , thank you for the feedback. I have pushed a change which moves the KILLED cleanup into the recipes themselves. Please let me know if this is preferable or you have further suggestions.

rob-luke:
@akoumpa the CI revealed two existing tests that I needed to update,  this is now done and I've merged the latest main in to this branch. It is ready to test again 🤞 thank you.

rob-luke:
Thank you @HuiyingLi for kicking off the tests. The failures look unrelated to the changes I made and seem to be NCCL flaking, could you please try kicking the tests off again or let me know if you think the errors are related to the code changes in this PR.

akoumpa (approved):
LGTM, thanks a lot @rob-luke !

rob-luke:
Thank you @akoumpa and @HuiyingLi

PR #2232, opened 2026-05-14
fix(megatron): make helpers_cpp compile portable across env activations

hawkoli1987:
# What does this PR do ?

Makes the Megatron C++ helpers' Makefile honor the active Python interpreter (via `$(PYTHON)` / `sys.executable`) instead of hard-coding `uv run python`, so the first-time build of `helpers_cpp.so` succeeds on the pip-install and container/HPC paths that `CONTRIBUTING.md` already promises to support.

`CONTRIBUTING.md` (section *"Developing with uv sync/pip install"*) states:

> *"Uv sync and pip install are both supported in Automodel."*

The Makefile contradicts that promise — it shells out to `uv run python` to probe `pybind11` includes and `EXT_SUFFIX`. On any environment activated via `source .venv/bin/activate` (the pip-install path) or running inside a container where `uv` is not on `PATH`, `make` resolves `uv` to "command not found", `CPPFLAGS` is left empty, `gcc` fails with `Python.h not found` / `pybind11/pybind11.h not found`, `compile_helper()` returns non-zero, and the recipe aborts before any training starts.

For users running everything through `uv run` (the recommended path), `$(PYTHON)` resolves to the interpreter `uv run` already activated — behaviour is unchanged. For pip-install / container / HPC users, the build now works on the first try.

# Changelog

- `nemo_automodel/components/datasets/llm/megatron/Makefile`
  - Introduced `PYTHON ?= python` and substituted `$(PYTHON)` for the hard-coded `uv run python` in both the `pybind11 --includes` probe and the `EXT_SUFFIX` lookup.
- `nemo_automodel/components/datasets/llm/megatron/megatron_utils.py`
  - `compile_helper()` now passes `PYTHON=sys.executable` explicitly into `subprocess.run(["make", ...])` so the build always uses the currently-loaded interpreter regardless of how the env was activated.
- `tests/unit_tests/datasets/llm/test_megatron_compile_helper.py`
  - New unit test (mocked `subprocess.run`) verifying `compile_helper` threads `PYTHON=sys.executable` into the `make` argv.

**Pre checks**:

# Additional Information

**Repro of the current failure**

In an env created via `pip install` (no `uv` on `PATH`), trigger the helpers compile:

```bash
python -c "
from nemo_automodel.components.datasets.llm.megatron.megatron_utils import compile_helper
compile_helper()
"
# /bin/sh: uv: command not found
# make: *** [Makefile:6: helpers_cpp.cpython-3xx-x86_64-linux-gnu.so] Error 127
```

**Verification**

- New unit test passes against the mocked subprocess (`tests/.../test_megatron_compile_helper.py`).
- End-to-end: compiled `helpers_cpp.so` on a PBS HPC compute node where the env was activated via `source .../bin/activate` (no `uv` on `PATH`); build completed, the `.so` loaded, `MegatronPretraining.__init__` no longer tripped on the missing helper.
- `ruff format` and `ruff check --fix` run clean on touched files.

**Backward compatibility**

Strictly additive. Users running everything through `uv run` see no behavioural change because `$(PYTHON)` resolves to the same interpreter `uv run` already activated. Users on the pip-install / container path stop hitting the `uv: command not found` error.


athitten (approved):
LGTM thank you @hawkoli1987 ! @akoumpa please take a look and approve as well, if evrything looks good to you. Thank you!

akoumpa (approved):

PR #2233, opened 2026-05-14
feat(megatron): accept flat-list blend JSON files

hawkoli1987:
# What does this PR do ?

Lets `MegatronPretraining` accept Megatron-LM's canonical flat-zipped blend form when the blend is supplied via a JSON file, not only when supplied as an inline list.

The `MegatronPretraining.paths` docstring already documents the flat form `["w1", "p1", "w2", "p2", ...]` as a valid *inline* list, and `get_blend_from_list` handles it correctly downstream. But the JSON loader (`try_load_blend_from_json`) rejects any JSON whose top-level isn't a dict:

```
ValueError: Blend JSON file must contain a dictionary, got <class 'list'>
```

This forces users emitting blends from Megatron-LM ecosystem tooling (Megatron-LM itself, Megatron-Bridge, NeMo-RL's blend tooling, NVIDIA preprocessing scripts) to manually wrap their blend JSON in `{"train": [...]}` before passing it to `MegatronPretraining`. This PR makes the JSON loader symmetric with the inline-list parser: if the JSON root is a list, return it as-is and let the existing downstream code (`get_blend_from_list` + the `split=` argument) route it through the same path the inline form already uses.

# Changelog

- `nemo_automodel/components/datasets/llm/megatron_dataset.py`
  - `try_load_blend_from_json`: return the list unchanged when JSON root is a list.
  - Updated return type annotation: `Optional[Union[Dict[str, List], List]]` (was `Optional[Dict[str, List]]`, which was already inaccurate for the `None` return on non-`.json` paths).
  - Rewrote the `try_load_blend_from_json` docstring to enumerate both accepted shapes (dict-of-splits + flat-list).
  - Updated the `MegatronPretraining.__init__` docstring's "Example JSON file format" block to document the flat-list option alongside the existing dict-of-splits form.
  - Relaxed the catch-all `ValueError` message from "must contain a dictionary" to "must contain a list or dictionary" to match the new behaviour.
- `tests/unit_tests/datasets/llm/test_blend_json_flat_list.py`
  - New unit test (~100 lines, 8 cases) covering: flat list of `[weight, prefix]` pairs, flat list of prefixes only, flat list with `s3://` prefixes, dict-of-splits regression (including the `valid → validation` alias normalization), scalar-root rejection, non-`.json` extension returns `None`, missing-file `FileNotFoundError`, malformed-JSON `ValueError`.

**Pre checks**:

# Additional Information

**Motivation**

Ecosystem tooling around Megatron-LM emits blend configs in the flat-list form:

```json
[
  "0.117",  "s3://bucket/EN_Fineweb-Edu",
  "0.201",  "s3://bucket/EN_MegaMath",
  "0.116",  "s3://bucket/EN_FinePDFs"
]
```

The wrapper-as-dict workaround adds nothing — the inline-list path already handles flat-zipped blends correctly via `get_blend_from_list`. The JSON loader was just narrower than its inline-list sibling.

**Backward compatibility**

Strictly additive. The dict-of-splits form (including `valid` / `val` / `dev` → `validation` alias normalization) is unchanged. The catch-all `ValueError` for unsupported root types (scalars, etc.) still fires; only its message text is relaxed.

**Verification**

- New unit tests cover the flat-list path plus the dict-of-splits regression to guard against breakage of the existing form.
- End-to-end: fed a Megatron-LM-format flat-list blend JSON (13 sources from S3) into `MegatronPretraining` for a real CPT run; the recipe parses the blend correctly and per-source weighting matches the JSON input.
- `ruff format` and `ruff check --fix` run clean on touched files.


athitten (dismissed):
LGTM thank you @hawkoli1987 for the detailed PR desc!

athitten (approved):
Re-approving after the CI test fix

PR #2244, opened 2026-05-15
fix(distributed): error on default TP plan fallthrough at tp_size>1

### LINKED ISSUE #2243: TP > 1 with HF custom-code arch (trust_remote_code) crashes in DTensor redistribute on shard_order=None
### Bug

Loading an HF custom-code architecture (a model with `auto_map.AutoModelForCausalLM` in its `config.json`, loaded via `trust_remote_code=True`) and parallelizing with `tensor_parallel_size > 1` crashes during the first DTensor redistribute (weight load or refit) with:

```
File ".../torch/distributed/tensor/_redistribute.py", line 752, in _gen_transform_infos_non_cached
    assert src_shard_order is not None and dst_shard_order is not None
AssertionError
```

The same model runs fine with `tensor_parallel_size=1`. Architectures explicitly registered in `PARALLELIZE_FUNCTIONS` (Qwen2/3, Llama, Gemma3, Phi3, CustomLlama, CustomQwen2; see `nemo_automodel/components/distributed/optimized_tp_plans.py:303`) are unaffected because they take a different branch and produce placements that carry the `shard_order` metadata that PyTorch DTensor now asserts on.

### Where it falls through

`_get_parallel_plan` in `nemo_automodel/components/distributed/parallelizer.py:825` selects a TP plan in priority order:

1. Custom `tp_shard_plan` if provided.
2. `use_hf_tp_plan` if explicitly set (relies on `model_cls._tp_plan`).
3. `model_cls in PARALLELIZE_FUNCTIONS`.
4. Otherwise the default base plan at `parallelizer.py:888`, which uses module-name patterns like `model.layers.*.self_attn.q_proj`.

A HF custom-code arch that is not in `PARALLELIZE_FUNCTIONS` and does not define `_tp_plan` falls into path 4. The base plan's placements appear not to populate `shard_order`, which trips the assertion the next time PyTorch tries to redistribute one of the resulting DTensors.

### Reproduction

Any HF model that ships custom modeling code via `auto_map`, with TP > 1. Example I hit in practice was a Kanana2Tiny student (vendored `modeling_kanana2_tiny.py` / `configuration_kanana2_tiny.py`) inside an on-policy distillation run; the same precondition applies to any custom-code arch.

```yaml
dtensor_cfg:
  enabled: true
  tensor_parallel_size: 2
policy:
  model_name: /path/to/custom-code-arch-model
```

Crash hits on the first weight redistribute. Hardware independent.

### Workaround

Set `tensor_parallel_size: 1` on the policy's `dtensor_cfg` (and keep the matching `generation.vllm_cfg.tensor_parallel_size=1` for clean refit, when colocated with vLLM). Acceptable for small students that fit FSDP only.

### Suggested fix

Either of these would be a real improvement, in rough order of preference:

1. Make the default base plan in `_get_parallel_plan` produce placements that carry `shard_order`, so unknown archs that happen to follow the default `model.layers.*.self_attn.q_proj` naming work at TP > 1 out of the box.
2. Failing that, have `_get_parallel_plan` refuse `tp_size > 1` for archs that hit path 4 with a clear error message naming the missing arch and pointing at `PARALLELIZE_FUNCTIONS`, `_tp_plan`, and `tp_shard_plan` as the three supported ways to register a working plan. Strictly better than the current PyTorch internal assertion.

Happy to send a PR for option 2 if maintainers agree on the surface; option 1 needs more local knowledge of how the per-arch plans set `shard_order`.

### Environment

* `nemo_automodel` at the version vendored in `NVIDIA-NeMo/RL` `3rdparty/Automodel-workspace/Automodel` (commit `1d42deb`).
* PyTorch version recent enough that `_redistribute._gen_transform_infos_non_cached` asserts on `shard_order is not None`.

bzantium:
# What does this PR do?

Closes #2243.

When `_get_parallel_plan` falls through to the default base plan (path 4) at `tp_size > 1`, raise a clear `ValueError` instead of returning a plan that will later crash inside PyTorch's DTensor `_redistribute` with an opaque `AssertionError` on `shard_order is not None`.

This was hit in practice when running tensor parallelism over an HF custom-code architecture (a model with `auto_map.AutoModelForCausalLM` loaded via `trust_remote_code=True`) that is neither registered in `PARALLELIZE_FUNCTIONS` nor exposes a `_tp_plan` on the model class. The base plan produces `Colwise/RowwiseParallel` placements that do not populate `shard_order`, so newer PyTorch versions assert on the first weight redistribute (weight load or refit). The error surface up to here is the internal PyTorch assert, which is hard to debug without reading `parallelizer.py`.

The new `ValueError` names the offending class and points at the three supported registration paths:

1. Add a plan function to `nemo_automodel.components.distributed.optimized_tp_plans.PARALLELIZE_FUNCTIONS`.
2. Define `_tp_plan` on the model class (HF-native per-model TP plan).
3. Pass `tp_shard_plan` (dict or import path) when constructing the parallelizer.

At `tp_size == 1` no real sharding happens, so the default plan is still returned and existing single-GPU / DP-only callers are unaffected.

# Changelog

- `nemo_automodel/components/distributed/parallelizer.py`
  - `_get_parallel_plan`: new `tp_size: int = 1` parameter; raise `ValueError` with an actionable message when path 4 would apply at `tp_size > 1`; updated docstring to reflect that HF-native `_tp_plan` is preferred over the default base plan.
  - `DefaultParallelizationStrategy.parallelize`: thread `tp_size=tp_mesh.size()` through to `_get_parallel_plan`.
- `nemo_automodel/components/distributed/megatron_fsdp.py`
  - `MegatronFSDPManager.parallelize`: pass `tp_size=self.device_mesh["tp"].size()` to `_get_parallel_plan` so the MegatronFSDP path gets the same fail-fast behavior.
- `tests/unit_tests/distributed/test_get_parallel_plan.py`
  - `test_default_plan_fallthrough_raises_at_tp_size_gt_1` — verifies the new `ValueError` (with and without sequence parallel) and asserts that the message names the class and all three registration paths.
  - `test_default_plan_fallthrough_tp_size_1_still_returns_base_plan` — pins the backwards-compatible behavior at `tp_size == 1`.
  - `test_hf_native_plan_unaffected_at_tp_size_gt_1` — verifies models that expose `_tp_plan` are not affected by the new guard.

**Pre checks**:

- [x] Did you add or update any necessary documentation? (docstring of `_get_parallel_plan` updated)

Verification:

- `pytest tests/unit_tests/distributed/test_get_parallel_plan.py` — 24/24 pass (3 new + 21 existing).
- `pytest tests/unit_tests/distributed/test_parallelization_strategies.py` — 32/32 pass.
- `pytest tests/unit_tests/models/qwen3_5/test_cp_linear_attn_patch.py` — 36/36 pass.
- `ruff format` / `ruff check` clean on changed files.

# Additional Information

- Picked option 2 from the issue ("refuse `tp_size > 1` for archs that hit path 4 with a clear error message") per @HuiyingLi's response on #2243. Option 1 (make the default base plan carry `shard_order`) is a strictly better fix but, as called out in the issue, needs more local knowledge of how per-arch plans populate the field — happy to follow up in a separate PR.
- Workaround that no longer crashes silently: users who hit the new error can either (a) register the arch via one of the three paths the message lists, or (b) drop to `tp_size=1` as before.

### DIFF
--- a/nemo_automodel/components/distributed/parallelizer.py
+++ b/nemo_automodel/components/distributed/parallelizer.py
@@ -1424,6 +1424,12 @@ def _get_parallel_plan(
     https://github.com/NVIDIA-NeMo/Automodel/issues/2243); known HF
     architectures that happen to fall through (e.g. Mixtral) are left on the
     default plan with a warning, since they have been working in practice.
+
+    When the model *did* define a ``_tp_plan`` but ``get_hf_tp_shard_plan``
+    raised while translating it (e.g. styles nemo does not recognize), the
+    translator's error message is folded into the ``ValueError`` as a
+    diagnostic so the user can tell whether to add a ``_tp_plan`` from
+    scratch or fix the styles in the one they already have.
     """
     model_parallel_plan = None
     model_cls = type(model)
@@ -1485,9 +1491,11 @@ def _get_parallel_plan(
         # live under model.language_model.layers.* and would be missed by the
         # hardcoded llama-style wildcards below.
         hf_plan = None
+        hf_plan_error: Exception | None = None
         try:
             hf_plan = get_hf_tp_shard_plan(model)
         except Exception as e:
+            hf_plan_error = e
             logger.info(f"HF tp plan not available ({e}). Falling back to default base plan.")
 
         if hf_plan:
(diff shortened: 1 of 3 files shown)


HuiyingLi:
Thanks @bzantium ci is triggered again on latest commit.

bzantium:
@HuiyingLi Thanks for the fast reply!

nemo_automodel/components/distributed/parallelizer.py
```diff
@@ -1478,6 +1494,36 @@ def _get_parallel_plan(
             model_parallel_plan = hf_plan
             logger.info(f"Using HF-native tp plan for {model_cls.__name__}.")
         else:
+            # HF places dynamic classes loaded via ``trust_remote_code=True`` under the
+            # ``transformers_modules.*`` namespace. Those are the only archs known to
+            # actually crash inside ``_redistribute`` with the default base plan, so we
+            # only fail-fast for them. See https://github.com/NVIDIA-NeMo/Automodel/issues/2243.
+            is_remote_code = (model_cls.__module__ or "").startswith("transformers_modules.")
+            if tp_size > 1 and is_remote_code:
```
HuiyingLi, line 1510: Would there be remote hub models that comes with _tp_plan?

bzantium: Good catch, yes there can be. Walked through the cases:

If a remote hub model defines `_tp_plan` and the styles are recognized by `translate_to_torch_parallel_style`, `get_hf_tp_shard_plan` returns a non-empty plan and we use path 3, the guard never fires. Two real examples I checked are Llama and Mistral3, both of which ship with `_tp_plan = {"lm_head": "colwise_rep"}` and are happily handled by path 3.

The corner case is a remote-hub model whose `_tp_plan` uses styles nemo cannot translate (e.g. `colwise_gather_output`, HF MoE styles like `local_colwise` / `ep_router`). There `get_hf_tp_shard_plan` raises during translation, we land at path 4, and the previous wording (`"the model does not expose a HuggingFace _tp_plan"`) was misleading, `_tp_plan` *was* there, just unusable.

Fixed in `46a15b57`: the translator's exception is now folded into the ValueError as a `Note: ...` line, and suggestion (2) points at the recognized style names so the author knows what to switch to. Unit test added.

HuiyingLi (approved):
LGTM thank you so much!

PR #2247, opened 2026-05-15
fix(checkpoint): exclude TE _extra_state keys from load-time mismatch warning

adil-a:
## What

When loading an HF safetensors checkpoint into a model that contains
TransformerEngine modules, the checkpoint loader emits a noisy
`Checkpoint key mismatch` warning. The "missing" keys are all
`_extra_state` entries that TE attaches to its own modules for internal
bookkeeping (FP8 amax history, etc.) and that are not part of the HF
safetensors checkpoint. The framework already neutralizes these via the
`set_extra_state` shim in `components/checkpoint/stateful_wrappers.py`,
so reporting them as missing weights is misleading.

This PR filters `*_extra_state` keys out of both sides of the
mismatch summary so the warning only fires when real weights are
actually missing or unexpected.

## Observed warning (8-rank FSDP2 + EP load of NVIDIA-Nemotron-3-Nano-30B-A3B-BF16)

```
WARNING:root:Checkpoint key mismatch for FSDPNemotronHForCausalLM: missing=77 unexpected=0 (missing examples=['lm_head._extra_state', 'model.layers.1.mixer.shared_experts.down_proj._extra_state', 'model.layers.1.mixer.shared_experts.up_proj._extra_state', 'model.layers.10.mixer.shared_experts.down_proj._extra_state', 'model.layers.10.mixer.shared_experts.up_proj._extra_state', 'model.layers.12.mixer.attn_module._extra_state', 'model.layers.12.mixer.k_proj._extra_state', 'model.layers.12.mixer.o_proj._extra_state', 'model.layers.12.mixer.q_proj._extra_state', 'model.layers.12.mixer.v_proj._extra_state'], unexpected examples=[])
```

All 77 missing keys end in `._extra_state`.

## Changelog

- Exclude `*_extra_state` keys from the load-time
  `Checkpoint key mismatch` warning so it focuses on real weight
  mismatches.

## Pre-checks

- [x] Linting: no formatting/style changes required
- [x] DCO sign-off

### DIFF
--- a/nemo_automodel/components/checkpoint/checkpointing.py
+++ b/nemo_automodel/components/checkpoint/checkpointing.py
@@ -554,7 +554,9 @@ class Checkpointer:
             state_dict[lm_head_param_name] = state_dict.pop(compat_tied_lm_head_source_key)
 
         state_dict = _maybe_adapt_state_dict_from_hf(model_state.model[0], state_dict, moe_mesh=self.moe_mesh)
-        key_diff = _summarize_state_dict_key_diff(expected_keys, set(state_dict.keys()))
+        expected_keys_for_diff = {k for k in expected_keys if not k.endswith("_extra_state")}
+        loaded_keys_for_diff = {k for k in state_dict if not k.endswith("_extra_state")}
+        key_diff = _summarize_state_dict_key_diff(expected_keys_for_diff, loaded_keys_for_diff)
         if key_diff["missing_count"] or key_diff["unexpected_count"]:
             logging.warning(
                 "Checkpoint key mismatch for %s: missing=%d unexpected=%d "


nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -554,7 +554,9 @@ def load_model(
             state_dict[lm_head_param_name] = state_dict.pop(compat_tied_lm_head_source_key)
 
         state_dict = _maybe_adapt_state_dict_from_hf(model_state.model[0], state_dict, moe_mesh=self.moe_mesh)
-        key_diff = _summarize_state_dict_key_diff(expected_keys, set(state_dict.keys()))
+        expected_keys_for_diff = {k for k in expected_keys if not k.endswith("_extra_state")}
```
akoumpa, line 557: Thanks @adil-a , i think for bf16 this should be fine, but thinking ahead, I'm wondering if that would break any fp8 workflows? Please let me know what you think.

adil-a: how about we put a check to ensure the model is in bf16 as well?

akoumpa: IMHO, i would remove `_extra_state` if it's empty on the model side

adil-a: response from CC:

Good question — I dug into this and it turns out FP8 workflows aren't affected, because the codebase already treats _extra_state as ephemeral on every save/load
  path. Quick summary of why:

  TE-based FP8. TE's FP8 amax history lives in _extra_state, but it's never round-tripped through our checkpointer:
  - _maybe_adapt_state_dict_to_hf in checkpointing.py calls adapter.to_hf(..., exclude_key_regex=r".*_extra_state.*", ...) unconditionally on save (line 1590).
  - stateful_wrappers.py monkey-patches TransformerEngineBaseModule.set_extra_state and BasicOperation.set_extra_state to no-op on DCP's _EXTRA_STATE sentinel (lines
   31–49), gated only on HAS_TE — not on FP8 being enabled.
  - te_attention.py stashes the TE attention module via object.__setattr__ specifically so attn_module._extra_state never enters the state_dict (lines 548–553, with
  a comment to that effect).
  - On load, when a TE module has a custom get_extra_state, an empty torch.tensor([], dtype=uint8) placeholder is injected so DCP doesn't complain (lines 1437–1443).

  Net result: TE FP8 amax history is rebuilt from observed activations after load. Filtering _extra_state from the mismatch warning only hides keys that the rest of
  the framework is already silently dropping.

  torchao Float8Linear. Doesn't use _extra_state at all — Float8Linear.{get,set}_extra_state is nn.Module.{get,set}_extra_state, its state_dict() is ['weight', 
  'bias'], and the weight stays as a bf16/fp32 nn.Parameter (FP8 conversion happens dynamically in forward).

  So unconditionally filtering _extra_state from the warning is consistent with how all four other code paths handle it, and won't mask any real mismatch for FP8
  workflows.

adil-a: > IMHO, i would remove _extra_state if it's empty on the model side

It comes with the TE nn.Modules as a parameter. We'd have to do unnecessary plumbing for this. TBH checkpointing is already in a finicky spot I'd rather we avoid changing things where possible in the current state.

akoumpa (approved):

PR #2258, opened 2026-05-18
fix(eagle3): avoid UnboundLocalError on empty train dataloader

khazic:
## Summary

``TrainEagle3Recipe.run_train_validation_loop`` referenced ``batch_idx + 1`` in the end-of-epoch summary log, but ``batch_idx`` is only bound inside the ``for batch_idx, batch in enumerate(self.train_dataloader)`` loop. If the dataloader yields zero batches — empty split, an over-aggressive filter, or a misconfigured data path — the log line raises ``UnboundLocalError`` before the user sees any actionable error.

Fix: initialize ``batches_seen = 0`` before the loop, update it inside the loop, and read that counter in the summary. The log now honestly reports ``total_batches_seen=0`` instead of crashing.

Minimal, local change to ``recipes/llm/train_eagle3.py`` only.

## Test plan

- [x] AST-parses cleanly
- [x] Manual reasoning: ``for ... in iter([])`` never binds the loop var; the new ``batches_seen`` is bound unconditionally so the post-loop log never references an unbound name

### DIFF
--- a/nemo_automodel/recipes/llm/train_eagle3.py
+++ b/nemo_automodel/recipes/llm/train_eagle3.py
@@ -301,6 +301,12 @@ class TrainEagle3Recipe(BaseRecipe):
             running_steps = 0
             self.optimizer.zero_grad(set_to_none=True)
 
+            # Track batches-seen explicitly so the epoch-summary log below is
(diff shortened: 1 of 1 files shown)


HuiyingLi (approved):
LGTM thank you!

PR #2277, opened 2026-05-19
fix(deepseek-v4): align HCA backward graph under FSDP2

HaloWorld:
# What does this PR do ?

This PR fixes a DeepSeek V4 HCA backward collective mismatch in the current 1D PyTorch FSDP2
DeepSeek V4 wrapper path.

DeepSeek V4 HCA only creates compressed KV entries after a complete 128-token HCA window is available.
When ranks in the same HCA parameter sync group receive mixed short and long sequences, short ranks can
skip the HCA compressor path while long ranks execute it. That makes the backward graphs differ across
ranks and can trigger FSDP2 collective mismatch.

The fix makes the HCA compressor path group-aware:

- ranks with a complete HCA window keep the original real HCA path;
- short ranks add one fully masked synthetic HCA window only when another rank in the same HCA parameter
  sync group has a complete HCA window;
- all-short groups keep the original no-HCA path, so HCA compressor parameters keep `grad=None`.

This preserves optimizer-visible all-short semantics while aligning the backward graph in mixed
short/long groups.

# Changelog

- Add a DeepSeek V4-only helper to derive the HCA parameter sync group from a named or unnamed 1D FSDP2
  mesh.
- Attach that group to `DeepseekV4Compressor` during DeepSeek V4 FSDP wrapping.
- Add a scalar max-reduce in the HCA compressor to detect whether any rank in the parameter sync group
  has a complete HCA window.
- Add a fully masked synthetic HCA window only for mixed short/long HCA groups.
- Preserve all-short optimizer semantics by avoiding synthetic HCA windows when no rank has a complete
  HCA window.
- Add CPU unit tests for:
  - all-short no-HCA behavior and `grad=None` semantics;
  - synthetic HCA window masking;
  - zero-valued local gradients for synthetic short ranks;
  - no-mask calls keeping the original no-HCA behavior;
  - HCA boundary lengths around 128;
  - named/unnamed 1D mesh group extraction, non-1D fallback behavior, and all-fp32 wrapping.

**Pre checks**:

# Additional Information

- Related to # (issue)
  - #1364 reports a different distributed variable-token-count deadlock in NemotronV3 EP. This PR fixes a
separate DeepSeek V4 HCA/FSDP2 conditional-compressor participation issue.
  - #2216 is related to DeepSeek V4 FSDP2 wrapping, but fixes fp32 parameter preservation rather than HCA
mixed short/long backward graph alignment.

# Implementation Notes

The correctness invariant is the same for FSDP and FSDP2: all ranks that synchronize the same sharded
HCA compressor parameters must either all create the HCA compressor backward graph or all skip it.

This patch wires that invariant through the existing DeepSeek V4 PyTorch FSDP2 wrapping path, where the
parameter sync domain is available as a 1D device mesh process group. The included distributed
validation uses PyTorch FSDP2, which is the DeepSeek V4 training path this issue was reproduced on.
Other FSDP backends can use the same HCA compressor alignment hook if they provide the HCA parameter
sync group, but this PR does not add or validate a separate MegatronFSDP/classic FSDP integration path.

For multi-dimensional meshes, choosing the correct HCA parameter sync domain requires explicit ownership
information. The helper returns `None` for non-1D meshes instead of guessing a broader or wrong process
group. This PR does not claim full HSDP/CP/TP/PP/EP coverage.

The scalar collective is one max-reduce per HCA compressor forward when alignment is enabled. It only
runs for DeepSeek V4 HCA training with the normal model-path attention mask and `compress_ratio == 128`.

Synthetic HCA graph alignment requires the explicit 4D additive attention mask built by the normal
DeepSeek V4 model path, so the synthetic compressed position can be fully masked. Direct/custom training
calls without an attention mask intentionally keep the original no-HCA behavior and may still see the
original mixed short/long mismatch under FSDP2.

The short-path output-equivalence tests use `attention_dropout=0.0`, matching the DeepSeek V4 default.
Nonzero custom attention dropout can change RNG consumption even though the synthetic position remains
zero-probability and zero-gradient. Zero-token local HCA inputs are also outside this narrow fix.

---

2-GPU PyTorch FSDP2 repro validation:

Run it with two GPUs:

```text
PATH=. torchrun --standalone --nproc_per_node=2 /tmp/reproduce_hca_fsdp2.py
```

<details>
<summary>reproduce_hca_fsdp2.py</summary>

```python
"""Minimal DeepSeek-V4 HCA/FSDP2 before/after reproducer.

Run with two GPUs using ``torchrun --standalone --nproc_per_node=2``. Before this fix,
``short_long`` can fail during backward; after the fix, all three cases pass.
"""

from datetime import timedelta
import os
import sys
import traceback

import torch
import torch.distributed as dist
import torch.nn as nn
from torch.distributed.device_mesh import init_device_mesh
from torch.distributed.fsdp import MixedPrecisionPolicy, fully_shard

from nemo_automodel.components.models.deepseek_v4.config import DeepseekV4Config
from nemo_automodel.components.models.deepseek_v4.layers import (
    DeepseekV4Compressor,
    DeepseekV4RotaryEmbedding,
    DeepseekV4TrainCache,
)


def make_inputs(config, seq_len, device):
    values = torch.arange(seq_len * config.hidden_size, device=device, dtype=torch.float32)
    hidden_states = (values.view(1, seq_len, config.hidden_size) % 251) / 251.0
    position_ids = torch.arange(seq_len, device=device).unsqueeze(0)
    partial_rotary_factor = float(config.qk_rope_head_dim) / float(config.head_dim)
    rotary_compress = DeepseekV4RotaryEmbedding(
        float(config.compress_rope_theta),
        int(config.head_dim),
        partial_rotary_factor,
        rope_scaling=config.rope_scaling,
    ).to(device)
    return hidden_states, rotary_compress, rotary_compress(hidden_states, position_ids)


def run_backward_case(name, seq_lens, config, mesh, mp_policy, rank, world_size, device):
    seq_len = seq_lens[rank]
    print(f"[rank {rank}] {name}: seq_len={seq_len}", flush=True)

    torch.manual_seed(20260518)
    base = nn.Linear(config.hidden_size, config.hidden_size, bias=False).to(device)
    compressor = DeepseekV4Compressor(config, 128, config.head_dim).to(device=device, dtype=torch.float32)
    compressor.train()
    if hasattr(compressor, "_set_hca_param_sync_group"):
        compressor._set_hca_param_sync_group(mesh.get_group())
    fully_shard(compressor, mesh=mesh, mp_policy=mp_policy)

    hidden_states, rotary_compress, pos_emb_compress = make_inputs(config, seq_len, device)
    kwargs = {
        "hidden_states": hidden_states,
        "q_residual": None,
        "rotary": rotary_compress,
        "position_embeddings": pos_emb_compress,
        "cache": DeepseekV4TrainCache(),
        "layer_idx": 0,
        "start_pos": 0,
    }
    if "enable_hca_fsdp_graph_alignment" in compressor.forward.__code__.co_varnames:
        kwargs["enable_hca_fsdp_graph_alignment"] = True
    pooled, _ = compressor(**kwargs)
    loss = base(hidden_states).float().square().mean() + pooled.float().square().sum()
    print(f"[rank {rank}] {name}: loss={loss.item():.8g}; backward", flush=True)
    loss.backward()

    params = [param for _, param in compressor.named_parameters()]
    presence = torch.tensor([0 if p.grad is None else 1 for p in params], device=device, dtype=torch.int32)
    gathered = [torch.empty_like(presence) for _ in range(world_size)]
    dist.all_gather(gathered, presence)
    by_rank = torch.stack(gathered).cpu().tolist()
    if rank == 0:
        print(f"[rank 0] {name}: hca_grad_presence_by_rank={by_rank}", flush=True)
        if any(row != by_rank[0] for row in by_rank[1:]):
            raise RuntimeError(f"{name}: HCA backward participation differs across ranks")
        print(f"[rank 0] {name}: PASS", flush=True)


def main():
    rank = int(os.environ["RANK"])
    world_size = int(os.environ["WORLD_SIZE"])
    local_rank = int(os.environ["LOCAL_RANK"])
    torch.cuda.set_device(local_rank)
    dist.init_process_group("nccl", timeout=timedelta(seconds=int(os.environ.get("DIST_TIMEOUT_SECONDS", "180"))))
    try:
        if world_size != 2:
            raise RuntimeError(f"expected exactly 2 ranks, got {world_size}")
        device = torch.device("cuda", local_rank)
        print(f"[rank {rank}] torch={torch.__version__}; device={torch.cuda.get_device_name(local_rank)}", flush=True)
        config = DeepseekV4Config(compress_ratios=[128])
        mesh = init_device_mesh("cuda", (world_size,), mesh_dim_names=("dp",))
        mp_policy = MixedPrecisionPolicy(torch.float32, torch.float32, torch.float32, cast_forward_inputs=False)
        for name, seq_lens in (("short_short", (7, 7)), ("long_long", (129, 129)), ("short_long", (7, 129))):
            run_backward_case(name, seq_lens, config, mesh, mp_policy, rank, world_size, device)
            dist.barrier()
        if rank == 0:
            print("[rank 0] PASS dsv4 hca fsdp2 minimal repro", flush=True)
        return 0
    except Exception:
        print(f"[rank {rank}] FAILED", flush=True)
        traceback.print_exc()
        return 1
    finally:
        dist.destroy_process_group()


if __name__ == "__main__":
    sys.exit(main())
```

</details>

```text
before patch, Slurm job 183017:
[rank 0] short_short: PASS
[rank 0] long_long: PASS
[rank 0] short_long: loss=0.11253042; backward
[rank 1] short_long: loss=512.10651; backward
[rank 0] FAILED
[rank 1] FAILED
RuntimeError: Detected mismatch between collectives on ranks.
```

```text
after patch, Slurm job 183043:
[rank 0] short_short: hca_grad_presence_by_rank=[[0, 0, 0, 0], [0, 0, 0, 0]]
[rank 0] short_short: PASS
[rank 0] long_long: hca_grad_presence_by_rank=[[1, 1, 1, 1], [1, 1, 1, 1]]
[rank 0] long_long: PASS
[rank 0] short_long: hca_grad_presence_by_rank=[[1, 1, 1, 1], [1, 1, 1, 1]]
[rank 0] short_long: PASS
[rank 0] PASS dsv4 hca fsdp2 minimal repro
```

### DIFF
--- a/nemo_automodel/components/models/deepseek_v4/fsdp.py
+++ b/nemo_automodel/components/models/deepseek_v4/fsdp.py
@@ -39,6 +39,42 @@ _DSV4_FP32_MODULE_SUFFIXES = (
 )
 
 
+def _hca_param_sync_group_from_1d_mesh(mesh):
+    """Return the 1D PyTorch FSDP2 group used for HCA graph alignment.
+
+    HCA graph alignment is an FSDP/FSDP2 parameter-sync invariant: ranks that
+    synchronize the same sharded HCA parameters must agree on whether the HCA
+    compressor path participates in backward. This DeepSeek-V4 wrapper gets
+    that domain from its 1D PyTorch FSDP2 mesh. The mesh may be named or
+    unnamed; multi-dimensional meshes need an explicit owner dimension to avoid
+    reducing across unrelated parallel groups. Until that is available, disable
+    HCA graph alignment instead of using a broader or wrong group.
+    """
+    if mesh is None:
+        return None
+
+    mesh_ndim = getattr(mesh, "ndim", None)
+    mesh_shape = getattr(mesh, "shape", None)
+    mesh_dim_names = getattr(mesh, "mesh_dim_names", None)
+    if mesh_ndim is not None:
+        is_1d_mesh = mesh_ndim == 1
+    elif mesh_shape is not None:
+        is_1d_mesh = len(mesh_shape) == 1
+    elif mesh_dim_names is not None:
+        is_1d_mesh = len(mesh_dim_names) == 1
+    else:
+        return None
+    if not is_1d_mesh:
+        return None
+
+    try:
+        if mesh.size() <= 1:
+            return None
+        return mesh.get_group()
+    except (AttributeError, RuntimeError, TypeError, ValueError):
+        return None
+
+
 def _matches_suffix(name: str, suffix: str) -> bool:
     return name == suffix or name.endswith(f".{suffix}")
 
@@ -131,6 +167,17 @@ def _iter_dsv4_fp32_modules(module: nn.Module):
         yield submodule
 
 
+def _attach_hca_param_sync_group(module: nn.Module, mesh) -> None:
+    process_group = _hca_param_sync_group_from_1d_mesh(mesh)
+    for submodule in module.modules():
+        setter = getattr(submodule, "_set_hca_param_sync_group", None)
+        if submodule.__class__.__name__ == "DeepseekV4Compressor" and setter is not None:
+            # The FSDP2 mesh is only known while wrapping. Bind its parameter
+            # sync group narrowly to the DeepSeek-V4 HCA compressor instead of
+            # adding public config.
+            setter(process_group)
+
+
 def fully_shard_deepseek_v4(module: nn.Module, mesh, mp_policy, offload_policy=None, **fsdp_kwargs):
     """Apply FSDP2 to DeepSeek-V4 without mixing fp32 and bf16 params in one unit.
 
@@ -138,6 +185,10 @@ def fully_shard_deepseek_v4(module: nn.Module, mesh, mp_policy, offload_policy=N
     reference-sensitive tensors in fp32, while the existing DeepEP path expects
     the transformer block itself to remain the main FSDP unit.
     """
+    is_dsv4 = _is_deepseek_v4_module(module)
+    if is_dsv4:
+        _attach_hca_param_sync_group(module, mesh)
+
     if _floating_param_dtypes(module) == {torch.float32}:
         return _fully_shard_once(
             module,
@@ -148,7 +199,7 @@ def fully_shard_deepseek_v4(module: nn.Module, mesh, mp_policy, offload_policy=N
             **fsdp_kwargs,
         )
 
-    if not _is_deepseek_v4_module(module):
+    if not is_dsv4:
         return _fully_shard_once(
             module,
             mesh=mesh,
--- a/nemo_automodel/components/models/deepseek_v4/layers.py
+++ b/nemo_automodel/components/models/deepseek_v4/layers.py
@@ -56,6 +56,7 @@ from __future__ import annotations
 from typing import Any, Optional
 
 import torch
+import torch.distributed as dist
 import torch.nn as nn
 import torch.nn.functional as F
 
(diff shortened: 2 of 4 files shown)


HuiyingLi:
@HaloWorld  Thank you for your PR!

HuiyingLi (approved):
Thank you for the fix

PR #2283, opened 2026-05-20
feat: Add late interaction model training support for retrieval

rnyak:
# What does this PR do ?

Updated `train_bi_encoder.py` to support local ColBERT-style (multi_vector) pooling by adding `colbert_scores_and_labels()`, which computes MaxSim scores with query and passage attention-mask handling. The train and validation paths now route ColBERT models through this scoring function instead of standard pooled embedding contrastive scoring. 

The changes in the MR add support for `multi_vector` pooling and maxsim scoring with and without distributed in-batch neg training.

The latest commits:

- Added `detach_distributed_inbatch_negatives` arg (default True),  so distributed in-batch negatives can use the previous efficient detached behavior by default, or preserve remote passage gradients when set to false.

- Renamed ColBERT-style scoring to multi_vector/MaxSim, kept colbert as a backward-compatible alias, and made distributed MaxSim scoring more memory efficient by scoring one passage slot at a time.

**Pre checks**:


nemo_automodel/recipes/retrieval/train_bi_encoder.py
```diff
@@ -348,22 +392,38 @@ def _forward_backward_step(self, idx, batch, *, loss_buffer, num_batches, is_tra
...
                 local_bs = q_reps.shape[0]
                 dist_initialized = torch.distributed.is_available() and torch.distributed.is_initialized()
                 rank = torch.distributed.get_rank() if dist_initialized else 0
                 world_size = torch.distributed.get_world_size() if dist_initialized else 1
-                all_p = dist_gather_tensor(p_reps)
-                expected_p = world_size * local_bs * n_passages
-                assert all_p.shape[0] == expected_p, f"Gathered passage count {all_p.shape[0]} != expected {expected_p}"
-                scores = torch.mm(q_reps, all_p.t())
-                labels = (torch.arange(local_bs, device=q_reps.device) + rank * local_bs) * n_passages
+                if getattr(model, "pooling", None) == "colbert":
```
oliverholworthy: wondering if `pooling: maxsim` be clearer than pooling: colbert here?

nemo_automodel/recipes/retrieval/train_bi_encoder.py
```diff
@@ -465,7 +533,15 @@ def _run_validation_epoch(self, val_dataloader):
                     q_reps = model(query)
                     p_reps = model(passage)
 
-                    scores, labels = contrastive_scores_and_labels(q_reps, p_reps, self.val_n_passages)
+                    if getattr(model, "pooling", None) == "colbert":
+                        scores, labels = colbert_scores_and_labels(
```
oliverholworthy: if we rename pooling to maxsim, this function and the distributed variant would make sense to rename accordingly too. (e.g. `maxsim_scores_and_labels`

nemo_automodel/components/models/common/inbatch_neg_utils.py
```diff
@@ -45,6 +45,34 @@ def dist_gather_tensor(t: Optional[torch.Tensor]) -> Optional[torch.Tensor]:
     return torch.cat(gathered, dim=0)
 
 
+def dist_gather_tensor_with_dim1_padding(
```
oliverholworthy, line 61: What do you think about updating this to avoid materializing the full [local_B, global_B * n_passages, q_len, p_len]. One option would be to score one passage slot at a time, with [local_B, global_B, q_len, p_len] temporaries, then stack back to the same query-major [local_B, global_B * n_passages] output. We could also consider a mini batch size option to chunk local queries further if we want another lever for larger batches. This would keep memory lower and enable longer sequences and larger batches to test with

nemo_automodel/recipes/retrieval/train_bi_encoder.py
```diff
@@ -348,22 +392,38 @@ def _forward_backward_step(self, idx, batch, *, loss_buffer, num_batches, is_tra
...
                 dist_initialized = torch.distributed.is_available() and torch.distributed.is_initialized()
                 rank = torch.distributed.get_rank() if dist_initialized else 0
                 world_size = torch.distributed.get_world_size() if dist_initialized else 1
-                all_p = dist_gather_tensor(p_reps)
-                expected_p = world_size * local_bs * n_passages
-                assert all_p.shape[0] == expected_p, f"Gathered passage count {all_p.shape[0]} != expected {expected_p}"
-                scores = torch.mm(q_reps, all_p.t())
-                labels = (torch.arange(local_bs, device=q_reps.device) + rank * local_bs) * n_passages
+                if getattr(model, "pooling", None) == "colbert":
+                    all_p = dist_gather_tensor_with_dim1_padding(p_reps)
```
oliverholworthy: I'm wondering if remote passages should receive gradients when they are used as distributed in-batch negatives? It looks like the current gather uses remote passages in the softmax denominator, but only keeps an autograd edge for the local rank’s passage slice? 

If that's true, it would mean distributed training may not be equivalent to a single-process global batch: local queries learn from remote passages, but remote passage embeddings do not learn from local-query losses. If that stop-gradient behavior is intentional, maybe we should document it explicitly; otherwise we might consider updating.

nemo_automodel/recipes/retrieval/train_bi_encoder.py
rnyak: @oliverholworthy thanks for the comment, I think may be even "multivector" is more informative for the pooling technique, instead of "colbert"?

nemo_automodel/recipes/retrieval/train_bi_encoder.py
rnyak: I think the current way is more efficient/scalable, but I can modify this to be equivalent to a single-process global batch. but it'd make the training slower, especially for the colbert training with distributed in batch neg = True

nemo_automodel/recipes/retrieval/train_bi_encoder.py
oliverholworthy: yes that makes more sense for a pooling value i was thinking of the scoring when i suggested maxsim because that appears to be more along the lines of what this condition is checking.

oliverholworthy (dismissed):

akoumpa (approved):

PR #2285, opened 2026-05-21
feat(speculative): add checkpoint resume for EAGLE recipes

khazic:
## What

Adds a proper save/load checkpoint cycle to the EAGLE-1, EAGLE-2, and EAGLE-3
training recipes. The previous ``_save_checkpoint`` only persisted the draft
model state dict plus a small meta blob, and there was no load path -- a
crashed run could not resume and discarded all optimizer / scheduler / RNG
progress.

## Changelog

- ``train_eagle1.py`` / ``train_eagle3.py``: build a ``Checkpointer`` +
  ``CheckpointingConfig`` in ``setup()`` and call
  ``load_checkpoint(self.cfg.get("checkpoint.restore_from"))`` before the
  training loop. Resolves ``"LATEST"``, named subdirs, and explicit paths
  with config-compat validation.
- Override ``save_checkpoint`` (not delegated to
  ``BaseRecipe.save_checkpoint``) because EAGLE holds multiple ``nn.Module``
  attributes -- ``target_model``, ``target_wrapper``, and the DDP-wrapped
  ``trainer_module`` -- and the inherited auto-tracker would pick the
  alphabetically-last module as the "main" model. Only ``draft_model`` is
  persisted; the frozen target is reloaded from HF on each run.
- EAGLE-3 ``_save_extra_state`` / ``_load_extra_state`` carry the
  ``selected_token_ids`` / ``selected_token_mask`` vocab mapping tensors
  alongside ``global_step`` and ``epoch``, with backward-compatible reading
  of the legacy ``eagle3_meta.pt`` filename.
- Training loop honours ``_resume_epoch`` and skips already-finished epochs.
- ``best_val_metric`` lookup uses ``.get()`` and the train loop passes
  ``best_metric_key="val_loss"`` so EAGLE's multi-metric eval dict does not
  KeyError on the default ``"default"`` key.
- Async + ``LATEST`` / ``LOWEST_VAL`` symlink machinery, ``losses.json``,
  and ``save_config`` snapshot all match ``BaseRecipe`` behavior.
- Example YAMLs (eagle1 / eagle2 / eagle3) gain a ``checkpoint:`` block with
  ``restore_from`` documentation.
- New ``test_eagle_checkpoint_resume.py`` (9 cases) covers meta round-trip,
  vocab-mapping persistence, legacy filename fallback, ``LATEST`` /
  auto-detect / missing-dir resolution, disabled-checkpoint no-op, on-disk
  artifact layout, and the resume-skips-completed-epochs guard. Existing
  flush / coverage tests are updated to the new save API.

## Test plan

- [x] ``ruff check`` + ``ruff format`` clean on all touched files.
- [x] ``pytest tests/unit_tests/recipes/llm/test_eagle_checkpoint_resume.py``
      -- 9 passed.
- [x] ``pytest tests/unit_tests/recipes/llm/test_train_eagle3_flush.py
      tests/unit_tests/recipes/llm/test_train_eagle3_grad_accum.py
      tests/unit_tests/speculative/`` -- 91 passed, 2 skipped, 0 failed.
- [ ] CI green (will monitor with ``gh pr checks`` after PR creation).
- [ ] End-to-end run: train EAGLE-1 N epochs, kill, restart with
      ``checkpoint.restore_from: LATEST``, verify optimizer / scheduler /
      ``global_step`` continue cleanly. (To be exercised on a real GPU.)

### DIFF
--- a/nemo_automodel/recipes/llm/train_eagle1.py
+++ b/nemo_automodel/recipes/llm/train_eagle1.py
@@ -325,9 +325,6 @@ class TrainEagle1Recipe(BaseRecipe):
             os.path.join(path, "eagle_meta.pt"),
         )
 
-    def _update_latest_symlink(self, path: str) -> None:
-        self._update_checkpoint_symlink("LATEST", path)
-
     def load_checkpoint(self, restore_from: str | None = None) -> None:
         """Resolve and restore a checkpoint produced by ``save_checkpoint``.
 
--- a/nemo_automodel/recipes/llm/train_eagle1.py
+++ b/nemo_automodel/recipes/llm/train_eagle1.py
@@ -387,6 +387,9 @@ class TrainEagle1Recipe(BaseRecipe):
     def _load_extra_state(self, ckpt_dir: str) -> None:
         """Restore EAGLE-recipe-specific scalars. Subclasses extend this."""
         meta_path = os.path.join(ckpt_dir, "eagle_meta.pt")
+        if not os.path.exists(meta_path):
+            legacy = os.path.join(ckpt_dir, "eagle1_meta.pt")
+            meta_path = legacy if os.path.exists(legacy) else meta_path
         if os.path.exists(meta_path):
             meta = torch.load(meta_path, weights_only=False, map_location="cpu")
             self.runtime.global_step = int(meta.get("global_step", 0))
--- a/nemo_automodel/recipes/llm/train_eagle1.py
+++ b/nemo_automodel/recipes/llm/train_eagle1.py
@@ -16,18 +16,26 @@
 
 from __future__ import annotations
 
+import json
 import logging
 import math
+import os
 import pathlib
 from types import SimpleNamespace
 
 import torch
 import torch.distributed as dist
+from huggingface_hub import constants as hf_constants
 from torch.nn.parallel import DistributedDataParallel
 from transformers import AutoConfig, LlamaConfig
 
 from nemo_automodel._transformers import NeMoAutoModelForCausalLM
 from nemo_automodel._transformers.auto_tokenizer import NeMoAutoTokenizer
+from nemo_automodel.components.checkpoint.checkpointing import (
+    Checkpointer,
+    CheckpointingConfig,
+    save_config,
+)
 from nemo_automodel.components.config._arg_parser import parse_args_and_load_config
 from nemo_automodel.components.datasets.llm.eagle3 import build_eagle3_dataloader
 from nemo_automodel.components.distributed.init_utils import initialize_distributed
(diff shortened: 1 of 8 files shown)


nemo_automodel/recipes/llm/train_eagle3.py
```diff
@@ -244,22 +294,184 @@ def _module(self):
...
+                "epoch": int(epoch),
                 "selected_token_ids": self._module().selected_token_ids.cpu(),
                 "selected_token_mask": self._module().selected_token_mask.cpu(),
-                "global_step": self.runtime.global_step,
             },
-            save_dir / "eagle3_meta.pt",
+            os.path.join(path, "eagle_meta.pt"),
         )
 
+    def _update_latest_symlink(self, path: str) -> None:
```
HuiyingLi: Can we reuse what's already existing in base_recipe? 
https://github.com/NVIDIA-NeMo/Automodel/blob/780da30926325bee83eef20194f1a09e2263932a/nemo_automodel/recipes/base_recipe.py#L431

nemo_automodel/recipes/llm/train_eagle1.py
```diff
@@ -182,14 +232,168 @@ def _module(self):
...
+                dist.barrier()
+
+    def _save_extra_state(self, path: str, epoch: int) -> None:
+        """Persist EAGLE-recipe-specific scalars. Subclasses extend this."""
+        torch.save(
+            {"global_step": self.runtime.global_step, "epoch": int(epoch)},
+            os.path.join(path, "eagle_meta.pt"),
+        )
+
+    def _update_latest_symlink(self, path: str) -> None:
```
HuiyingLi: same here

nemo_automodel/recipes/llm/train_eagle3.py
khazic: Good catch — both overrides were identical to `BaseRecipe._update_latest_symlink` and added nothing. Removed them in the latest commit (21c26654) so the subclasses simply inherit the base-class implementation.

nemo_automodel/recipes/llm/train_eagle1.py
khazic: Same fix applied here — removed in the same commit (21c26654).

nemo_automodel/recipes/llm/train_eagle1.py
```diff
@@ -182,14 +232,168 @@ def _module(self):
...
+
+        self._load_extra_state(ckpt_dir)
+
+    def _load_extra_state(self, ckpt_dir: str) -> None:
+        """Restore EAGLE-recipe-specific scalars. Subclasses extend this."""
+        meta_path = os.path.join(ckpt_dir, "eagle_meta.pt")
+        if os.path.exists(meta_path):
+            meta = torch.load(meta_path, weights_only=False, map_location="cpu")
+            self.runtime.global_step = int(meta.get("global_step", 0))
+            self._resume_epoch = int(meta.get("epoch", 0))
```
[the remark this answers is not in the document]
khazic: Good catch. Added the legacy `eagle1_meta.pt` fallback to `_load_extra_state` in commit a5b95a12, mirroring the existing EAGLE-3 probe. Also added a unit test (`test_eagle1_load_extra_state_accepts_legacy_filename`) to cover this path — 34 tests pass locally.

HuiyingLi (approved):
LGTM Thank you!

PR #2307, opened 2026-05-23
feat(speculative): add offline dataset regeneration with EAGLE target model

khazic:
# What does this PR do ?

Adds an offline dataset-regeneration CLI for EAGLE training: given a
chat-style dataset, replay each sample's prompt context through the EAGLE
target model (served via SGLang's OpenAI-compatible endpoint) and write
out a new dataset whose assistant turns are produced by the target
itself. This closes a real gap — most public chat datasets were generated
by other models, so their assistant tokens are off-distribution for an
EAGLE drafter trained against a different target.

The output parquet directory has a ``messages`` column that
``ChatDataset`` (used by ``build_eagle3_dataloader``) consumes directly,
so it plugs into the existing EAGLE-3 recipe with no further changes.

# Changelog

- Add ``nemo_automodel/components/speculative/regenerate.py``:
  - Streaming dataset iteration with shard-by-shard parquet writes
    (atomic ``.tmp`` + ``os.replace``) — works on corpora that do not
    fit in memory.
  - ``--resume`` skips shards already on disk, guarded by a
    ``manifest.json`` that pins input dataset / model / sampling
    params / shard size so resumes cannot silently mix incompatible
    runs.
  - Non-resume runs refuse to clobber a non-empty output directory.
  - Bounded async concurrency via ``asyncio.Semaphore``; retries with
    exponential backoff on 5xx / 429 / transport errors.
  - Optional deps (``aiohttp``, ``pyarrow``) guarded via
    ``safe_import`` with install hints.
- Add ``tests/unit_tests/speculative/test_regenerate.py`` — 20 unit
  tests covering prompt extraction edge cases, manifest validation
  (mismatch / missing / clobber-guard / operational-field exclusion),
  shard discovery, atomic write, bounded concurrency, argument
  validation, and an end-to-end resume scenario that simulates a crash
  after one shard and verifies the second pass only regenerates the
  missing shard while leaving the first byte-identical.
- Add ``examples/speculative/regenerate_with_target.md`` documenting
  the two-step flow (spin up SGLang → regenerate) and the tuning knobs.

# Before your PR is \"Ready for review\"

**Pre checks**:

# Additional Information

- Companion to the SGLang serve helper in
  ``nemo_automodel/components/speculative/serve_sglang.py`` — that
  helper lets you serve a trained drafter; this PR lets you build the
  drafter's training data against the corresponding target.
- Linting: ``ruff format`` and ``ruff check`` pass.
- Tests: ``pytest tests/unit_tests/speculative/test_regenerate.py`` — 20
  passed locally.


HuiyingLi:
Looks good aside from the claude comments. Thank you @khazic

HuiyingLi (approved):
LGTM thank you

PR #2310, opened 2026-05-25
fix(speculative): make EAGLE recipes complete a checkpoint save end-to-end

khazic:
# What does this PR do?

Make the EAGLE-1 / EAGLE-3 training recipes actually reach and complete the
end-of-epoch consolidated safetensors checkpoint save. Two related bugs were
blocking it.

## Bug 1 — `TypeError: 'NoneType' object is not iterable` at checkpoint save

On an 8-rank EAGLE-3 / EAGLE-1 run with the default
`--checkpoint.model_save_format safetensors --checkpoint.save_consolidated True`,
training completes ~16 optimizer steps and then crashes at the first
checkpoint save:

```
File ".../components/checkpoint/checkpointing.py", line 924, in _maybe_build_consolidated_index
    keys_to_remove = list(set(fqn_to_file_index_mapping.keys()) - set(pre_shard_hf_state_dict_keys))
TypeError: 'NoneType' object is not iterable
```

Root cause:

- `_maybe_build_consolidated_index` reads
  `model._pre_shard_hf_state_dict_keys` (set by `apply_model_infrastructure`)
  with fallback to `CheckpointingConfig.model_state_dict_keys`.
- EAGLE recipes build the draft model directly (`LlamaEagle*DraftModel(...).to(device)`)
  and bypass `apply_model_infrastructure`, so the attribute is never set.
- The recipes also never passed `model_state_dict_keys` to
  `CheckpointingConfig`, so the fallback is `None` too.
- `set(None)` raises.

Fix: in `_build_checkpointer`, capture
`list(self.draft_model.state_dict().keys())` and pass it as
`model_state_dict_keys` on `CheckpointingConfig`. Mirrors what
`recipes/diffusion/train.py:561` already does.

## Bug 2 — EAGLE-3 `main()` raised before `-c` could be parsed

`train_eagle3.py:main(config_path=None)` rejected `config_path is None`
before calling `parse_args_and_load_config`, but that helper is what reads
`-c <config.yaml>` from `sys.argv`. As a result
`torchrun -m nemo_automodel.recipes.llm.train_eagle3 -c <config>` was
unusable; only the direct-python-import call path worked.

Fix: remove the premature `ValueError` so argv parsing can resolve
`config_path` itself.

# Changelog

- `recipes/llm/train_eagle1.py`: pass `model_state_dict_keys` from the draft
  model to `CheckpointingConfig` in `_build_checkpointer`. EAGLE-2 inherits
  via `TrainEagle1Recipe` so it is covered automatically.
- `recipes/llm/train_eagle3.py`: same fix; additionally remove the premature
  `ValueError` in `main()` so `torchrun -m ... -c <yaml>` works.

# Verification

End-to-end run on 8x H100 with the user-reported repro config (target
`Llama-3.1-8B-Instruct`, PerfectBlend-Regenerated dataset, sliced to 200
samples, 1 epoch):

```
2026-05-25 15:11:44 | INFO | __main__ | Epoch 0 done: total_batches_seen=25 global_step=25
2026-05-25 15:11:48 | INFO | ...consolidate_hf_safetensors | Rank 0/8: Consolidating safetensors files ...
2026-05-25 15:11:50 | INFO | ...consolidate_hf_safetensors | Total time taken: 2.41 secs.
2026-05-25 15:11:52 | INFO | __main__ | Saved checkpoint to outputs/verify_eagle3_fix/checkpoints/epoch_1_step_25
2026-05-25 15:11:52 | INFO | __main__ | Training complete: global_step=25
```

Checks against the verification log:

- `grep -nE "TypeError|NoneType.* not iterable" "$LOG"` -> empty
  (`PASS: no TypeError`).
- `grep -n "Saved checkpoint to" "$LOG"` -> the line above.
- Consolidated output directory contents:

```
$ ls -la outputs/verify_eagle3_fix/checkpoints/epoch_1_step_25/model/consolidated/
config.json
chat_template.jinja
model-00001-of-00004.safetensors  (1.5 GB)
model-00004-of-00004.safetensors  (363 MB)
model.safetensors.index.json
special_tokens_map.json
tokenizer.json
tokenizer_config.json
```

Before this fix, the same command crashed at the first save and the output
directory contained only partial files (no `model.safetensors*` written).

## Known cosmetic follow-up

The two shard files are numbered `00001-of-00004` and `00004-of-00004`
(no `00002` / `00003`). This is because the consolidated-index path still
reads the **target** model's safetensors index (target has 4 shards) and
maps the draft's keys onto shards 1 and 4 (the target's `max(shard)` is used
as `default_index` for any key the target index does not cover). The
`model.safetensors.index.json` correctly points every key to an existing
file, so this loads fine via standard HF safetensors loaders — just a
slightly ugly shard layout. Worth a follow-up to either set
`model_repo_id=None` for EAGLE or build the consolidated index from the
draft's keys directly, but out of scope for this fix.

**Pre checks**:

- [x] Contributor guidelines followed
- [ ] Did you write any new necessary tests? No — bug repro requires a real
  multi-GPU run with consolidated safetensors save; not feasible as a unit
  test. End-to-end repro + verification log attached above.
- [ ] Did you add or update any necessary documentation? No — behavior of
  the public API is unchanged; only a crash on a previously-broken path is
  removed.

### DIFF
--- a/nemo_automodel/recipes/llm/train_eagle1.py
+++ b/nemo_automodel/recipes/llm/train_eagle1.py
@@ -201,6 +201,12 @@ class TrainEagle1Recipe(BaseRecipe):
         """Build the checkpointer using the same plumbing as the standard recipes."""
         ckpt_cfg = self.cfg.get("checkpoint", None)
         default_dir = str(self.output_dir / "checkpoints")
+        # EAGLE recipes construct the draft model directly and bypass
+        # `apply_model_infrastructure`, which is where `_pre_shard_hf_state_dict_keys`
+        # would normally be attached. Capture the pre-shard keys here so the
+        # consolidated-safetensors path in `_maybe_build_consolidated_index`
+        # has something to diff against instead of `None`.
+        draft_state_dict_keys = list(self.draft_model.state_dict().keys())
         ckpt_kwargs = dict(
             enabled=True,
             checkpoint_dir=default_dir,
@@ -209,11 +215,14 @@ class TrainEagle1Recipe(BaseRecipe):
             model_cache_dir=hf_constants.HF_HUB_CACHE,
             save_consolidated=True,
             is_peft=False,
+            model_state_dict_keys=draft_state_dict_keys,
         )
         if ckpt_cfg is not None:
             user_cfg = ckpt_cfg.to_dict() if hasattr(ckpt_cfg, "to_dict") else dict(ckpt_cfg)
             user_cfg.pop("restore_from", None)
             ckpt_kwargs.update(user_cfg)
+        if ckpt_kwargs.get("model_state_dict_keys") is None:
+            ckpt_kwargs["model_state_dict_keys"] = draft_state_dict_keys
 
         self.checkpoint_config = CheckpointingConfig(**ckpt_kwargs)
         dp_rank = dist.get_rank() if dist.is_initialized() else 0
--- a/nemo_automodel/recipes/llm/train_eagle3.py
+++ b/nemo_automodel/recipes/llm/train_eagle3.py
@@ -681,8 +681,6 @@ class TrainEagle3Recipe(BaseRecipe):
 
 def main(config_path=None):
     """Main entry point for the EAGLE-3 recipe."""
-    if config_path is None:
-        raise ValueError("config_path is required for TrainEagle3Recipe")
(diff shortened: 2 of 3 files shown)


HuiyingLi (approved):
Thank you!

PR #2312, opened 2026-05-25
feat(speculative): add Phi-3 support for EAGLE-1/2/3 + correctness fixes

khazic:
# What does this PR do?

Add Phi-3 (``Phi3ForCausalLM``) as a supported target for EAGLE-1 / EAGLE-2 /
EAGLE-3 training, alongside the existing Llama support. Also lands several
correctness / usability bugs surfaced while bringing Phi-3 online -- many
of which also help any EAGLE run, not just Phi-3.

## Changelog

**Phi-3 target support:**

- Add ``components/speculative/eagle/registry.py``: a small dispatch table
  (``DraftSpec`` + ``EAGLE{1,3}_DRAFT_REGISTRY``) keyed by HF
  ``architectures`` string. Current entries: ``LlamaForCausalLM``,
  ``Phi3ForCausalLM``.
- ``recipes/llm/train_eagle{1,3}.py``: replace the inline
  ``{LlamaForCausalLM}``-only allowlist with a registry lookup; instantiate
  the draft config via ``type(target_config).from_dict(...)`` so the
  draft uses the target's concrete config class (LlamaConfig / Phi3Config)
  and the architecture-specific defaults flow through.
- ``components/speculative/eagle/draft_llama.py`` and
  ``draft_llama_v12.py``: relax ``config_class`` from ``LlamaConfig`` to
  ``PretrainedConfig``; read ``attention_bias`` / ``mlp_bias`` via
  ``getattr(config, '<field>', False)`` so families that don't define those
  fields (Phi-3) don't crash with ``AttributeError``.
- ``components/speculative/eagle/target.py``: add ``model.transformer.h``
  layer-tree fallback to ``_get_transformer_layers``.
- Add example YAMLs: ``phi_eagle{1,2,3}_perfectblend.yaml``.
- Rename ``llama_eagle{1,2}_mvp.yaml`` -> ``llama_eagle{1,2}_perfectblend.yaml``
  (filename only; content unchanged).

**Correctness / usability fixes:**

- ``recipes/llm/train_eagle{1,3}.py``: pass ``model_state_dict_keys`` to
  ``CheckpointingConfig`` so the consolidated safetensors save at epoch
  end no longer crashes with
  ``TypeError: 'NoneType' object is not iterable``.
- ``recipes/llm/train_eagle{1,2,3}.py``: drop the premature
  ``raise ValueError('config_path is required')`` so
  ``torchrun -m ... -c <yaml>`` can have argv parsing read ``-c`` itself.
- ``recipes/llm/train_eagle{1,2}.py``: add the missing
  ``if __name__ == '__main__': main()`` guard. Without it,
  ``torchrun -m`` imports the module and exits silently with no training
  output.
- ``components/datasets/llm/formatting_utils.py``: in
  ``format_chat_template``, shift the assistant loss mask right by
  ``pad_len`` when ``tokenizer.padding_side == 'left'``. Otherwise
  ``_build_multiturn_assistant_mask`` computes spans from unpadded
  (left-aligned) lengths, the attention-mask zeroing step wipes them out,
  and ~80% of samples end up with an all-zero loss_mask. The change is
  triple-guarded and is a no-op for right-padding tokenizers (the
  majority -- Llama / Qwen / Mistral / Gemma / ...).

## Verification

End-to-end smoke test on 8 x H100:

- ``Llama-3.1-8B-Instruct`` + PerfectBlend, EAGLE-3: train loss decreases
  ``11.2 -> 7.2`` over 25 optimizer steps; consolidated safetensors
  checkpoint written.
- ``Phi-3-mini-4k-instruct``: surfaces the ``getattr`` fix (was crashing
  with ``AttributeError: 'Phi3Config' object has no attribute 'mlp_bias'``),
  the ``__main__`` guard fix (was silent-dying after grouped_gemm import),
  and the left-padding loss-mask fix (was getting all-zero loss for ~80%
  of samples). Each fix is in this PR.

**Pre checks**:

- [x] Contributor guidelines followed
- [ ] Did you write any new necessary tests? Not in this PR -- end-to-end
  repro requires multi-GPU + real HF target; smoke-test evidence above.
- [x] Example configs added under ``examples/speculative/eagle{1,2,3}/``.


nemo_automodel/components/speculative/eagle/target.py
```diff
@@ -96,10 +96,16 @@ def _validate_aux_layer_ids(self, aux_layer_ids: Sequence[int]) -> list[int]:
         return aux_layer_ids
 
     def _get_transformer_layers(self) -> Iterable[nn.Module]:
```
HuiyingLi, line 98: Would reusing _extract_model_layers be feasible

khazic: I looked into reusing `_extract_model_layers` from `parallelizer.py` but decided against it for a few reasons:

1. **Different return contract**: `_extract_model_layers` returns a flattened `List[nn.Module]` of individual layers (for FSDP wrapping), while `_get_transformer_layers` needs the `nn.ModuleList` container itself (to hook `register_forward_hook` on specific layer indices for aux hidden state capture).

2. **Cross-component dependency**: importing from `components.distributed.parallelizer` into `components.speculative.eagle.target` would introduce a coupling between the speculative and distributed packages.

3. **Scope difference**: `_extract_model_layers` has a large class-keyed dispatch table (VLM, diffusion, etc.) — heavy machinery for what is a 3-line hasattr chain here.

Happy to reconsider if you feel the dedup is worth the coupling.

HuiyingLi:
@khazic could you please share a running log for reference? Thank you~

HuiyingLi (approved):

PR #2323, opened 2026-05-27
fix(datasets): serialize dict tool arguments to JSON string in xlam

khazic:
## Summary

OpenAI tool calling format requires `function.arguments` to be a JSON
string, but `xlam._convert_tool_calls` passed `call["arguments"]`
through verbatim. Upstream `Salesforce/xlam-function-calling-60k`
happens to store arguments as JSON strings so this worked in practice,
but derivative xLAM-style datasets (Chinese variants,
instruction-rewritten splits) often emit a dict and were either
rendered as Python repr by lenient chat templates or rejected by
stricter ones.

Match the sibling `agent_chat` adapter: when `arguments` is not already
a string, call `json.dumps(arguments, ensure_ascii=False)`. The
`ensure_ascii=False` flag keeps CJK characters readable in the rendered
chat instead of escaping them to `\uXXXX`.

## Files

| File | Change |
|---|---|
| `nemo_automodel/components/datasets/llm/xlam.py` | +6 / -1 |
| `tests/unit_tests/datasets/llm/test_xlam.py` | +22 / -0, two new cases |

## Test plan

- [x] `pytest tests/unit_tests/datasets/llm/test_xlam.py` — 7/7 pass
  (5 existing + 2 new: dict args serialize, unicode preserved)
- [x] No behavior change for the upstream xLAM dataset path (string args
      are passed through unchanged, covered by the existing
      `test_convert_tool_calls_preserves_arguments_and_ids` test)

### DIFF
--- a/nemo_automodel/components/datasets/llm/xlam.py
+++ b/nemo_automodel/components/datasets/llm/xlam.py
@@ -102,6 +102,12 @@ def _convert_tool_calls(raw_calls: List[Dict], example_id: Optional[int] = None)
             logger.warning("Skipping call without name: %s", call)
             continue
         arguments = call.get("arguments", "")
+        if not isinstance(arguments, str):
+            # OpenAI tool calling format requires `function.arguments` to be a
+            # JSON-encoded string. Upstream xLAM uses string arguments, but
+            # derivative datasets may store them as dicts; serialize so the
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
Thank you so much for the fix

PR #2347, opened 2026-05-29
feat(datasets): add train_on_last_turn_only to agent chat SFT dataset

khazic:
## What

Adds an opt-in `train_on_last_turn_only` (mask_history) mode to `make_agent_chat_dataset`. When enabled, only the **final assistant turn** of each multi-turn agent dialogue is supervised; all earlier assistant turns are excluded from the loss. Default is `False`, which preserves the existing all-turn supervision behavior.

This mirrors the `mask_history` / `train_only_last_turn` option found in LLaMA-Factory, ms-swift, SpecForge, and Megatron-Bridge, which is a common requirement for multi-turn agent SFT.

## How

- New pure helper `_mask_labels_to_last_turn(labels)`: keeps only the last contiguous supervised run of `labels` (the final assistant message, since the chat template renders each assistant message — including tool_calls — as one contiguous span) and masks every earlier supervised token in place.
- `_format_example` / `make_agent_chat_dataset` gain a `train_on_last_turn_only: bool = False` flag, threaded through.
- The change is contained to the agent-chat dataset; it post-processes the labels returned by `format_chat_template` and does not touch the shared chat-template utilities.

## Changelog

- `make_agent_chat_dataset(..., train_on_last_turn_only=True)` supervises only the last assistant turn of each dialogue.

## Usage

```yaml
dataset:
  _target_: nemo_automodel.components.datasets.llm.agent_chat.make_agent_chat_dataset
  dataset_name: llamafactory/glaive_toolcall_en
  train_on_last_turn_only: true
```

## Pre-checks

- `ruff format` + `ruff check`: clean
- `pytest tests/unit_tests/datasets/llm/test_agent_chat.py`: 22 passed (4 new tests covering the helper and the flag)
- Sign-off: included (DCO)

### DIFF
--- a/nemo_automodel/components/datasets/llm/agent_chat.py
+++ b/nemo_automodel/components/datasets/llm/agent_chat.py
@@ -187,6 +187,38 @@ def _convert_messages(
     return out
 
 
+def _mask_labels_to_last_turn(labels: List[int], ignore_index: int = -100) -> List[int]:
+    """Restrict the loss to the final assistant turn (``mask_history``).
+
+    ``labels`` come from :func:`format_chat_template` with every assistant
+    turn supervised; non-assistant tokens are already ``ignore_index``.
+    Because the chat template renders each assistant message as a single
+    contiguous span, supervised tokens form one maximal run per assistant
+    turn separated by ``ignore_index`` runs. This keeps only the last such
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
LGTM thank you so much.

PR #2348, opened 2026-05-29
feat(datasets): preserve and optionally mask reasoning_content in agent SFT

khazic:
## What

Adds reasoning/thinking support to the multi-turn agent SFT dataset (`agent_chat.py`):

1. **Preserve reasoning traces.** `_convert_messages` and `_sharegpt_to_chatml` now carry an assistant turn's `reasoning_content` field through conversion (previously only `content` and `tool_calls` survived, so any thinking trace was silently dropped before the chat template saw it). Reasoning is preserved across the tool-call merge, so an assistant "think then call tool" turn keeps its reasoning.
2. **Optional reasoning loss masking.** New `mask_reasoning_content` flag on `make_agent_chat_dataset` is forwarded to `format_chat_template` (which already supports building a reasoning-token mask). When True, reasoning tokens are rendered into the prompt but excluded from the loss.

Default is `False`, which preserves the current behavior (reasoning trained as ordinary assistant content). The change is contained to the agent-chat dataset and does not modify the shared chat-template utilities.

## Why

Reasoning-interleaved agent data (`<think>...</think>` followed by a tool call) is common for reasoning models. Without this, such traces are dropped during conversion. This mirrors the reasoning/thinking handling in SpecForge (ThinkingParser), LLaMA-Factory, ms-swift, and verl (`enable_thinking`).

## Changelog

- `make_agent_chat_dataset(..., mask_reasoning_content=True)` excludes assistant reasoning_content tokens from the loss.
- Assistant `reasoning_content` is preserved through agent message conversion (chatml and ShareGPT schemas).

## Pre-checks

- `ruff format` + `ruff check`: clean
- `pytest tests/unit_tests/datasets/llm/test_agent_chat.py`: 22 passed (4 new tests covering reasoning preservation, the tool-call merge case, ShareGPT pass-through, and flag forwarding)
- Sign-off: included (DCO)

### DIFF
--- a/nemo_automodel/components/datasets/llm/agent_chat.py
+++ b/nemo_automodel/components/datasets/llm/agent_chat.py
@@ -95,7 +95,12 @@ def _sharegpt_to_chatml(conversations: List[Dict[str, Any]]) -> List[Dict[str, A
         src_role = turn.get("from")
         if src_role not in _SHAREGPT_ROLE_MAP:
             raise ValueError(f"Unsupported sharegpt role: {src_role!r}")
-        out.append({"role": _SHAREGPT_ROLE_MAP[src_role], "content": turn.get("value", "")})
+        chatml_turn = {"role": _SHAREGPT_ROLE_MAP[src_role], "content": turn.get("value", "")}
+        # Carry an explicit reasoning/thinking field through if the export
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
Thank you so much!

PR #2350, opened 2026-05-29
feat(model_init): add FP8 pre-flight check for the force_hf path

### LINKED ISSUE #2114: FineGrainedFP8Config(dequantize=true) materializes full BF16 model per rank before sharding
## What

The standard FP8-checkpoint load path

\`\`\`yaml
quantization_config:
  _target_: transformers.FineGrainedFP8Config
  dequantize: true
\`\`\`

dequantizes the **entire** FP8 checkpoint into BF16 **on every rank** before any TP / FSDP / PP sharding kicks in. The full BF16 model is the per-rank peak.

This is fine on small models (24B BF16 ≈ 48 GB, fits 80 GB H100) but OOMs at scale.

## Why it matters

| Model | Full BF16 footprint | Fits 80 GB H100? | Fits 141 GB H200? |
|---|---|---|---|
| Devstral-Small-2-24B (FP8) | ~48 GB | yes | yes |
| Devstral-2-123B (FP8) | ~246 GB | no | no |
| Mistral-Medium-3.5-128B (FP8) | ~256 GB | no | no |

For 123B / 128B+ FP8 checkpoints there's *no* per-rank configuration — even TP=8 PP=8 across 64 GPUs — that lets the HF dequant path run, because the OOM happens *before* the model is sharded. Cranking up node count does nothing.

## Today's solution

\`nemo_automodel/components/models/mistral3_vlm/state_dict_adapter.py\` (see \`Mistral3FP8StateDictAdapter.for_vlm_full()\`) hooks the DCP load so each rank only reads its own shard from disk: pull the FP8 weight + per-tensor \`weight_scale_inv\` for the local shard, dequantize to BF16 in place, emit the local DTensor. Per-rank peak ≈ shard size, not full model.

This works but is a one-off — it lives inside the \`mistral3_vlm\` package and only kicks in when the resolver hook claims the config (i.e. for the dawn-ridge 128B specifically). Any other FP8-native dense / MoE checkpoint at this scale would need to ship its own adapter.

Possible shapes:
1. Generalize the \`Mistral3FP8StateDictAdapter\` factory pattern into a \`StreamingFP8StateDictAdapter\` keyed off \`config.quantization_config.quant_method == \"fp8\"\`, registered for any model whose checkpoint declares per-tensor or per-block FP8.
2. Push the streaming dequant upstream into \`transformers/quantizers/quantizer_finegrained_fp8.py\` (the loader that processes \`FineGrainedFP8Config\`) so HF itself reads + dequants per shard rather than per full tensor.
3. Document the failure mode loudly: today the OOM happens deep in the HF loader with no log line saying \"this can't work at this scale, use a streaming adapter\". A pre-flight check in \`nemo_automodel/_transformers/model_init.py\` that estimates BF16 footprint vs available HBM and refuses to start would at least convert the silent OOM into a useful error.

## Repro

Any FP8 checkpoint > ~80 GB BF16 with \`FineGrainedFP8Config(dequantize=true)\`, e.g. \`mistralai/Devstral-2-123B-Instruct-2512\` on 8 × H100-80GB:

\`\`\`yaml
model:
  pretrained_model_name_or_path: mistralai/Devstral-2-123B-Instruct-2512
  force_hf: true
  quantization_config:
    _target_: transformers.FineGrainedFP8Config
    dequantize: true
\`\`\`

OOMs at HF dequant before any sharding.

stanley1208:
# What does this PR do ?

Adds a pre-flight check so FP8 checkpoints that would OOM during the `force_hf` dequantize load fail fast with a clear message, instead of dying deep inside the HF loader.

# Changelog

* Loading an FP8 HF checkpoint with `force_hf=True` and `FineGrainedFP8Config(dequantize=True)` makes transformers dequantize the whole model to bf16 on every rank inside `_from_pretrained_parent_class`, before any sharding. For models bigger than one GPU (Devstral 123B, Mistral Medium 128B) this just OOMs with no useful error.
* Added a check in the `force_hf` path that estimates the bf16 footprint and stops early with a clear message + the streaming workaround. Disable with `NEMO_AUTOMODEL_DISABLE_FP8_PREFLIGHT=1` if the estimate is ever wrong.
* Budgets against total GPU memory, not free (free is noisy across ranks, and the materialization is per-rank regardless of TP degree).
* Reads the exact param count from the safetensors headers when the checkpoint is local; falls back to a config formula that now accounts for GQA.
* Same guard runs on the non-force_hf fallback too, since it hits the same loader.
* Doesn't make these models loadable on its own, that needs the per-shard streaming path (option 1 in the issue). This just makes the failure obvious.
* Tests: `pytest tests/unit_tests/_transformers/test_fp8_preflight_check.py`, 32 passing (CUDA mocked).

**Pre checks**:

# Additional Information

* Related to NVIDIA-NeMo/Automodel#2114


stanley1208:
@akoumpa CI passed, ready for review.

akoumpa:
Hi @stanley1208 , thanks for working on this ! @HuiyingLi can you review / validate? 🙇

HuiyingLi:
@stanley1208 I will test it later this week. Thank you!

stanley1208:
Pushed a follow-up that hardens the memory check. CI flipped back to "approval required" after the force-push. 
@HuiyingLi this is the version to test whenever you get a chance, thanks!

akoumpa:
Hi @stanley1208 sorry for the delay, can i ask you to take a look at the conflict? then i can help with the ci. thank you

stanley1208:
Hey @akoumpa the conflicts have been resolved as well, thank you for your time again

nemo_automodel/_transformers/model_init.py
```diff
@@ -898,6 +1147,14 @@ def __init_model(
 
     # 1. if force_hf is True, use HF model class wrapped with mixin
     if force_hf:
+        # Refuse early if HF's loader would dequantize an FP8 checkpoint to a
+        # full BF16 model that won't fit in HBM (issue #2114). Both the
+        # user-supplied quant config and the one baked into the HF config
+        # (e.g. set by _maybe_dequantize_fp8_for_peft) can trigger this path.
+        _check_fp8_dequantize_will_fit(hf_config, quantization_config, pretrained_model_name_or_path)
```
yuhezhang-ai: [P1] These preflight calls run before is_pretrained_init gates the checkpoint loader. As a result, a config-only/from_config call carrying FP8 dequantization metadata can raise even though no checkpoint will be materialized; I reproduced this at the current head. Please run the guard only when is_pretrained_init is true here and in the non-force-HF fallback, and add config-only regression coverage for both paths.

nemo_automodel/_transformers/model_init.py
```diff
@@ -470,6 +470,255 @@ def _setup_bnb_loading_kwargs(kwargs: dict) -> None:
...
+
+    gib = 1024**3
+    raise RuntimeError(
+        f"FP8 dequantize pre-flight failed for {pretrained_model_name_or_path!r}.\n"
+        f"  Estimated BF16 footprint: {bf16_bytes / gib:.1f} GB "
+        f"(~{n_params / 1e9:.1f}B parameters x 2 bytes).\n"
+        f"  Total CUDA memory: {total_bytes / gib:.1f} GB "
+        f"(safe budget at {int(_FP8_PREFLIGHT_FOOTPRINT_THRESHOLD * 100)}%: {budget_bytes / gib:.1f} GB; "
+        f"free right now: {free_bytes / gib:.1f} GB).\n"
+        f"  With force_hf=True and quant_method='fp8' + dequantize=True, transformers "
```
yuhezhang-ai: [P2] This error hard-codes force_hf=True and recommends disabling it, but the helper is now also invoked from the non-force-HF fallback. That path therefore reports a false cause and suggests a no-op workaround. Please pass the load-path context into the diagnostic or make the wording generic, and assert the real fallback error text in a test.

nemo_automodel/_transformers/model_init.py
```diff
@@ -470,6 +470,255 @@ def _setup_bnb_loading_kwargs(kwargs: dict) -> None:
...
+    budget_bytes = int(total_bytes * _FP8_PREFLIGHT_FOOTPRINT_THRESHOLD)
+    if bf16_bytes <= budget_bytes:
+        return None
+
+    gib = 1024**3
+    raise RuntimeError(
+        f"FP8 dequantize pre-flight failed for {pretrained_model_name_or_path!r}.\n"
+        f"  Estimated BF16 footprint: {bf16_bytes / gib:.1f} GB "
+        f"(~{n_params / 1e9:.1f}B parameters x 2 bytes).\n"
+        f"  Total CUDA memory: {total_bytes / gib:.1f} GB "
```
yuhezhang-ai: [P3] These values are divided by 1024**3, so the displayed unit is GiB rather than GB. Please label them GiB or divide by 1e9 so the diagnostic is numerically accurate.

yuhezhang-ai (commented):
Reviewed the current head. The fail-fast direction is useful, and the focused 32-test suite passes, but I found two behavior issues in the new guard plus one diagnostic unit mismatch.

stanley1208:
Thanks you so much for the review @yuhezhang-ai, all three fixes  are now landed

yuhezhang-ai:
/ok to test [f245f30](https://github.com/NVIDIA-NeMo/Automodel/pull/2350/commits/f245f30a41eb52f979b050488aa1bc3b79079f66)

yuhezhang-ai:
/ok to test [ffd9037](https://github.com/NVIDIA-NeMo/Automodel/pull/2350/commits/ffd9037b8c6c65b313fdccc557f94ed27546210f)

yuhezhang-ai (approved):
Thanks! LGTM.

PR #2357, opened 2026-05-30
refactor(vlm): gate build_model via recipe-side model-target allowlist

### LINKED ISSUE #2355: Refactor gemma4 drafter specific from recipes/vlm/finetune.py
Move `_is_gemma4_joint_target` to a model specific file since we want to keep `recipes/vlm/finetune.py` generic. 
Plan is to add a marker `__nemo_recipe_target__ = True` to class `Gemma4WithDrafter`  and also to `_BaseNeMoAutoModelClass`. This way we remove [_is_gemma4_joint_target](https://github.com/NVIDIA-NeMo/Automodel/blob/athitten/gemma4_drafter_support/nemo_automodel/recipes/vlm/finetune.py#L189-L204) from finetune.py and as a bonus we can also get rid of NeMoAutoModel* tuples in [is_nemo_auto_model](https://github.com/NVIDIA-NeMo/Automodel/blob/athitten/gemma4_drafter_support/nemo_automodel/recipes/vlm/finetune.py#L166-L173) and just have a generic  is `__nemo_recipe_target__ True` check. `build_model's` decision is just one line instead of this check: [finetune.py#L166-L173](https://github.com/NVIDIA-NeMo/Automodel/blob/athitten/gemma4_drafter_support/nemo_automodel/recipes/vlm/finetune.py#L166-L173)

Will submit this in a follow-up after #2240 gets merged to avoid super large PRs and also this change would touch autoclass base

athitten:
# What does this PR do ?

`build_model` in `recipes/vlm/finetune.py` used to decide whether a YAML `_target_` may receive the recipe's infrastructure kwargs (`device_mesh`, `distributed_config`, `peft_config`, `freeze_config`, `pipeline_config`) by enumerating every accepted callable inline, plus a gemma4-specific helper (`_is_gemma4_joint_target`) that lazily imported `Gemma4WithDrafter` just to compare classmethod equality. Every new composite required recipe-side edits scattered across two places.

This replaces that with a single, explicit **recipe-side allowlist**:

* `_accepted_targets()` -- returns the set of accepted model entrypoints: the `NeMoAutoModelFor{CausalLM,ImageTextToText,MultimodalLM}` `.from_pretrained` / `.from_config` classmethods, plus `Gemma4WithDrafter.from_pretrained` added behind a `try/except` (optional `transformers.models.gemma4_assistant` dep).
* `_is_recipe_target(target)` -- one-line membership check (relies on bound-classmethod `__eq__`/`__hash__`, not identity).
* `build_model`'s gate collapses to `if _is_recipe_target(...)`, with a clearer error message that points onboarders at `_accepted_targets()`.

### Why the allowlist lives in the recipe (not a marker on components)

An earlier iteration added a `__nemo_recipe_target__ = True` marker to the component/wrapper classes. Per feedback from @hemildesai, components should stay independent and carry no knowledge of the recipes that consume them. So the allowlist is kept entirely inside `recipes/vlm/finetune.py`, preserving the one-directional `recipes/ -> components/` layering. New infra-aware composites opt in by adding their entrypoint to `_accepted_targets()`.

# Changelog

* Remove the inline `_target_` whitelist and the `_is_gemma4_joint_target` helper from `recipes/vlm/finetune.py`.
* Add `_accepted_targets()` + `_is_recipe_target()`; use them in `build_model`.
* Update the `build_model` rejection message.
* Rewrite `tests/unit_tests/recipes/test_vlm_drafter_helpers.py` to cover the new contract (allowlist membership, NeMoAuto entrypoints, lazy gemma4 composite, bare-class / lambda / unrelated-classmethod / vanilla-HF-autoclass rejection, bound-classmethod equality-not-identity) and add `TestAcceptedTargets`.
* Update the ValueError regex in `test_finetune_vlm_helpers.py`.

No behavior change for existing models: the same set of `_target_` callables flows through `build_model` with the same infrastructure kwargs.

# Additional Information

* Related to NVIDIA-NeMo/Automodel#2355


hemildesai:
Can we do this some other way? Ideally components are independent pieces and shouldn't have any information about the recipes.

athitten:
> Can we do this some other way? Ideally components are independent pieces and shouldn't have any information about the recipes.

Thanks @hemildesai for flagging this. I have addressed in b96a88a58844cabd3ee3cbd4a19c3410cd647bed, could you please check if it looks okay ?

hemildesai (approved):

PR #2359, opened 2026-05-31
fix(gemma4): cast dense params without casting buffers

DOGEUNNKIM:
# What does this PR do?

This PR preserves HF-initialized non-parameter buffers for dense Gemma4 models so that `NeMoAutoModel` matches HuggingFace parity.

## Changelog

* Skip the `self.to(dtype)` cast in `Gemma4ForConditionalGeneration.initialize_weights()` when `enable_moe_block=False`.
* Keep the dense Gemma4 path fully HF-native, including computed buffers such as:

  * `model.language_model.embed_tokens.embed_scale`
  * `model.language_model.rotary_emb.full_attention_inv_freq`
  * `model.language_model.rotary_emb.sliding_attention_inv_freq`
* Prevent bf16 rounding of non-parameter buffers during dense Gemma4 initialization.
* Fix HF parity failure where all shared parameters matched but logits diverged.

## Root Cause

Dense Gemma4 models such as `google/gemma-4-E4B-it` do not use the NeMo MoE replacement path because `enable_moe_block=False`.

However, `initialize_weights()` still cast the whole model to `bf16` before returning:

```python
if not getattr(text_config, "enable_moe_block", False):
    self.to(dtype)
    return
```

That cast changed computed non-parameter buffers initialized by the HF model. For example, E4B has:

```text
sqrt(hidden_size=2560) = 50.596443...
```

After bf16 casting, `embed_scale` became approximately:

```text
50.5
```

This changed the input embeddings and propagated through the model, causing logit parity to fail even though all checkpoint parameters were identical.

## Validation

Ran Gemma4 E4B parity check:

```bash
python tests/unit_tests/models/gemma4/parity_check_gemma4.py \
    --hf-dir path/to/gemma-4-E4B-it/model \
    --atol 0.01
```

Before this fix:

```text
Weight check PASSED (all shared weights identical)
max  |diff| : 6.098891
mean |diff| : 0.299839
--> FAILED
```

After this fix:

```text
Weight check PASSED (all shared weights identical)

Language-model hidden-state parity:
  max  |diff| : 0.000000
  mean |diff| : 0.000000
  --> language_model backbone MATCHES HF

Manual lm_head parity from language_model outputs:
  max  |diff| : 0.000000
  mean |diff| : 0.000000

max  |diff| : 0.000000
mean |diff| : 0.000000
--> PASSED
```

**Pre checks**:

# Additional Information


athitten:
@DOGEUNNKIM thank you for the fix! Could you please fix the failing unit test ? It needs to be updated to reflect the change in your PR

DOGEUNNKIM:
> @DOGEUNNKIM thank you for the fix! Could you please fix the failing unit test ? It needs to be updated to reflect the change in your PR

@athitten Thanks for checking. I updated the dense initialize_weights path to keep the existing parameter dtype-cast behavior while avoiding casts on dense model buffers.

Specifically, instead of calling self.to(dtype) in the dense path, it now casts only parameters to the requested dtype. This keeps test_initialize_weights_dense_only_casts_dtype valid and prevents precision-sensitive buffers from being rounded to bf16.

DOGEUNNKIM:
@athitten Hi, I fixed the Ruff I001 issue by removing the extra blank line that split the first-party import block in `nemo_automodel/components/models/gemma4_moe/model.py`.

I also verified it locally with:

`ruff check --select I nemo_automodel/components/models/gemma4_moe/model.py`

and it passed. Could you please rerun the test/CI?

Thank you.

DOGEUNNKIM:
@akoumpa 

Hi, all CI checks are passing now.

Could you please review this again when you have a chance?

Thank you.

athitten (approved):
Thank you @DOGEUNNKIM ! LGTM

PR #2360, opened 2026-05-31
fix(datasets): warn when agent SFT seq_length has no effect

khazic:
# What does this PR do ?

`make_agent_chat_dataset` forwards `seq_length` to `apply_chat_template(max_length=...)`, which the tokenizer only honors when truncation is enabled (to cap) or when `padding="max_length"` (to pad). With the dataset defaults (`truncation=False`, `padding=False`) — and the shipped agent example configs — `max_length` is ignored, so `seq_length` silently has no effect and over-long dialogues pass through uncapped (misleading config + OOM risk). This warns instead of failing silently, and documents the requirement in the example configs.

# Changelog

- Warn in `make_agent_chat_dataset` when `seq_length` is set but neither truncation nor `padding="max_length"` will apply it.
- Add a clarifying comment next to `seq_length` in both `examples/llm_finetune/agent/qwen2_5_3b_function_calling*.yaml` configs.
- Add unit tests for the warning case and the no-warning cases (truncation on / `padding="max_length"` / `seq_length` unset).

**Pre checks**:

# Additional Information

- No behavior change: warning only — it does not enable truncation or drop any tokens.

### DIFF
--- a/examples/llm_finetune/agent/qwen2_5_3b_function_calling.yaml
+++ b/examples/llm_finetune/agent/qwen2_5_3b_function_calling.yaml
@@ -82,6 +82,9 @@ dataset:
   _target_: nemo_automodel.components.datasets.llm.agent_chat.make_agent_chat_dataset
   dataset_name: llamafactory/glaive_toolcall_en
   split: train
+  # NOTE: seq_length only caps length when truncation is enabled (or pads when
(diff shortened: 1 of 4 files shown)


HuiyingLi (dismissed):
LGTM, thank you!

HuiyingLi (approved):

PR #2361, opened 2026-05-31
fix(datasets): tag agent SFT row errors with the example id

khazic:
# What does this PR do ?

`_format_example` renders agent SFT rows lazily inside the dataloader. A single malformed row — bad `tools`/`tool_call` JSON, or missing required fields — surfaced as an opaque `JSONDecodeError`/`AssertionError` deep in the stack with no indication of which row caused it, aborting the whole training run. This re-raises any parse/render failure as a `ValueError` tagged with the example id so the offending row is actionable.

# Changelog

- Wrap `_format_example` so any parsing/rendering failure is re-raised as a `ValueError` tagged with the example `id` (original error chained via `raise ... from e`).
- Move the renderer body into `_format_example_impl`; the public `_format_example` (used by the lazy dataloader map) is the thin error-context wrapper.
- Add unit tests: malformed `tools` JSON and malformed `tool_call` content both report the example id; existing field-validation messages are preserved.

**Pre checks**:

# Additional Information

- No happy-path behavior change.
- Touches the same `_format_example` as the "reject examples with no supervised tokens" fix (separate PR); whichever merges first, the other needs a trivial rebase.

### DIFF
--- a/nemo_automodel/components/datasets/llm/agent_chat.py
+++ b/nemo_automodel/components/datasets/llm/agent_chat.py
@@ -260,6 +260,43 @@ def _format_example(
     train_on_last_turn_only: bool = False,
     drop_history_reasoning_content: bool = False,
 ) -> Dict[str, List[int]]:
+    """Render one agent example into tokenized ``input_ids`` / ``labels``.
+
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
LGTM thanks

PR #2362, opened 2026-05-31
fix(datasets): warn on agent SFT rows with no supervised tokens

khazic:
# What does this PR do ?

When truncation drops every assistant turn (or last-turn masking leaves
nothing), `_format_example` produces an agent SFT sample whose `labels` are all
`ignore_index` (-100). The pre-existing guard only catches "no assistant message
present", not "assistant tokens truncated away".

A *single* such sample is harmless: the loss normalizes by the batch's
supervised-token count, so an all-masked example contributes nothing to the
gradient. A NaN only arises if a *whole* step is empty, which signals a
misconfigured `seq_length`/`truncation`. So rather than hard-failing the run on
one example, this detects the all-masked case and **warns** with the example id
(pointing at `seq_length`/`truncation`) while **keeping** the sample — it is
already effectively skipped from the gradient, and the map-style
dataset/collator cannot drop a row in place at this stage.

# Changelog

- After rendering (and after optional last-turn masking), emit a warning tagged
  with the example `id` when a sample has no supervised tokens (all labels
  masked), pointing at `seq_length`/`truncation` as the likely cause. The sample
  is kept, not dropped or raised on.
- Add unit tests for the all-masked (warns, kept) and partially-supervised
  (passes through) cases.

**Pre checks**:

# Additional Information

- Rebased onto current `main` (after #2361 moved the renderer body into
  `_format_example_impl` and #2360 added the `seq_length`-inert warning); both
  warnings coexist in `agent_chat.py`.

### DIFF
--- a/nemo_automodel/components/datasets/llm/agent_chat.py
+++ b/nemo_automodel/components/datasets/llm/agent_chat.py
@@ -338,6 +338,23 @@ def _format_example_impl(
     )
     if train_on_last_turn_only:
         _mask_labels_to_last_turn(tokenized["labels"])
+
+    # Truncation (or over-aggressive last-turn masking) can leave a sample with no
+    # supervised tokens at all — every label is ``ignore_index`` (-100). A single
+    # such sample is harmless: the loss normalizes by the batch's supervised-token
+    # count, so it just contributes nothing (a NaN only arises if a *whole* step is
+    # empty, which signals a misconfigured seq_length/truncation). Rather than
+    # hard-failing the run on one example, warn (with the id) and let it through — it
+    # is already effectively skipped from the gradient, and the map-style
+    # dataset/collator cannot drop a row in place at this stage.
+    labels = tokenized.get("labels")
+    if labels is not None and all(label == -100 for label in labels):
+        logger.warning(
+            "Agent SFT example (id=%r) has no supervised tokens (all labels masked); it "
+            "contributes nothing to the loss. This usually means truncation dropped every "
+            "assistant turn — increase seq_length or disable truncation.",
+            example.get("id"),
+        )
     return tokenized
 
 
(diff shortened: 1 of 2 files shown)


nemo_automodel/components/datasets/llm/agent_chat.py
```diff
@@ -301,6 +301,18 @@ def _format_example(
     )
     if train_on_last_turn_only:
         _mask_labels_to_last_turn(tokenized["labels"])
+
+    # Truncation (or over-aggressive last-turn masking) can leave a sample with no
+    # supervised tokens at all — every label is ``ignore_index`` (-100). Such a
+    # sample contributes no gradient and can drive a NaN loss for the batch, so
+    # fail loudly with the example id instead of emitting it silently.
```
HuiyingLi: What do you think about skipping example(s) with warning v.s. hard fail the run?

HuiyingLi (dismissed):
Thank you so much

nemo_automodel/components/datasets/llm/agent_chat.py
khazic: Thanks — good call, switched from hard-fail to a warning in 302a6d39.

Reasoning: a single all-masked example is actually harmless. The loss normalizes by the batch's supervised-token count (`loss = loss / num_label_tokens`, e.g. `linear_ce.py`), so an empty sample just contributes 0 — a NaN only arises if a *whole* step has zero supervised tokens, which signals a systematic `seq_length`/`truncation` misconfig rather than one bad row. So aborting the run on a single example was too aggressive.

On skip vs keep: I warn (with the example id) and keep the sample rather than truly dropping it. Detection happens inside the lazy `map_fn`, where the row can't be dropped in place — `LazyMappedDataset` is fixed-length and `default_collater` doesn't accept `None` (the existing `skip_invalid_samples` only skips at JSONL-read time). Since the kept sample contributes nothing to the gradient, warn-and-keep is the clean equivalent of skipping here, and the per-sample warning still surfaces the systematic case (a flood of warnings → check `seq_length`).

The test now asserts the warning fires (carrying the id) and that the sample passes through unchanged.

HuiyingLi (dismissed):
LGTM thank you so much.

HuiyingLi (approved):

PR #2365, opened 2026-06-01
fix(agent-sft): harden tool-call eval and fix last-turn loss masking

khazic:
# What does this PR do ?

Correctness fixes for multi-turn agent (tool-calling) SFT, all on that path.

## Tool-call eval: deadlock-safe, honestly documented, and DP-sharded

**Deadlock.** `_run_validation` reduced `ToolCallAccuracyEvaluator` metrics with one DP all-reduce per metric key, inside a `try`. Two rank-divergent conditions desynced the collectives and hung surviving ranks: (a) `evaluate()` raising on a subset of ranks (e.g. a divergent `generate()` CUDA OOM) skipped that rank's all-reduces while peers blocked; (b) the per-key loop iterated a **data-dependent** key set (`_skip_<reason>` differs per rank). Now reduced as a single packed all-reduce over a **fixed** vector, with a local `evaluate()` failure tolerated as an empty result that still participates.

**Honest docs + actually running.** The example YAML claimed the eval "Runs at every val step", but the recipe is FSDP2, which **skips** in-loop generation eval unless `run_on_fsdp2: true` (running `generate()` on the sharded training model repeatedly unshards params and can OOM the next step). The comment now says so, with a commented opt-in. (The default-skip is intentional and unchanged.)

**DP sharding.** The recipe never set `sample_shard`, so every DP rank redundantly scored the full set and `_count` was summed into `dp_size * n`. Now `sample_shard=(dp_rank, dp_size)` is injected **only under DDP** (model replicated per rank → each can generate a disjoint subset). Under FSDP2/MegatronFSDP the sharded model's `generate()` issues per-layer all-gathers that must stay in lockstep, so sharding there would desync and hang — those keep scoring the full set. The reduction sums across ranks only when sharded; when replicated it uses the identical local result (true `_count`, no needless collective).

## Last-turn loss mask: apply before reasoning is masked

`train_on_last_turn_only` post-processed labels with a "keep the last contiguous supervised run" heuristic. With `mask_reasoning_content=True`, masking reasoning punched a hole into the final assistant turn, so the heuristic dropped supervised content before it. The restriction now runs inside `format_chat_template`, on the hole-free assistant mask, before reasoning is masked.

# Changelog

- Reduce agent SFT tool-call eval metrics with one packed, failure-tolerant all-reduce over a fixed key set (fixes a multi-GPU deadlock); expose `ToolCallAccuracyEvaluator.METRIC_KEYS`.
- Inject `sample_shard` for DDP eval (disjoint per-rank scoring, correct `_count`); keep FSDP2/MegatronFSDP replicated in lockstep. New `_dp_eval_sample_shard` helper + public `sample_shard` property.
- Make the example YAML's tool-call eval comment honest (FSDP2 default-skip + `run_on_fsdp2` opt-in).
- Fix `train_on_last_turn_only` dropping in-turn content before a masked reasoning span; move the restriction into `format_chat_template`. `_mask_labels_to_last_turn` moved to `formatting_utils`, generalized over `ignore_index`.
- Tests: fixed-key reduction contract, `sample_shard` property + DDP-shard decision, last-turn mask (multi-turn + content-around-masked-reasoning regression).

**Pre checks**:

- [x] New tests added
- [ ] Docs: example YAML comment updated; no public API doc change

# Additional Information

- DCO signed-off. `ruff format` + `ruff check` clean; affected unit tests pass locally.
- The DP-sharding builds directly on the deadlock-safe reduction (sharded per-rank subsets are only safe to all-reduce once the key set is fixed), which is why both land here together.

### DIFF
--- a/nemo_automodel/recipes/llm/train_ft.py
+++ b/nemo_automodel/recipes/llm/train_ft.py
@@ -61,7 +61,7 @@ from nemo_automodel.components.config._arg_parser import parse_args_and_load_con
 from nemo_automodel.components.datasets.llm.megatron.sampler import create_megatron_sampler
 from nemo_automodel.components.datasets.llm.megatron_dataset import MegatronPretraining
 from nemo_automodel.components.datasets.llm.packed_sequence import pack_dataset
-from nemo_automodel.components.distributed.config import DDPConfig, FSDP2Config, MegatronFSDPConfig
+from nemo_automodel.components.distributed.config import FSDP2Config, MegatronFSDPConfig
 from nemo_automodel.components.distributed.cp_utils import make_cp_batch_and_ctx
 from nemo_automodel.components.distributed.init_utils import (
     initialize_distributed,
@@ -69,7 +69,7 @@ from nemo_automodel.components.distributed.init_utils import (
 from nemo_automodel.components.distributed.megatron_fsdp import fully_shard_optimizer
 from nemo_automodel.components.distributed.mesh import MeshContext
 from nemo_automodel.components.distributed.pipelining import AutoPipeline
-from nemo_automodel.components.distributed.utils import FirstRankPerNode, get_sync_ctx
+from nemo_automodel.components.distributed.utils import FirstRankPerNode, dp_eval_sample_shard, get_sync_ctx
 from nemo_automodel.components.loggers.comet_utils import build_comet
 from nemo_automodel.components.loggers.log_utils import setup_logging
 from nemo_automodel.components.loggers.metric_logger import MetricsSample, build_metric_logger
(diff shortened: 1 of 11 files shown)


nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -174,6 +174,20 @@ def _get_num_thd_chunks(pp_enabled, cfg):
     return 1
 
 
+def _dp_eval_sample_shard(distributed_config: Any, dp_rank: int, dp_size: int) -> tuple | None:
```
HuiyingLi: What about moving this helper to  components/distributed/utils.py

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -904,6 +918,18 @@ def __init__(self, cfg):
         """
         self.cfg = cfg
 
+    def _shard_tool_call_eval_across_dp(self):
```
HuiyingLi: Is this wrapper func necessary? Can we call _dp_eval_sample_shard directly?

nemo_automodel/recipes/llm/train_ft.py
khazic: Moved to `components/distributed/utils.py` as `dp_eval_sample_shard` in bbee2728.

nemo_automodel/recipes/llm/train_ft.py
khazic: Agreed. Dropped the wrapper and inlined the guarded call at the single `setup()` call site (bbee2728).

tests/unit_tests/recipes/test_train_ft.py
```diff
@@ -1983,3 +1984,149 @@ def test_mixed_strings_and_enums(self):
...
+        "args_json_valid",
+        "args_field_recall",
+        "args_field_precision",
+        "args_exact_match",
+    )
+
+    def __init__(self, *, sample_shard=None, run_on_fsdp2=False, result=None, raises=False):
+        self.sample_shard = sample_shard
+        self.run_on_fsdp2 = run_on_fsdp2
+        self._result = result if result is not None else {}
```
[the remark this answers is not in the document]
khazic: Done. The fake now reuses `ToolCallAccuracyEvaluator.METRIC_KEYS` so it can't drift from the real evaluator (bbee2728).

HuiyingLi (approved):

PR #2366, opened 2026-06-01
feat(datasets): add turn-aware history truncation to agent SFT dataset

khazic:
# What does this PR do ?

Adds **turn-aware history truncation** to the multi-turn agent (tool-calling) SFT dataset, so long dialogues don't lose their supervised target.

Today, a dialogue longer than `seq_length` is handled only by token-level `truncation`, which clips from the tokenizer's `truncation_side` (typically the right). For chat SFT that drops the **final assistant turn** — exactly the token span being supervised — producing all-masked or mistargeted samples. (#2360/#2362 added warnings for these cases; this PR lets you avoid them.)

`make_agent_chat_dataset(..., truncate_history=True)` (default `False`): when `seq_length` is set, drop the **oldest** conversation exchanges — keeping any leading `system` message, the tool definitions, and the **final exchange** (the user turn that produces the last assistant answer) — until the rendered dialogue fits. Exchanges are cut at `user` boundaries so `tool_call`/`tool_response` pairing among the survivors stays intact. It composes with the existing token-level `truncation`, which still clips the remainder if even the final exchange overflows.

# Changelog

- Add `truncate_history` to `make_agent_chat_dataset` (and the `_format_example` render path): turn-level, drop-oldest truncation that preserves system + tools + the final supervised turn. Default `False` (no behavior change).
- New `_truncate_messages_to_fit` helper (reuses `_tokenized_chat_length`); applied to the converted messages before chat-template rendering.
- Document the option (commented) in the `qwen2_5_3b_function_calling{,_lora}.yaml` examples.
- Tests: drop-oldest selection, already-fits no-op, no-`user`-boundary no-op, minimal-suffix fallback when nothing fits, and `_format_example` truncate-before-render wiring.

**Pre checks**:

- [x] Did you add or update any necessary documentation? (docstring + example config note)

# Additional Information

- Orthogonal to and stacks with the existing token-level `truncation`; `truncate_history` defaults off, so existing configs are unaffected.
- DCO signed-off. `ruff format` + `ruff check` clean; `pytest tests/unit_tests/datasets/llm/test_agent_chat.py` → 45 passed.


HuiyingLi (dismissed):
Thank you for handling the details.

HuiyingLi (approved):

PR #2369, opened 2026-06-01
fix(speculative): rescale EAGLE-1/2 trailing grad-accum window

khazic:
## What

The EAGLE-1/2 recipe (`train_eagle1.py`, and `train_eagle2.py` via its
`TrainEagle1Recipe` subclass) hand-rolls its gradient-accumulation loop and
flushes the trailing partial accumulation window at each epoch end. That flush
was mis-scaled in two ways — both already fixed in EAGLE-3 and backported here:

1. **LR schedule was truncated.** `total_optim_steps` used floor division
   (`(num_epochs * num_batches) // accum`) while the loop actually performs
   `ceil(num_batches / accum)` optimizer steps per epoch (the trailing flush is
   a real step). The cosine `progress` therefore saturated early and the
   trailing steps trained at `min_lr_ratio` instead of following the intended
   decay.
2. **Trailing gradient was undersized.** Each micro-batch divides its loss by
   `grad_accumulation_steps`. A trailing window of `r < accum` micro-batches
   thus stepped with only `r / accum` of a normal step's gradient magnitude —
   the window was never rescaled.

EAGLE-3 fixed both in #2257 (`_optim_steps_per_epoch` ceil helper) and the
`pending_micro_batches` gradient rescale; this PR brings EAGLE-1/2 to parity.

It also marks the **EAGLE-3 target model `requires_grad_(False)`** explicitly,
matching EAGLE-1/2. The target is supervision-only and `generate_batch` already
runs under `@torch.no_grad()`, so this changes no behavior — it just makes the
frozen-target intent unambiguous and guards against future code paths.

## Changes

- `nemo_automodel/recipes/llm/train_eagle1.py`
  - Add `_optim_steps_per_epoch()` (ceil), use it for `total_optim_steps`.
  - Track `pending_micro_batches`; rescale the trailing window's gradient by
    `grad_accumulation_steps / pending_micro_batches` before clip/step.
- `nemo_automodel/recipes/llm/train_eagle3.py`
  - Explicit `self.target_model.requires_grad_(False)` in `setup()`.
- `tests/unit_tests/recipes/llm/test_train_eagle1_grad_accum.py` (new)
  - Parametrized ceil-division coverage for `_optim_steps_per_epoch`.
  - Regression test asserting every optimizer step — full window **and**
    trailing flush — consumes a full-scale (unit) gradient. Without the rescale
    the trailing step sees `pending/accum · ones` and the test fails.

## Testing

- `ruff format` / `ruff check`: clean.
- `pytest tests/unit_tests/speculative/ tests/unit_tests/recipes/llm/test_eagle_*.py
  tests/unit_tests/recipes/llm/test_train_eagle*` → **200 passed, 2 skipped**
  (skips are the GPU/flash-attn functional cases).

## Scope

Contained bugfix + regression test. Extracting a shared `EagleRecipeBase` to
de-duplicate the EAGLE-1/3 training loops is a worthwhile but separate refactor
and intentionally out of scope here.


nemo_automodel/recipes/llm/train_eagle1.py
```diff
@@ -534,13 +560,31 @@ def run_train_validation_loop(self):
...
+            # gradients would otherwise be wiped by the next epoch's
+            # ``zero_grad`` and the samples wasted.
+            #
+            # Each micro-batch divided its loss by ``grad_accumulation_steps``
+            # in anticipation of a full window. With only ``pending_micro_batches``
+            # contributors, the accumulated gradient magnitude is
+            # ``pending_micro_batches / grad_accumulation_steps`` of a normal
+            # step; rescale by the inverse so the trailing step's gradient is on
+            # the same scale as every other step.
+            if pending_micro_batches > 0:
```
HuiyingLi, line 587: Non-blocking: this trailing flush assumes **equal batch counts across DP ranks**. If ranks ever had unequal trailing windows, the per-rank `grad_accumulation_steps / pending_micro_batches` scale would diverge — and a rank with `pending_micro_batches == 0` would skip the flush entirely (including its `clip_grad_norm_` / any collective inside it) while peers step, which is the same class of collective desync fixed in #2365. In practice the default `DistributedSampler` pads to equal length so this holds; a one-line note here making that assumption explicit would help (and flag it if a non-padding sampler is ever used). Not a blocker.

khazic: Added the assumption note in d1500be5. Confirmed the recipe's loader uses `DistributedSampler(drop_last=False)` (eagle3.py:60) plus `DataLoader(drop_last=False)` (eagle3.py:69), so every rank is padded to an equal sample count -> equal batches/epoch -> equal trailing windows; the per-rank `scale` matches and no rank lands on `pending_micro_batches == 0` while peers step. So no behavior change is needed today. The comment documents that assumption and flags the desync/hang risk if a non-padding sampler is ever introduced. Thanks for the careful catch.

HuiyingLi (approved):

PR #2390, opened 2026-06-02
refactor(datasets): typed Config + build per dataset, config-driven dataloader

akoumpa:
## What

Completes typed config-owned construction for the dataset and dataloader layer. Recipes now resolve declarative YAML at the RecipeConfig boundary and compose component config build results with explicit runtime arguments.

- Adds typed Config + build coverage across LLM, retrieval, VLM, diffusion, audio-backed VLM, and BAGEL multimodal datasets.
- Routes VLM processor, dataset, pretokenization, neat packing, sampler, collator, pipeline-media wrapping, and StatefulDataLoader construction through VlmDataloaderConfig.build.
- Routes every diffusion dataloader target through a typed config while retaining the existing public builder functions as compatibility delegates.
- Adds BagelDatasetConfig and BagelDataloaderConfig; serialized config state is copied before BAGEL packing mutates runtime data, and runtime tokenizer, rank, worker, and seed values remain explicit build arguments.
- Keeps legacy YAML targets working through typed resolvers; the old VLM build_dataloader entry point is a tested deprecation wrapper.

## Validation

- All affected example schemas resolve: 76 VLM/VLM-KD, 15 diffusion, and 2 BAGEL configs.
- Domain suite: 267 passed, 3 skipped.
- Config/dataset/recipe regression suite: 266 passed, 7 skipped.
- Ruff format and check, compileall, and import-linter all pass.
- ty is not installed in the local environment.


nemo_automodel/recipes/retrieval/train_bi_encoder.py
```diff
@@ -294,7 +294,9 @@ def setup(self):
             self.tokenizer.padding_side = "left"
 
         self.dataloader = build_dataloader(
-            self.cfg.dataloader,
+            # Raw ``dataloader`` config node (retrieval nests dataset under dataloader.dataset);
+            # RecipeConfig.dataloader is the typed LLM-recipe property and resolves to None here.
+            self.cfg.get("dataloader"),
```
oliverholworthy: This change keeps retrieval working, but it leaves retrieval outside the main config-driven dataloader refactor, which seems to be the purpose of this PR.

Since `RetrievalDatasetConfig` is already registered and `DataloaderConfig` now owns the `StatefulDataLoader` construction, is the intended direction to migrate retrieval examples to top-level `dataset:` and have the retrieval recipe use `self.cfg.dataloader.build(tokenizer=...)` in this PR, or as a follow-up? Or is typed `RecipeConfig.dataloader` intended to be LLM-only?

It looks like class-based collator construction is the remaining issue: `make_collate_fn()` currently partial-binds targets as per-batch callables, while the retrieval collators need to be instantiated once with the runtime tokenizer.

akoumpa: Hi @oliverholworthy , thanks for the feedback, i'm revisiting the current code and will incorporate your feedback in next commits; expect to be pinged in the next few days; 🙇

nemo_automodel/recipes/_typed_config.py
```diff
@@ -158,6 +176,382 @@ def optimizer(self) -> "OptimizerConfig" | None:
         factory, kwargs = _callable_and_kwargs(node)
         return build_optimizer_config(factory, kwargs)
 
+    @cached_property
+    def dataloader(self) -> "DataloaderConfig" | None:
+        node = self._raw.get("dataset", None)
```
oliverholworthy: I wonder if it would be helpful to support backward compatibility of nested dataset under dataloader. or if this enforcment of dataset being top-level is prefered?

oliverholworthy: The cross-encoder example configs still need migrating to the new form.

tests/unit_tests/datasets/vlm/test_vlm_loader.py
oliverholworthy, line 1: This seems to collide with the existing tests/unit_tests/config/test_loader.py under pytest’s current import mode. Could we rename it to test_vlm_loader.py, or make the directories packages with __init__.py?

nemo_automodel/recipes/_typed_config.py
akoumpa: Thank you @oliverholworthy , these are both good suggestions that I've followed. In particular, i've added support for backward compatibility for nested dataset declaration and updated the configs to use the top-level dataset yaml layout.

oliverholworthy (dismissed):
I’ve reviewed the latest changes and everything looks good to me from the retrieval model perspective

akoumpa:
CI is fully green on `841bda17432b8d681626c568de114ac7f5c8cf6b` (75 passed, 6 skipped, 0 pending/failing). The original failing `L2_HF_Transformer_VLM` lane now passes. Re-requesting review because the prior approval was dismissed by the branch update.

nemo_automodel/components/datasets/llm/chat_dataset.py
```diff
@@ -315,6 +319,56 @@ def _conversations_to_messages(conversations: Any) -> List[Dict[str, Any]]:
...
+    """Truncation mode for ``format_chat_template``."""
+    start_of_turn_token: str | None = None
+    """Optional token marking assistant turns for answer-only loss."""
+    chat_template: str | None = None
+    """Optional Jinja template string overriding ``tokenizer.chat_template``."""
+    shuffle_seed: int | None = None
+    """If set, shuffles Hub/Parquet data before applying a split slice."""
+    mask_reasoning_content: bool = False
+    """If ``True``, exclude rendered reasoning traces from the loss mask."""
+    unshifted: bool = False
```
yuhezhang-ai, line 350: `ChatDatasetConfig` must preserve and forward the existing `mask_history` argument. Both `examples/dllm_sft/diffusion_gemma_sft.yaml` and `diffusion_gemma_lora.yaml` set `mask_history: true`; they now fail during `RecipeConfig.dataloader` resolution with `TypeError: ChatDatasetConfig got unexpected dataset config field(s): mask_history`. This contradicts the PR's stated legacy-YAML compatibility.

nemo_automodel/components/datasets/llm/squad.py
```diff
@@ -139,3 +146,38 @@ def make_squad_dataset(
...
+    fp8: bool = False
+    """Flag reserved for future mixed-precision use (currently unused)."""
+    split: str = "train"
+    """Which split of the dataset to load (e.g. ``train``, ``validation``)."""
+    dataset_name: str = "squad"
+    """Identifier for the HuggingFace dataset to load."""
+    padding: bool | str = False
+    """Optional padding strategy."""
+    truncation: bool | str = False
+    """Optional truncation strategy."""
```
yuhezhang-ai, line 170: `SquadConfig` must preserve and forward the existing `chat_template` argument. Both OLMo SQuAD example configs use it and now fail during typed config resolution with `TypeError: SquadConfig got unexpected dataset config field(s): chat_template`. Add the field and pass it to `make_squad_dataset`.

nemo_automodel/components/datasets/llm/agent_chat.py
```diff
@@ -659,3 +665,50 @@ def make_agent_chat_dataset(
...
+    """If set, keep only the first N examples."""
+    padding: str | bool = False
+    """Padding strategy forwarded to the tokenizer."""
+    truncation: str | bool = False
+    """Truncation strategy forwarded to the tokenizer."""
+    mask_reasoning_content: bool = False
+    """If True, exclude assistant ``reasoning_content`` tokens from the loss."""
+    train_on_last_turn_only: bool = False
+    """If True, supervise only the final assistant turn of each dialogue."""
+    drop_history_reasoning_content: bool = False
```
yuhezhang-ai, line 697: `AgentChatConfig` must preserve and forward `truncate_history`, which remains a public argument of `make_agent_chat_dataset` and is documented in the agent example configs. Enabling that documented option now produces an unexpected-field error.

yuhezhang-ai (commented):
Please add a parameter-parity regression test for every legacy-target → typed-config registration. Strict field validation means an omitted legitimate legacy argument is rejected exactly like a typo. The test should verify that each public legacy parameter is represented by the typed dataclass, excluding explicit runtime-only arguments such as tokenizer, rank, and training schedule.

nemo_automodel/components/datasets/llm/agent_chat.py
akoumpa: Thanks Yuhe for the sharp eyes, I'm implementing your suggestion.

yuhezhang-ai (approved):

PR #2399, opened 2026-06-03
feat(speculative): add gpt-oss EAGLE-3 draft model

khazic:
## What

Adds EAGLE-3 speculative-decoding draft support for gpt-oss targets
(`GptOssForCausalLM`, both 20B and 120B), wired through the existing
`TrainEagle3Recipe` and the EAGLE-3 draft registry.

## Why a dedicated draft (RoPE)

The draft is structurally a Llama-style dense draft — it consumes only the
target's post-block auxiliary hidden states, so the target's experts, attention
sinks, and sliding-window mask never reach it. RoPE is the exception, and it
must match the target. During speculative decoding the draft runs at the **same
token positions** as the target, and gpt-oss is a long-context model whose
rotary frequencies are reshaped by YaRN (base 150000, `factor=32`, NTK-by-parts
with a concentration scale, extending 4096 → 131072). A draft trained with a
different rotary schedule is positionally inconsistent with the target: it can
converge at the short context used during training, but its acceptance rate
degrades at the long positions gpt-oss is built for.

The shared `LlamaRotaryEmbedding` implements only `{default, llama3}` and
silently falls back to the llama3 NTK schedule for `rope_type="yarn"`. So
`GptOssEagle3DraftModel` swaps in `GptOssDraftRotaryEmbedding`, which reproduces
gpt-oss's exact YaRN `inv_freq` and concentration (reusing the target's own
`components/models/gpt_oss/rope_utils.RotaryEmbedding`) but returns `(cos, sin)`
in the duplicated `head_dim` layout. gpt-oss's interleaved `apply_rotary_emb`
and the draft's `rotate_half`-based `apply_rotary_pos_emb` are algebraically
identical under that layout, so the draft's rotation is bit-faithful to the
target's.

## Changes

- `components/speculative/eagle/draft_gpt_oss.py`: `GptOssDraftRotaryEmbedding`
  + `GptOssEagle3DraftModel` (a thin `LlamaEagle3DraftModel` subclass).
- Registered `GptOssForCausalLM` → `GptOssEagle3DraftModel` in
  `EAGLE3_DRAFT_REGISTRY`.
- Example config `examples/speculative/eagle3/gpt_oss_20b_eagle3.yaml`.
- Unit tests: draft↔target RoPE equivalence, YaRN-engaged check, partial-rotary
  guard, registry wiring, forward shapes.

The on-disk state-dict layout and the saved
`architectures: ["LlamaEagle3DraftModel"]` string are unchanged, so checkpoints
load into SGLang exactly like the Llama draft. The draft is config-driven, so it
covers both gpt-oss-20b and -120b — only the target sharding in the YAML differs.

## Validation

Unit tests pass; the full `tests/unit_tests/speculative` suite is green with no
regressions. The RoPE equivalence test asserts the draft's rotation matches the
gpt-oss target's `RotaryEmbedding` exactly (so a silent fallback to plain/llama3
RoPE would fail the test).

End-to-end long-context acceptance-rate logs from an 8×8 run to follow — I'll
add them.


khazic:
Logs to follow — I'll add the end-to-end long-context acceptance-rate logs from an 8×8 run shortly.

HuiyingLi (approved):

PR #2413, opened 2026-06-04
feat(speculative): add P-EAGLE parallel-drafting training

kashif:
Adds P-EAGLE training: an EAGLE-3 drafter that predicts several tokens in parallel in one forward pass instead of unrolling autoregressively ([arXiv:2602.01469](https://arxiv.org/abs/2602.01469)).

It reuses the EAGLE-3 target, dataset, and train loop, and swaps in a parallel core: COD down-sampling packs all prediction depths onto one axis, depth>=1 positions share a learnable hidden state, and a FlexAttention block mask handles causality. The draft also gets the EAGLE-3.1 `fc_norm` toggle.

Small shared change: a `target_attn_implementation` knob on the EAGLE-3 recipe so the frozen target can run on sdpa when a backend's flash path is broken.

Tested: unit tests for COD sampling / mask / registry / forward, and a 2-GPU smoke run (loss drops, clean exit). Docs updated in the EAGLE guide.

### DIFF
--- a/examples/speculative/peagle/llama_peagle_mvp.yaml
+++ b/examples/speculative/peagle/llama_peagle_mvp.yaml
@@ -1,0 +1,66 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# P-EAGLE (Parallel-Drafting EAGLE) draft training — arXiv:2602.01469.
+#
+# Trains a parallel-drafting EAGLE draft that predicts num_depths tokens per
+# position in one forward pass, conditioned on the frozen target's aux hidden
+# states (same online supervision as the EAGLE-3 recipe).
+#
+# Launch (8-GPU single node):
+#   python -m torch.distributed.run --nproc-per-node=8 \
+#       nemo_automodel/recipes/llm/train_peagle.py \
+#       -c examples/speculative/peagle/llama_peagle_mvp.yaml
+
+recipe: TrainPEagleRecipe
+
+dist_env:
+  backend: nccl
+  timeout_minutes: 30
+
+recipe_args:
+  target_model_name_or_path: meta-llama/Llama-3.1-8B
+  train_data_path: /path/to/train.jsonl
+  val_data_path: null
+  train_split: null
+  val_split: null
+  output_dir: ./outputs/peagle_llama_mvp
+  seq_length: 2048
+  micro_batch_size: 1          # P-EAGLE packs one sequence across depths per step
+  grad_accumulation_steps: 1
+  num_workers: 0
+  num_epochs: 1
+  draft_vocab_size: 32000
+  # Parallel-drafting knobs (COD sampling)
+  num_depths: 8                # parallel prediction groups K
+  down_sample_ratio: 0.7       # geometric retention decay across depths
+  down_sample_ratio_min: 0.2   # retention floor
+  # mask_token_id: null        # defaults to tokenizer pad/eos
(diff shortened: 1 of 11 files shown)


kashif:
Heads up — there are two earlier open PRs for the same feature that I missed before opening this: #2376 (P-EAGLE parallel-drafting) and #2409 (adds sequence partitioning on top). After reading both, they're further along than this PR and I'd lean toward closing this one in favor of #2409:

- They wire P-EAGLE into the existing `TrainEagle3Recipe` via a `parallel_drafting` toggle and extend the existing `LlamaEagle3DraftModel`, instead of adding a separate recipe + draft class like I did here. Less duplication, and merged EAGLE checkpoints still round-trip.
- They support a multi-layer draft (the paper's capacity compensation); this PR is single-layer only.
- #2409 implements the paper's sequence partitioning (Algorithm 1, §3) for long-context training — this PR doesn't have that at all, and it's one of the main P-EAGLE contributions.
- They serialize the COD config into `config.json` with a checkpoint round-trip test so the draft loads straight into vLLM's parallel-drafting runtime.

The COD sampler and the flex-attention mask are essentially the same port in both (we landed on the same reference independently).

Two bits here that aren't in those PRs and are worth keeping regardless of which P-EAGLE lands:
1. `target_attn_implementation` — lets the frozen target run on sdpa, which sidesteps a transformers FA2 `s_aux` crash on Qwen3 targets. It's independent of P-EAGLE.
2. The EAGLE guide docs section.

I'll pull those two into a small standalone PR and close this one in favor of #2409 unless anyone wants the separate-recipe structure instead.

kashif:
Closing in favor of #2409 (and #2376), which cover the same parallel-drafting feature more completely — multi-layer draft, sequence partitioning for long context, and the vLLM checkpoint contract — and are integrated into the existing EAGLE-3 recipe rather than a separate one.

The two independent bits from this branch (the `target_attn_implementation` knob that dodges the Qwen3 FA2 `s_aux` crash on the frozen target, and the EAGLE guide docs section) I can send as a small standalone PR if useful.

HuiyingLi:
Hi @kashif — thanks for the contribution!

Heads up: we already have an open PR for P-EAGLE parallel-drafting training in #2376, which covers largely the same scope (same paper, same feature). There's significant overlap in the files touched — `peagle_attention.py`, `peagle_data.py`, `train_eagle3.py`, `eagle/__init__.py`, and `test_peagle.py` are modified in both PRs.

Could you take a look at #2376 and coordinate so we don't end up with duplicate work or merge conflicts? It may make sense to consolidate into one PR or build on top of the other.

Also, if you haven't already, please check out our existing speculative decoding directory at `nemo_automodel/components/speculative/` and the examples under `examples/speculative/` — there's existing infrastructure (EAGLE-3 core, registry, data pipeline, serving) that both PRs should align with.

kashif:
yes @HuiyingLi  just saw... i'll coordinate, closing this

kashif:
Thanks! I completely agree, and you'll see I just closed this PR in favor of #2376 / #2409 for exactly that reason. I went through both: they cover the same paper/feature more completely (multi-layer draft, the Algorithm 1 sequence partitioning in #2409, and the vLLM checkpoint contract) and, importantly, they build on the existing `nemo_automodel/components/speculative/` infra — the `parallel_drafting` toggle on `TrainEagle3Recipe` and the extension of `LlamaEagle3DraftModel`, rather than the separate recipe/draft class I had here. That's the better shape, so no need to consolidate mine in.

Only one piece from my branch is independent of P-EAGLE and might be worth keeping: a `target_attn_implementation` knob on the EAGLE-3 recipe that lets the frozen target run on sdpa — it sidesteps a transformers FA2 `s_aux` crash I hit with Qwen3 targets. Happy to send that (plus the EAGLE guide docs touch-up) as a small standalone PR if it's useful; otherwise, I'll leave #2376/#2409 to it.

kashif:
Done: opened #2415 with just the target_attn_implementation knob (plus a one-line docs entry). It defaults to None so it's a no-op for existing configs, and it's independent of P-EAGLE, so it should rebase cleanly under #2376/#2409 whenever they land. No rush on my end if you'd rather take it after those merge.

And yes, I'd love to keep working on speculative decoding here. Send over whatever ideas you're refining whenever they're ready and I'll dig in. Happy to help push the SD stack forward.

khazic:
Thanks @kashif! I'm the author of #2376 (P-EAGLE parallel-drafting) and #2409 (sequence partitioning on top). Really glad to see your PR — looks like we landed on the same paper independently.

It'd be great to build out Automodel's speculative-decoding stack together. For now it probably makes sense to go with #2376 + #2409 as the base, but I'd love to keep collaborating with you from here.

PR #2415, opened 2026-06-04
feat(speculative): add target_attn_implementation knob for EAGLE-3 target

kashif:
Small, P-EAGLE-independent fix split out of #2413.

The frozen target in EAGLE-3 training loads with HF's auto-selected attention backend. On current transformers builds that picks FlashAttention-2 for Qwen3 targets, which crashes in flash_attention_forward when s_aux is None, so target supervision dies before the first step no matter which drafter you train.

This adds an optional recipe_args.target_attn_implementation that, when set, is passed straight through to the target's from_pretrained. Setting it to sdpa sidesteps the broken FA2 path. It defaults to None, so existing configs behave exactly as before. The target is frozen and only produces supervision, so there's no throughput cost to running it on sdpa.

Tested with a 2-GPU Qwen3 EAGLE-3 run: without the knob the run dies in the FA2 path; with target_attn_implementation: sdpa it trains cleanly.


kashif:
cc @khazic

HuiyingLi:
Hi @kashif this PR looks good. Could you please resolve the codecov? Thank you!

HuiyingLi (approved):
LGTM thank you!

PR #2431, opened 2026-06-05
fix: load Qwen2.5-VL vision tower weights under FSDP2 - Do not merge

yuekaizhang:
The PR fixed a bug:

transformers >=5.5 nests the Qwen2.5-VL vision tower under `model.visual.*`, but HF checkpoints store it flat (`visual.*`). get_combined_key_mapping only picked up the transformers WeightRenaming for the language model (`model.*` -> `model.language_model.*`) and had no rule for `visual.*`, so the entire vision tower silently failed to load (set_model_state_dict strict=False) and was left randomly initialized. 

Add a non-destructive `^visual\.` -> `model.visual.` rule for the Qwen VL family (qwen2_5_vl, qwen2_vl), augmenting rather than overriding the transformers language-model renaming.

### DIFF
--- a/nemo_automodel/components/checkpoint/conversion_mapping.py
+++ b/nemo_automodel/components/checkpoint/conversion_mapping.py
@@ -183,6 +183,16 @@ _VLM_KEY_MAPPINGS: dict[str, dict[str, str]] = {
 }
 
 
+# VLM model types whose vision tower is nested under `model.visual.*` in
+# transformers >=5.5, while HF checkpoints store it flat (`visual.*`). The
+# transformers WeightRenaming for these only remaps the language-model keys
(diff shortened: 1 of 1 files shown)


sharonyu-115:
The problem should have been addressed already by transformers 5.6: https://github.com/huggingface/transformers/blob/e933550ec03334b47d2f6f3c92125290c4cd5295/src/transformers/conversion_mapping.py#L134 (PR https://github.com/huggingface/transformers/pull/45358). We ran into this problem with NeMo RL (https://github.com/NVIDIA-NeMo/RL/pull/2224) because NeMo RL pins at an earlier Automodel commit which uses transformers 5.5. Applying a temporary patch in NeMo RL could be a workaround.

PR #2433, opened 2026-06-05
refactor(speculative): reuse shared dflash mask and loss in the trainer

kashif:
The speculative DFlash trainer (`components/speculative/dflash/core.py`) shipped its own copy of the block attention mask and an inline block-wise cross-entropy. Both already exist on the dllm side, so this points the speculative trainer at `components/attention/dflash_mask.py` and `DFlashDecayLoss` and deletes the copies. Net -116 lines.

To keep the speculative trainer's behavior identical, `DFlashDecayLoss` gets a `normalize` option: `"mean"` for its local per-micro-batch decay-weighted mean, `"tokens"` (default) for the dllm recipe's global-token normalization. `loss_gamma=None` now disables decay. The dllm recipe is unchanged.

The speculative trainer also picks up the shared mask's compiled `create_block_mask` cache for free.

Tests: the shared mask test now covers both verticals (dropped the redundant speculative copy), and a numerical check confirms `normalize="mean"` matches the old inline loss for gamma 7 and None. Full speculative + loss + attention suites pass (318). The dllm GPU smoke passes; the speculative GPU smoke is blocked by an unrelated target flash-attention env issue, so it is not exercised here.


kashif:
cc @khazic

kashif:
cc @HuiyingLi

nemo_automodel/recipes/llm/train_dflash.py
```diff
@@ -108,11 +108,16 @@ def setup(self):
         )
         self.compute_dtype = torch.bfloat16 if self.device.type == "cuda" else torch.float32
 
+        target_attn_implementation = recipe_cfg.get("target_attn_implementation", None)
```
khazic, line 111: This `target_attn_implementation` knob is about how the frozen target is loaded, which is orthogonal to the mask + loss sharing this PR is about, so it reads as unrelated scope here. It's harmless. Was it added to unblock the GPU smoke (forcing the target to SDPA)? A one-line note in the PR description, or splitting it into its own small PR, would keep this one focused.

khazic (approved):

khazic:
LGTM, fine to merge. I verified the refactor is behavior-preserving: the shared decay weights align position-for-position with the old inline loss, normalize="mean" reproduces the old denominator, the SDPA mask is semantically equivalent (additive vs boolean), and the CPU forward test exercises the new path. The target_attn_implementation note above is non-blocking.

HuiyingLi (approved):

nemo_automodel/recipes/llm/train_dflash.py
kashif: yes i needed it to test.

PR #2440, opened 2026-06-06
refactor(datasets): unify reasoning_content coercion in agent chat

khazic:
## Summary

The agent-chat reasoning_content support landed across two converters that read and coerce the field inconsistently:

- `_sharegpt_to_chatml` carried `reasoning_content` through verbatim (no coercion).
- `_convert_messages` coerced non-string values to `str` for assistant turns.

So a non-string `reasoning_content` was handled differently depending on the entry point. This PR extracts a single `_reasoning_content` helper that both converters use, so the field is read and coerced identically: non-string values are stringified, and an empty/falsy trace is treated as absent.

No behavior change for the common string case. It also composes cleanly with the existing `mask_reasoning_content` and `drop_history_reasoning_content` paths, which are untouched.

## Testing

Added tests covering coercion through both converters and the empty-as-absent rule; the full `test_agent_chat.py` suite passes (48 tests). ruff clean.


HuiyingLi (approved):
LGTM thank you!

PR #2450, opened 2026-06-08
fix(deepseek_v3): initialize weights in fp32 and default router to fp32

yuhezhang-ai:
## What

**Primary fix: initialize DeepSeek-V3 weights in fp32, then cast to the resident dtype.**

### The issue we observed

From-scratch pretraining of Moonlight-16B (DeepSeek-V3-style MoE) with a bf16 resident dtype (`model.torch_dtype: bfloat16`) **failed to converge**: a huge `grad_norm` already at **step 1**, followed by flat / diverging loss. This reproduced with both torch `AdamW` and TE `FusedAdam`, and with fp32 optimizer states — so it was not an optimizer-state precision problem. The same run with an **fp32 resident dtype converged fine**.

Issue was tracked in linear: https://linear.app/nvidia/issue/AM-396/moonlight16b-moe-model-doesnt-converge-with-modeltorch-typebf16-even

### Root cause

The random init was being **sampled directly in bf16**. bf16's 8-bit mantissa quantizes the small init magnitudes and biases the `normal_` / `trunc_normal_` truncation/scaling arithmetic, distorting the per-tensor variance/mean schedule. In a deep residual stack this compounds into genuinely large first-step gradients. Sampling in fp32 and **rounding** to the resident dtype is an unbiased per-element perturbation that preserves the init statistics — the standard "init in fp32, then cast" practice — while keeping reduced-precision storage.

### Fix

Add an `init_weights_in_fp32(model, resident_dtype)` context manager that casts the model to fp32, runs the init body, and casts back on exit. Wrap DeepSeek-V3's `initialize_weights` in it. No-op when the resident dtype is already fp32. Validated: bf16-resident Moonlight pretraining now converges (matches the fp32-resident run).

### Scope / broader impact

This is **only wired for DeepSeek-V3** here, because it is the repo's main meaningfully-exposed from-scratch MoE pretraining path (the repo is overwhelmingly finetune-oriented; finetuning loads pretrained weights so it is not affected). **Other model families that pretrain from scratch in bf16 are likely affected by the same bf16-init issue** and should adopt the same `init_weights_in_fp32` wrapper in their `initialize_weights` — but auditing/validating every family is a larger scope, so it is intentionally left as a follow-up. The helper is shared and documented in the model-onboarding skill so other models can opt in.

## Secondary change (bookkeeping)

Default DeepSeek-V3's router `gate_precision` to fp32 when unset, to match the HF reference (which scores routing in fp32). Routing is selection-sensitive and the gate is tiny, so this is cheap; it mirrors the existing idiom in `nemotron_v3` / `mimo_v2_flash` / `step3p5`. Note this also makes existing DeepSeek-V3 **finetune** configs route in fp32 (strictly closer to HF) — worth a quick sanity check.

## Changelog

- Add `init_weights_in_fp32` context manager (`components/models/common/utils.py`).
- `deepseek_v3`: wrap `initialize_weights` in `init_weights_in_fp32`; default `gate_precision` to fp32 when unset.
- Document the init-in-fp32 and fp32-router patterns in the model-onboarding MoE skill.
- Add unit tests for the gate-precision default + explicit override.

## Pre-checks

- [x] ruff clean
- [x] unit tests pass (`tests/unit_tests/models/deepseek_v3/test_dsv3_model.py`)
- [x] DCO sign-off

## Validation
Real pretraining short runs before and after the fix:
https://wandb.ai/Nemo-automodel/yuhez_workspace/groups/moonlight16b-dtype-optimizer-10m/workspace?nw=nwuseryuhez
Note that, for the after fix run, the base is this branch, which doesn't include the master wegitht bug fixes for moe, that's why the adamw+f32 converge slower than TE.


nemo_automodel/components/models/common/utils.py
```diff
@@ -456,6 +455,47 @@ def cast_model_to_dtype(model: nn.Module, dtype: torch.dtype = torch.bfloat16) -
             _restore_fp32_modules(model, fp32_keywords)
 
 
+@contextmanager
+def init_weights_in_fp32(model: nn.Module, resident_dtype: torch.dtype):
```
hemildesai: Since this is a general purpose context manager, can we rename it to something like `yield_fp32_model`?

yuhezhang-ai: good point, I renamed it.

yuhezhang-ai:
/ok to test [e0d7461](https://github.com/NVIDIA-NeMo/Automodel/pull/2450/commits/e0d74615dd1b82a130495ec93535e38da48da6cf)

akoumpa (approved):

PR #2458, opened 2026-06-09
feat(speculative): add activation checkpointing for P-EAGLE draft layers

khazic:
## What

Adds an opt-in activation (gradient) checkpointing knob to the EAGLE-3 draft model. When enabled, each P-EAGLE draft layer is recomputed in the backward pass instead of keeping its activations resident, lowering the activation-memory peak of the long flattened COD (chain-of-draft) sequence. It complements `sequence_partitions`: that knob slices the sequence in time (one segment resident at a time), this one trades a recompute for the per-layer activation storage along depth. The two stack.

## Why

P-EAGLE flattens all `num_depths` COD depths into one long parallel-drafting sequence, so the draft forward's activation footprint grows with depth and sequence length. On larger targets / longer sequences this is the memory wall (the draft itself is small; it is the activations that OOM). `sequence_partitions` alone is a single-rank time-slice; activation checkpointing is the orthogonal depth-axis lever and the standard tool for exactly this.

## How

* `LlamaEagle3DraftModel` gains `gradient_checkpointing_enable()` / `gradient_checkpointing_disable()` and a `gradient_checkpointing` flag (off by default, training-only, never serialized into the draft `config.json`, mirroring how `sequence_partitions` is treated).
* `_PeagleDraftMixin.forward_peagle` routes every draft layer through a small `_run_draft_layer` helper that wraps the call in `torch.utils.checkpoint.checkpoint(..., use_reentrant=False)` when the flag is set and the module is training. `use_reentrant=False` is required so the non-tensor `block_mask` passes through and the recompute composes with the flex-attention path.
* Scope: only the parallel-drafting `forward_peagle` path. The standard EAGLE-3 single-layer TTT `forward()` is intentionally left unchanged (one layer is little benefit, and its stateful `cache_hidden` recurrence does not compose with recompute).
* Recipe arg `draft_gradient_checkpointing` wires it through `train_eagle3`.

## Usage

```yaml
recipe_args:
  parallel_drafting: true
  draft_gradient_checkpointing: true   # off by default
```

## Tests

`tests/unit_tests/speculative/test_peagle.py`:
* CPU: flag default/toggle; `_run_draft_layer` recomputes only when enabled AND training (call-count), and is numerically equal (output + input grad) to the direct path; `forward_peagle` routes every layer through the helper (flex-free stubbed layers).
* CUDA-gated: on the real flex-attention forward, checkpointing reproduces the uncheckpointed loss, valid-token count, and per-parameter gradients exactly (up to float summation order).

ruff lint/format clean; `/simplify` applied.

### DIFF
--- a/nemo_automodel/components/speculative/eagle/draft_llama.py
+++ b/nemo_automodel/components/speculative/eagle/draft_llama.py
@@ -665,6 +665,10 @@ class LlamaEagle3DraftModel(_PeagleDraftMixin, PreTrainedModel):
         self.config = config
         self.target_hidden_size = getattr(config, "target_hidden_size", config.hidden_size)
         self.draft_vocab_size = getattr(config, "draft_vocab_size", config.vocab_size)
+        # Activation checkpointing for the P-EAGLE draft layers (training-only
+        # memory knob; toggled via ``gradient_checkpointing_enable`` and read by
+        # ``_PeagleDraftMixin.forward_peagle``). Off by default; never serialized.
+        self.gradient_checkpointing = False
 
         self.model = Eagle3LlamaModel(config)
         self.lm_head = nn.Linear(config.hidden_size, self.draft_vocab_size, bias=False)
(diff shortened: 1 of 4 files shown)


khazic:
Validated on a CUDA GPU, so the flex-attention tests that CI skips on CPU all run and pass.

`pytest tests/unit_tests/speculative/test_peagle.py` → **28 passed**, 0 skipped, ~91s.

New activation-checkpointing tests (the first is CUDA-only and skipped on CPU CI):

* `test_gradient_checkpointing_matches_uncheckpointed_forward` **passed**: on the real compiled flex-attention P-EAGLE forward, checkpointing on vs off yields identical loss, valid-token count, and per-parameter gradients (up to float summation order). This also confirms the recompute composes with the compiled flex-attention path under `use_reentrant=False`.
* `test_run_draft_layer_checkpoints_only_when_enabled_and_training`, `test_run_draft_layer_matches_direct_numerically`, `test_forward_peagle_routes_every_layer_through_checkpoint_helper`, `test_gradient_checkpointing_default_off_and_toggle`: all pass.

The full P-EAGLE suite is green (every CUDA-gated test ran), so there is no regression on the existing parallel-drafting paths.

HuiyingLi (approved):
LGTM

PR #2459, opened 2026-06-09
fix(speculative): retry the remote EAGLE-3 /generate wire path

khazic:
## Problem

`_ServerClient.generate()` is the per-step hot path of remote EAGLE-3 target training, but it issued a single `session.post()` with no retry ([client.py](nemo_automodel/components/speculative/eagle/remote/client.py)). Meanwhile the lower-frequency control endpoints (`init_nccl`, `set_vocab_mapping`, `model_info`, `disconnect`) all go through `request()`, which has exponential-backoff retry. So a single transient timeout or connection reset on `/generate` propagates up and aborts a long remote-training run (nothing in the prefetch pipeline / training loop catches it; `_AsyncHandle.result()` re-raises).

## Fix

Route the **wire-format path** through `request()` so it inherits the retry/backoff. The wire path is an idempotent HTTP round-trip (the target forward is read-only), so retry is safe and correct.

The **NCCL path stays single-shot by design**. The POST triggers a server-side NCCL send paired with the client `recv_tensors`, so a blind retry would issue a second send and desync the 2-process data-plane group (the client's one recv vs the server's two sends would hang). Recovering the NCCL path safely needs a transport resync (tear down + re-init, or fall back to wire with server-side handling of the abandoned send), which is a larger change left as a follow-up. The asymmetry is documented in the method.

## Scope

* In scope: wire-format remote target (the default when sglang is absent on either end, and any NCCL-disabled deployment).
* Follow-up: NCCL data-plane resync/fallback on transient error.

## Tests

`tests/unit_tests/speculative/test_eagle3_remote_coverage.py`:
* `test_generate_wire_path_retries_then_succeeds`: a transient failure on the wire path recovers (post issued twice, tensors decoded correctly).
* `test_generate_wire_path_exhausts_retries`: a persistent failure surfaces as `RuntimeError` after exhausting retries.
* `test_generate_nccl_path_does_not_retry`: the NCCL branch issues the POST exactly once and propagates the error (pins the deliberate no-retry contract so it is not naively "fixed" into a desync).

Full remote suite green (39 passed); ruff clean; `/simplify` applied (no changes needed).


HuiyingLi (dismissed):
LGTM thank you

HuiyingLi (approved):

PR #2465, opened 2026-06-09
feat(speculative): add context parallelism for the EAGLE-3 target model

khazic:
## What

Adds Context Parallelism (CP) for the **colocated EAGLE-3 target model**. The frozen
target's forward is sharded along the sequence dimension across the `cp` device submesh
(ring attention), and the aux hidden states / logits are gathered back to the full
sequence before the draft consumes them, so the draft, TTT loop, loss, and the
offline-cache / remote paths stay completely CP-unaware. This lets long-sequence target
attention scale across `cp_size` GPUs.

## Why

In EAGLE-3 the frozen target is the large model and its `O(T^2)` attention is the
long-sequence memory/compute bottleneck. CP shards that work across `cp_size` ranks while
keeping the small draft on the full sequence.

## Design

CP is an internal optimization of the target forward, encapsulated in
`HFEagle3TargetModel.generate_batch` (`make_target_cp_ctx` / `gather_cp_seq` in
`cp_utils.py`, built on torch's `context_parallel` + `context_parallel_unshard`,
load-balancing disabled for a deterministic ordered gather).

Because the recipe previously assumed `world == dp`, three call sites are made cp-aware:

* the dataloader keys its `DistributedSampler` on the **dp sub-axis** so the `cp_size` ranks
  in a dp group receive the same sample (CP shards its sequence across them);
* the draft DDP gradient all-reduce uses the **dp process group**, not the full world;
* the checkpointer shards by **dp position** instead of global rank.

`cp_size == 1` degrades to today's exact behavior. CP with sequence packing and CP with the
remote/sglang target backend are gated off with a clear `NotImplementedError`.

## Validation

Three layers, the last two on 2 GPUs:

**1. CPU unit tests** cover the gates, the target wrapper's `cp_mesh` / hook handling, the
dp-axis sampler (including `dp_size == 1`), and the helpers.

**2. Operator parity** (`scripts/verify_eagle3_target_cp_parity.py`): the CP gathered
aux/logits vs a non-CP reference on identical weights/inputs:

| tensor | max_diff (bf16) | cosine |
|---|---|---|
| `aux_hidden_states` | 7.8e-03 | 0.999991 |
| `logits` | 1.2e-02 | 0.999985 |

`input_ids` is verified uncorrupted by the in-place CP sharding.

**3. End-to-end loss parity**: real EAGLE-3 recipe, Qwen3-4B target + PerfectBlend data,
`cp_size=2` vs `cp_size=1`, dp fixed at 1, shuffle disabled (same data order), 64 steps.
Per-step `train_loss` tracks within bf16 / ring-attention noise and does not diverge:

| step | cp=1 | cp=2 | \|diff\| |
|---|---|---|---|
| 1 | 10.926428 | 10.927055 | 0.0006 |
| 5 | 9.362926 | 9.425581 | 0.063 |
| 27 | 5.204263 | 5.204389 | 0.0001 |
| 37 | 4.738049 | 4.898445 | 0.160 |
| 50 | 6.180621 | 6.218976 | 0.038 |
| 64 | 6.753325 | 6.813065 | 0.060 |

The end-to-end run also caught and fixed a real integration bug: under pure CP
(`dp_size == 1`) the dataloader was splitting the batch across the cp ranks instead of
feeding them the same shared sequence.

## Out of scope (follow-ups)

Draft-side CP (TTT diagonal-cache / P-EAGLE flex / DFlash), CP + sequence packing (TE THD),
CP + remote/sglang target, and CP + MoE/EP target.


HuiyingLi (dismissed):
LGTM thank you!

HuiyingLi (approved):

PR #2470, opened 2026-06-09
feat(speculative): add EAGLE feature-noise augmentation to EAGLE-1/2 training

khazic:
## What

Adds the EAGLE **feature-noise data augmentation** to EAGLE-1/2 training, which the recipe
was missing.

## Why — paper reference

The EAGLE paper trains the draft with uniform noise added to the target features to mitigate
the error accumulation that compounds across the autoregressive drafting steps. From the
original paper:

> "we employ data augmentation by adding random noise sampled from a uniform distribution
> 𝒰(−0.1, 0.1) to features of the target LLM during training."

— EAGLE (Li et al., 2024), [arXiv:2401.15077](https://arxiv.org/abs/2401.15077).

Automodel's `EagleTrainerModule` (`components/speculative/eagle/core_v12.py`) fed the target
features straight into the draft with no noise, so EAGLE-1/2 training did **not** reproduce
the paper. The rest of the objective (SmoothL1 on hidden states + soft cross-entropy on
tokens, shifted-token targets) was already aligned; this closes the remaining gap.

## How

- `EagleTrainerModule` gains a `feature_noise` knob (the symmetric magnitude). In `forward`,
  **training-only**, it adds `U(-feature_noise, feature_noise)` to the draft's **input**
  features (`input_hidden_states`). The SmoothL1 regression target (`target_hidden_states`)
  is left clean. It is a no-op in eval (`self.training` is False) and at `feature_noise == 0`.
  `(rand_like - 0.5) * 2 * feature_noise` is exactly `U(-feature_noise, feature_noise)`.
- The EAGLE-1/2 recipe (`train_eagle1.py`, reused by `train_eagle2.py`) reads
  `recipe_args.feature_noise`, **defaulting to `0.1` to match the paper** (set `0` to disable).

## Test

`tests/unit_tests/speculative/test_eagle12_feature_noise.py`:
- `feature_noise=0` leaves the draft input clean;
- in training the input is perturbed and every element stays within `U(-fn, fn)`;
- in eval no noise is applied even with `feature_noise>0`.

The existing EAGLE-1/2 trainer tests (which construct the module without `feature_noise`,
so default `0.0`) are unaffected.

### DIFF
--- a/nemo_automodel/components/speculative/eagle/core_v12.py
+++ b/nemo_automodel/components/speculative/eagle/core_v12.py
@@ -46,6 +46,7 @@ class EagleTrainerModule(nn.Module):
         target_lm_head: nn.Module,
         hidden_loss_weight: float = 1.0,
         token_loss_weight: float = 0.1,
+        feature_noise: float = 0.0,
     ):
         super().__init__()
         self.draft_model = draft_model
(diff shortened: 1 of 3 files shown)


HuiyingLi (approved):
Thank you for aligning the implementation details with the paper.

PR #2482, opened 2026-06-09
fix: resolve nightly CI failures (FP8, ckpt, gemma3n, benchmark)

akoumpa:
## Summary

Fixes the **akoumpa-owned failures** in nightly CI [pipeline 54175836](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/54175836) (2026-06-09): 8 LLM/VLM finetune tests across 5 root causes, plus the 2 benchmark "empty pipeline" CI-infra failures. The failing container runs **transformers 5.8.1**.

Each fix was reproduced and confirmed on an H100 (cw-dfw) node (nightly image + `transformers==5.8.1`, patched code mounted at `/opt/Automodel`), demonstrating the failure on `main` and a pass with the fix. Unit tests added per fix; the repo ruff hook passes on every commit. One commit per root cause. The fixes are scoped to the model component / recipe wherever possible (no generic `_transformers/` changes).

### 1. `fix(mistral3)`: per-tensor FP8 Mistral3/Ministral checkpoints
**Failing jobs:** `devstral2_small_2512_squad`, `ministral3_3b_squad`, `ministral3_3b_squad_peft`
**Issue:** Devstral-24B and Ministral-3B are the same architecture (`Mistral3ForConditionalGeneration`, `ministral3` backbone, **per-tensor** FP8 → `weight_block_size: null`, 0-D scalar `weight_scale_inv`). transformers 5.8.1's `Fp8Dequantize._dequantize_one` does `scale_rows, scale_cols = scales.shape[-2:]`, which raises on a scalar → `RuntimeError: ...issues during automatic conversion of the weights...`. Automodel already ships the correct model-scoped path (`Mistral3FP8VLMForConditionalGeneration` + `Mistral3FP8StateDictAdapter` + a resolver hook), but the recipes **over-specified** (`force_hf: true` + an explicit `FineGrainedFP8Config(dequantize=true)`), diverting the load to HF's loader.
**Fix (no generic-code / monkeypatch changes):** drop the explicit `quantization_config` from both ministral recipes and drop `force_hf` + `quantization_config` from both devstral recipes, so all four route to `Mistral3FP8VLMForConditionalGeneration` via the resolver hook (`apply_model_infrastructure` auto-enables `dequantize_base_checkpoint` from `config.json`); and make the `mistral3_vlm` resolver hook lightweight (import-free claim predicate + lazy model import). `_transformers/auto_model.py` & `model_init.py` are unchanged from `main`.
**Confirmed (Slurm 12679914):** before → all three FAIL with the exact CI error; after → all load as `Mistral3FP8VLMForConditionalGeneration` (Devstral 24.0B, Ministral 3.85B SFT, Ministral PEFT + 366 LoRA tensors).

### 2. `fix(distributed)`: custom-code architectures with TP at `tp_size>1`
**Failing jobs:** `llama3_3_nemotron_super_49B_squad` (tp=4) + `nemotron_flash_1b_squad_peft` (tp=2), and their cascading `*_vllm_deploy` stages.
**Issue:** both raise `No tensor-parallel plan is registered for the custom-code architecture '<…>ForCausalLM'` at parallelize — the archs aren't registered and the base-plan fall-through fail-fasts for `trust_remote_code` at tp>1.
**Fix:** register a `DeciLMForCausalLM` plan in `optimized_tp_plans.py` (reuses `get_decilm_nemotron_tp_plan`); downgrade the fail-fast in `parallelizer.py` to a warning + graceful fall-through; make the PEFT reload in `stateful_wrappers.py` reshard via `distribute_tensor` (robust to FSDP2's `(_StridedShard, Shard)` composite, `shard_order=None`).
**Confirmed:** DeciLM resolves a 9-entry plan at tp=4 on the **real 49B**; an unregistered remote-code arch at tp=2 completes a real fwd+bwd+optimizer step via the base plan (no `shard_order` assert). *Caveat: real `Nemotron-Flash-1B` wouldn't build under the available image (`huggingface_hub` rejects its `layer_types`), so NemotronFlash was confirmed via a faithful synthetic stand-in. NemotronFlash GLA/FFN stay replicated at tp>1 — correct & crash-free; a tuned plan is an optional perf follow-up.*

### 3. `fix(checkpoint)`: transitive custom-code imports in consolidated checkpoints
**Failing jobs:** `nemotron_flash_1b_squad` + its `_vllm_deploy`
**Issue:** consolidated-checkpoint reload fails with `FileNotFoundError: .../consolidated/<hash>/triton_attention.py`. The aux-code copy only grabbed the HF dynamic-module cache (modeling file + its **direct** imports), missing NemotronFlash's **transitively**-imported `triton_attention.py`.
**Fix:** `addons.py` copies the full transitive relative-import closure (and uses the model's HF snapshot dir as a source) so a `trust_remote_code` consolidated checkpoint is self-contained on reload.
**Confirmed:** before → missing-file error (matching hash `980dc150…`); after → reload succeeds.

### 4. `fix(distributed)`: gemma3n KV-sharing under FSDP2
**Failing job:** `gemma3n_vl_4b_medpix` (VLM SFT)
**Issue:** first forward dies with `KeyError: 18` at `modeling_gemma3n.py shared_kv_states[self.kv_shared_layer_index]`. FSDP2's `cast_forward_inputs` rebuilds dict kwargs per layer (`_apply_to_tensors`), so each decoder layer gets its own throwaway `shared_kv_states` and the producer layer's KV writes are discarded.
**Fix:** `activation_checkpointing.py` adds `stabilize_kv_sharing_across_fsdp()` (folds the per-layer rebuilt dict into one canonical dict, cleared per step), wired into `DefaultParallelizationStrategy.parallelize`. No-op when the kwarg is absent.
**Confirmed:** 2×H100 — before → `KeyError`; after → trains (step 0 loss 3.03 + validation).

### 5. `fix(peft)`: gemma3n per-layer LoRA output safe for in-place use (model-scoped)
**Failing job:** `gemma3n_vl_4b_medpix_peft` (VLM PEFT)
**Issue:** backward raises `Output 0 of LoRATritonFunctionBackward is a view and is being modified inplace`. gemma3n's `project_per_layer_inputs` does an in-place op on the output of `per_layer_model_projection`, which under the memory-efficient LoRA path is a view of a custom-Function output.
**Fix (model-scoped, not a blanket clone):** `_peft/lora.py` adds `patch_gemma3n_inplace_lora_views()` (called from `apply_lora_to_linear_modules`) that structurally detects gemma3n's text model and wraps **only** its `per_layer_model_projection` LoRA forward to return a non-view clone — one clone per forward instead of ~200+, and **zero cost for non-gemma3n LoRA** (the previous blanket `.clone()` in `LoRATritonFunction` is removed). `clone()` is an autograd identity; guarded + idempotent.
**Confirmed:** before (model-side patch + clone both removed) → exact view+inplace `RuntimeError`; after (model-side only) → fwd+bwd complete, finite grads. *(torch-2.12-specific failure — the HPC run was load-bearing.)*

### 6. `fix(ci)`: valid pipeline when a test folder has no nightly recipes
**Failing jobs:** `automodel_llm_benchmark_tests`, `automodel_vlm_benchmark_tests`
**Issue:** `llm_benchmark`/`vlm_benchmark` have no `configs/<folder>/nightly_recipes.yml`, so `generate_ci_tests.py` produced a child pipeline with only an `include:` and **0 jobs** → GitLab rejects it (`jobs config should contain at least one visible job`).
**Fix:** `generate_ci_tests.py` emits a single no-op placeholder job (`<folder>_no_nightly_recipes`) when a folder/scope yields zero visible jobs. Non-empty folders unchanged.
**Confirmed:** generator `0 → 1 job` for both; no regression across all (folder, scope) combos. *Caveats: live `ci/lint` not run (token lacked `api` scope); the placeholder job is untagged — if nemo-ci requires tagged runners, a tag may be needed.*

## Out of scope
The two retrieval-eval failures (`llama3_2_1b_rerank_eval`, `llama3_2_1b_embed_eval`) are `oliverholworthy`'s recipes — small golden-value drift (one is an *improvement* tripping the symmetric ±0.01 tolerance) — left for that owner to triage (investigate whether the down-drift is a real regression before re-baselining).


pthombre (approved):

akoumpa:
stale

PR #2486, opened 2026-06-10
feat(moe): MTP FLOPs accounting, inline shared experts, checkpoint warning fix

adil-a:
## What

Four related framework changes extracted from MoE benchmarking work:

1. **MTP / repeated-layer FLOPs accounting.** Adds `_nemotronh_mtp_flops` (handles Multi-Token-Prediction heads, including the repeated-layer "duplication" used by Nemotron-3 Ultra) and wires it into the benchmark recipe (`_mtp_tflops`) so reported MFU includes the MTP head. New unit test at `tests/unit_tests/utils/test_mtp_flops.py`.
2. **Inline MoE shared experts.** Runs shared experts inline on the main stream and removes the side-stream overlap path, which raced under expert parallelism and corrupted gradients (grad-norm explosion at higher tokens/microbatch). Trades a small overlap speedup for correctness.
3. **Weight-tied MTP head FSDP fix.** For repeated-layer (`use_repeated_layer`) MTP heads, skip activation checkpointing on the head blocks and keep their EP-sharded experts un-resharded (`reshard_after_forward=False`). FSDP2 cannot re-unshard the shared experts param group on the 2nd+ recompute/use, which otherwise makes `grouped_gemm` read a resharded `Shard(1)` weight and raise a shape error. The head is tiny, so the memory cost is negligible; non-tied heads are unaffected.
4. **Checkpoint warning fix.** Excludes view-loaded native MoE expert keys from the load-time key-mismatch warning (cosmetic; the model still loads correctly).

## Changelog
- feat: MTP/repeated-layer FLOPs in `flops_utils` + benchmark recipe MFU
- fix(moe): run shared experts inline (remove racy side-stream overlap)
- fix(moe): skip AC and keep experts un-resharded for weight-tied MTP heads
- fix(checkpoint): suppress spurious key-mismatch warning for view-loaded MoE experts
- test: unit tests for MTP FLOPs

## Pre-checks
- [x] `ruff check` passes on changed files
- [x] `py_compile` clean
- [x] DCO sign-off


nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -2080,6 +2086,9 @@ def _maybe_adapt_state_dict_from_hf(
     """
     adapter = getattr(model_part, "state_dict_adapter", None)
     if adapter:
+        # Reset the per-load record of in-place (view) loaded keys so the caller's key-diff
+        # reflects only this load; MoE adapters repopulate it during from_hf.
+        adapter._view_loaded_native_keys = set()
         ep_mesh_dims = [dim for dim in moe_mesh.mesh_dim_names if dim != "pp"] if moe_mesh is not None else []
```
akoumpa: is it possible to keep all changes to the adatper?

adil-a: Good call — done in 1599a6cd5. The whole view-loaded-key lifecycle now lives in the adapter:

- `_from_hf_w_merged_experts` resets the record itself via a new `reset_view_loaded_keys` param (default `True`). `nemotron_v3.from_hf` passes `False` on its second (MTP) merge so the backbone + MTP view-loaded keys accumulate across one logical `from_hf`.
- The mixin exposes a public `view_loaded_native_keys` property.
- `checkpointing.py` no longer touches the adapter's private state: the explicit `adapter._view_loaded_native_keys = set()` reset is gone, and the key-diff just reads the public property — the only remaining touch-point, since the key-diff itself is computed loader-side.

Resetting per-`from_hf` also closes a latent cross-load leak (an init-time partial load could previously bleed view keys into a later load's key-diff). Verified with the mixin/adapter/checkpoint unit tests (484 passing) and an adversarial trace of the init-load and backbone+MTP paths.

akoumpa (approved):

PR #2551, opened 2026-06-13
feat: CP support for MiniMax M3

### LINKED ISSUE #2617: MiniMax-M3 sparse attention: SDPA crashes with "invalid dtype for bias"
## Summary

For MiniMax-M3 models with sparse-attention layers, the indexer produces an additive
attention bias in **float32**, and for sparse layers this bias is passed to
`F.scaled_dot_product_attention` as `attn_mask` **without matching the query dtype** (bf16).
cuDNN's SDPA backend tolerates this dtype mismatch, so standard SFT works by default.

 However, this issue occurs when I rebase AutoModel and run training in NeMo-RL.
*nemo-rl globally disables the cuDNN SDPA backend —
`torch.backends.cuda.enable_cudnn_sdp(False)` — for activation-checkpointing recompute
correctness. With cuDNN off, PyTorch selects the **memory-efficient SDPA backend**, which
enforces `attn_mask.dtype == query.dtype` and raises on the float32 bias:

```
RuntimeError: invalid dtype for bias - should match query's dtype
```

## Environment

- **Automodel: rebase with main branch, commit id:** `2d946b00286213448a15f1814c0cd7de76457d87`
- **Model:** MiniMax-M3 VL
- **Backend:** `backend.attn: sdpa`

## What happens

The model forward crashes inside `F.scaled_dot_product_attention` — **only when the cuDNN
SDPA backend is unavailable/disabled**. With cuDNN enabled (the default), the identical
model and config train fine. This makes the bug backend-dependent and easy to miss: a
plain SFT run passes, while any harness that disables cuDNN SDPA fails at the first forward.

A common way to hit this in practice: a trainer that disables cuDNN SDPA for
activation-checkpointing recompute correctness (cuDNN may not be re-selected during
recomputation, which otherwise raises *"Recomputed values have different metadata than
during forward pass"*), e.g. `torch.backends.cuda.enable_cudnn_sdp(False)`.

## Root cause

The sparse bias is constructed in a **hardcoded `torch.float32`** and is never returned in the
model's compute (activation) dtype.

`nemo_automodel/components/models/minimax_m3_vl/layers.py`, `build_block_sparse_attn_bias`:

```python
q = idx_q.permute(0, 2, 1, 3).float()      # index scores upcast to fp32 (numerically sensitive — fine)
...
bias = torch.where(key_sel, 0.0, neg_inf).to(torch.float32)   # <-- final bias HARDCODED fp32
```

## Full-model reproduction

Run any MiniMax-M3 (sparse-attention) recipe with `backend.attn: sdpa`, `bfloat16`, and the
cuDNN SDPA backend disabled.

## Traceback

```
File "nemo_automodel/recipes/vlm/finetune.py", line 942, in _forward_backward_step
    out = model(**batch)
File "nemo_automodel/components/models/minimax_m3_vl/model.py", line 638, in forward
    hidden = self.model(None, ...)
File "nemo_automodel/components/models/minimax_m3_vl/model.py", line 186, in forward
    h = layer(x=h, ...)
File ".../torch/distributed/algorithms/_checkpoint/checkpoint_wrapper.py", line 169, in forward
File "nemo_automodel/components/models/minimax_m3_vl/layers.py", line 471, in forward
    attn_out = self.self_attn(x=self.input_layernorm(x), ...)
File "nemo_automodel/components/models/minimax_m3_vl/layers.py", line 392, in forward
    out = self.attn_func(q, k, v, **_attn_kwargs)
File "nemo_automodel/components/attention/utils.py", line 77, in attn_func
    return F.scaled_dot_product_attention(q, k, v, attn_mask=attn_mask, ..., enable_gqa=enable_gqa)
RuntimeError: invalid dtype for bias - should match query's dtype
```

## Affected code

- **Primary** — `nemo_automodel/components/models/minimax_m3_vl/layers.py`,
`build_block_sparse_attn_bias`: the bias is built as `float32` and assigned to `attention_mask`
for sparse layers (`MiniMaxM3Attention.forward`).
- **Secondary** — `nemo_automodel/components/attention/utils.py`,
`preprocess_args_and_kwargs_for_attn` (`explicit_mask` branch): moves the mask to device but does not
align its dtype to the query.

## Proposed fix

**Primary (root cause): emit the bias in the activation dtype at the source**, in
`build_block_sparse_attn_bias` (`nemo_automodel/components/models/minimax_m3_vl/layers.py`):

```python
- bias = torch.where(key_sel, 0.0, neg_inf).to(torch.float32)
+ bias = torch.where(key_sel, 0.0, neg_inf).to(idx_q.dtype)   # match the model's compute dtype
```

`idx_q` shares the activation dtype of the main attention query, so the bias now matches `q` for  
*every* consumer (SDPA, TE `core_attention_bias`, flex, future backends)

athitten:
# What does this PR do ?

Adds **context-parallel (CP) support to MiniMax M3 VL** so it can train at long context by sharding the sequence across ranks — including the model's **block-sparse DSA attention**. Sparse layers use a CP-aware FlexAttention path (all-gather K/V across the CP group → reconstruct global token order → global block selection → block-sparse `BlockMask`); dense layers use the standard DTensor-SDPA CP path. The PR also fixes a bf16 masking bug in the **regular (non-CP) sparse path**. As part of this correction the issue mentioned in #2617 should also be resolved.

These are the final and correct SFT and LoRA loss curves for Minimax M3: [SFT loss curve](https://wandb.ai/nvidia-nemo-fw-public/Automodel/runs/vnbok2ch?nw=nwuserathittenaman) and [LoRA loss curve](https://wandb.ai/nvidia-nemo-fw-public/Automodel/runs/2ss5w1pc?nw=nwuserathittenaman
)
# Changelog

- **CP-aware sparse attention** (`MiniMaxM3CPSparseAttention`): all-gathers K/V
  across the CP group, reconstructs the load-balanced (`2*cp`) global token order,
  runs global block selection, and attends with FlexAttention over a block-sparse
  `BlockMask`. Dense layers keep the standard DTensor-SDPA CP path.
- **VLM CP pre-embed** (`prepare_model_inputs_for_cp` + `_pre_embed_only`): splices
  vision features into `inputs_embeds` before sequence sharding; returns a
  **detached** `inputs_embeds` (torch `context_parallel` shards buffers in place and
  rejects grad-requiring tensors).
- **Parallelizer wiring**: `apply_cp` attaches the CP mesh to each sparse `self_attn`
  post-FSDP.
- Declare **`supports_cp=True`** in `ModelCapabilities`.
- **bf16-safety fix (non-CP path)**: the regular sparse path built an additive `-inf`
  SDPA bias, which leaks past the causal/sparse mask under bf16. Switched to a
  **boolean keep-mask** (`build_block_sparse_attn_mask`, `_padding_mask_to_keep_mask`,
  combined via boolean AND) — matching the CP FlexAttention masking.
- **Packed-sequence (no-CP) fix**: derive a 4-D-aware `[B, T]` MoE `padding_mask`
  from the packed block-causal mask.
- **Example CP recipes**: `cp2_medpix_2k` (cp1-vs-cp2 real-data parity) and
  `tulu3_text_cp8_16k` (real-data long-context, sequence-packed 16k @ cp8).

# Testing and Validation

**Unit tests** (`tests/unit_tests/models/minimax_m3_vl/`, all passing):
- `test_cp_permutation_inverse` — load-balanced shard reorder round-trips to identity.
- `test_cp_sparse_selection` — CP local-query block selection matches the eager selection.
- `test_cp_packing_docmask` — per-document (doc-id) masking for packed sequences.
- `test_cp_forward_equivalence` — `cp_size==1` CP path matches the eager forward.
- `test_minimax_m3_sparse` — block-sparse indexer/selection, padding combine, **plus a
  CUDA `bf16`-vs-`fp32` regression test** that pins the boolean-mask property so the
  additive-`-inf` leak can't silently return.
- Existing M3 coverage (`model`, `mtp`, `pp`, `vlm`, `state_dict_adapter`, `processing`).

**Functional CP forward-equivalence** (`tests/functional_tests/context_parallel/`, 2 GPUs):
`run_minimax_m3_sparse_cp.py` (isolated sparse layer) and
`run_minimax_m3_full_model_cp.py` (full model via the real `make_cp_batch_and_ctx`
sharding) assert the CP forward matches the eager full-sequence forward at real
`cp_size>=2` — exact in fp32, within rounding in bf16.

**Real-data parity (MedPix-VQA, 16 nodes / 128 GPUs)**: 
A matched-regime `cp_size=1` vs `cp_size=2` comparison (unpacked, `max_length=2048`, same global
batch and seed) using `examples/vlm_finetune/minimax_m3/minimax_m3_vl_sft_ep32pp4.yaml` and `examples/vlm_finetune/minimax_m3/minimax_m3_vl_sft_cp2_medpix_2k.yaml` respectively. The loss curves **match step-for-step within bf16 noise** (e.g.
step 0 ≈ 2.03 / grad_norm ≈ 3.4 for both; tracking together throughout), confirming
CP is numerically transparent for MiniMax M3. Loss curves below:
Screenshot 2026-06-24 at 4 13 35 PM

**Real-data long-context (Tulu-3 SFT, 16 nodes / 128 GPUs)**: 
Sequence-packed to 16k with `cp_size=8` (`pp4*dp4*cp8*tp1=128`, `ep32`), using
`examples/vlm_finetune/minimax_m3/minimax_m3_vl_sft_tulu3_text_cp8_16k.yaml`.
Unlike the short MedPix parity run (mostly padding at long seqlen), packed Tulu-3
fills ~126k–149k real supervised tokens/batch, so the block-sparse DSA CP path is
exercised across a genuinely long sequence (cp8 shards 16384 → 2048 tokens/rank).
Trains cleanly end-to-end with a descending, converging loss (loss ≈ 0.82 → ~0.57
over the first ~10 steps, then plateauing as the strong base model converges on
in-distribution data; grad_norm 3.4 → 0.14; ~875 tok/s/GPU). Loss curve below:
Screenshot 2026-06-22 at 6 49 25 PM

**Multi-node scale (mock-data dev smoke runs)**: 
During development, mock-data CP runs exercised cp4/pp4/ep32 at 128 GPUs (incl. sequence-packed) and cp16 at 256 GPUs / 128k context, all training cleanly end-to-end. Those dev smoke recipes have
since been consolidated into the real-data recipes above.

**Pre checks**:

# Additional Information

- Builds on the MiniMax M3 VL model PR (#2538).


examples/vlm_finetune/minimax_m3/minimax_m3_vl_sft_cp2_medpix_2k.yaml
```diff
@@ -0,0 +1,140 @@
...
+    # The bf16 checkpoint ships no MTP weights, so this must stay 0.
+    num_mtp_modules: 0
+  backend:
+    _target_: nemo_automodel.components.models.common.BackendConfig
+    attn: sdpa
+    linear: te
+    rms_norm: torch_fp32
+    rope_fusion: false
+    experts: gmm
+    dispatcher: deepep
```
akoumpa: ep_size > 8 requires hybridep @athitten

athitten: > ep_size > 8 requires hybridep @athitten

@akoumpa the recipes are updated and validated with hybridep. The SFT loss curves are shared above. This is the LoRA one:

Screenshot 2026-06-24 at 4 15 09 PM

nemo_automodel/components/moe/parallelizer.py
```diff
@@ -581,7 +581,18 @@ def apply_cp(model: torch.nn.Module, cp_mesh: DeviceMesh, cp_comm_type: str = "p
         layer_type = getattr(block, "layer_type", getattr(block, "attention_type", "full_attention"))
 
         if layer_type in ("full_attention", "sliding_attention"):
-            attn_module = getattr(block.self_attn, "attn_module", None)
+            # CP-aware custom attention (e.g. MiniMax M3 block-sparse DSA, which
```
HuiyingLi: Is it possible to use the setup_cp_attention below?

nemo_automodel/recipes/vlm/finetune.py
```diff
@@ -389,7 +389,17 @@ def build_dataloader(
...
+                # sdpa/eager but keeps a cheap indexed [B, S] mask for flash_attention_2.
+                # At long context (e.g. 128k) the dense mask is ~S^2 bytes/sample and
+                # OOMs the dataloader workers. ``packed_sequence.attn_implementation``
+                # overrides the mask form: set it to "flash_attention_2" for
+                # context-parallel runs, where the attention mask is stripped before the
+                # model anyway (document boundaries are recovered from position_ids), so
+                # the indexed form is both sufficient and orders of magnitude cheaper.
+                # Attn computation still uses sdpa, the attn_implementation attribute in packing_cfg
+                # is only used to switch the mask format to [B, S] when cp>1.
+                _attn_impl = packing_cfg.get("attn_implementation", None) or get_attn_implementation(cfg_model)
```
HuiyingLi: What if for some models pack.attn is different from model.attn?

nemo_automodel/components/moe/parallelizer.py
athitten: Yes that is possible and is cleaner, thanks for the suggestion. Let me add it.

nemo_automodel/recipes/vlm/finetune.py
athitten: Yeah thats a good point, thanks @HuiyingLi . We can honor the packing override when cp_size > 1, else use the model's attn impl (or assert). wdyt ?

nemo_automodel/components/moe/parallelizer.py
athitten: Addressed in 07bcae49aa3f4b42e7ae369a69e257a3d44b92b0

nemo_automodel/recipes/vlm/finetune.py
athitten: Addressed in 4f3def594ef2fc7619d7bc499044acc8f5ff4a86

examples/vlm_finetune/minimax_m3/minimax_m3_vl_sft_tulu3_text_cp8_16k.yaml
```diff
@@ -0,0 +1,162 @@
...
+# Pack full Tulu-3 conversations to 16k. pretokenize uses M3's processor (chat template
+# + assistant-only label masking); no Gemma-specific thinking-prefix hook. 16384 is
+# divisible by 2*cp=16.
+packed_sequence:
+  pretokenize: true
+  max_length: 16384
+  pack_size: 16384
+  collate_max_length: 16384
+  packing_ratio: 0.9
+  drop_long_samples: true
```
HuiyingLi, line 133: Does ti need attn?

athitten: Yes good find. Dint realize it uses the default [B, 1, S, S]. It works at 16k but longer we might need [B, S]. Addressed in a0f247f3f09fd6302867c0094bd5b22b0580640c

HuiyingLi (approved):

PR #2559, opened 2026-06-15
test: fix all 5 vllm_deploy tests (token drift, nemotron OOM + mamba merge)

adil-a:
## What
Fixes **all 5** failing `*_vllm_deploy` checkpoint-robustness CI tests (consolidates #2560-#2563, which this PR subsumes):

| Test (job) | Failure |
|---|---|
| gemma-3-270m PEFT ([663](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/337980663)) | HF↔vLLM token mismatch |
| gemma-3-270m SFT ([589](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/337980589)) | HF↔vLLM token mismatch |
| llama-3.2-3b PEFT ([665](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/337980665)) | HF↔vLLM token mismatch |
| qwen2.5-7b PEFT ([670](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/337980670)) | HF↔vLLM token mismatch |
| nemotron-nano-9b PEFT ([666](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/337980666)) | vLLM OOM + mamba LoRA |

## Root causes & fixes
1. **Brittle assertion (all 5):** exact HF-vs-vLLM greedy token equality is not a valid cross-engine invariant — eager HF and compiled vLLM flip argmax on near-tied logits and diverge mid-sequence (CI showed divergence at tokens 12/12/13/42, never at the start). → require a matching **prefix** (`MIN_MATCH_PREFIX=5`): catches a broken checkpoint load (which corrupts token 0), tolerant of late fp drift.
2. **nemotron OOM:** the `PeftModel`↔base reference cycle keeps the HF weights resident, so vLLM (separate process) can't get enough free GPU memory. → `gc.collect()` before `empty_cache()` + `gpu_memory_utilization=0.7`.
3. **nemotron mamba LoRA:** vLLM can't serve LoRA on NemotronH's fused `MambaMixer2` (`conv1d`), independent of the adapter. → opt-in `ci.checkpoint_robustness.vllm_merge_lora` merges the adapter and deploys the merged model without `enable_lora`.

## Changelog
[...]

PR #2563, opened 2026-06-15
test: tolerate HF/vLLM token drift in qwen2.5-7b PEFT vllm_deploy

adil-a:
## What
Fixes the `qwen2_5_7b_squad_peft_vllm_deploy` CI test ([job 337980670](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/337980670)).

## Root cause
The test asserts **exact** HF-vs-vLLM greedy token equality, which is not a valid cross-engine invariant. HF runs eager bf16; vLLM runs its own compiled / CUDA-graph / fused-attention kernels and can pick a different argmax on near-tied logits, after which the autoregressive sequences diverge and never re-converge. In this job they diverge at token 42 — far from the start. A checkpoint that loaded *incorrectly* would instead diverge at the very first token.

## Changelog
- Replace full-sequence equality with a matching-prefix check (`MIN_MATCH_PREFIX=5`): catches a broken checkpoint load (diverges at token 0) while tolerating expected late fp drift between engines.

## Validation
- Targeted GitLab pipeline (release scope): https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/54775707
- Comparison logic unit-checked against the real CI divergence tokens from this job.

## Pre-checks
- `ruff format` / `ruff check` clean; DCO signed off.

> The prefix-comparison change is shared with the sibling vllm_deploy fixes (nemotron-9b, gemma PEFT/SFT, llama-3.2-3b, qwen2.5-7b); the diff is identical across these PRs, so merging any one resolves the shared assertion.

### DIFF
--- a/tests/functional_tests/checkpoint_robustness/test_checkpoint_vllm_deploy.py
+++ b/tests/functional_tests/checkpoint_robustness/test_checkpoint_vllm_deploy.py
@@ -236,15 +236,26 @@ def test_vllm_greedy_matches_hf():
         vllm_outputs.append(tokens)
         print(f"[vLLM] Prompt {idx}: generated {len(tokens)} tokens")
 
-    # Token-for-token comparison
+    # HF (eager) and vLLM (compiled) can flip argmax on near-tied logits and diverge
+    # mid-sequence; a broken checkpoint load instead diverges at the first token. So
+    # require a matching prefix, not full equality.
+    MIN_MATCH_PREFIX = 5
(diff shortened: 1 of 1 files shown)


adil-a:
Consolidated into #2559, which carries the identical comparison fix plus the nemotron-specific changes and is validated by pipeline 54861165 (all 5 vllm_deploy jobs green). Closing in favor of the single PR.

PR #2572, opened 2026-06-15
feat(dflash): add dpace loss

kashif:
Adds D-PACE loss variants to the shared DFlash loss and threads the knobs through both DFlash training paths.

Also adds focused unit coverage for block-local D-PACE weights, fused CE parity, and recipe/strategy wiring.

Tests:
- python -m pytest tests/unit_tests/loss/test_dllm_loss.py tests/unit_tests/speculative/test_dflash_core.py tests/unit_tests/recipes/dllm/test_strategy.py tests/unit_tests/recipes/llm/test_train_dflash.py -q
- srun -p hpc-mid --gres=gpu:h100:1 --ntasks=1 --cpus-per-task=4 --time=00:10:00 python -m pytest tests/unit_tests/attention/test_dflash_mask.py::test_flex_block_mask_matches_reference -q


docs/guides/dllm/finetune.mdx
jgerh:
```suggestion
- **LLaDA/LLaDA2 (MDLM):** Bidirectional masked diffusion. The model receives corrupted tokens and predicts the clean token at each masked position (see [LLaDA2 paper](https://arxiv.org/abs/2602.08676)).
```

jgerh:
```suggestion
- **Nemotron-Labs-Diffusion (Hybrid):** Combines diffusion with an autoregressive loss. During training, the model processes clean tokens plus a `masked_indices` sidecar and learns both a diffusion objective and an AR objective simultaneously.
```

jgerh:
```suggestion
- **DFlash:** Speculative block diffusion. A small draft model proposes tokens for a block conditioned on frozen target LM hidden states; a decay-weighted loss trains it to predict the target's distribution (see [DFlash paper](https://arxiv.org/abs/2602.06036)).
```

jgerh:
```suggestion
1. A **recipe script** ([`train_ft.py`](https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/recipes/dllm/train_ft.py)) that orchestrates the training loop with dLLM-specific corruption, loss, and batch handling.
```

jgerh:
```suggestion
2. A **YAML configuration file** that specifies the model, data, optimizer, dLLM-specific settings, and distributed training strategy.
```

docs/guides/dllm/finetune.mdx
```diff
@@ -163,6 +165,14 @@ dflash:
 | `dflash.num_blocks_per_sample` | Number of anchor blocks processed per sequence per step (paper default: 512, Appendix A.1) |
 | `dflash.attention_backend` | `flex_attention` (sparse, scales to 512 anchors) or `sdpa` (dense, OOMs above ~64). Default `sdpa` for backward compat — set to `flex_attention` for production runs |
 | `dflash.overlap_anchors` | `true` (paper, independent sampling) or `false` (non-overlapping stars-and-bars, caps at `seq_len // block_size`) |
+| `dflash.loss_type` | `dflash` (decay-weighted CE, default) or a D-PACE variant: `dpace`, `dpace-cumulative-confidence-only`, `dpace-continuation-value-only` (detached confidence-aware weights) |
+| `dflash.dpace_alpha` | Smoothing α in `[0, 1]` for the D-PACE confidence products; ignored when `loss_type: dflash` |
+
+<Note>
+The `dpace*` variants implement D-PACE (Dynamic Position-Aware Cross-Entropy),
+which reweights each position by detached cumulative draft confidence — see
```
jgerh:
```suggestion
which reweights each position by detached cumulative draft confidence; see
```

docs/guides/dllm/finetune.mdx
jgerh:
```suggestion
| `draft_acc` | Overall fraction of valid block positions where `argmax(draft_logits) == target_token` | Console line and wandb / mlflow / comet and file logger |
```

jgerh:
```suggestion
| `draft_acc_k{k}` | Same fraction restricted to block offset `k` (`k = 1..block_size-1`), the acceptance-length curve | wandb / mlflow / comet and file logger (one panel per offset); intentionally omitted from the console line to keep it readable |
```

jgerh, line 279: suggest adding a lead-in sentence to avoid stacked headings

```suggestion
Run the following commands to launch fine-tuning for each supported model family.

```

jgerh (commented):
Completed tech pubs review of docs/guides/dllm/finetune.mdx and provided a few copyedits.

nemo_automodel/components/loss/dllm_loss.py
```diff
@@ -426,10 +471,35 @@ def _reduce(
...
+        block_mask = block_mask.to(token_nll.dtype)
+        if self.loss_type == "dflash":
+            w = self._decay_weights(T, block_size, token_nll.device, token_nll.dtype)
+            weights = w.unsqueeze(0) * block_mask  # [B, T]
+        elif self.loss_type in _DPACE_LOSS_TYPES:
+            dpace_nll = token_nll
+            dpace_mask = block_mask
+            if block_size is not None and block_size > 1:
+                t_per = block_size - 1
+                if T % t_per == 0:
```
pthombre: **D-PACE silently skips the per-block reset when blocks don't partition `T`.**

The block reshape only happens when `T % t_per == 0`; otherwise `dpace_nll`/`dpace_mask` stay `[B, T]` and the confidence product in `_dpace_weight` runs across the *entire* sequence instead of resetting per block.

For the `dflash` path the analogous fallback only changes a decay curve, but here `torch.cumprod` over a long unsegmented sequence underflows toward zero, so the resulting weights become effectively meaningless rather than just mis-scaled — and it happens silently.

In practice `T = N*(block_size-1)` always partitions, so this is an edge case, but it would be good to make the fallback explicit: either a short comment documenting it, or a `logger.warning`/assert when `block_size` is set but doesn't divide `T`.

tests/unit_tests/loss/test_dllm_loss.py
```diff
@@ -381,7 +467,8 @@ def test_none_when_block_size_does_not_partition_tokens(self):
         assert correct_per_pos is None
         assert count_per_pos is None
 
-    def test_fused_matches_nonfused(self):
+    @pytest.mark.parametrize("loss_type", ["dflash", "dpace"])
```
pthombre: **Fused-path parity isn't covered for the two `*-only` variants.**

This parametrization (and `test_forward_supports_dpace_loss...` in `test_dflash_core.py`, which only uses `"dpace"`) leaves the chunked/fused linear-CE path untested for `dpace-cumulative-confidence-only` and `dpace-continuation-value-only`. They share the same reduction plumbing so the risk is low, but extending this to all four `loss_type`s is a cheap way to close the gap:

```python
@pytest.mark.parametrize(
    "loss_type",
    ["dflash", "dpace", "dpace-cumulative-confidence-only", "dpace-continuation-value-only"],
)
```

khazic:
I will review this portion shortly.

nemo_automodel/components/speculative/dflash/core.py
```diff
@@ -136,17 +145,27 @@ def __init__(
...
+        # and the D-PACE variants). ``normalize="mean"`` gives a local per-micro-batch
+        # decay-weighted mean; ``loss_decay_gamma=None`` disables decay (uniform
+        # weights). The variable-prefix objective needs per-block data-dependent
+        # weights and computes its loss inline instead (see _variable_prefix_loss).
+        self.loss_fn = (
+            None
+            if loss_type == "variable_prefix"
+            else DFlashDecayLoss(
+                loss_gamma=loss_decay_gamma,
+                normalize="mean",
```
khazic, line 237: D-PACE's published objective and reference implementation divide the weighted CE sum by batch size. Using `normalize="mean"` instead divides by the sample-dependent sum of D-PACE weights, which changes the objective and cancels the intended variation in total weight. Please preserve a batch-based normalization for the D-PACE variants.

akoumpa:
Hi @kashif I apologize for the long delay here.

Would you mind taking a look at the conflicts and i can help run CI. 

Thank you.

kashif:
not a problem! fixing

nemo_automodel/components/loss/dllm_loss.py
```diff
@@ -426,13 +472,55 @@ def _reduce(
...
-            loss = loss / (weights.sum() + 1e-6)
+            if self.loss_type in _DPACE_LOSS_TYPES:
+                # D-PACE is a weighted *sum*: the weight magnitude is the signal
+                # (the expected accepted-prefix length), so dividing by the weight
+                # sum would cancel exactly the variation the objective encodes.
+                # It would also desync the DP gradient average -- that stays exact
+                # only when every rank divides by the same constant, and the D-PACE
+                # weights are data-dependent. Normalize per sequence instead, which
+                # is what the reference implementation does.
+                loss = loss / max(float(token_nll.shape[0]), 1.0)
```
khazic: This changes the loss scale by orders of magnitude when `loss_type` flips. `DFlashTrainerModule` builds this loss with `normalize="mean"`, so `dflash` divides by `weights.sum()` (a per-token weighted mean over all `n` anchor blocks) while D-PACE divides only by the batch size. With the config defaults (`num_anchors: 512`, `block_size: 16`) that is roughly a 512*15 = 7.7k factor, i.e. switching `loss_type` silently multiplies the effective LR by ~3 orders of magnitude.

The reference implementation's "per sequence" normalization is per sequence with a single block, not 512 sampled anchors, so `token_nll.shape[0]` is not the equivalent denominator here. Dividing by `bsz * n_blocks` would keep the two objectives comparable.

nemo_automodel/components/loss/dllm_loss.py
```diff
@@ -426,13 +472,55 @@ def _reduce(
...
+        exact and the objective's weight magnitudes are preserved.
+        """
         _, T = token_nll.shape
-        w = self._decay_weights(T, block_size, token_nll.device, token_nll.dtype)
-        weights = w.unsqueeze(0) * block_mask.to(token_nll.dtype)  # [B, T]
+        block_mask = block_mask.to(token_nll.dtype)
+        if self.loss_type == "dflash":
+            w = self._decay_weights(T, block_size, token_nll.device, token_nll.dtype)
+            weights = w.unsqueeze(0) * block_mask  # [B, T]
+        elif self.loss_type in _DPACE_LOSS_TYPES:
```
khazic, line 550: The per-block partition guard only runs when `block_size is not None and block_size > 1`. `DFlashStrategy` passes `block_size=self.block_size if N > 1 else None`, so with `N == 1` D-PACE falls into the un-guarded path. It happens to be correct there (T is exactly one block), but the guard's own comment says the cumprod must never run across blocks, and nothing enforces that if a future caller passes `block_size=None` with multiple blocks.

nemo_automodel/components/speculative/dflash/core.py
```diff
@@ -560,7 +584,7 @@ def forward(
         pred_mask = block_mask[:, :, 1:].reshape(bsz, n * (bs - 1))
 
         loss_fn = self.loss_fn
-        assert loss_fn is not None, "loss_fn is always constructed for loss_type='dflash'"
+        assert loss_fn is not None, "loss_fn is constructed for every loss_type except 'variable_prefix'"
```
khazic, line 589: Follow-up on the block just below: `loss_weight` (line 596) is still computed from the decay weights even when `loss_type` is a `dpace*` variant, so the reported metric no longer describes the weights the loss actually used. Anything downstream that weights by `loss_weight` (or just logs it) will be inconsistent with the D-PACE objective.

nemo_automodel/components/loss/dllm_loss.py
kashif: I checked this against the D-PACE reference implementation. It divides the weighted sum by the **batch size**, not `bsz * n_blocks`: it samples up to `num_anchors` (default 512) anchors per sequence and still normalizes by `bsz` alone — its own loss test asserts `sum / float(bsz)` with `n_blocks = 2`. So the per-sequence normalization isn't single-block; the anchor sum is intended to scale with the number of sampled blocks, and only the batch dimension is averaged out.

I've kept the `/bsz` denominator to stay faithful to the reference objective, and fixed the two related points instead: `block_size` is now required on the D-PACE path so the confidence `cumprod` always resets per block, and `loss_weight` now reports `bsz` (the actual denominator) so the token-weighted validation average is consistent.

Happy to switch to a scale that's decoupled from the anchor count if you'd prefer it for LR stability — just flagging that it would diverge from the reference.

khazic: Thanks for checking the reference. Two things though.

The faithfulness argument doesn't bind here: `bsz` and `bsz * n_blocks` differ by a fixed constant, so both give the same gradient direction and the same optimum. Only the printed loss value and the lr you need change, not the objective. Staying faithful to the reference and keeping the two `loss_type` values on a comparable scale aren't in conflict.

The DP argument doesn't apply to what I proposed. `n_blocks` is `token_nll.shape[1]`, fixed by `num_anchors` in the config, identical on every rank. It isn't data dependent, so it can't desync the gradient average. That objection is valid against a `weights.sum()` denominator, not against this one.

What stays broken is at the recipe level: `core.py` hardcodes `normalize="mean"` for every fixed-anchor variant, so one config and one `optimizer.lr` now cover both a per-token weighted mean (denominator around `bsz * 3379` at gamma=7, k=15, 512 anchors) and a per-sequence sum divided by `bsz` alone. Counting the D-PACE weight magnitudes on top, that's roughly four orders of magnitude of effective-lr jump from flipping one string, and it fails by diverging mid-run rather than by erroring out. The new `finetune.mdx` and yaml text describe the variants and `dpace_alpha` but say nothing about lr.

Either resolution works for me:

1. Keep `/bsz`, and state in `finetune.mdx` and the yaml comment that the `dpace*` variants run about four orders of magnitude hotter than `dflash` and the lr must be scaled down accordingly, with a suggested starting point.
2. Scale by `n_blocks` at the recipe level, leaving the loss in its reference form, so both `loss_type` values are usable at the same default lr.

nemo_automodel/components/loss/dllm_loss.py
```diff
@@ -394,80 +413,166 @@ def __init__(
...
+
+        suffix = torch.flip(
+            torch.cumsum(torch.flip(prefix * binary_mask, dims=[-1]), dim=-1),
+            dims=[-1],
+        )
+
+        if self.loss_type == "dpace":
+            return suffix
+        if self.loss_type == "dpace-continuation-value-only":
+            return suffix / prefix.clamp_min(torch.finfo(prefix.dtype).tiny)
```
khazic: `dpace_alpha` is validated down to `0.0`, and at that end this division is unsafe. With alpha 0 the smoothing term vanishes, so `prefix` is a plain product of k confidences: at `block_size: 16` that is 15 factors, and a draft that is merely lukewarm (say 0.3 per position) already lands near 1e-8, well past bf16 subnormals. `prefix` then clamps to `tiny` and the ratio explodes into the weights. The default 0.5 keeps `prefix >= 0.5**15`, which is why this doesn't show up in the tests. Either tighten the lower bound to something that keeps the product representable, or compute this branch in log space (`suffix.log() - prefix.log()` accumulated as a cumsum of `log(smooth)`).

kashif:
thanks @HuiyingLi should be ready

HuiyingLi:
I checked the D-PACE recurrence against the paper and reference implementation. The core weighting formula looks consistent, but I found several integration issues at the current head (`d1345215`).

1. **The D-PACE denominator is rank-local and data-dependent.**

   `DFlashTrainerModule._sample_anchor_positions()` computes `max_n = min(num_anchors, valid_counts.max())` independently for each local batch/rank, but `DFlashDecayLoss` normalizes by the resulting local `batch_size * n_blocks`. DDP then averages gradients that were divided by different denominators.

   I reproduced this with the same examples having anchor capacities `[1, 1, 4, 4]`: partitioning them as `[1,1] | [4,4]` gives a DDP-averaged loss scale of `1.0`, while repartitioning the same examples as `[1,4] | [1,4]` gives `0.625`. The update therefore depends on rank partitioning. The denominator needs to use a common configured sampling budget, or the numerator/denominator need correct global reduction and DDP compensation. This also needs a two-rank gradient/optimizer-step parity test with uneven local anchor capacities.

2. **A DFlash2 top-k miss changes the D-PACE schedule for later positions.**

   `selector_mask = pred_mask * has_target` is passed into the D-PACE weight construction. A missing target candidate is consequently treated as padding/confidence 1, rather than merely being excluded from selector CE, so it changes the suffix weights of later hits.

   For base confidence `[0.5, 0.5, 0.5]` and `alpha=0.5`, the base weights are `[1.734375, 0.984375, 0.421875]`. With `has_target=[0,1,1]`, the current selector path produces `[0,1.3125,0.5625]`; preserving the base schedule and masking only selector supervision should produce `[0,0.984375,0.421875]`. Please construct position weights from `pred_mask`, then apply `has_target` only to selector supervision.

   The current tests use `selector_top_k == vocab_size`, so `has_target` is always true and this case is not exercised.

3. **DFlash2 redundantly computes a full-vocabulary FP32 CE on every step.**

   The base loss already computes full-vocabulary CE, but `dflash2_core.py:266-269` recomputes it from `pred_logits.float()`. This also happens for the default `loss_type="dflash"`, where `weighted_mean()` does not use `base_token_nll` at all.

   With the shipped Qwen3 DFlash2 shape, this scans about 1.17B logits again and materializes an approximately 4.35 GiB FP32 tensor. Please skip it entirely for fixed-decay DFlash and expose/reuse the base per-token NLL or computed position weights for D-PACE.

4. **The documented generic DFlash path and specialized trainer use different D-PACE normalization.**

   The generic strategy keeps `normalize="tokens"`, giving a denominator of `B * N * (block_size - 1)` for full blocks, while the specialized trainer divides by `B * N`. For block size 16, the same loss/gradient differs by a factor of 15, contrary to the documented denominator and “no LR change” guidance. A parity test between the two training paths would catch this.

There are also backward-compatibility issues worth addressing:

- New constructor arguments were inserted before existing positional arguments in both trainer modules, so old positional callers can silently misbind `prefix_weight_base`, `selector_loss_weight`, or `sliding_window`.
- `DFlashDecayLoss.forward()` no longer accepts the previous flattened input plus `block_size` contract.
- Adding `loss_denominator` changes `DLLMLossOutput` tuple arity for every dLLM loss.
- `DFlashDecayLossConfig` is not exported from `nemo_automodel.components.loss`.

The existing 254 relevant unit tests pass and current CI is green, but their inputs avoid the candidate-miss and uneven-rank cases above. The most useful additional validation would be a realistic BF16 DFlash2 D-PACE forward/backward test with `top_k << vocab_size`, forced target misses, and peak-memory/step-time measurement.

kashif:
Thanks for finding this, especially the selector masking case; that one wasn't caught by anything we had!

Fixed all four:
  - denominator now uses `total_blocks` (configured num_anchors, not the achieved count) so it's rank-consistent under DDP
  - selector weights are built from the full pred_mask and only narrowed by has_target after, not fed into the D-PACE recursion
  - skip the redundant full-vocab CE when `loss_type="dflash"`
  - `strategy.py` and `train_dflash.py` do use different normalization conventions (`"tokens"` vs `"mean"`). I didn't want to unify without knowing which configs you've already tuned to one or the other, so I documented it in both places. Let me know if I should just pick one?

Also made `loss_type`/`dpace_alpha` keyword-only and exported `DFlashDecayLossConfig` from the package root per your notes.

Tests cover the actual reported behavior (the denominator partition dependence, your exact numeric selector example, the redundant CE call count), not a rehash of what was already covered. Unit tests are green, plus a real 2 process CPU/gloo DDP run to check that gradients actually sync across ranks.

HuiyingLi:
Thanks — I re-reviewed the new head (`5e845cdf`). The configured-`total_blocks` denominator and the separation of the D-PACE schedule mask from the selector value mask both fix the two correctness repros. The package export and the default-`dflash` CE fast path also look good.

I still found the following issues:

1. **The D-PACE path still recomputes the full-vocabulary FP32 CE.**

   `loss_fn(...)` at `dflash2_core.py:273` already computes the backbone token NLL, then lines 285–288 compute the same full-vocabulary CE again for the selector weights. The new test explicitly expects three CE calls for `dpace`, so it currently codifies the remaining duplicate rather than preventing it.

   With the shipped Qwen3-8B DFlash2 shape, this is another scan over about 1.17B logits. Please compute the token NLL once and reuse it for both the base reduction and the detached D-PACE selector schedule.

2. **The two D-PACE training paths still have different normalization, and the public documentation now contradicts the implementation documentation.**

   The generic strategy divides by `B * N * (block_size - 1)`, while the dedicated trainer divides by `B * N`; with block size 16, I reproduced an exact 15x loss/gradient-scale difference. `strategy.py:560-569` documents that the LR is not portable, but `docs/guides/dllm/finetune.mdx:280-284` still says the denominator is batch times block count and that no LR change is needed.

   Since D-PACE itself is new here, there should not be an existing tuned D-PACE compatibility constraint. I would prefer one canonical D-PACE normalization across both paths. If the different conventions are intentionally retained, the user-facing guide must at least describe the actual path-dependent behavior and the ratio should have an explicit test.

3. **The positional compatibility change is still incorrect for `DFlashTrainerModule`.**

   Before this PR, its trailing parameters were:

   ```python
   ..., loss_type="dflash", prefix_weight_base=0.9, sliding_window=None
   ```

   They are now:

   ```python
   ..., sliding_window=None, *, loss_type="dflash", dpace_alpha=0.5, prefix_weight_base=0.9
   ```

   Thus an old positional caller still fails (or binds the former `loss_type` value to `sliding_window`). Please preserve all existing argument positions and make only the new argument keyword-only, e.g.:

   ```python
   ..., loss_type="dflash", prefix_weight_base=0.9, sliding_window=None, *, dpace_alpha=0.5
   ```

   The corresponding `DFlash2TrainerModule` ordering now looks correct.

4. The two other compatibility concerns from the previous review remain:

   - `DFlashDecayLoss.forward()` / `forward_fused()` no longer accept the existing flattened input plus `block_size` contract.
   - Adding `loss_denominator` changes the tuple arity of the shared `DLLMLossOutput` from four to five for every dLLM loss.

Local validation on this head: all 262 relevant tests pass, and Ruff/format checks pass. The latest head has not run the full CI suite yet; currently only DCO and PR-title checks are reported.

kashif:
thanks again @HuiyingLi can you kindly review again!

HuiyingLi:
I re-reviewed the latest head (`75e18902`). Most of the previous issues are fixed, but one denominator wiring issue remains.

`DFlashStrategy.pre_step()` returns `(num_loss_units, num_predicted)` for D-PACE. However, the strategy still inherits `normalization_mode="supervised"`, so `DiffusionLMSFTRecipe._compute_loss_denominators()` selects the second value, `num_predicted`. Because the loss uses `normalize="tokens"`, `total_blocks` is ignored: the generic path still divides by `B * N * (block_size - 1)`, while the dedicated recipe divides by `B * N`.

For `B=2, N=2, block_size=4`, the values are `(4, 12)`; the recipe selects 12 instead of 4, producing an exact 3x loss/gradient-scale difference. With `block_size=16`, this becomes 15x. The new distributed test passes the intended denominator directly, so it bypasses this wiring and does not catch the regression.

Please route `B * N` through the denominator that the recipe actually selects and add an integration test covering `pre_step() -> _compute_loss_denominators()`.

Local validation: 308 relevant tests passed; Ruff and format checks passed; BF16 forward/backward stayed finite for all three D-PACE variants at `alpha=0/0.5/1`.

docs/guides/dllm/finetune.mdx
jgerh:
```suggestion
Install NeMo AutoModel with `uv`:
```

jgerh, line 115:
```suggestion
instead of adding it only at inference. Its forward process mixes the usual absorbing `[MASK]` noise
with uniform token transitions, so the model is trained on contexts that contain
```

jgerh:
```suggestion
every visible position on top of the familiar denoising term at `[MASK]` positions. That extra term
```

jgerh:
```suggestion
only other released setting differs solely in `uniform_ratio: 0.2`, so that is the first hyperparameter to
```

jgerh:
```suggestion
  torch_dtype: float32          # fp32 master weights; compute is bf16 using mp_policy
```

docs/guides/dllm/finetune.mdx
```diff
@@ -257,6 +257,8 @@ dflash:
   num_blocks_per_sample: 512                    # Paper default (Appendix A.1)
   attention_backend: flex_attention             # required for N > ~64; sdpa OOMs
   overlap_anchors: true                         # paper samples anchors independently
```
jgerh:
```suggestion
  attention_backend: flex_attention             # Required for N > ~64; sdpa OOMs
  overlap_anchors: true                         # Paper samples anchors independently
```

docs/guides/dllm/finetune.mdx
```diff
@@ -257,6 +257,8 @@ dflash:
...
+  loss_type: dflash                             # dflash | dpace | dpace-cumulative-confidence-only | dpace-continuation-value-only
+  dpace_alpha: 0.5                              # smoothing alpha for the D-PACE variants
```
jgerh:
```suggestion
  dpace_alpha: 0.5                              # Smoothing alpha for the D-PACE variants
```

docs/guides/dllm/finetune.mdx
```diff
@@ -269,6 +271,23 @@ The following table describes the key configuration fields for DFlash fine-tunin
...
+divided by the globally reduced batch size times `num_blocks_per_sample`. This
+is equivalent to the local mean used by the dedicated DFlash recipe
+(`train_dflash.py`), including across data-parallel ranks and gradient
+accumulation, so `dpace_alpha` and `optimizer.lr` are portable between paths.
+
+Switching `loss_type` from `dflash` to a `dpace*` variant here does change the
+scale: `dflash` uses `normalize="tokens"` (a supervised-token count) in this
+strategy, not `normalize="mean"`, so it is on a different scale than either
+D-PACE convention. Retune `optimizer.lr` when switching `loss_type`.
+</Note>
```
jgerh, line 291:
```suggestion
D-PACE convention. Retune `optimizer.lr` when switching `loss_type`.

</Note>
```

docs/guides/dllm/finetune.mdx
jgerh:
```suggestion
Both metrics are computed from the logits that the chunked linear cross-entropy path already produces, without an additional model forward pass. They are reduced across data-parallel and context-parallel ranks using per-rank raw `(correct, count)` sums. An all-reduce operation sums these values before division. This process ensures that the values are correct across arbitrary per-rank token distributions in any NeMo AutoModel distributed mode.
```

docs/guides/dllm/finetune.mdx
```diff
@@ -336,6 +355,7 @@ dataset:
 The mask is built for `sdpa` or `eager` (dense additive) or `flex_attention` (sparse `BlockMask`, preferred at scale). FlashAttention-2 is unsupported because it ignores arbitrary masks, and context parallelism is also unsupported. The paper trains a `block_length` 1→2→3 curriculum (one epoch each). Run the stages as successive fine-tunes, and enable `auto_balance_clean_loss` at the b3 stage.
 
 ## Fine-Tune the Model
+Run the following commands to launch fine-tuning for each supported model family.
```
jgerh, line 360:
```suggestion
## Fine-Tune the Model

Run the following commands to launch fine-tuning for each supported model family.
```

docs/guides/dllm/finetune.mdx
jgerh:
```suggestion
The sampler selected with the `--sampler` argument supplies these preset values. LLaDA uses the standalone NeMo AutoModel sampler. LLaDA2, Nemotron, and DiffusionGemma call the built-in `generate()` methods of their models. DiffusionGemma only forwards `--steps` (as `max_denoising_steps`, default 48) and `--max_new_tokens` (default 256). The remaining flags keep their upstream Hugging Face defaults. The `--remasking` and `--no_kv_cache` arguments do not affect LLaDA2 or DiffusionGemma generation.
```

jgerh (commented):
Completed tech pubs review of docs/guides/dllm/finetune.mdx and provided a few copyedits.

nemo_automodel/components/loss/dllm_loss.py
```diff
@@ -786,118 +818,419 @@ def __init__(
...
+        ``num_anchors`` whenever the batch supplies that many valid anchors --
+        i.e. a constant, leaving the gradient direction and the optimum unchanged.
+        What it buys is scale: without it, flipping ``loss_type`` in the YAML
+        moves the effective learning rate by orders of magnitude and the run
+        diverges instead of erroring. The reported value is correspondingly
+        ``n_blocks`` times smaller than the published one.
+        """
+        if self.loss_type in _DPACE_LOSS_TYPES:
+            blocks = n_blocks if total_blocks is None else total_blocks
+            return weights.new_tensor(max(float(bsz * blocks), 1.0))
```
khazic: `weights` carries `token_nll`'s dtype and the unfused path (`_compute_per_token_nll`) does not upcast, so with bf16 logits this integer denominator is materialized in bf16: `bsz * num_anchors = 1022` becomes `1024.0` (checked with `torch.tensor(1., dtype=torch.bfloat16).new_tensor(1022.)`). The whole D-PACE loss and gradient pick up that rounding, the unfused and fused paths (fp32 `_chunk_nll`) then disagree on scale for identical inputs, and the rounded value is what `loss_weight` reports. Building the scalar as fp32 (`torch.tensor(..., dtype=torch.float32, device=weights.device)`) fixes all three.

PR #2417, opened 2026-06-04
feat(models): add Qwen3.5 MTP support

HuiyingLi:
## Summary

Adds Qwen3.5 MTP support for dense and MoE Automodel implementations.

- Adds Qwen3.5 dense MTP module and registry support.
- Adds Qwen3.5-MoE MTP module matching the checkpoint architecture: fusion projection, embedding/hidden norms, MTP decoder layer, final norm, and base LM head reuse.
- Adds HF state-dict conversion for MTP weights, including Qwen3.5-MoE per-expert checkpoint tensors to/from grouped native tensors.
- Exposes `num_nextn_predict_layers` and `mtp_loss_scaling_factor` knobs in Qwen3.5 35B recipes.

## Validation
- 27B dense: https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/krsqgchv?nw=nwuserhuiyingl
- 35B moe: https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/35bkflbr?nw=nwuserhuiyingl

PR #2595, opened 2026-06-16
fix(qwen3_5_moe): convert MTP experts as grouped tensors (AM-442)

HuiyingLi:
## Summary

Fixes **AM-442**: `RuntimeError: Missing key in checkpoint state_dict: mtp.layers.0.mlp.experts.224.down_proj.weight` when loading the Qwen3.6-35B-A3B MoE checkpoint under expert parallelism (recipes `qwen3_6_35b`, `qwen3_6_35b_lora`, `qwen3_6_35b_medpix_ep8cp2_4k`; Automodel main + transformers 5.8.1).

## Root cause

The MTP block's MoE experts are stored in the HF checkpoint in the **same grouped layout** as the main decoder layers — confirmed against the real checkpoint index (`Qwen/Qwen3.6-35B-A3B/model.safetensors.index.json`):

```
mtp.layers.0.mlp.experts.down_proj        # grouped 3-D tensor
mtp.layers.0.mlp.experts.gate_up_proj
```
(zero per-expert keys.)

But `Qwen3_5MoeStateDictAdapter.convert_single_tensor_to_hf` (the `to_hf` direction) **split the MTP experts into per-expert keys** (`mtp.layers.0.mlp.experts.{id}.down_proj.weight`), while leaving the main decoder layers grouped. When the checkpoint loader builds its load destinations by running the model state-dict through `to_hf`, it asked DCP to load into per-expert MTP keys that don't exist in the grouped checkpoint → missing-key error.

Under EP=8 with 256 experts, each rank owns 32 experts; rank 7 owns 224–255, so `split_experts_weights_dtensor_aware` emits expert **224** first on that rank — which is why 224 is the first key to fail.

## Fix

Remove the per-expert MTP branches in `convert_single_tensor_to_hf` so MTP experts fall through to the **generic grouped** rename+transpose already used for the main decoder layers, producing `mtp.layers.0.mlp.experts.{gate_up_proj,down_proj}` — exactly matching the on-disk keys. The grouped `from_hf` path already handles loading these (including EP-shard slicing).

## Testing

- `tests/unit_tests/models/qwen3_5_moe/test_qwen3_5_moe_state_dict_adapter.py` — updated the two `to_hf` MTP tests to assert the grouped output, and added `test_mtp_experts_emit_no_per_expert_keys` (AM-442 regression: `to_hf` must not fabricate `...experts.<id>....` keys). **47/47 pass**; `ruff` clean.
- Root cause byte-confirmed against the real checkpoint's safetensors index and against the transformers 5.8.1 `Qwen3_5MoeExperts` grouped-parameter definition.

## W&B validation run (end-to-end on the real 35B checkpoint)

Ran the exact failing recipe — `qwen3_6_35b` (Qwen3.6-35B-A3B, **EP=8 + MTP enabled**, `num_nextn_predict_layers: 1`) — on **8 GPUs** with transformers 5.8.1 and the fixed adapter, loading the grouped HF-hub checkpoint:

🔗 https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/amkgnw1h

- The 72 GB checkpoint **loads cleanly** — previously this exact path raised `RuntimeError: Missing key in checkpoint state_dict: mtp.layers.0.mlp.experts.224.down_proj.weight` on the EP rank that owns expert 224.
- **100/100 steps completed, 0 `Missing key` / 0 `KeyError`**; loss 2.06 → ~1.5 (decreasing).

### DIFF
--- a/nemo_automodel/components/models/qwen3_5_moe/state_dict_adapter.py
+++ b/nemo_automodel/components/models/qwen3_5_moe/state_dict_adapter.py
@@ -287,41 +287,13 @@ class Qwen3_5MoeStateDictAdapter(StateDictAdapter):
 
         new_fqn = fqn
         value = tensor
-        mtp_gate_up_match = re.match(r"mtp\.layers\.(\d+)\.mlp\.experts\.gate_and_up_projs$", fqn)
-        mtp_down_match = re.match(r"mtp\.layers\.(\d+)\.mlp\.experts\.down_projs$", fqn)
-        if mtp_gate_up_match:
-            layer_num = mtp_gate_up_match.group(1)
-            splits, expert_ids = state_dict_utils.split_experts_weights_dtensor_aware(
-                tensor, self.moe_config.n_routed_experts
-            )
-            result = []
-            inter_dim = self.moe_config.moe_inter_dim
-            for expert_tensor, expert_id in zip(splits, expert_ids):
-                gate = expert_tensor[:, :inter_dim].transpose(0, 1)
-                up = expert_tensor[:, inter_dim:].transpose(0, 1)
-                if not state_dict_utils.is_dtensor(gate):
-                    gate = gate.contiguous()
-                if not state_dict_utils.is_dtensor(up):
-                    up = up.contiguous()
-                result.append((f"mtp.layers.{layer_num}.mlp.experts.{expert_id}.gate_proj.weight", gate))
-                result.append((f"mtp.layers.{layer_num}.mlp.experts.{expert_id}.up_proj.weight", up))
-            if exclude_key_regex:
-                result = [(key, val) for key, val in result if not re.match(exclude_key_regex, key)]
-            return result
-        if mtp_down_match:
-            layer_num = mtp_down_match.group(1)
-            splits, expert_ids = state_dict_utils.split_experts_weights_dtensor_aware(
-                tensor, self.moe_config.n_routed_experts
-            )
-            result = []
-            for expert_tensor, expert_id in zip(splits, expert_ids):
-                down = expert_tensor.transpose(0, 1)
-                if not state_dict_utils.is_dtensor(down):
-                    down = down.contiguous()
-                result.append((f"mtp.layers.{layer_num}.mlp.experts.{expert_id}.down_proj.weight", down))
-            if exclude_key_regex:
-                result = [(key, val) for key, val in result if not re.match(exclude_key_regex, key)]
-            return result
+        # MTP experts use the SAME grouped layout as the main decoder layers in the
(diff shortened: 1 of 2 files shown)


HuiyingLi:
### Additional corroboration (independent reproduction)

- **Reproduced the exact error** by simulating EP=8 rank 7 (experts 224–255) with the real `Qwen3.6-35B-A3B` config/checkpoint metadata: `to_hf` generated 96 per-expert MTP destination keys, the first being `mtp.layers.0.mlp.experts.224.down_proj.weight` — none of which exist among the checkpoint's 2 grouped MTP keys → the reported `RuntimeError`. Post-fix, `to_hf` emits exactly the 2 grouped keys and round-trips cleanly through `from_hf`.
- **Not a transformers regression.** The per-expert MTP split was introduced in #2417 ("feat(models): add Qwen3.5 MTP support", commit `76bb18f3`) — the same PR that added MTP. The grouped expert layout (main *and* MTP) has been stable in the checkpoint.
- **Why earlier runs passed:** only `qwen3_6_35b.yaml` enables MTP (`num_nextn_predict_layers: 1`); the `lora` and `medpix_ep8cp2_4k` recipes leave MTP disabled, so the buggy branch was never hit there. The trigger is **MTP enabled + EP sharding + grouped checkpoint load**.

(The dead per-expert `from_hf` reader is left in place as harmless backward-compat; can remove for symmetry if preferred.)

HuiyingLi:
### Design note: two MTP checkpoint layouts exist; this fix targets the canonical (grouped) one

Investigation surfaced that MTP expert weights appear in **two** layouts in practice:

- **Grouped** — HF hub `Qwen/Qwen3.6-35B-A3B`: `mtp.layers.0.mlp.experts.{gate_up_proj,down_proj}` (2 keys), identical to the main decoder layers. This is what the shipped recipes load, and what AM-442 needs.
- **Per-expert** — a local conversion artifact (`qwen35_35B`) that PR #2417 was originally validated against: `mtp.layers.0.mlp.experts.{N}.{gate,up,down}_proj.weight`. The per-expert `to_hf` code matched *that* checkpoint, which is why #2417's run passed.

`from_hf` already handles **both** (grouped `mtp_match` + per-expert `mtp_split_match`), so loading either checkpoint works. This PR changes only the `to_hf` / DCP-destination direction to emit **grouped** keys — consistent with the main decoder layers and the canonical hub checkpoint, and round-trip-verified via `from_hf`.

Tradeoff: the old per-expert *local* checkpoint would no longer match through the `to_hf` destination path. Given the hub checkpoint is the canonical format the recipes ship with (and main-layer experts are always grouped), grouped is the correct single target rather than special-casing MTP to be format-aware.

akoumpa (approved):

PR #2616, opened 2026-06-17
fix(config): validate pp_size against distributed.pipeline

akoumpa:
## Summary

Validate `distributed.pipeline` against `pp_size` **asymmetrically** — warn on
the recoverable mismatch, hard-error on the unambiguous one:

- a `pipeline:` block with `pp_size <= 1` → inert (PP disabled); the block is
  dropped and a **warning** is logged.
- `pp_size > 1` with no `pipeline:` block → **`ValueError`**. PP is requested
  but the schedule / microbatch size are unspecified. An explicit empty
  `pipeline: {}` opts into defaults.

## Why the split

Keeping a `pipeline:` section at `pp_size=1` is intentional and useful: it's a
ready reference for enabling PP, and handy for `ep8/pp1` vs `ep4/pp2` parity
debugging (re-adding the section by hand is a pain). The L2 finetune tests also
load such configs and set `--distributed.pp_size 1` on the CLI — a hard error
there can't be fixed by a static YAML edit when `pp_size` is overridden at
launch. So that direction only warns.

The reverse — `pp_size > 1` with no `pipeline:` block — is an unambiguous
misconfiguration (PP requested, but the schedule is never specified), so it
fails fast instead of silently running with `PipelineConfig` defaults.

**No recipe changes** — earlier `pp_size` bumps were dropped; every config keeps
its pipeline section at `pp_size=1`. Verified across all example +
functional-test configs with a `distributed` section: **0 trip the new error**,
and the 38 `pipeline:`+`pp_size:1` template recipes only warn.

## Testing

`tests/unit_tests/recipes/test_dist_utils.py`: **71 passed.** Covers both
directions — warn-and-drop for `pipeline:`+`pp<=1`, and
`pytest.raises(ValueError)` for `pp>1` with no block (empty `pipeline: {}` still
opts into defaults).

### DIFF
--- a/nemo_automodel/recipes/_dist_utils.py
+++ b/nemo_automodel/recipes/_dist_utils.py
@@ -160,11 +160,22 @@ def parse_distributed_section(cfg_dict: dict) -> dict:
     if activation_checkpointing == "selective" and strategy_name != "fsdp2":
         raise ValueError("selective activation checkpointing is supported only for FSDP2 configs.")
 
-    # YAML-level sanity: silently discard sub-configs that don't apply to the
-    # current parallelism sizes (e.g. pipeline section present but pp_size=1,
-    # which is common when a YAML template is overridden via CLI).
+    # `distributed.pipeline` and `pp_size` are validated asymmetrically:
+    #   * pipeline block with pp_size<=1 -> WARN (inert; block ignored). This is
+    #     often intentional -- a recipe keeps the pipeline section as a reference
+    #     and toggles pp_size via CLI (e.g. ep8/pp1 vs ep4/pp2 parity debugging),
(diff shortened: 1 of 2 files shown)


HuiyingLi:
These recipes fit in single node without PP. That's the more desirable setup. Why do we make them multinode by adding PP? We can keep the pipeline block as an reference/example

examples/convergence/tulu3/models/gpt-oss-20b/gptoss_20b_ep8_te_fusedadam.yaml
akoumpa, line 61: @HuiyingLi all recipes have a pipeline section but pp_size=1; this pr chooses either; I'm open to suggestions

HuiyingLi (approved):

PR #2634, opened 2026-06-18
fix(examples): enable activation checkpointing for phi_4_squad

akoumpa:
## What

Fixes the recurring OOM in the `phi_4_squad` **`sft_ckpt_robustness`** functional test.

Two commits:
1. **`fix(examples)`** — the actual fix: enable `activation_checkpointing: true` for `phi_4_squad`.
2. **`test(checkpoint)`** — memory hygiene in the robustness test (reclaim each phase's model + optimizer before the next builds its own). Useful on its own, but **not** what fixes this OOM.

## Root cause (it was not a leak)

Profiled per-step CUDA memory on 8×H100. The OOM is **not** cross-phase accumulation, **not** FSDP, **not** validation, **not** the optimizer/`_supports`:

```
step=0 TRAIN alloc=10.33 peak=68.03 GiB
step=1 TRAIN alloc=10.33 peak=73.94 GiB
...
step=4 VAL   alloc=12.88 peak=15.44 GiB
→ step 5 OOM
```

- **`alloc` is flat at ~10 GiB every step** → no leak / no accumulation.
- **Every training step peaks at ~70–74 GiB** — phi-4 (14B) keeps ~60 GiB of activations per fwd/bwd, running right at the 80 GiB edge. SQuAD's variable-length batches spike some steps past 80 GiB.
- The **resume phase** trains `max_steps + 3` (8) steps vs the base 5, so it gets more rolls of the dice and reliably hits a spiking batch → OOM. (Base phase survives by a hair.)

Enabling **activation checkpointing** bounds the per-step peak so it fits.

## Verification

Reproduced and fixed end-to-end on EOS (8×H100), running only the `phi_4_squad` checkpoint-robustness test in the CI container:

- **Without** `activation_checkpointing`: OOM at step 5 (`75 GiB allocated`), matching CI.
- **With** `activation_checkpointing: true`: **passes** — all phases incl. resume, peak ~70 GiB, `1 passed in 624s`.

Failing CI before the fix (reference): https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/343744188

### DIFF
--- a/examples/llm_finetune/phi/phi_4_squad.yaml
+++ b/examples/llm_finetune/phi/phi_4_squad.yaml
@@ -49,6 +49,10 @@ checkpoint:
 
 distributed:
   strategy: fsdp2
+  # phi-4 (14B) stores ~60 GiB of activations per training step; without
+  # recomputation a single fwd/bwd peaks at ~74 GiB and OOMs on long batches
+  # (notably the checkpoint-robustness resume phase, which runs extra steps).
+  activation_checkpointing: true
   dp_size: none
   tp_size: 1
   cp_size: 1
--- a/nemo_automodel/_transformers/capabilities.py
+++ b/nemo_automodel/_transformers/capabilities.py
@@ -29,6 +29,7 @@ from __future__ import annotations
 import functools
 import inspect
 import logging
+import weakref
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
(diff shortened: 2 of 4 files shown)


HuiyingLi (dismissed):

akoumpa:
Verified on EOS (8×H100), running just the `phi_4_squad` checkpoint-robustness test in the CI container (`automodel:pipe.55165794`):

- **without** `activation_checkpointing`: OOMs at step 5 of the resume phase (~75 GiB allocated), reproducing the CI failure (ref [job 343744188](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/343744188)).
- **with** `activation_checkpointing: true` (this PR): **passes** — `1 passed in 624s`, per-step peak ~70 GiB, all phases incl. resume.

Per-step profiling: steady `alloc` flat at ~10 GiB, peak ~70–74 GiB — i.e. the OOM is a per-step activation peak, not a leak or cross-phase accumulation.

HuiyingLi (approved):

PR #2638, opened 2026-06-19
fix(parallelizer): resolve NemotronH decoder blocks for Nemotron-V3

akoumpa:
# What does this PR do ?

Fixes `AttributeError: 'NemotronHForCausalLM' object has no attribute 'backbone'` that crashes FSDP2 parallelization for the native Nemotron-V3 model (e.g. `nemotron_nano_v3_singlegpu_lora`).

Two distinct classes share the name `NemotronHForCausalLM`:
- the **HF** model keeps decoder blocks at `model.backbone.layers` (an `nn.ModuleList`), while
- the **native** Nemotron-V3 model (`NemotronV3Model`) keeps them at `model.model.layers` (an `nn.ModuleDict` keyed `"0".."N-1"`).

`NemotronHParallelizationStrategy` (selected by the class-name string) and the `LLM_MODEL_CLS_TO_LAYERS` registry assumed the HF layout, so the native model raised at `layers = model.backbone.layers`.

# Changelog

- Add `_nemotronh_decoder_blocks(model)` resolving the decoder-block container for both layouts (`backbone.layers` ModuleList vs `model.layers` ModuleDict) and returning the ordered block list.
- Use it in `NemotronHParallelizationStrategy.parallelize` (TP iteration, CP setup, FSDP wrap), and make the activation-checkpointing rewrap write back into the real container by ModuleDict key / ModuleList index.
- Add `"model.layers"` to the `NemotronHForCausalLM` entry of `LLM_MODEL_CLS_TO_LAYERS`. `_reduce_attrs` already skips the absent `backbone.layers` via `getattr(..., None)`, so both paths are covered and neither raises.
- Add regression tests (`TestNemotronHLayoutResolution`) for both layouts (helper + `_extract_model_layers`).

- [x] Did you write any new necessary tests? — `tests/unit_tests/distributed/test_parallelization_strategies.py::TestNemotronHLayoutResolution` (10 NemotronH tests pass locally)

# Additional Information

- Affected recipe: `nemotron_nano_v3_singlegpu_lora` (single-GPU LoRA), surfaced in nemo-ci job 343744282.
- Scope note: for native Nemotron-V3 with `tp>1`/`ep>1`, this strategy's TP plan does not shard MoE experts; proper multi-GPU native-v3 parallelism should use the MoE parallelizer. This PR fixes the single-GPU / DP-only crash.
- Fixes AM-448.

### DIFF
--- a/nemo_automodel/components/distributed/parallelizer.py
+++ b/nemo_automodel/components/distributed/parallelizer.py
@@ -380,6 +380,24 @@ class DefaultParallelizationStrategy(ParallelizationStrategy):
         return model
 
 
+def _nemotronh_decoder_blocks(model: nn.Module) -> tuple[nn.Module, list[nn.Module]]:
+    """Return ``(container, blocks)`` for a NemotronH model's decoder blocks.
+
+    Two distinct classes share the name ``NemotronHForCausalLM``:
+
+    * the HF model keeps its blocks in ``model.backbone.layers`` (an ``nn.ModuleList``), while
+    * the native Nemotron-V3 model (``NemotronV3Model``) keeps them in ``model.model.layers``
+      (an ``nn.ModuleDict`` keyed ``"0".."N-1"``).
+
+    ``container`` is the underlying ``ModuleList``/``ModuleDict`` (so callers can write rewrapped
+    blocks back into the model), and ``blocks`` is the ordered list of block modules.
+    """
+    inner = model.backbone if hasattr(model, "backbone") else model.model
+    container = inner.layers
+    blocks = list(container.values()) if isinstance(container, nn.ModuleDict) else list(container)
+    return container, blocks
+
+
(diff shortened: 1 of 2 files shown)


nemo_automodel/components/distributed/parallelizer.py
```diff
@@ -380,6 +380,24 @@ def parallelize(
         return model
 
 
+def _nemotronh_decoder_blocks(model: nn.Module) -> tuple[nn.Module, list[nn.Module]]:
```
HuiyingLi, line 383: any other locations than here?

akoumpa: I'm leaving parallelizer's (needed) refactor for another PR.

HuiyingLi (approved):

PR #2645, opened 2026-06-19
fix(datasets): decode MedPix images on demand instead of up front

akoumpa:
## What

`make_medpix_dataset` built its conversation list with `[format(ex) for ex in dataset]`, and `format` referenced `example["image_id"]` (an HF `Image` feature). Iterating therefore **decoded every image in the split up front** and held all 17,420 decoded images in memory before training started.

For short runs this is pure overhead: a 50-step `qwen3_5_4b_neat_packing` job on 8×H100 spent ~160 s in a GPU-idle dataset-build phase decoding the whole split, even though only ~50×GBS samples are consumed.

## Change (`datasets.py` only)

- Defer formatting to `__getitem__` via `dataset.with_transform(...)` — the dataset stays Arrow-backed, so images are **not** copied into a Python list.
- Load images undecoded (`cast_column("image_id", Image(decode=False))`) and wrap the bytes in a **lazy** `PIL.Image.open(BytesIO(...))` (header-only read). The pixel decode then happens on demand in the DataLoader workers; the existing `_preload_media` calls `.convert("RGB")` per sample.
- A lazy `PIL` handle is the form every consumer already accepts (training collate via `_preload_media`, and the validation collate/processor), so **no other files change**.

## Validation (qwen3_5_4b_neat_packing, Qwen3.5-4B, neat-packing, FSDP2, 8×H100, full dataset)

|  | dataset-build | end-to-end | result |
|---|---|---|---|
| before (eager) | ~160 s (0% GPU) | 11:53 | 50 steps + val + consolidated ckpt |
| after (this PR) | ~12 s | 7:51 | 50 steps + val + consolidated ckpt |

Loss unchanged (full before/after in the comment below). fd-safe: `BytesIO` holds no OS file descriptor — measured open-fd count stays flat (8) whether 0 or all 17,420 lazy handles are built.

## Note

The same eager-decode pattern exists in the other `make_*` VLM dataset builders (e.g. `make_cord_v2_dataset`); this PR fixes `make_medpix_dataset` (the one exercised by the timing-sensitive CI recipe). Happy to generalize.

### DIFF
--- a/nemo_automodel/components/datasets/vlm/datasets.py
+++ b/nemo_automodel/components/datasets/vlm/datasets.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import io
 import json
 import logging
 import math
@@ -22,6 +23,7 @@ import time
 from concurrent.futures import ThreadPoolExecutor
 
 import torch.utils.data
+from datasets import Image as HfImage
 from datasets import load_dataset
 from PIL import Image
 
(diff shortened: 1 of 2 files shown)


akoumpa:
### Loss: before vs after (no change)

Reproduced `qwen3_5_4b_neat_packing` (Qwen3.5-4B, full dataset, neat-packing, FSDP2, 8×H100, seed 1234) — 50 steps + validation:

| step | before (eager) | after (this PR) |
|---|---|---|
| 0 | 2.3573 | **2.3573** (identical) |
| 49 | 1.3622 | 1.3594 |
| val | 1.5254 | 1.5231 |

- **Step 0 is bit-identical** → the lazy decode yields the same pixels and packing is unchanged (same first batch + same init).
- Over all 50 steps: **max |Δloss| = 0.0431, mean = 0.0066** — within the run-to-run FSDP/bf16 nondeterminism present between any two runs (atomic / NCCL reduction order), not a systematic shift. (The "after" run also used `num_workers=4` vs the baseline's `0`; DataLoader order is sampler-deterministic regardless, which the step-0 identity confirms.)

No regression in training, validation, or checkpointing.

HuiyingLi (approved):
Thank you!

Issue #2674, opened 2026-06-20
Fused TE RoPE produces NaN on MLA k_pe in THD/packed-sequence mode (Moonlight, GLM-4.7-Flash)


akoumpa:
## Summary

For **MLA** models trained with `backend.attn: te` + **packed sequences (THD)**, the fused Transformer-Engine RoPE path produces **non-finite values on the single-head `k_pe`** at the very first forward. The resulting `NaN` loss/grad at step 0 corrupts all weights through the optimizer step, and at step 1 the all-`NaN` router collapses MoE routing onto one expert-parallel rank, which surfaces as a **CUDA OOM** in the token-permute. The OOM is a symptom; the root cause is the RoPE `NaN`.

This was originally reported as an OOM (`moonlight_16b_te_packed_sequence`), but it is **not** a memory/capacity problem and **not** a checkpoint-loading problem (weights load 100% clean).

## Affected configs (confirmed)

| Recipe | Model | Arch |
|---|---|---|
| `examples/llm_finetune/moonlight/moonlight_16b_te_packed_sequence.yaml` | moonshotai/Moonlight-16B-A3B | DeepSeek-V3 MLA |
| `examples/llm_finetune/glm/glm_4.7_flash_te_packed_sequence.yaml` | zai-org/GLM-4.7-Flash | GLM-4.x MLA |

Both reproduce `loss nan | grad_norm nan` at step 0 → step-1 MoE-dispatch OOM. Any other **MLA + TE + packed-sequence** MoE recipe is expected to hit this too (e.g. `glm_moe_dsa`, DeepSeek-V3.2).

## Root cause

`rope_fusion` defaults to `True` when TE+CUDA are present (`nemo_automodel/components/models/common/utils.py:199`), and these recipes do not override it. The MLA forward then calls:

```
deepseek_v3/layers.py: apply_rotary_emb_qk(q_pe, k_pe, ...)
  -> rope_utils.apply_rotary_emb_qk(rope_fusion=True)
     -> transformer_engine.pytorch.attention.rope.apply_rotary_pos_emb(tensor_format="thd", interleaved=True, fused=True)
```

In THD/packed mode this corrupts **only the single-head `k_pe`** (shape `(T, 1, 64)`); `q_pe` (multi-head) under the *identical* `freqs` is fine.

Instrumented repro on the exact failing commit (`d94e9110` = `r0.5.0` HEAD), 8×H100, `ep_size=8`:

```
[...]

PR #2675, opened 2026-06-20
fix(recipe): disable fused RoPE for MLA packed-sequence MoE recipes

akoumpa:
## What

Disable fused RoPE (`model.backend.rope_fusion: false`) in the two **MLA + TE + packed-sequence** MoE finetune recipes:

- `examples/llm_finetune/moonlight/moonlight_16b_te_packed_sequence.yaml`
- `examples/llm_finetune/glm/glm_4.7_flash_te_packed_sequence.yaml`

## Why

In THD (packed-sequence) mode the fused Transformer-Engine RoPE path emits **NaN/Inf on the MLA single-head `k_pe`** at the first forward (`q_pe`, multi-head, is fine under the identical `freqs`). That NaN-s step-0 loss/grad, the optimizer step then corrupts every weight, and at step 1 the all-`NaN` router collapses MoE routing onto a single EP rank — which surfaces as a **CUDA OOM** in the token-permute. The OOM is a symptom, not the cause; weights load 100% clean.

Repro evidence (instrumented, exact CI commit `d94e9110`, 8×H100, `ep_size=8`):
```
ROPE OUT  q_pe[nan=0 finmax=1.0e1]  k_pe[nan=30 finmax=3.390e38]   # bf16 max + NaN
```

The non-fused RoPE path does the rotation as a unit-magnitude complex multiply in fp32, so it is bounded.

## Validation (8×H100)

| Recipe | before (`rope_fusion: true` default) | after (this PR) |
|---|---|---|
| Moonlight-16B | `loss nan` → step-1 OOM | loss 2.57 → 2.47 → 2.16, val 2.07, no NaN, balanced routing, no OOM |
| GLM-4.7-Flash | `loss nan` → step-1 OOM | loss 3.03 → 2.84 → 2.67, val 2.50, no NaN, balanced routing, no OOM |

## Notes

- Workaround for #2674 — the proper TE-side fix (fused RoPE on single-head `k_pe` in THD, and/or per-sequence positions from `cu_seqlens`) is tracked there. This PR only flips a recipe flag to unblock these CI configs.
- Other MLA + packed recipes (e.g. `glm_moe_dsa`, DeepSeek-V3.2) likely need the same guard once their configs run.

### DIFF
--- a/examples/llm_finetune/glm/glm_4.7_flash_te_packed_sequence.yaml
+++ b/examples/llm_finetune/glm/glm_4.7_flash_te_packed_sequence.yaml
@@ -45,6 +45,11 @@ model:
     dispatcher: deepep
     fake_balanced_gate: false
     enable_hf_state_dict_adapter: true
+    # WAR for https://github.com/NVIDIA-NeMo/Automodel/issues/2674:
+    # fused TE RoPE emits NaN/Inf on the MLA single-head k_pe in THD (packed-sequence)
+    # mode, NaN-ing step-0 loss/grad and collapsing MoE routing -> OOM.
+    # Non-fused RoPE (fp32 unit-magnitude complex multiply) is bounded.
+    rope_fusion: false
 
 autopipeline:
   _target_: nemo_automodel.components.distributed.pipelining.config.PipelineConfig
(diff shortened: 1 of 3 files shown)


akoumpa:
both configs confirmed working : https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/55331542

PR #2676, opened 2026-06-20
fix(recipe): minimax/glm *_hellaswag_pp PP+EP MoE deadlock (WIP: loss diagnosis)

akoumpa:
**Draft / WIP — carries diagnostic commits; the loss issue is not yet resolved. Not for merge.**

## Fixed — the PP+EP MoE NCCL deadlock
`minimax_m2.{1,7}` and `glm_5.1` `*_hellaswag_pp` deadlock ~step 12 (NCCL watchdog: PP SEND/RECV ↔ EP all-gather circular wait). The interleaved1f1b schedule precomputes **static** inter-stage shapes, but the default collater pads each batch to a **variable** length, so the first microbatch longer than the precomputed length mismatches the P2P buffers → hang. Static shapes fix it: m2.1/m2.7 now run all 50 steps, and `glm_5.1` runs once off the THD path (its DSA indexer doesn't support THD packing — `pad_to_max_length` BSHD avoids it).

## Under investigation — garbage loss
With static shapes the deadlock is gone but the loss starts at ~13–15 instead of the correct ~3 for a pretrained model. Two independent corrupters found (table in the comment below): `rope_fusion: true`, and the **static-shape padding itself poisoning the MoE** (prime suspect: `pad_token_id=None` → token-0 padding routed through experts / aux loss). Activation checkpointing was ruled out. Also includes a RoPE fp32-rotation fix (`apply_rotary_emb` was downcasting cos/sin to bf16).

The branch will be squashed to the real fixes once the loss is resolved.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -78,9 +78,12 @@ RUN if [ "$INSTALL_TE" = "True" ]; then \
 # Install HybridEP / DeepEP. Build with apt rdma-core v60 (so build==runtime libibverbs)
 # and RDMA_CORE_HOME symlinked to the system install; create the libnvshmem_host.so->.so.3
 # symlink required for DeepEP fabric handle ops; use the nvshmem wheel 3.6.5 and pin
-# DEEPEP_COMMIT to 17cfb817.
+# DEEPEP_COMMIT to 42144303 (== 17cfb817 + DeepEP #638), which pads the HybridEP
+# max_num_of_tokens_per_rank up to the combine-kernel chunk size so a per-rank token
+# count that is not a multiple of 64 (e.g. 672) no longer fails the combine-kernel
+# static_assert (MAX_NUM_OF_TOKENS_PER_RANK % NUM_OF_TOKENS_PER_CHUNK == 0).
 COPY docker/common/deepep.patch /opt/deepep.patch
-ARG DEEPEP_COMMIT=17cfb817bccec3a9c247013360cc550c2bac441e
+ARG DEEPEP_COMMIT=42144303752422ade37f24bca9e2dde12df70e09
 ENV HYBRID_EP_MULTINODE=1
 ENV RDMA_CORE_HOME=/opt/rdma-core/build
 ENV LD_LIBRARY_PATH=/usr/local/cuda/lib64/:$LD_LIBRARY_PATH
--- a/examples/llm_finetune/glm/glm_4.7_flash_te_packed_sequence.yaml
+++ b/examples/llm_finetune/glm/glm_4.7_flash_te_packed_sequence.yaml
@@ -45,6 +45,11 @@ model:
     dispatcher: deepep
     fake_balanced_gate: false
     enable_hf_state_dict_adapter: true
+    # WAR for https://github.com/NVIDIA-NeMo/Automodel/issues/2674:
+    # fused TE RoPE emits NaN/Inf on the MLA single-head k_pe in THD (packed-sequence)
+    # mode, NaN-ing step-0 loss/grad and collapsing MoE routing -> OOM.
+    # Non-fused RoPE (fp32 unit-magnitude complex multiply) is bounded.
+    rope_fusion: false
 
 autopipeline:
   _target_: nemo_automodel.components.distributed.pipelining.config.PipelineConfig
--- a/examples/llm_finetune/glm/glm_5.1_hellaswag_pp.yaml
+++ b/examples/llm_finetune/glm/glm_5.1_hellaswag_pp.yaml
@@ -52,9 +52,6 @@ model:
     rope_fusion: false
     experts: gmm
     dispatcher: hybridep
-    # Static per-rank token budget so HybridEP's num_permuted_tokens is rank-uniform; every EP
-    # rank sizes its permute buffer identically, fixing the PP+EP NCCL deadlock at ~step 12.
-    moe_expert_rank_capacity_factor: 2.0
     fake_balanced_gate: false
     enable_hf_state_dict_adapter: true
     enable_fsdp_optimizations: true
--- a/examples/llm_finetune/glm/glm_5.1_hellaswag_pp.yaml
+++ b/examples/llm_finetune/glm/glm_5.1_hellaswag_pp.yaml
@@ -67,25 +67,29 @@ dataset:
   _target_: nemo_automodel.components.datasets.llm.hellaswag.HellaSwag
   path_or_dataset: rowan/hellaswag
   split: train
-  pad_to_max_length: false
+  pad_to_max_length: true
 
 packed_sequence:
-  packed_sequence_size: 1024
+  packed_sequence_size: 0
 
 dataloader:
   _target_: torchdata.stateful_dataloader.StatefulDataLoader
-  collate_fn: nemo_automodel.components.datasets.utils.packed_sequence_thd_collater
+  collate_fn:
+    _target_: nemo_automodel.components.datasets.utils.default_collater
+    pad_seq_len_divisible: 160
   shuffle: true
 
 validation_dataset:
   _target_: nemo_automodel.components.datasets.llm.hellaswag.HellaSwag
   path_or_dataset: rowan/hellaswag
   split: train
-  pad_to_max_length: false
+  pad_to_max_length: true
 
 validation_dataloader:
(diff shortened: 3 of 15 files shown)


akoumpa:
## Loss-corruption diagnosis — MiniMax/GLM `*_hellaswag_pp`

A pretrained model should start hellaswag SFT at loss **~3**. Observed **step-0 loss** across configs (8-node eos):

| run | `rope_fusion` | static shapes | AC | start loss |
|---|---|---|---|---|
| m2.7 vanilla | false | **none (variable)** | no | **3.18** ✅ |
| m2.5 vanilla | **true** | none (variable) | no | 15.6 ❌ |
| m2.7 packed / BSHD / BSHD-no-AC | false | packed **or** `pad_to_max_length` | yes/no | ~13.6 ❌ |
| m2.1 (all `rope_fusion:false`) | false | packed **or** `pad_to_max_length` | yes | ~15.3 ❌ |

**Two independent corrupters:** `rope_fusion: true`, and the **static-shape padding itself**.

The *only* clean config is **variable** shapes — which is exactly what deadlocks, so the deadlock fix and the loss corruption are coupled through padding. `num_label_tokens` is identical clean-vs-broken (15349), so the real tokens are simply predicted worse (grad_norm explodes 10 → ~6000) → the extra padding is poisoning the forward. Prime suspect: the standing `pad_token_id=None` warning ("incorrect MoE auxiliary loss … if valid tokens share token ID 0") — `pad_to_max_length` routes far more token-0 padding through the MoE.

- **Activation checkpointing ruled out** (m2.7 no-AC still 13.66 ≈ AC's 13.59).
- **RoPE**: `freqs_cis` buffer is genuine fp32, but `apply_rotary_emb` was downcasting cos/sin to bf16 in the rotation (~0.5%, washed out by bf16 q/k → not the cause; fp32-rotation fix included anyway).
- **In progress**: m2.5 with `rope_fusion: false` to isolate rope_fusion vs the MiniMax-common padding corruption.

akoumpa:
## Update — rope_fusion ruled out; m2.1/m2.5 garbage is **checkpoint-specific**

Ran m2.5 with **only** `rope_fusion: false` (else original: variable shapes, no AC, torch_mm, ep32) → still **15.74** (grad_norm 1690), job `344418203`. That's the **identical config** to m2.7 vanilla (**3.18**) — so the only difference is the model checkpoint. Revised conclusion (two clean isolations):

- **m2.1 / m2.5 — checkpoint/cache issue.** They load garbage (~15) regardless of rope / static-shapes / AC / experts. M2.7 loads correctly (~3) through the *identical* MiniMaxM2 code path, so it's the **M2.1/M2.5 fp8 checkpoints/cache**, not the recipe (likely incomplete/corrupt eos cache shards — cf the nemotron_ultra_v3 pattern; needs a cluster-side cache/weight-stats check). `rope_fusion: true` was a **red herring** — the earlier m2.5-vs-m2.7 gap was the *model*, not rope.
- **m2.7 — static-shape padding poisons the MoE.** Good checkpoint: 3.18 with variable shapes, ~13.6 with static shapes (pad_to_max_length *or* packed), same model. Prime suspect: `pad_token_id=None` → token-0 padding routed through experts / aux-loss. This is the real recipe-level loss bug.

So the corrected table is: **the only thing that decides loss is (a) the model checkpoint [M2.7 good, M2.1/M2.5 bad] and (b) static-shape padding [corrupts the good one].** rope_fusion / gmm / AC / capacity-factor are all ruled out.

akoumpa:
## ROOT CAUSE FOUND: `ep_size=32` corrupts the MoE forward

Controlled isolation on **M2.5** — only `ep_size` changed, everything else fixed (same checkpoint, tokenizer, config, recipe):

| m2.5 | step-0 loss | grad_norm |
|---|---|---|
| `ep_size: 32` | **15.74** ❌ | 1690 |
| `ep_size: 16` | **3.22** ✅ | 10.2 |

So minimax_m2.1/m2.5 have garbage loss **because they run `ep_size=32`**; the working m2.7 runs `ep_size=16`. `ep_size` is supposed to be loss-neutral, so this is an **EP-path correctness bug at 32-way (4-node internode) expert parallelism** — token mis-routing in the deepep/hybridep dispatch, or wrong expert sharding at 8 local-experts/rank. (Same internode EP path as the original deadlock — it both *hangs* and silently *corrupts* the forward at ep32.)

**5 hypotheses ruled out first:** chat-template (SFT never calls apply_chat_template — raw ctx+tgt), tokenizer (byte-identical), model-identity (all same interleaved-thinking family), fp8-dequant (sane/verified on the eos cache), config/code-path (identical except inert max_pos/dtype). Checkpoints verified intact on eos (the M2.1 "extra shards" are empty + ignored).

(ep16 OOMs in backward — 16 experts/GPU vs 8 — but logs step-0 first; needs pp/AC for memory.)

akoumpa:
## Reverse confirmation — bidirectional proof, model-independent

Forced the **good** model (m2.7, normally 3.18 at ep16) to `ep_size=32`, nothing else changed → step-0 loss **14.96** (then 11.9, 11.5, … 7.8; ran clean to step 19, no OOM since ep32 = 8 experts/rank fits). So ep32 corrupts even the model that works.

| run | ep16 | ep32 |
|---|---|---|
| m2.5 | **3.22** ✅ | 15.74 ❌ |
| m2.7 | 3.18 ✅ | **14.96** ❌ |

Both directions isolate `ep_size` as the sole cause ⇒ **pure ep_size, model-independent** — not the M2.1/M2.5 checkpoints.

**Localized to the dispatch path (not sharding):** expert sharding/placement is provably ep-degree-independent — 256 experts divides evenly by both 16 and 32, so the global→local expert mapping uses the identical contiguous-block formula at both degrees (verified byte-identical to the clean tree, no ep-conditional branch anywhere in placement). The corruption is in the **`GroupedExpertsDeepEP` token-dispatcher all-to-all** (`token_dispatcher.py`, data-dependent buffer sizing) at the 32-rank / 4-node EP group — the same HybridEP machinery that previously *deadlocked* there. Decisive confirm-test: ep32 + `fake_balanced_gate:true` vs ep16-balanced.

**Fix direction:** localize the data-dependent all-to-all in the flex token-dispatcher, or run these recipes at ep16 (+ pp/AC for memory headroom).

akoumpa:
## Code localization + fix (final)

Two independent code audits — expert placement, and the dispatch/combine path — both come back clean and **byte-identical to `main`**, which pins the defect precisely:

- **Expert placement / sharding** — ep-degree-independent. `256 % 16 == 256 % 32 == 0`, so the global→local expert mapping is the same contiguous-block formula at both degrees; no ep-conditional branch anywhere. ✔ not the bug.
- **Python dispatch/combine wrapper** (`_HybridEPManager`, `megatron/token_dispatcher.py`) — ep-invariant (only `num_local_experts` changes, passed through correctly); combine inverts dispatch via the per-call opaque `handle`. md5-identical to the clean tree. ✔ not the bug.
- **⇒ The corruption is in the compiled HybridEP _internode_ dispatch/combine kernel** (`HybridEPBuffer.dispatch_with_permute` / `combine_with_unpermute`, built from DeepEP `17cfb817` + `HYBRID_EP_MULTINODE=1`) — the only component whose behavior differs between a 2-node (ep16 ✅) and 4-node (ep32 ❌) EP group, and the same internode path that previously *deadlocked* at ep32.
- **Buffer-sizing excluded:** the receive buffer is sized from the local `seq_len` (`fused_a2a.py:361,393`) and reallocs only when a *later* batch exceeds the first — but the corruption is already at **step 0**, where the buffer is freshly sized for that exact batch. So it's the kernel's 4-node dispatch math, not buffer realloc.

### Fix
- **Immediate / proven:** run these MiniMax-M2 recipes at **`ep_size=16`** (correct loss ~3.2), using `pp_size` / activation-checkpointing / extra nodes for the memory headroom (ep16 = 16 experts/GPU vs 8). **Avoid `ep_size=32` / 4-node EP** until the kernel is fixed.
- **Root fix (kernel, outside this repo's Python):** correct or re-pin the DeepEP HybridEP internode build so 4-node dispatch/combine is functionally correct.

### CI gap
CI only asserts the job *completes*, so ep32 silently false-passes while mistraining (loss ~15). A step-0 loss sanity bound (e.g. `assert start_loss < 6`) would catch this whole class.

PR #2686, opened 2026-06-21
fix(moe): free DeepEP buffer before process-group destroy to avoid hang

akoumpa:
## What

DeepEP keeps a process-global communication `Buffer` (NVSHMEM symmetric memory plus its own NCCL sub-groups) that was never torn down. As a result `torch.distributed.destroy_process_group()` **hangs** on the leftover collective state at process exit, so callers work around it by skipping `destroy` entirely (e.g. the checkpoint-robustness test `atexit.unregister`s it).

This wires up a clean DeepEP teardown:

- `fused_a2a.get_buffer()` constructs the `Buffer` with `explicitly_destroy=True` so it can be released.
- New `fused_a2a.free_buffer()` destroys the buffer, freeing the NVSHMEM/cpp runtime.
- `destroy_global_state()` frees the DeepEP buffer **before** `destroy_process_group()`, which lets teardown complete without hanging. It is a no-op when DeepEP was never used.

## Validation

- **8×H100**: a DeepEP finetune now exits cleanly through the `atexit` teardown — `destroy_global_state()` completes on all ranks. Previously the bare `destroy_process_group()` hung; destroying the buffer first makes it return.
- **Unit tests** (new): `free_buffer` behavior (destroy + clear, no-op when unset, error-swallowing) and `destroy_global_state` ordering (buffer freed before PG destroy; PG destroy still runs if `free_buffer` raises).

## Scope

This fixes the DeepEP **teardown hang** only. It does **not** turn the `qwen3_moe_30b_lora` checkpoint-robustness CI job green — that failure is a separate issue: DeepEP training leaves GPU-allocation-level state that corrupts a *same-allocation* HF reload (it survives process exit, `Buffer.destroy()`, and `destroy_process_group()`, and is cleared only by a fresh Slurm job). Making that test reliable requires running its HF-reload comparison in a separate Slurm job, which is tracked separately.

### DIFF
--- a/nemo_automodel/components/distributed/init_utils.py
+++ b/nemo_automodel/components/distributed/init_utils.py
@@ -155,9 +155,22 @@ def destroy_global_state():
 
     This function is registered to execute at exit to ensure the process group is properly destroyed.
     It temporarily ignores SIGINT to avoid interruption during cleanup.
+
+    For MoE runs that use DeepEP, the process-global DeepEP ``Buffer`` (NVSHMEM symmetric memory
+    plus its own NCCL sub-groups) is destroyed *before* ``destroy_process_group()``. Tearing the
+    buffer down first releases that pending collective state; otherwise ``destroy_process_group()``
+    hangs on it. Freeing the buffer is a no-op when DeepEP was never used.
     """
     # Don't allow Ctrl+C to interrupt this handler
     signal.signal(signal.SIGINT, signal.SIG_IGN)
+    try:
+        from nemo_automodel.components.moe.megatron.fused_a2a import free_buffer
+
+        if torch.cuda.is_available():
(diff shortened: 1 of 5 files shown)


akoumpa:
Exercised the DeepEP teardown on real CI — nemo-ci pipeline on this branch running `qwen3_moe_30b_hellaswag` (DeepEP, `ep_size=8`):

https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/55374481

This runs the new teardown path (`free_buffer` → `destroy_process_group`) at exit on a real DeepEP run as no-regression evidence. The fix itself is additionally covered by the new unit tests here and by an 8×H100 finetune that now exits cleanly through the patched `atexit` handler.

PR #2690, opened 2026-06-21
perf(checkpoint): mmap HF DCP read_data to avoid host-RAM OOM on large loads

akoumpa:
## What does this PR do ?
mmap-based `read_data` in the HF DCP storage reader to cut **host RAM** during checkpoint load and avoid host-OOM-kills on very large (fp8 MoE) checkpoints.

## Problem
`_HuggingFaceStorageReader.read_data` (`_backports/hf_storage.py`) read **every full tensor** into an anonymous host `bytearray` (plus a second copy from `bytearray(...)`), then narrowed to the rank's shard *afterward*:
```python
tensor_bytes = bytearray(stream.read(item_md.length))   # whole tensor -> anonymous host RAM (x2)
tensor = torch.frombuffer(tensor_bytes, ...).reshape(shape)
tensor = narrow_tensor_by_index(tensor, ...)            # narrow AFTER the full read
```
For very large checkpoints (e.g. GLM-5.1 355B fp8, 8 ranks/node) this exhausts **host** RAM and the kernel OOM-killer SIGKILLs ranks mid-load — while GPU memory is still ~half free.

## Fix
mmap local files and `torch.frombuffer` **only each request's narrowed slice**, so only those pages fault in — and they are **file-backed (reclaimable)** rather than anonymous. No full-tensor read, no second copy.
- Aligned offsets -> zero-copy `frombuffer`.
- Unaligned offset (safetensors header can misalign e.g. bf16) -> per-tensor copy fallback (correctness preserved).
- Remote / non-local files -> keep the existing streaming path.

## Verification (eos)
- **GLM-5.1 (355B, ep64 / 32 nodes):** previously SIGKILL'd (host OOM) *during load* (~4 min, before step 0); with mmap it loads through to step 0. Load-time host-OOM cleared.
- **MiniMax-M2.5 (ep32, 8 nodes):** loads **byte-identically** -> step-0 loss **3.22** (same as without mmap). No regression.
- Unit-level: aligned zero-copy + unaligned fallback + `narrow`->`copy_` verified byte-correct.

## Notes
- The reader is shared by all DCP HF loads, so this benefits any large sharded load (fp8 or bf16), not just GLM.
- Independent of the fp8 expert-load correctness fix (#2682); orthogonal change.

### DIFF
--- a/nemo_automodel/components/checkpoint/_backports/hf_storage.py
+++ b/nemo_automodel/components/checkpoint/_backports/hf_storage.py
@@ -19,6 +19,7 @@
 import dataclasses
 import json
 import logging
+import mmap
 import os
 import queue
 import re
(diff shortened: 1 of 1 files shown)


akoumpa:
## Verification jobs (eos)

- **GLM-5.1 (355B, ep64/32-node)** — previously host-OOM-killed *mid-load*; with mmap it loads through and trains to completion (loss 9.76→~0.0005): [glm job 344567935](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/344567935)
- **MiniMax-M2.5 (ep32, 8-node)** — mmap load is byte-identical, step-0 loss 3.22 (no regression): [m2.5 job 344552070](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/344552070)

PR #2703, opened 2026-06-22
fix(distributed): register DeciLM Nemotron TP plan

akoumpa:
## Summary
- register the remote-code `DeciLMForCausalLM` class name to use the existing DeciLM/Nemotron TP plan automatically
- keep the fix in Python code, so Llama 3.3 Nemotron Super recipes do not need `distributed.tp_plan` overrides
- add unit coverage for the automatic bare-class-name path used by `trust_remote_code=True` models

## Context
This addresses NeMo-CI job https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/337980628, which failed because `DeciLMForCausalLM` was loaded from `transformers_modules.*` at `tp_size=2` and no registered plan was found.

## Validation
- `pytest -q tests/unit_tests/distributed/test_get_parallel_plan.py`
- direct `_get_parallel_plan()` smoke check for a class named `DeciLMForCausalLM` with `model_type="nemotron-nas"`
- `git diff --check origin/main..HEAD`

PR #2626, opened 2026-06-17
fix(checkpoint): super-49B consolidated reload and vllm_deploy

adil-a:
## What

Makes the **llama-3.3-nemotron-super-49B** (DeciLM / nemotron-nas) `checkpoint_robustness` and `vllm_deploy` CI jobs pass. The finetune died first (custom-code arch, no default TP plan → `tp_size>1` DTensor assert → cascade to "No checkpoint"), and wiring the TP plan exposed two further blockers, plus a deploy sizing wall.

## Fixes

1. **recipe `tp_plan`** — `distributed.tp_plan: llama_nemotron_super_tp_plan` selects the registered DeciLM/nemotron-nas plan so the finetune trains and saves.

2. **`model_init.py` — full custom-code closure into the HF module cache.** HF's `get_cached_module_file` copies a custom-code file plus its *direct* relative imports, but `get_class_in_module` validates the *transitive* closure. DeciLM reaches `transformers_4_44_2__configuration_llama.py` only via `configuration_decilm`, so it was never copied into the model's cache dir and the multi-rank robustness reload died with `FileNotFoundError`. `_prepopulate_remote_code_cache` copies every `.py` from the checkpoint dir into the resolved cache dir under `FirstRankPerNode`.

3. **checkpoint `addons.py` — guard removed transformers symbols.** The deploy container (`vllm:25.12`, transformers 4.57) removed `NEED_SETUP_CACHE_CLASSES_MAPPING`, which DeciLM's `modeling_decilm.py` imports. Consolidation now guards such imports in the copied code (defines the symbol if absent). Verified it's the only import-time removal in 4.57.

4. **deploy on native vLLM across 4 GPUs.** The 49B (98 GB) OOMs loading the HF reference on the deploy harness's single pinned GPU, and forcing `model_impl="transformers"` + TP is the *unsupported* path for DeciLM (no `base_model_tp_plan`). vLLM ships a **native** `nemotron_nas`/`DeciLMForCausalLM` backend that tensor-parallels cleanly, so the 49B deploy now runs through the existing `vllm_smoke_test` path (native backend, no HF reference) on 4 GPUs:
   - recipe: `ci.vllm_smoke_test: true` + `ci.vllm_deploy_gpus: 4`
   - `vllm_launcher.sh`: expose `ci.vllm_deploy_gpus` GPUs (default 1; every other deploy unchanged)
[...]

PR #2719, opened 2026-06-22
ci: raise super-49B robustness time limit to 1h15m

adil-a:
## What

Raise the `ci.time` limit for the `llama3_3_nemotron_super_49B_squad` checkpoint-robustness recipe from `00:55:00` to `01:15:00`.

**Targets #2703** (`akoumpa/am-436-llama33-super`) so this rides along with the DeciLM Nemotron TP-plan work rather than as a standalone change. (The original 49B fix landed in #2626, which merged with `ci.time` set to `00:55:00`.)

## Why

`00:55:00` is **too tight**: the finetune + robustness producer runs **~51–55 min wall-clock** across measured runs:
- pipeline 55194011 (the run behind #2626): ~51 min
- pipeline 55467887 (1h-limit experiment): **55.1 min**

A 55-min limit risks SLURM-cancelling the job right at the finish line (the 55.1-min run would have been killed at 55:00). `01:15:00` gives ~20 min of headroom for Lustre/queue variance while staying far tighter than a blanket 2h.

`ci.time` only sets the SLURM wall-clock ceiling — it does not change test behavior.

### DIFF
--- a/examples/llm_finetune/nemotron/llama3_3_nemotron_super_49B_squad.yaml
+++ b/examples/llm_finetune/nemotron/llama3_3_nemotron_super_49B_squad.yaml
@@ -119,7 +119,7 @@ lr_scheduler:
 ci:
   recipe_owner: akoumpa
   nodes: 2
-  time: "00:55:00"  # 49B robustness (Phase 1-4 reloads of a 49B on 2 nodes) needs >45min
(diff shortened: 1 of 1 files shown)


akoumpa (approved):
thanks @adil-a

PR #2687, opened 2026-06-21
fix(qwen3_moe): disable fused RoPE for the packed-sequence LoRA recipe

akoumpa:
## Problem

`qwen3_moe_30b_te_packed_sequence_lora` (nemo-ci `peft`, eos) fails at step 1 with a per-rank **CUDA OOM** in the DeepEP token-permute + a `DeepEP error: CPU recv timeout` cascade. The OOM is a **symptom, not the cause**.

## Root cause

**Fused TE RoPE emits NaN/Inf on the query under THD (packed-sequence) mode.** The chain:
1. Step 0: the fused-RoPE backward produces a non-finite **query** gradient (`grad_norm=nan` while `loss` is finite ~3.38).
2. `clip_grad_norm` (global norm = NaN) smears it across **all** LoRA grads; the unconditional optimizer step writes NaN into every LoRA weight.
3. Step 1: NaN weights → routing collapses onto a single EP rank → that rank floods → **OOM** in the token-permute; other ranks hang at the DeepEP combine barrier → recv timeout.

This is the **same root cause and fix as #2675** ("disable fused RoPE for MLA packed-sequence MoE recipes", which fixed Moonlight-16B and GLM-4.7-Flash). qwen3_moe was not covered there, and its recipe doesn't set `rope_fusion`, so it defaults to `True` (TE present).

### How it was localized
Per-parameter and per-module **backward grad-finiteness** traces on eos (8×H100):
- First non-finite grad = `model.layers.N.self_attn.q_proj.lora_A` — **q-only** (`k_proj`/`v_proj`/`o_proj` and the MoE experts all finite); origin layer varies run-to-run (non-determinism).
- `q_proj`/`q_norm` forward inputs, outputs, and weights are all **finite**; only the **grad-input** (gradient arriving from the upstream RoPE backward) is NaN — so the NaN is manufactured upstream of `q_norm`, in RoPE.
- Independent of the GEMM kernel (`gmm` == `torch_mm`), the EP dispatcher (`deepep` == `hybridep`), and the SDPA backend (forcing `NVTE_FUSED_ATTN`/`NVTE_FLASH_ATTN` left it unchanged) — i.e. **not** the MoE/dispatcher and **not** the attention kernel.

## Fix

Set `rope_fusion: false` on this recipe → the bounded fp32 non-fused RoPE path.

## Validation (eos, 8×H100)

Before: step 0 `grad_norm nan`, step 1 OOM + DeepEP recv timeout → job fails.
[...]

PR #2723, opened 2026-06-22
fix(qwen3_moe): disable fused RoPE on packed-sequence recipes (step-0 NaN grad)

hemildesai:
## Problem

Fused TransformerEngine RoPE is numerically unstable on THD/packed sequences:
the forward stays finite, but the **backward emits NaN gradients** at step 0
(`grad_norm nan` with a finite loss), which NaNs training from step 1 onward.

The qwen3-moe-30b packed-sequence recipes don't set `rope_fusion`, so they
inherit the default `rope_fusion = HAVE_TE and torch.cuda.is_available()` →
`True` on a GPU+TE build, and hit the bad path.

## Fix

Pin `rope_fusion: false` in the `model.backend` block of the three affected
packed recipes so they use the bounded non-fused RoPE path:

- `qwen3_moe_30b_te_hybridep_mxfp8.yaml`
- `qwen3_moe_30b_te_hybridep.yaml`
- `qwen3_moe_30b_te_packed_sequence.yaml`

This mirrors #2687, which applied the identical one-line fix to the sibling
`qwen3_moe_30b_te_packed_sequence_lora` recipe. See #2674 for the underlying
fused-RoPE-on-THD issue. (The other packed qwen3-moe recipes — `magi_*`,
`flashoptim`, `lora` — already set `rope_fusion: false`.)

## Verification

GB200 Qwen3-MoE-30B (Qwen3-30B-A3B) finetune, `ep_size=8`, packed sequences.
Before: step-0 `grad_norm nan` (bf16) / step-0 `loss nan` (MXFP8). After
`rope_fusion: false`, all of bf16, TE-MXFP8, and torchao-MXFP8 train cleanly
and converge identically:

| precision | step-0 loss | step-0 grad_norm | → step-2 loss |
|---|---|---|---|
| bf16 | 3.04 | 13.88 | 2.23 |
| TE-MXFP8 | 3.04 | 13.81 | 2.22 |
| torchao-MXFP8 | 3.05 | 15.69 | 2.23 |

(bf16 ran the full 50 steps: loss 3.04 → 1.89, grad_norm → ~1.1, zero NaN.)

Refs nvbug 6350293. Companion to #2722 (a separate checkpoint-load bug that
also surfaced on these recipes).

### DIFF
--- a/examples/llm_finetune/qwen/qwen3_moe_30b_te_hybridep.yaml
+++ b/examples/llm_finetune/qwen/qwen3_moe_30b_te_hybridep.yaml
@@ -48,6 +48,7 @@ model:
     rms_norm: torch_fp32
     experts: te
     dispatcher: hybridep
+    rope_fusion: false  # fused TE RoPE produces NaN gradients on packed (THD) sequences
     fake_balanced_gate: false
     enable_hf_state_dict_adapter: true
 
--- a/examples/llm_finetune/qwen/qwen3_moe_30b_te_hybridep_mxfp8.yaml
+++ b/examples/llm_finetune/qwen/qwen3_moe_30b_te_hybridep_mxfp8.yaml
@@ -51,6 +51,7 @@ model:
     rms_norm: torch_fp32
     experts: te
     dispatcher: hybridep
+    rope_fusion: false  # fused TE RoPE produces NaN gradients on packed (THD) sequences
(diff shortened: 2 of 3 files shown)


hemildesai:
Folded into #2722 (the two fixes for nvbug 6350293 are now in a single PR).

Issue #2454, opened 2026-06-08
Training side support for Router Replay for Automodel


snowmanwwg:
Similar to https://github.com/NVIDIA/Megatron-LM/pull/4256

PR #2797, opened 2026-06-26
feat(moe): add Rollout Routing Replay (R3) for MoE RL training

khazic:
## What

Adds opt-in **Rollout Routing Replay (R3)** to the MoE subsystem.

In on-policy RL on a Mixture-of-Experts model, the rollout (inference) engine and the training engine compute the router's top-k expert selection independently. Small numerical differences flip a fraction of routing decisions per layer, and that compounds across layers until most tokens train under a different expert assignment than they were rolled out with. The mismatch breaks the importance-sampling assumption behind GRPO/GSPO and destabilizes training.

R3 removes the mismatch: capture the top-k expert **selection** during the rollout-equivalent forward, then replay that exact selection during the training forward. Only the discrete selection is replayed. The router logits and their softmax/sigmoid are still recomputed from the live router weights, so the gradient continues to flow into the router. This mirrors Megatron-LM's `moe_enable_routing_replay`.

## Changes

- `components/moe/router_replay.py` (new): a `RouterReplay` handle (RECORD / REPLAY modes) with a process-global per-layer registry plus `record()` / `replay()` context managers, and the `replay_selection` helper.
- `MoEConfig.enable_routing_replay` flag (default `False`).
- Hooks in the shared `Gate` (every `score_func` branch) and the custom `Gemma4Gate`, so both record and replay their final expert selection.
- Unit tests at 100% coverage of `router_replay.py`.

## Default path stays a no-op

When `enable_routing_replay=False` (the default), no `RouterReplay` handle is created and the gate forward is unchanged: `replay_selection` returns the original index tensor. The two branches that previously consumed `topk`'s returned values directly re-gather only when replay actually swapped the selection, guarded by an identity check, so the replay-off and record paths add no extra tensor work.

## Usage

```python
from nemo_automodel.components.moe.router_replay import RouterReplay

# Capture the selection on the rollout-equivalent forward.
with RouterReplay.record():
    model(batch)
captured = RouterReplay.collect()      # one tensor per MoE layer, in layer order

# Replay it on the training forward over the same tokens.
with RouterReplay.replay(captured):
    loss = model(batch)
loss.backward()
```

## Testing

Unit tests: `tests/unit_tests/moe/test_router_replay.py`, **27 tests, 100% coverage** of `router_replay.py`. They cover the record/replay roundtrip across every `score_func` branch, the default-path no-op, gradient flow to the router under replay, multi-layer distribute/collect, error handling, and the `Gemma4Gate` path.

Beyond the unit tests, I exercised the real Automodel model classes on GPU (record on batch A, replay on a different batch B):

| Check | Qwen3-MoE (`Qwen3MoeForCausalLM`, 3 MoE layers) | Gemma4-MoE (`Gemma4MoE` + `Gemma4Gate`) |
|---|---|---|
| Natural routing of A and B differs | yes | yes |
| Replay reproduces A's selection across all layers | yes | yes |
| Replay overrides B's natural selection | yes | yes |
| Router weight receives a gradient under replay | yes (CE-loss backward) | yes |

### Test environment

- 1x NVIDIA A800-80GB
- Python 3.12, PyTorch 2.12.0+cu130, Transformers 5.13.0.dev0
- `ruff format` and `ruff check` clean

Addresses #2454.

### DIFF
--- a/nemo_automodel/components/models/gemma4_moe/model.py
+++ b/nemo_automodel/components/models/gemma4_moe/model.py
@@ -83,6 +83,7 @@ from nemo_automodel.components.models.common.hf_checkpointing_mixin import HFChe
 from nemo_automodel.components.models.common.utils import cast_model_to_dtype
 from nemo_automodel.components.moe.fsdp_mixin import MoEFSDPSyncMixin
 from nemo_automodel.components.moe.layers import MoE, MoEConfig
+from nemo_automodel.components.moe.router_replay import RouterReplay, replay_selection
 from nemo_automodel.shared.utils import dtype_from_str as get_dtype
 
 from .cp_attention import attach_gemma4_cp_ring_attention, gemma4_vision_group_ids
@@ -194,12 +195,16 @@ class Gemma4Gate(nn.Module):
     This class reproduces that logic but returns (weights, indices, aux_loss) as expected by GroupedExperts.
     """
 
-    def __init__(self, config: Gemma4TextConfig):
+    def __init__(self, config: Gemma4TextConfig, enable_routing_replay: bool = False):
         super().__init__()
         hidden_size = config.hidden_size
         num_experts = config.num_experts
         self.topk = config.top_k_experts
         self.num_experts = num_experts
+
+        # Rollout Routing Replay (R3): owns a handle only when enabled so the
+        # default routing path stays a no-op.
+        self.router_replay: Optional[RouterReplay] = RouterReplay() if enable_routing_replay else None
 
         # Thread dtype explicitly from config.torch_dtype so the router's own
         # params (proj weight + scale) stay aligned with the rest of the model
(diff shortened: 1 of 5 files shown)


HuiyingLi:
I tested PR #2797 locally on 8x H100s.

Validation done:
- `tests/unit_tests/moe/test_router_replay.py -q`: 27 passed.
- Qwen3-MoE 30B-A3B EP8 training smoke test with the existing 8k mock config: 5/5 steps passed.
- Qwen3-MoE EP8 with real `Gate` and `enable_routing_replay=true`: 1/1 step passed.
- Added a temporary hook-based EP8 replay check on `MoE + ExpertParallel`:
  - recorded routing from rollout input
  - verified train input naturally routed differently
  - replayed rollout routing on train input
  - verified in-forward gate indices matched recorded rollout indices
  - verified replay changed the routed output
  - backward completed with finite grads

HuiyingLi (approved):

PR #2806, opened 2026-06-26
fix(docs): catch Fern MDX syntax errors

akoumpa:
## Summary
- add a no-secret MDX syntax validation step to the Fern docs PR check
- wire the same validation into `make docs-check`
- self-close raw `` tags that broke the publish job, plus the same latent issue in DiffusionGemma docs

## Evidence
- Failing publish job: https://github.com/NVIDIA-NeMo/Automodel/actions/runs/28267737434/job/83758291397
- Verified locally: `make -C docs/fern docs-check`
- Verified locally: `git diff --check`


chtruong814 (dismissed):

docs/fern/Makefile
akoumpa:
```suggestion
.PHONY: help docs docs-login docs-check docs-preview docs-publish docs-generate-library docs-stitch docs-mdx-check
```

PR #2820, opened 2026-06-28
chore: CI proxy for #2819

HuiyingLi:
CI-only proxy PR. Do not merge / do not review. Points at the exact head commit of #2819 (0c2973716444c0fb870af5c4441e145d944fc488) so internal CI runs under the internal-contributor queue; results post back to #2819 via the shared SHA. Source of truth: #2819. Close once CI completes.


HuiyingLi:
Closing CI-only proxy because source PR #2819 is merged.

PR #2822, opened 2026-06-28
chore: CI proxy for #2449

HuiyingLi:
CI-only proxy PR. Do not merge / do not review. Points at the exact head commit of #2449 (b2b8e970de43945c9525169fbcb793f80feec4f0) so internal CI runs under the internal-contributor queue; results post back to #2449 via the shared SHA. Source of truth: #2449. Close once CI completes.


HuiyingLi:
Closing CI-only proxy because source PR #2449 is merged.

PR #2823, opened 2026-06-28
fix(speculative): gather sharded target lm_head in EAGLE-1/2 token loss

khazic:
## What

`EagleTrainerModule.compute_logits` projects the predicted draft hidden states through the **target** `lm_head`:

```python
return F.linear(hidden_states, self._target_lm_head.weight)
```

When the EAGLE-1/2 recipe shards the target via its optional `distributed:` path (the documented route for targets that do not fit on a single GPU, e.g. a Qwen3-30B-A3B MoE target), `lm_head.weight` becomes a `DTensor`. The draft runs under DDP, so `hidden_states` is a plain tensor, and the projection fails with:

```
RuntimeError: aten.mm.default got mixed torch.Tensor and DTensor, need to convert all torch.Tensor to DTensor
```

The EAGLE-3 path already handles the analogous case in `copy_embeddings_from_target` by gathering with `.full_tensor()`; the EAGLE-1/2 token-loss head was missing the same guard.

## Fix

Gather the frozen target weight to a local full tensor before `F.linear`:

```python
weight = self._target_lm_head.weight
if hasattr(weight, "full_tensor"):
    weight = weight.full_tensor()
return F.linear(hidden_states, weight)
```

It is a no-op when the target is unsharded, so the common single-GPU-per-rank dense path is unaffected.

## Test

Adds `test_compute_logits_gathers_sharded_dtensor_target_lm_head`: inits a single-rank gloo group, shards the tiny target's `lm_head.weight` into a `DTensor`, and asserts `compute_logits` returns finite logits equal to the dense `F.linear` reference. Verified the test fails on the old code path (the mixed Tensor/DTensor error above) and passes with the fix.

```
tests/unit_tests/speculative/test_eagle12.py ....  4 passed
```

### DIFF
--- a/nemo_automodel/components/speculative/eagle/core_v12.py
+++ b/nemo_automodel/components/speculative/eagle/core_v12.py
@@ -69,7 +69,16 @@ class EagleTrainerModule(nn.Module):
 
     def compute_logits(self, hidden_states: torch.Tensor) -> torch.Tensor:
         """Project predicted hidden states through the frozen target lm_head."""
-        return F.linear(hidden_states, self._target_lm_head.weight)
+        weight = self._target_lm_head.weight
+        # The target may be FSDP2-sharded (the EAGLE-1/2 recipe's optional
+        # ``distributed:`` path, used for targets that do not fit on one GPU),
+        # which makes ``weight`` a ``DTensor``. The draft runs under DDP, so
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
Thank you!

PR #2824, opened 2026-06-28
fix(speculative): validate DFlash mask_token_id on checkpoint resume

khazic:
## What

DFlash fills every non-anchor slot of a `[anchor, MASK, MASK, ...]` block with `mask_token_id`, and the draft's `embed_tokens` row at that id becomes the learned "predict here" signal. The inference runtime must fill block slots with the **same** id.

`_save_extra_state` already persists `mask_token_id` into `dflash_meta.pt`, but `_load_extra_state` restored only `global_step` and `epoch`:

```python
self.runtime.global_step = int(meta.get("global_step", 0))
self._resume_epoch = int(meta.get("epoch", 0))
```

On resume `mask_token_id` is re-read solely from the YAML (`_resolve_mask_token_id`). If the resume YAML carries a different `mask_token_id` than the checkpoint was trained with, the mask slots silently point at an untrained embedding row and acceptance degrades with no error.

## Fix

Validate the saved `mask_token_id` against the configured one in `_load_extra_state` and raise on mismatch. Legacy checkpoints saved before this field existed (`None`) skip the check, so older checkpoints still resume.

## Test

Adds four cases to `test_train_dflash.py`: step/epoch restore on match, `ValueError` on mismatch, legacy meta without `mask_token_id` accepted, and no-op when the meta file is absent.

```
tests/unit_tests/recipes/llm/test_train_dflash.py  24 passed
```

### DIFF
--- a/nemo_automodel/recipes/llm/train_dflash.py
+++ b/nemo_automodel/recipes/llm/train_dflash.py
@@ -472,12 +472,28 @@ class TrainDFlashRecipe(BaseRecipe):
         self._load_extra_state(ckpt_dir)
 
     def _load_extra_state(self, ckpt_dir: str) -> None:
-        """Restore DFlash meta: global_step and epoch."""
+        """Restore DFlash meta: global_step and epoch, and validate mask_token_id."""
         meta_path = os.path.join(ckpt_dir, "dflash_meta.pt")
         if os.path.exists(meta_path):
             meta = torch.load(meta_path, weights_only=False, map_location="cpu")
             self.runtime.global_step = int(meta.get("global_step", 0))
             self._resume_epoch = int(meta.get("epoch", 0))
+            # ``mask_token_id`` comes only from the resume YAML (it is not
+            # restored from the checkpoint); the draft's ``embed_tokens`` row at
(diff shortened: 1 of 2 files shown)


khazic:
The only failing check here is codecov/patch (33.33% of the 3 new lines in _load_extra_state). This looks like a spurious coverage-collection result rather than a real gap:

- All three new lines are exercised by the four unit tests added in this PR, and they pass in L0_Unit_Tests_CPU of this run, including test_load_extra_state_raises_on_mask_token_id_mismatch, which triggers the raise.
- The per-line coverage Codecov has for train_dflash.py is internally inconsistent: several function bodies are marked covered while the same function's def/guard lines are marked uncovered (for example _maybe_save_step_checkpoint and _load_extra_state), and module-level imports come back mixed hit/miss. That is impossible for straight-line execution, so the 33.33% patch number comes from misaligned line data, not a missing test.

Could you re-run the Codecov job (or waive codecov/patch for this PR)? I am happy to push a no-op commit to re-trigger if that is easier.

HuiyingLi (approved):

PR #2835, opened 2026-06-29
chore: CI proxy for #2827

HuiyingLi:
CI-only proxy PR. Do not merge / do not review. Points at the exact head commit of #2827 (452dd57b407391cdbcff47b1815488cc00302f7a) so internal CI runs under the internal-contributor queue; results post back to #2827 via the shared SHA. Source of truth: #2827. Close once CI completes.


HuiyingLi:
Closing CI-only proxy because source PR #2827 is merged.

PR #2836, opened 2026-06-29
chore: CI proxy for #2826

HuiyingLi:
CI-only proxy PR. Do not merge / do not review. Points at the exact head commit of #2826 (81c067985a9126bf8b1c27ab5cfb7fe669a84049) so internal CI runs under the internal-contributor queue; results post back to #2826 via the shared SHA. Source of truth: #2826. Close once CI completes.


HuiyingLi:
Closing CI-only proxy because source PR #2826 is merged.

PR #2841, opened 2026-06-29
feat(speculative): add serve_vllm for EAGLE-3 / P-EAGLE drafts

khazic:
## What

Adds `nemo_automodel/components/speculative/serve_vllm.py`, the vLLM companion to
`serve_sglang`. It serves an Automodel-trained EAGLE-3 / P-EAGLE drafter under
vLLM's parallel-drafting runtime.

## Why

The P-EAGLE recipe (`parallel_drafting: true`) produces a head that SGLang cannot
serve today, so `serve_sglang` rejects it and points at vLLM. There was no vLLM
entry point, so a trained P-EAGLE draft could not actually be served or
benchmarked (the `qwen_peagle_perfectblend.yaml` recipe comment already notes
that acceptance can only be measured by running the head under vLLM). This closes
that gap.

## How

`serve_vllm` resolves a recipe checkpoint (`epoch_*/model[/consolidated]`),
applies three fixups so vLLM can load the draft, then shells out to
`vllm.entrypoints.openai.api_server` with the right `--speculative-config`:

1. `architectures`: rewrite Automodel's `LlamaEagle3DraftModel` to vLLM's
   registered `LlamaForCausalLMEagle3`.
2. `pard_token` (P-EAGLE): vLLM's parallel-drafting proposer reads the masked-slot
   id from `pard_token` / `ptd_token_id` / `dflash_config.mask_token_id`, but the
   recipe writes only the top-level `mask_token_id`. The value is copied over,
   mirroring vLLM's own speculators `update_peagle` mapping.
3. weight keys: Automodel wraps the draft as `self.model`, so weights are saved as
   `model.*`; vLLM re-adds `model.` to every non-top-level weight, double-prefixing
   them. The prefix is stripped into a one-time `vllm_export/` copy, leaving the
   source checkpoint untouched. A draft whose weights are already vLLM-standard
   skips the export and gets the config fixups in place.

The draft already stores the `d2t` / `t2d` vocab-remap buffers in the offset form
vLLM expects, so (unlike the SGLang path) no separate speculative-token-map file
is emitted.

## Validation

Trained a standard P-EAGLE Qwen3-8B draft and served it through `serve_vllm` on
vLLM's parallel-drafting runtime: the draft loads and parallel drafting runs,
producing a measurable acceptance length end to end. New unit tests cover artifact
resolution, all three fixups, the export vs in-place split, export caching, and
CLI dispatch (`--print-only`, missing-vLLM exit, subprocess / execv dispatch).

## Notes

- vLLM is intentionally not bundled with the container and not declared in
  `pyproject.toml`; the script exits with an install hint when it is missing.
- P-EAGLE inference requires vLLM >= 0.16 (parallel-drafting runtime).

### DIFF
--- a/nemo_automodel/components/speculative/serve_vllm.py
+++ b/nemo_automodel/components/speculative/serve_vllm.py
@@ -1,0 +1,542 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Serve an Automodel-trained EAGLE-3 / P-EAGLE drafter with vLLM.
+
+This is the vLLM companion to ``serve_sglang``. It exists because the
+parallel-drafting (P-EAGLE) head produced by the EAGLE-3 recipe with
+``parallel_drafting: true`` is *not* servable by SGLang today (``serve_sglang``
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
Thank you!

PR #2852, opened 2026-06-29
fix(data): allow text collate without Qwen VL utilities

hemildesai:
## Stack

4 of 7. Previous: #2851. Next: #2853.

## Summary

- Removes the Qwen VL utility requirement from the generic text-capable VLM collator.
- Keeps Qwen Omni-specific utility validation on the path that actually needs it.
- Adds coverage for text collation when `qwen_vl_utils` is absent.

## Validation

- `tests/unit_tests/datasets/vlm/test_collate_fns.py`: `127 passed`
- Commit hooks and `git diff --check`

### DIFF
--- a/nemo_automodel/components/datasets/vlm/collate_fns.py
+++ b/nemo_automodel/components/datasets/vlm/collate_fns.py
@@ -19,15 +19,12 @@ from unittest.mock import MagicMock
 import torch
 from PIL import Image as PILImage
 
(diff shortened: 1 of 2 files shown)


hemildesai:
Removing this unrelated collator change from the DeepEP v2 stack. We will keep the current main behavior for now and revisit it separately if needed.

PR #2850, opened 2026-06-29
feat(moe): DeepEP v2 dispatcher, Torch 2.12 PP, and selectable EP variants

hemildesai:
# What does this PR do ?

Collapses the six-PR DeepEP v2 upgrade stack into a single PR against `main`. It
adds the DeepEP v2 (ElasticBuffer) expert-parallel dispatcher, the dependency
machinery to select DeepEP v1 / DeepEP v2 / HybridEP at build time, Torch 2.12
pipeline and FSDP synchronization fixes, tuned communication defaults, benchmark
and finetuning recipes, a HybridEP JIT-cache bridge, and a selectable CUDA
development image. Preserved as 13 signed-off commits.

# Changelog

- **build(moe):** selectable DeepEP v1 / DeepEP v2 / HybridEP dependency variants as
  conflicting `uv` extras, with updated pins and lockfiles (`deepep_v1`, `deepep_v2`,
  `hybridep`; `moe` selects v2 by default).
- **feat(moe):** DeepEP v2 ElasticBuffer dispatcher — one process-global buffer sized
  from model/batch shape, reused across layers/steps, with checkpoint-safe handle
  tensor save/restore for non-reentrant activation checkpointing; tuned SM/QP defaults
  and QP configuration propagated through dispatch and backward combine.
- **fix(pp):** adapt pipeline execution and FSDP MoE gradient synchronization to Torch
  2.12 (final-microbatch reduction without duplicate gradient reduction; guards that
  keep older-torch unit tests passing).
- **test/perf(moe):** DeepEP v2 benchmark and finetuning recipes; Qwen3 30B DeepEP SM
  tuning.
- **perf(moe):** stable HybridEP JIT kernel-cache bridge (canonical cache staged into
  the per-process kernel dir, hard-linked back after compilation; cache errors never
  fail training).
- **build(docker):** selectable EP CUDA development image with hardlink-deduped
  per-backend venvs (three backends in one image at ~+0.9 GiB over a single backend).

# Validation

- DeepSeek V3 PP4/EP64 on 256 H100: DeepEP v2 ~29.1% MFU, at parity with DeepEP v1 and
  HybridEP.
- Qwen3 MoE 235B finetuning: 100-step convergence matching v1/HybridEP loss and
  grad-norm trends; numerical parity probes clean.
- Unit tests: dispatcher, PP, backend config, and HybridEP cache suites pass; CI on this
  branch is green (it validated the full cumulative tree).

# Before your PR is "Ready for review"

Pre checks:
[...]

PR #2854, opened 2026-06-29
perf(moe): reuse HybridEP JIT cache

hemildesai:
**Consolidated into #2850.**

This PR was one layer of the DeepEP v2 upgrade stack. The stack has been collapsed
into a single PR against `main` — see **#2850**, which contains all of these commits
(signed off). This PR is closed for history only; there is no longer a stack.


hemildesai:
Superseded by #2855, which is being re-scoped as the single umbrella PR for the DeepEP v2 upgrade (this layer's commits are preserved in it).

PR #2855, opened 2026-06-29
feat(moe): DeepEP v2 dispatcher, Torch 2.12 PP, and selectable EP variants

hemildesai:
**Consolidated into #2850.**

This PR was one layer of the DeepEP v2 upgrade stack. The stack has been collapsed
into a single PR against `main` — see **#2850**, which contains all of these commits
(signed off). This PR is closed for history only; there is no longer a stack.


hemildesai:
Consolidated into #2850, now the single umbrella PR for the DeepEP v2 upgrade (base `main`, all 13 signed-off commits preserved).

PR #2859, opened 2026-06-30
fix(speculative): keep DSpark draft RoPE inv_freq in fp32 under bf16

khazic:
## What

The DSpark Qwen3 and Gemma4 draft models are cast to bfloat16 on the training
build path (`train_dspark.py` builds the draft with
`.to(device=..., dtype=compute_dtype)`, and `compute_dtype` is bf16 on CUDA).
That cast rounds the rotary `inv_freq` buffer to bf16, and the rotary forward
only up-casts it (`self.inv_freq.float()`), which cannot recover the lost bits.
The rounded frequencies dephase with absolute position, so the train/inference
RoPE diverges (worse with longer context) and draft acceptance erodes, while the
serving runtime keeps an fp32 RoPE cache.

The sibling DFlash draft already guards against this
(`Qwen3DFlashDraftModel._apply` + `test_rope_inv_freq_stays_fp32_after_bf16_cast`);
the DSpark drafts did not.

## Fix

- Add a shared `pin_rope_inv_freq_fp32` helper in `dspark/common.py` that
  recomputes fresh fp32 frequencies from the rotary config via `rope_init_fn`
  (the same recompute pattern already used in `checkpointing.py` and
  `hf_utils.py`), so it covers both `Qwen3RotaryEmbedding` and
  `Gemma4TextRotaryEmbedding`.
- Call it from an `_apply` override on `Qwen3DSparkModel` and
  `Gemma4DSparkModel`, so any dtype cast keeps `inv_freq` in fp32.

## Tests

- `test_rope_inv_freq_stays_fp32_after_bf16_cast` and
  `test_rope_inv_freq_fp32_survives_chained_casts` (Qwen3) plus
  `test_gemma4_rope_inv_freq_stays_fp32_after_bf16_cast` (Gemma4) assert that
  `inv_freq` stays fp32 and bit-exact against the fp32 reference after a bf16
  cast, while the rest of the model still casts to bf16. They fail before this
  change (the buffer ends up bf16).

### DIFF
--- a/nemo_automodel/components/speculative/dspark/common.py
+++ b/nemo_automodel/components/speculative/dspark/common.py
@@ -217,6 +217,33 @@ def create_noise_embed(
     return embed_tokens(noise_ids)
 
 
+def pin_rope_inv_freq_fp32(rotary_emb: Optional[nn.Module]) -> None:
+    """Keep a RoPE module's ``inv_freq`` buffer in fp32 after a dtype cast.
+
+    ``module.to(bfloat16)`` (the training build path) rounds the rotary
+    ``inv_freq`` buffer to bf16. The rounded frequencies dephase with absolute
+    position, so the train/inference RoPE diverges (worse with longer context)
+    and draft acceptance erodes, while the serving runtime keeps an fp32 RoPE
+    cache. A bf16 round-trip cannot be undone by upcasting, so recompute fresh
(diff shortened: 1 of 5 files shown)


HuiyingLi (approved):
LGTM thank you!

PR #2867, opened 2026-06-30
feat(speculative): add JetSpec causal parallel drafting training

khazic:
## What

Adds **JetSpec** ([arXiv:2606.18394](https://arxiv.org/abs/2606.18394), "Breaking the Scaling Ceiling of Speculative Decoding with Parallel Tree Drafting") as a new draft-training method. JetSpec reuses the existing DFlash parallel draft backbone but trains it as a *causal* parallel tree drafter:

1. **Block-causal attention.** DFlash drafts a block bidirectionally, so each node's distribution is branch-agnostic and a constructed tree can be internally inconsistent. JetSpec makes the in-block attention causal (a query at within-block offset `i` attends only to offsets `j <= i`), so each branch is conditioned on its own prefix and the draft factorization mirrors the target's autoregressive order.
2. **Forward-KL distillation.** Instead of DFlash's hard-label decay-weighted cross-entropy, JetSpec distills against the target model's per-position soft distribution with a temperature-scaled forward-KL loss (reuses the existing `KDLoss`).

The draft architecture is unchanged (`Qwen3DFlashDraftModel`): JetSpec differs from DFlash only in the attention mask and the loss, so it is implemented as a thin subclass, mirroring how Domino extends DFlash.

## Changes

- `components/attention/dflash_mask.py`: add a `causal` flag to `create_dflash_sdpa_mask` / `create_dflash_block_mask` (in-block lower-triangular attention; default `False` preserves DFlash).
- `components/speculative/dflash/target.py`: optional `capture_logits` on `HFDFlashTargetModel` so the frozen target's already-computed logits are exposed as the KD teacher (off by default, no cost to DFlash).
- `components/speculative/dflash/core.py`: extract the shared block label/target/mask construction into `DFlashTrainerModule._build_block_targets`, reused by both DFlash and JetSpec.
- `components/speculative/dflash/jetspec_core.py` (new): `JetSpecTrainerModule` = causal block mask + forward-KL against the gathered teacher logits.
- `recipes/llm/train_jetspec.py` (new): `TrainJetSpecRecipe` (subclass of `TrainDFlashRecipe`) + `examples/speculative/jetspec/qwen3_jetspec.yaml`.
- `recipes/llm/train_dflash.py`: add a `_build_target_wrapper` seam and route eval through `_run_trainer_step`, so subclasses inject their extra inputs uniformly (behavior-preserving for DFlash and Domino).

## Scope

This PR is the **training** side, producing a causal draft-head checkpoint. The tree drafting and tree-attention verification are an inference-engine concern (vLLM) and are intentionally out of scope here.

## Tests

New unit tests cover the causal mask (vs. an element-level reference, strict-subset and no-fully-masked-row properties), the JetSpec trainer (teacher-position alignment, forward-KL, padding blocks, skip-on-no-supervision, chunked vs. unchunked equivalence, the flex-attention branch), the optional logit capture, and the recipe seams. The shared `_build_block_targets` extraction is covered by the existing DFlash/Domino suites.

```
pytest tests/unit_tests/attention/test_dflash_mask.py tests/unit_tests/speculative/ \
       tests/unit_tests/recipes/llm/test_train_dflash.py \
       tests/unit_tests/recipes/llm/test_train_domino.py \
       tests/unit_tests/recipes/llm/test_train_jetspec.py \
       tests/unit_tests/loss/test_kd_loss.py
# 530 passed, 12 skipped
```


khazic:
Two intentional configuration choices worth calling out, with the paper references and the rationale:

**1. `kd_temperature` defaults to 1.0 (configurable), not hardcoded.**

The paper (§2.3, Eqs. 8-9) defines the distillation over temperature-normalized distributions `p̃ = softmax(z_p / T_KD)` (target) and `q̃ = softmax(z_q / T_KD)` (draft), with the objective `L_train = T_KD² · Σ_m w_m · L_FKL^(m) / Σ_m w_m`, and states only that `T_KD > 0` (the main text does not give a concrete value). I therefore exposed it as `recipe_args.kd_temperature` (default `1.0`), forwarded into `KDLoss`, which already applies the matching `T²` scaling so gradient magnitudes stay temperature-independent. Default `1.0` is the standard forward-KL setting and does not affect the correctness of the objective; if the paper's specific `T_KD` is confirmed (e.g. in the Appendix F ablations), it can be set from YAML without any code change.

**2. No DFlash-style depth-weighting knob (uniform `w_m`), by design.**

The paper's main results deliberately drop DFlash-style depth weighting to isolate the effect of the causal mask. Appendix D: "we remove it when isolating the effect of causal masking versus the native block-diffusion head design", and the `w_m` in Eq. 9 are uniform in that setting. `JetSpecTrainerModule` computes the loss with `KDLoss` as a plain mean over active draft positions (anchor and non-loss positions excluded), which equals uniform `w_m` and matches the paper's main configuration exactly. The DFlash-style weighted variant is an ablation (§3.4); reproducing that later would need an explicit per-depth `w_m` schedule, which is intentionally not added here to keep the default faithful to the main results.

khazic:
### End-to-end GPU training run (early validation)

Ran the recipe end-to-end on real hardware to confirm the training path works and the draft actually learns. This is a **single-epoch validation run on general chat data, not a paper reproduction** (the paper's headline τ≈10 is Qwen3-8B over 6 epochs with math-specialized evaluation), but it exercises the full production path.

**Setup**
- Recipe: `TrainJetSpecRecipe`, target `Qwen3-8B` (the paper's target), draft = 5-layer Qwen3 stack, fused target layers `{1, 9, 17, 25, 33}`.
- `block_size=16`, `num_anchors=256`, forward-KL distillation (`kd_temperature=1.0`), `attention_backend=flex_attention`.
- Training data: chat prompts with the assistant turns **regenerated by the target model** (in-distribution for the draft), `seq_length=2048`, `micro_batch_size=2`.
- 8× A100-80GB, single node, DDP. ~1.0 step/s.

**Trajectory** (`accuracy` = draft-vs-target greedy agreement; `accept_len` = expected accepted-prefix length per block, the τ proxy added in this PR):

| step | loss | accuracy | accept_len (τ~) | lr |
|-----:|-----:|---------:|----------------:|------:|
| 10   | 14.74 | 0.000 | 1.000 | 1.9e-6 |
| 80   | 7.21  | 0.059 | 1.092 | 1.4e-5 |
| 400  | 6.32  | 0.069 | 1.244 | 6.8e-5 |
| 790  | 5.46  | 0.106 | 1.441 | 1.34e-4 |
| 860  | 5.36  | 0.110 | 1.816 | 1.46e-4 |
| 930  | 5.04  | 0.126 | 1.695 | 1.58e-4 |
| 980  | 4.92  | 0.142 | 1.691 | 1.66e-4 |

**Observations**
- Loss drops from `log(vocab) ≈ 11.9` (the untrained forward-KL value; the 14.7 at step 10 is a random-init student that is worse than uniform) down to `~4.9`, and is still falling.
- `accept_len` (τ proxy) climbs monotonically from `1.0` (anchor only) to `~1.7` (peak 1.82 so far), i.e. the draft is starting to get whole drafted tokens accepted by the target. All of this is still **inside the LR warmup** (`lr ≈ 1.7e-4` of the `3e-4` peak), so the curve has not plateaued.
- The flex-attention block-causal mask (the torch.compile'd `causal=True` BlockMask kernel), the forward-KL `KDLoss`, and the optional teacher-logit capture all run stably across 8-way DDP. No NaN / OOM / NCCL issues.

The run was stopped early once the trend was clear (loss ↓, τ ↑); the point was to validate the training-side end-to-end, which it does. A full quality run (more epochs, longer LR schedule, task-specialized eval) and the inference-side tree drafting / verification in vLLM are follow-ups.

HuiyingLi (approved):

PR #2436, opened 2026-06-05
perf: FFPA D=512 attention backend for Gemma4 (3× fwd / 6× bwd vs SDPA)

Butterfingrz:
# What does this PR do ?

Introduces `attn_implementation="ffpa"` for the Gemma4 family: full-attention layers with `head_dim=512` are routed to the
[ffpa-attn](https://github.com/xlite-dev/ffpa-attn) CuTeDSL kernel (both dense and varlen, fwd + bwd). All other layers automatically fall back to the SDPA flash backend.

# Changelog

### 1. `components/attention/ffpa_attention.py` (New Module)
Core implementation for the FFPA attention interface:

* **`ffpa_attention_forward`**: Implements the HF attention interface.
  * The dense path dispatches to the high-level `ffpa_attn_func` autograd wrapper (CuTeDSL `_fwd_cute` kernel underneath); `torch.ops.ffpa_attn._fwd_cute` itself is the low-level readiness probe and the op the ring-CP path calls directly.
  * When a 2D bool padding mask is supplied, it routes through `_varlen_fwd_cute`, reusing transformers' `_upad_input` / `_pad_input` to share the flash-varlen protocol.
* **`ffpa_mask`**: Registered under `ALL_MASK_ATTENTION_FUNCTIONS["ffpa"]`.
  * No padding → `None` (dense fast path).
  * With padding → 2D bool (varlen path).
  * Sliding-window / non-causal (e.g. vision-bidirectional) patterns → a flex `BlockMask`, which `ffpa_attention_forward` routes to **FlexAttention** (not the SDPA/eager fallback) so the sliding sparsity is exploited. `ffpa_mask` never emits a 4D additive float; the 4D-mask handling in the forward only covers masks supplied by other code paths.
  * Vision / audio sub-encoders → `None` (ensures the shared registry is not polluted).
* **Fallback Priority**:
  1. **SDPA (Default)**: O(L) memory, dispatches to the flash backend on H100 / H200 / B200. Used for full-attention (`head_dim=512`) layers that are FFPA-ineligible (4D mask, fp32, kernel unavailable, …). *Note: sliding layers do NOT use this path — they carry a flex `BlockMask` and run on FlexAttention, avoiding a dense `(B, H, L, L)` materialization.*
  2. **Eager**: Triggered only when `softcap is not None`. Since SDPA lacks a `softcap` entry point and silently swallows the value via `**kwargs`, Gemma2 / 3 softcap layers must use eager for numerical correctness.
[...]

PR #2869, opened 2026-06-30
chore: CI proxy for #2436

HuiyingLi:
CI-only proxy PR. Do not merge / do not review. Points at the exact head commit of #2436 (87df8cf75a68cd53c84d87ce24728403a5d4aab0) so internal CI runs under the internal-contributor queue; results post back to #2436 via the shared SHA. Source of truth: #2436. Close once CI completes.


HuiyingLi:
Closing CI-only proxy because source PR #2436 is merged.

PR #2873, opened 2026-07-01
build(deps): upgrade transformers to 5.12.1

athitten:
Bump transformers 5.8.1 -> 5.12.1 in pyproject.toml, uv.lock and docker/common/uv-pytorch.lock, and fix API incompatibilities:

- create_causal_mask / create_sliding_window_causal_mask: drop the removed cache_position argument (the mask now derives from position_ids) across datasets/utils, distributed/pipelining/hf_utils (+ an input_embeds typo), common/packing, and the qwen2, gemma4_moe, mimo_v2_flash, llama and mistral3 models.
- gemma4_moe: replace the removed transformers create_causal_mask_mapping with the block_sequence_ids-based mask builders.
- diffusion_gemma parity test: skip on the incompatible multimodal upstream (native text-only model unaffected; tracked in AM-588).

# What does this PR do ?

Add a one line overview of what this PR aims to accomplish.

# Changelog

- Add specific line by line info of high level changes in this PR.

# Before your PR is "Ready for review"

**Pre checks**:

- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?

If you haven't finished some of the above items you can still open "Draft" PR.

# Additional Information

- Related to # (issue)

PR #2883, opened 2026-07-01
fix(diffusion_gemma): make parity test work on transformers >= 5.11

zyzhou5:
# What does this PR do ?

- DiffusionGemmaEncoderModel.__init__ dropped the 'vision_config is not None' guard, so the reference can no longer be constructed with vision_config=None. Build it with a tiny throwaway gemma4_vision tower instead -- it is never exercised (text-path-only comparison; the native model and adapter ignore/drop vision), so it cannot perturb the parity check.
- the block-diffusion forward renamed canvas_ids -> decoder_input_ids. Select the accepted kwarg via inspect so the test is version-agnostic (5.8-dev fork and mainline >=5.12).

Test-only change; no model/adapter/training code touched. Validated on transformers 5.12.1 (nemo_automodel_nightly container): the parity + model tests pass.

### DIFF
--- a/tests/unit_tests/models/diffusion_gemma/test_diffusion_gemma_parity.py
+++ b/tests/unit_tests/models/diffusion_gemma/test_diffusion_gemma_parity.py
@@ -28,12 +28,16 @@ native ``encode``/``decode`` building blocks with the fork's **inference** mask
 apples. The block-causal training mask is covered by
 ``test_diffusion_gemma_mask.py``.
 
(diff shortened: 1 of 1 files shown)


zyzhou5:
/ok to test [51b8168](https://github.com/NVIDIA-NeMo/Automodel/commit/51b81681cae3c6ce30025d79aa1908928539cb36)

athitten (approved):
LGTM, thank you @zyzhou5 !

athitten:
Will remove the skip guard for Diffusin gemma tests in #2873

PR #2887, opened 2026-07-02
chore: CI proxy for #2885

HuiyingLi:
CI-only proxy PR. Do not merge / do not review. Points at the exact head commit of #2885 (a6c182987120b61343e6f52cc74eefdd2dbee6d1) so internal CI runs under the internal-contributor queue; results post back to #2885 via the shared SHA. Source of truth: #2885. Close once CI completes.


HuiyingLi:
Closing CI-only proxy because source PR #2885 is merged.

PR #2901, opened 2026-07-03
chore: CI proxy for #2898

HuiyingLi:
CI-only proxy PR. Do not merge / do not review. Points at the exact head commit of #2898 (b9ff55e309349c75d6e35e3602277af05b71d590) so internal CI runs under the internal-contributor queue; results post back to #2898 via the shared SHA. Source of truth: #2898. Close once CI completes.


HuiyingLi:
Closing CI-only proxy because source PR #2898 is merged.

PR #2910, opened 2026-07-03
chore: CI proxy for #2909

HuiyingLi:
CI-only proxy PR. Do not merge / do not review. Points at the exact head commit of #2909 (897134674046f334d442eb0dae8f3126495b0a72) so internal CI runs under the internal-contributor queue; results post back to #2909 via the shared SHA. Source of truth: #2909. Close once CI completes.


HuiyingLi:
Closing CI-only proxy because source PR #2909 is merged.

PR #2912, opened 2026-07-03
feat(speculative): add bench_vllm acceptance/speedup benchmark

khazic:
## What does this PR do?

Adds `bench_vllm.py`, the vLLM companion to `bench_sglang`: an offline acceptance / speedup benchmark for any drafter served by vLLM (EAGLE-3, P-EAGLE, and the DFlash family once #2913 lands; works with plain `vllm serve --speculative-config` too).

- Acceptance statistics come from vLLM's Prometheus `/metrics` endpoint. The spec-decode counters (`vllm:spec_decode_num_drafts_total`, `..._num_draft_tokens_total`, `..._num_accepted_tokens_total`) are snapshotted before and after the workload and differenced, so the reported numbers cover exactly this benchmark's requests even on a server that has already handled other traffic (unlike SGLang's server-cumulative `avg_spec_accept_length`).
- Reports `accept_length` (`1 + accepted/drafts`, including the bonus token, matching the formula documented in vLLM's `SpecDecodingProm`), `acceptance_rate` (`accepted/draft_tokens`), output throughput, and an optional `speedup` against a `--baseline-server` running without speculation.
- The engine-agnostic workload machinery (prompt loading, the timed chat-completions pass, throughput/speedup math, shared CLI bounds) moves from `bench_sglang` into a new `bench_common` module now that it has two consumers. `bench_sglang` keeps only the SGLang-specific `/server_info` handling; no behavior change on the SGLang path.

## Verified end to end

Ran against a vLLM 0.24.0 server hosting Qwen3-4B with an Automodel-trained DFlash draft (16 prompts, concurrency 4): `accept_length 1.30`, `acceptance_rate 2.0%` (1555 drafts / 23325 draft tokens / 472 accepted), throughput 183.9 tok/s, 16/16 completed. The low acceptance matches the deliberately under-trained smoke draft (1000 optimizer steps); the point is the counters difference into exact per-run numbers.

## Tests

New `tests/unit_tests/speculative/test_bench_vllm.py` (24 tests: Prometheus parsing, delta/acceptance math, /metrics fetching, orchestration incl. baseline and output-json). `test_bench_sglang.py` retargeted to the moved helpers; full speculative suite passes (534 passed, 11 skipped).

### DIFF
--- a/nemo_automodel/components/speculative/bench_common.py
+++ b/nemo_automodel/components/speculative/bench_common.py
@@ -1,0 +1,196 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Engine-agnostic workload machinery shared by the drafter benchmarks.
+
+``bench_sglang`` and ``bench_vllm`` drive the same OpenAI-style
+chat-completions workload and report the same throughput/speedup numbers; they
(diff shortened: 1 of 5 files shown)


khazic:
The force-push above was a no-change amend (same tree as the vetted commit); it reset the copy-pr-bot vetting, so the pipeline needs a re-trigger when someone gets a chance.

HuiyingLi (approved):

PR #2918, opened 2026-07-03
feat(speculative): context parallelism for draft-model training

kashif:
Adds context parallelism (`cp_size>1`) to the DFlash/Domino/JetSpec and DSpark draft trainers for long-context training, and fixes the same path in EAGLE-3.

The approach follows EAGLE-3's existing seam: only the frozen target's forward is sharded across the `cp` mesh, and its captured hidden states are gathered back to the full sequence before the draft runs — so the draft's block/anchor masks are never sequence-split. The draft shards over the `dp` submesh; cp ranks in a dp group share data and draft weights.

Two things that bit me and are worth a careful look in review:

- The drafts sample anchor positions from the global RNG, so cp (and tp) replicas would diverge if seeded per global rank. Now seeded per dp rank so they stay in sync.
- torch's `context_parallel` ring doesn't actually fire for a plain HF target — q/k/v reach SDPA as local tensors, so each rank silently attended only to its own sequence shard. Since the target is frozen (forward-only), I all-gather K/V and attend the local Q against the full sequence with a global causal mask (new `attach_cp_kv_gather_hooks`). This needs the HF torch-SDPA path, so `cp_size>1` requires `target_force_hf=true`.

Validated on 2 GPUs (Qwen3-0.6B): with the fix, per-step draft loss matches the 1-GPU baseline within bf16 noise (mean signed +0.18%, symmetric above/below), versus a systematic +2–7% before. Unit tests pass.

Defaults (`cp_size=1`) are unchanged, and DSpark CP is gated to the dense Qwen3 target (the MoE targets already run under their own EP/FSDP mesh).


kashif:
One design question for reviewers on the `attach_cp_kv_gather_hooks` approach.

Right now the CP hook intercepts attention by swapping `torch.nn.functional.scaled_dot_product_attention` during each `self_attn` forward. That only works when the target lowers to torch SDPA, which is why `cp_size>1` requires `target_force_hf=true` — a NeMo-custom target defaults to TE attention (and flex/tilelang/`kernels`-provided kernels would also bypass the swap), so the hook would silently no-op and each rank would fall back to local-only attention.

A more robust alternative: register a dedicated `"cp_gather"` entry in HF's `ALL_ATTENTION_FUNCTIONS` and select it via `attn_implementation`, instead of monkeypatching the torch symbol. That intercepts at the attention-interface level regardless of the underlying kernel, and would let us drop the `force_hf` requirement (and compose CP with faster attention backends). The trade-off is a bit more surface tied to the transformers attention-interface API across versions.

Worth doing in this PR, or fine as a follow-up? Leaning follow-up since the current gated version is correct where it's enabled, but happy to fold it in if preferred.

khazic:
Taking a look at this now, will review shortly.

nemo_automodel/recipes/llm/train_eagle3.py
```diff
@@ -467,6 +482,14 @@ def _setup_online_target(self, recipe_cfg, target_path, target_config):
         tp_size = int(self.cfg.get("distributed.tp_size", 1) or 1)
         _validate_cp_gates(cp_size, backend, packed_sequence_size)
         _validate_tp_gates(tp_size, backend, cp_size)
+        # The CP K/V-gather hook attends via the target's torch SDPA call; a
+        # custom-attention (non-HF) target would silently skip it and each rank would
+        # see only its own shard, so require the HF target path.
+        if cp_size > 1 and not bool(recipe_cfg.get("target_force_hf", False)):
+            raise NotImplementedError(
+                "Context parallelism (cp_size>1) requires recipe_args.target_force_hf=true so the "
```
khazic: This gate checks `target_force_hf` but the hook's real precondition is that the target's attention routes through `F.scaled_dot_product_attention`. `force_hf` only picks the model class; `target_attn_implementation` still defaults to HF auto-select, which can pick FA2 when flash-attn is installed (and the CP draft ring requires flash-attn, so it will be). In that case `attach_cp_kv_gather_hooks` patches a function the model never calls and every rank silently attends only its local shard, so the draft trains on corrupted supervision with no error. Suggest forcing or asserting `attn_implementation == "sdpa"` on the target under cp_size>1 (same for the gates in train_dflash.py:320 and train_dspark.py:399). Only the eagle3 CP example yaml pins sdpa manually today.

nemo_automodel/components/speculative/eagle/ring_attention.py
```diff
@@ -0,0 +1,295 @@
...
+import torch.distributed as dist
+
+from nemo_automodel.shared.import_utils import safe_import
+
+HAVE_FLASH_ATTN, _flash_attn = safe_import("flash_attn.flash_attn_interface")
+
+
+def _fa_forward(q, k, v, softmax_scale, causal, dropout_p=0.0):
+    """flash_attn 2.8.3 fwd: returns ``(out[B,S,H,D], lse[B,H,S])``."""
+    out, lse, _s_dmask, _rng = _flash_attn._flash_attn_forward(
```
khazic, line 82: `_flash_attn` is None when flash-attn is not installed, so this fails with an opaque AttributeError deep in the ring. Nothing upstream gates on `HAVE_FLASH_ATTN` (the draft default is eager and `_validate_cp_gates` does not check it). A clear raise here or in the cp gate would save users the stack dive.

nemo_automodel/recipes/llm/train_eagle3.py
```diff
@@ -743,6 +766,45 @@ def _setup_cached_target(self, recipe_cfg, target_config):
...
+        so slice every ``[B, S, ...]`` tensor to this rank's cp shard and inject the
+        global ``position_ids``. No-op without CP.
+        """
+        if getattr(self, "cp_group", None) is None:
+            return inputs
+        cp_size = self.cp_mesh.size()
+        cp_rank = self.cp_mesh.get_local_rank()
+        seq_len = inputs["input_ids"].shape[1]
+        if seq_len % cp_size != 0:
+            raise ValueError(f"Context parallelism requires seq_length ({seq_len}) divisible by cp_size ({cp_size}).")
```
khazic: The dataloader pads every batch to the fixed `recipe_args.seq_length`, so this only ever fires for `seq_length % cp_size != 0`, and it fires after the target is loaded and has run a full forward. Checking `seq_length % cp_size == 0` in `_validate_cp_gates` would fail fast at setup instead.

nemo_automodel/components/speculative/eagle/zigzag_ring_attention.py
```diff
@@ -0,0 +1,431 @@
...
+            ctx.half0,
+            ctx.half1,
+            ctx.softmax_scale,
+            ctx.causal,
+            ctx.window_size,
+        )
+        return dq, dk, dv, None, None, None, None, None, None
+
+
+def zigzag_ring_flash_attn_varlen_func(
```
khazic, line 425: Nothing in the repo imports this module; the draft CP path uses `cached_ring_attention` from ring_attention.py. That leaves 431 lines of vendored kernel code with no caller and no test. If load-balanced sharding is planned as a follow-up it may be cleaner to add this file then; note `get_half_lse` (line 102) also has a latent shape mismatch for odd per-document lengths (`end//2 - start//2` vs the post-offset source slice length).

nemo_automodel/components/speculative/eagle/zigzag_ring_attention.py
```diff
@@ -0,0 +1,431 @@
...
+import torch
+import torch.distributed as dist
+import torch.nn.functional as F
+
+from nemo_automodel.shared.import_utils import safe_import
+
+HAVE_FLASH_ATTN, _fa = safe_import("flash_attn.flash_attn_interface")
+
+
+class RingComm:
```
khazic, line 45: This `RingComm` is byte-for-byte the one in ring_attention.py:84 (plus `send_recv_kv`). Two copies of the p2p ring in the same PR will drift; defining it once and importing would keep any future deadlock/ordering fix in one place.

nemo_automodel/components/distributed/cp_utils.py
```diff
@@ -212,6 +212,84 @@ def _self_attn_pre_forward_hook(_module, module_args, module_kwargs):
...
+        torch.distributed.all_gather(v_parts, value, group=group)
+        k_full = torch.cat(k_parts, dim=2)
+        v_full = torch.cat(v_parts, dim=2)
+        # Global causal mask: local query at global position ``cp_rank*seq_local + i``
+        # attends to every key position ``j`` up to and including it.
+        q_pos = torch.arange(seq_local, device=query.device) + cp_rank * seq_local
+        k_pos = torch.arange(k_full.shape[2], device=query.device)
+        mask = q_pos[:, None] >= k_pos[None, :]
+        return _original_sdpa(
+            query, k_full, v_full, attn_mask=mask, dropout_p=0.0, is_causal=False, scale=scale, enable_gqa=enable_gqa
```
khazic: Passing a materialized bool mask with `is_causal=False` pushes SDPA off the fused causal flash path onto the math/mem-efficient kernel for every target layer, and the `q_pos`/`k_pos`/`mask` tensors are rebuilt per layer although they only depend on the shard geometry. At 32k context and cp=4 that is a ~256 MB mask allocated per layer per step plus roughly 2x attention FLOPs on the frozen-target hot path. Computing the mask once per forward (or expressing the per-shard causal pattern with is_causal semantics / an FA varlen call) would keep the fast path.

nemo_automodel/components/speculative/dflash/target.py
```diff
@@ -113,23 +142,46 @@ def _hook(_module, _inputs, outputs):
...
+                # Shard the sequence, run the target as ring attention, then gather the
+                # captured layers (and logits) back to the full sequence. The self_attn
+                # hooks strip the mask and force is_causal, so attention_mask is None.
+                from nemo_automodel.components.distributed.cp_utils import gather_cp_seq, make_target_cp_ctx
+
+                if "position_ids" not in forward_params:
+                    raise ValueError(
+                        "Context parallelism requires the target model's forward to accept `position_ids`."
+                    )
+                cp_ctx, cp_input_ids, cp_position_ids, orig_len = make_target_cp_ctx(self.cp_mesh, input_ids)
```
khazic: This shard-forward-gather block (make_target_cp_ctx, attention_mask=None, gather_cp_seq with seq_dim=1/orig_len, gather before hook removal) is now repeated nearly verbatim in the eagle, dflash, and dspark target wrappers. A shared helper would keep the gather-inside-ctx invariant and the seq_dim/orig_len contract in one place; a fix applied to one copy will otherwise miss the other two silently.

khazic:
One PR-level item: none of the 13 commits carries a Signed-off-by trailer, and CONTRIBUTING requires DCO sign-off on every commit. A rebase with `--signoff` will fix it.

nemo_automodel/recipes/llm/train_eagle3.py
kashif: Good catch — real silent-corruption risk. Fixed in 5426232c9: `_validate_cp_gates` now also requires `target_attn_implementation=sdpa` under cp_size>1 (not just `target_force_hf`), and the same check is added to train_dflash.py and train_dspark.py.

nemo_automodel/components/speculative/eagle/ring_attention.py
kashif: Fixed in 5426232c9 — `_validate_cp_gates` now raises a clear error when flash-attn is missing under cp_size>1 (the draft ring needs it), instead of the opaque AttributeError.

nemo_automodel/recipes/llm/train_eagle3.py
kashif: Fixed in 5426232c9 — moved the `seq_length % cp_size` check into `_validate_cp_gates` so it fails fast at setup instead of mid-forward.

nemo_automodel/components/speculative/eagle/zigzag_ring_attention.py
kashif: Now addressed — da55384f9 adds `cached_zigzag_ring_attention` in ring_attention.py, which imports and drives this module as the load-balanced draft ring; validated 2-GPU (forward rel 1.7e-3, dq rel 1.76e-3). The `get_half_lse` odd-length path isn't exercised by the EAGLE single-doc even-length use, but noted for when varlen/odd docs land. Resolving the dead-code concern.

nemo_automodel/components/speculative/eagle/zigzag_ring_attention.py
kashif: Agreed — ring_attention.py now imports from this module, so folding RingComm down to one definition makes sense. Tracking as a follow-up cleanup (want to re-run the 2-GPU ring checks after the merge).

nemo_automodel/components/distributed/cp_utils.py
kashif: Partially addressed in 8bc8eaed4 — the mask depends only on shard geometry and is now cached across layers/steps (no per-layer realloc). The fast-path half (gather only the causal prefix + is_causal, or an FA varlen call) is a bigger rework of the gather hook — tracking as a follow-up. Note the MATH-inclusive context here is required: bool-mask + GQA has no fused flash/efficient kernel.

nemo_automodel/components/speculative/dflash/target.py
kashif: Agreed — this shard-forward-gather block is duplicated across the eagle/dflash/dspark target wrappers. Tracking a shared helper as a follow-up so the gather-inside-ctx invariant and the seq_dim/orig_len contract live in one place.

nemo_automodel/components/speculative/eagle/zigzag_ring_attention.py
kashif: Done in 9f5a48144 — `ring_attention.py` now imports the single `RingComm` from this module (canonical here since ring_attention already depends on it; the other direction would be a circular import). Kept this one's `send_recv` `.contiguous()` so it's a safe superset for both callers. Re-ran the 2-GPU ring validation after the merge: forward rel 1.7e-3, dq rel 1.76e-3, unchanged.

nemo_automodel/components/speculative/dflash/target.py
kashif: Done in 4596d4397 — added `run_target_cp_forward_and_gather` in cp_utils.py; the eagle3/dflash/dspark target wrappers now all go through it, so the gather-inside-context invariant and the seq_dim=1/orig_len contract live in one place. Bonus: it centralizes the 'target forward must accept position_ids' check as a fail-fast raise — dspark was silently missing that (filter_forward_kwargs would have dropped position_ids), exactly the drift you flagged. Verified byte-identical model call + gather, 14 target-CP unit tests pass, and re-ran the 2-GPU eagle3 recipe (trains clean end-to-end).

khazic:
There are a few more issues that I am still reviewing in depth.

nemo_automodel/components/speculative/eagle/draft_llama.py
```diff
@@ -477,6 +509,18 @@ def _flash_block1_varlen(
         return attn_output_bhtd, lse_fa
 
 
+def attach_eagle3_cp_attention(model: nn.Module, cp_group) -> None:
```
khazic: `attach_eagle3_cp_attention` silently no-ops for non-Llama drafts: `Eagle3DeepseekMLAAttention` is not an `Eagle3LlamaAttention`, so with a DeepSeek target and `cp_size>1` no ring is installed and each rank's draft attends only its own shard, while the trainer still left-shifts across rank boundaries and renormalizes the loss as if the sequence were sharded. Training is silently wrong, the same failure mode this PR fixes for the target. Please raise when no module matched (or gate the draft arch in `_validate_cp_gates`). gpt-oss is fine since it reuses `Eagle3LlamaAttention`.

nemo_automodel/recipes/llm/train_eagle3.py
```diff
@@ -1258,6 +1357,8 @@ def _train_epochs(self, start_epoch, batches_per_epoch, is_ddp, pbar=None):
                 pending_micro_batches += 1
 
                 if pending_micro_batches == self.grad_accumulation_steps:
+                    if getattr(self, "cp_group", None) is not None:
+                        self._all_reduce_draft_grads_over_cp()
```
khazic, line 1610: This cp grad sum is missing from the trailing partial-window flush (the `pending_micro_batches > 0` block near L1418). With `cp_size>1` and `batches_per_epoch` not a multiple of `grad_accumulation_steps`, that final step applies per-rank partial gradients, so the cp replicas of the draft desync permanently from that point.

nemo_automodel/recipes/llm/train_eagle3.py
```diff
@@ -743,6 +799,45 @@ def _setup_cached_target(self, recipe_cfg, target_config):
             )
         return selected_token_ids, selected_token_mask
 
+    def _maybe_shard_cp(self, inputs: dict) -> dict:
```
khazic, line 993: The cached-target branch bypasses CP entirely: `setup()` calls `_setup_cached_target` (L277) without passing through `_validate_cp_gates` (it lives in `_setup_online_target`), and `cp_mesh` is only built in `_setup_colocated_target`. With a cache and `cp_size>1`, `cp_group` stays `None` and this method no-ops, so the user silently gets plain DP. Worth a hard error for cache + `cp_size>1`.

nemo_automodel/recipes/llm/train_dspark.py
```diff
@@ -931,7 +987,10 @@ def _build_checkpointer(self, target_path: str) -> None:
             ckpt_kwargs["model_state_dict_keys"] = draft_state_dict_keys
 
         self.checkpoint_config = CheckpointingConfig(**ckpt_kwargs)
-        dp_rank = dist.get_rank() if dist.is_initialized() else 0
+        # Under CP the draft is replicated across cp ranks, so key the shard on the dp
+        # coordinate (identical for cp peers) rather than the global rank. Without a
+        # mesh (cp_size=1) this returns the global rank, unchanged.
+        dp_rank = self._get_dp_rank()
```
khazic, line 1019: Under `cp_size>1` the cp peers share this `dp_rank`, and the `Checkpointer` below is built with `tp_rank=0, pp_rank=0` for every rank, so `save_on_dp_ranks` has all cp peers concurrently `torch.save` the same `rng_dp_rank_N.pt`. Concurrent writes to one path on a shared FS can corrupt the file; please gate the write to one cp peer. `train_dflash.py` has the same pattern.

nemo_automodel/components/distributed/cp_utils.py
```diff
@@ -212,6 +277,106 @@ def _self_attn_pre_forward_hook(_module, module_args, module_kwargs):
...
 
 
+def attach_cp_kv_gather_hooks(model: torch.nn.Module, cp_mesh) -> None:
+    """Context-parallel self-attention for a FROZEN (forward-only) target.
+
+    Torch's ``context_parallel`` ring dispatch does not fire for a plain HuggingFace
+    forward -- q/k/v reach ``F.scaled_dot_product_attention`` as ordinary local
+    tensors, so each rank silently attends only to its own sequence shard and every
+    position past the first shard is wrong. Because the target is frozen (no
+    backward through attention), the correct fix is simple: all-gather K/V across the
```
khazic: As the comment above notes, flash/efficient reject a boolean mask under GQA, so this reliably lands on MATH, which materializes the `[B, H, S/cp, S]` score matrix in fp32. At the long contexts CP is meant for (e.g. S=32k, cp=8, 64 expanded heads) that is tens of GB per layer, transiently, exactly where the feature should be saving memory. An additive float mask (plus `repeat_kv` if needed) keeps the memory-efficient kernel eligible.

nemo_automodel/components/speculative/eagle/ring_attention.py
```diff
@@ -0,0 +1,374 @@
...
+    zigzag_ring_flash_attn_varlen_forward as _zz_forward,
+)
+from nemo_automodel.shared.import_utils import safe_import
+
+HAVE_FLASH_ATTN, _flash_attn = safe_import("flash_attn.flash_attn_interface")
+
+
+def _fa_forward(q, k, v, softmax_scale, causal, dropout_p=0.0):
+    """flash_attn 2.8.3 fwd: returns ``(out[B,S,H,D], lse[B,H,S])``."""
+    out, lse, _s_dmask, _rng = _flash_attn._flash_attn_forward(
```
khazic, line 82: These positional args pin flash-attn's private `_flash_attn_forward/_backward` to the 2.8.3 signature, but the setup gate only checks that `flash_attn` imports. Other 2.x versions reorder/add params (window_size, softcap, ...), so this either crashes mid-training or silently binds `causal`/`softmax_scale` to the wrong slots. Please assert the version, or use the public `flash_attn_func` API.

nemo_automodel/components/speculative/eagle/ring_attention.py
```diff
@@ -0,0 +1,374 @@
...
+            d_out_i = grad_out_f * wi.unsqueeze(-1)
+            d_lse_i = wi * (grad_out_f * (vi - out)).sum(-1)
+            dq = dq + d_lse_i.unsqueeze(-1) * scale * ki
+            dcache_k[:, i] = d_lse_i.unsqueeze(-1) * scale * q_f
+            dcache_v[:, i] = d_out_i
+
+        return dq.to(q.dtype), dcache_k.to(cache_k.dtype), dcache_v.to(cache_v.dtype), None, None
+
+
+def cached_zigzag_ring_attention(
```
khazic, line 388: `cached_zigzag_ring_attention` has no callers, and everything feeding the ring assumes contiguous shards (`_maybe_shard_cp`, `_cp_shift_left`, the injected position_ids), so the zigzag layout cannot be produced today. That leaves this class plus most of `zigzag_ring_attention.py` (~490 lines beyond `RingComm`) dead, while the live contiguous ring keeps the causal load imbalance zigzag was meant to fix (the last rank does cp x the flash work of rank 0). Suggest shipping it together with the zigzag data layout instead.

nemo_automodel/components/speculative/eagle/draft_llama.py
```diff
@@ -322,6 +328,32 @@ def forward(
...
+        diagonals stay local. Inputs are ``[B, H, T_local, D]``; the ring works in
+        FlashAttention ``[B, T_local, H, D]`` layout.
+        """
+        from nemo_automodel.components.distributed.ring_attention import cached_ring_attention
+
+        # RoPE upcasts q/k to fp32; the ring uses FlashAttention (fp16/bf16 only), so
+        # cast to the module's compute dtype (matches the plain FA2 path).
+        dt = self.o_proj.weight.dtype
+        qf = q.transpose(1, 2).contiguous().to(dt)
+        ckf = [t.transpose(1, 2).contiguous().to(dt) for t in cache_k]
```
khazic: This re-transposes/casts and re-stacks the whole cache every TTT step, so across `ttt_steps` the cache is copied O(steps^2) times. Storing blocks in FA layout and compute dtype once at append time (the diagonal loop already consumes per-block tensors) makes it O(steps).

nemo_automodel/components/speculative/eagle/ring_attention.py
```diff
@@ -0,0 +1,374 @@
...
+    return out, lse
+
+
+def _init_out_and_lse(block_out: torch.Tensor, block_lse: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+    out = block_out.to(torch.float32)
+    lse = block_lse.transpose(-2, -1).unsqueeze(dim=-1).to(torch.float32)
+    return out, lse
+
+
+def ring_flash_attn_forward(
```
khazic, line 118: None of the new ring/diagonal forward+backward math is unit-tested (the only test change in the PR is a mesh mock). A world_size=1 group test, where the ring reduces to a single FA block plus the TTT diagonals, compared against the eager reference in `draft_llama.py`, would pin the numerics; as-is the codecov patch gate will also flag this.

khazic:
@kashif  Verified the latest fixes and they all look correct. LGTM once the DCO check is green: the commits are missing Signed-off-by, so a `git rebase --signoff` and force-push should clear it.

kashif:
yes will do a final test and then move out of draft and fix DCO, thanks!

nemo_automodel/components/speculative/eagle/ring_attention.py
HuiyingLi, line 1: What do you think about moving this to eagle dir?

nemo_automodel/components/speculative/eagle/zigzag_ring_attention.py
HuiyingLi, line 1: Same here

nemo_automodel/components/distributed/cp_utils.py
```diff
@@ -189,6 +189,71 @@ def gather_cp_seq(cp_mesh: DeviceMesh, tensors: List[torch.Tensor], seq_dim: int
     return [t.narrow(seq_dim, 0, orig_len).contiguous() for t in full]
 
 
+def run_target_cp_forward_and_gather(
```
HuiyingLi: How about moving this to speculative/target_cp or something?

nemo_automodel/components/distributed/cp_utils.py
```diff
@@ -212,6 +277,107 @@ def _self_attn_pre_forward_hook(_module, module_args, module_kwargs):
             module.register_forward_pre_hook(_self_attn_pre_forward_hook, with_kwargs=True, prepend=True)
 
 
+def attach_cp_kv_gather_hooks(model: torch.nn.Module, cp_mesh) -> None:
```
HuiyingLi: Same here

PR #2923, opened 2026-07-04
chore: CI proxy for #2920

HuiyingLi:
CI-only proxy PR. Do not merge / do not review. Points at the exact head commit of #2920 (e160f0e494dc6f3372aec70cab44636210ab2523) so internal CI runs under the internal-contributor queue; results post back to #2920 via the shared SHA. Source of truth: #2920. Close once CI completes.


HuiyingLi:
Closing CI-only proxy because source PR #2920 is merged.

PR #2924, opened 2026-07-04
feat(speculative): add DSpark offline cache path

### LINKED ISSUE #2876: DSpark: add offline training support (precomputed target supervision)
## Summary

DSpark draft training (`nemo_automodel/components/speculative/dspark/`, `recipes/llm/train_dspark.py`) currently only supports the **online** path: the target model is colocated and its hidden states are captured live during draft training (see `dspark/target.py`, `dspark/core.py`).

Requesting an **offline** mode: precompute target supervision once and stream it from disk for draft training, without loading/running the target model during the draft training loop.

## Motivation

EAGLE-3 already has this duality in the repo: `precompute_eagle3.py` writes the target supervision to disk, and `train_eagle3.py` accepts `cached_target_path` to skip loading the target entirely and stream the cache instead (see `_setup_cached_target` vs `_setup_online_target` in `train_eagle3.py`).

DSpark would benefit from the same split:
- Decouples (often large) target model compute from draft iteration, so draft architecture/hyperparameter sweeps don't repeatedly pay for target forward passes.
- Lets the precomputed dataset be reused across multiple draft training runs.
- Reduces GPU memory pressure during draft training (no colocated target).

## Proposed scope

- A `precompute_dspark`-style script (mirroring `precompute_eagle3.py`) that runs the target once and writes the supervision DSpark's objective needs to disk.
- A `cached_target_path`-style option in `train_dspark.py` / `DSparkTrainerModule` to stream the precomputed cache instead of building a live target wrapper.

Happy to help scope/implement this if useful, just wanted to flag the gap first.

fkuner:
# What does this PR do ?

Adds a DSpark offline target-supervision cache path so draft training can reuse precomputed target outputs without loading or running the target model in the draft training loop.

The offline path is text-only in this PR: `precompute_dspark.py` supports HF-loadable single-process text targets, writes DSpark supervision shards plus target `embed_tokens` / `lm_head`, and `train_dspark.py` consumes the cache through `recipe_args.cached_target_path`.

# Changelog

- Add shared offline cache helpers and DSpark cache read/write utilities for sharded target supervision, manifest metadata, and cached target weights.
- Add `nemo_automodel.components.speculative.precompute_dspark` for text-only DSpark target precompute.
- Add `recipe_args.cached_target_path` support in `train_dspark.py` so DSpark training can stream cached target batches and skip constructing the live target wrapper.
- Validate cached DSpark manifests against target model type, vocab/hidden/layer shapes, target layer ids, sequence length, and compute dtype before training starts.
- Relax raw `target_model` path matching to a warning because cache and training paths may differ across machines, while keeping structural manifest validation strict.
- Skip required chat-template setup for cached DSpark training unless an explicit `recipe_args.chat_template` override is provided.
- Clone cached target `embed_tokens` / `lm_head` weights when loading them into the draft model to avoid tied-weight aliasing issues.
- Reject unsupported DSpark offline precompute targets with clear errors instead of silently producing incompatible caches.
- Update the Qwen3 DSpark example and speculative README for DSpark offline cache usage.
- Add unit and functional coverage for DSpark cache serialization, cached dataloading, manifest validation, target-free trainer setup, and cached target weight handling.

**Pre checks**:

# Additional Information

Related issue: #2876

Scope notes:

- This PR targets DSpark text offline caching: precompute target supervision once, then train the draft through `recipe_args.cached_target_path` without constructing the live target wrapper.
- DSpark offline precompute in this PR is intentionally limited to HF-loadable single-process text targets.
- DeepSeek V4, GLM-5.2, and MiniMax M3 are not enabled for DSpark offline precompute in this PR. They require separate large-model/distributed target precompute design and validation.
- MiniMax M3 multimodal offline precompute is out of scope because the current DSpark cache schema and precompute dataloader are text-only.
- Cached training still reads cached target `embed_tokens` / `lm_head` from the cache directory; it does not load or forward the target model.

Latest validation:

- `ruff format --check nemo_automodel/components/speculative/precompute_dspark.py nemo_automodel/recipes/llm/train_dspark.py tests/unit_tests/recipes/llm/test_train_dspark_helpers.py` passed.
- `ruff check nemo_automodel/components/speculative/precompute_dspark.py nemo_automodel/recipes/llm/train_dspark.py tests/unit_tests/recipes/llm/test_train_dspark_helpers.py` passed.
- `pytest tests/unit_tests/recipes/llm/test_train_dspark_helpers.py tests/unit_tests/speculative/test_dspark_offline_cache.py -q` passed: `81 passed`.
- `pytest tests/functional_tests/speculative/test_dspark_trainer_module.py -q` passed: `2 passed`.

Earlier validation:

- `pytest tests/functional_tests/speculative/test_dspark_train_tiny_qwen.py -q` passed: `1 passed`.
- `pytest tests/unit_tests/speculative/test_dspark_core.py -q` passed: `1 passed`.
- Manual real-HF `Qwen/Qwen3-0.6B` DSpark offline E2E smoke completed successfully.

### DIFF
--- a/nemo_automodel/components/datasets/llm/dspark_cache.py
+++ b/nemo_automodel/components/datasets/llm/dspark_cache.py
@@ -1,0 +1,250 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""On-disk format and reader for DSpark offline target supervision.
+
+The DSpark online recipe runs a frozen target model every step to capture the
+intermediate target features used by the draft and the final hidden state used
+for the TV / confidence losses. This module owns the disk format that lets a
+precompute job write those tensors once and lets training stream them later via
+``recipe_args.cached_target_path`` without loading the target model.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import re
+from types import SimpleNamespace
+from typing import Any, Callable
+
+import torch
+from torch.utils.data import DataLoader, Dataset
+from torch.utils.data.distributed import DistributedSampler
+
+from nemo_automodel.shared.import_utils import safe_import_from
+
+_MANIFEST_NAME = "manifest.json"
+_TARGET_WEIGHTS_NAME = "target_weights.safetensors"
+_SHARD_RE = re.compile(r"^shard-(\d{6})\.safetensors$")
+_FORMAT_VERSION = 1
+
+CACHE_KEYS = (
+    "input_ids",
+    "attention_mask",
+    "loss_mask",
+    "target_hidden_states",
+    "target_last_hidden_states",
+)
+
+DTYPE_MAP = {"bf16": torch.bfloat16, "fp16": torch.float16, "fp32": torch.float32}
+
+
+def _load_safetensors():
+    """Return ``(save_file, safe_open)`` or raise a clear error if safetensors is missing."""
+    has_save, save_file = safe_import_from("safetensors.torch", "save_file")
+    has_open, safe_open = safe_import_from("safetensors", "safe_open")
+    if not (has_save and has_open):
+        raise ImportError(
+            "The DSpark offline cache requires the 'safetensors' package. "
+            "Install it with `uv sync --locked` and re-run."
+        )
+    return save_file, safe_open
+
+
+def _atomic_write(path: str, write_fn: Callable[[str], None]) -> str:
+    """Write through a sibling temporary file, then atomically replace the target path."""
+    os.makedirs(os.path.dirname(path) or ".", exist_ok=True)
+    tmp_path = path + ".tmp"
+    write_fn(tmp_path)
+    os.replace(tmp_path, path)
+    return path
+
+
+def shard_path(cache_dir: str, shard_index: int) -> str:
+    """Return the path of shard ``shard_index`` inside ``cache_dir``."""
+    return os.path.join(cache_dir, f"shard-{shard_index:06d}.safetensors")
+
+
+def manifest_path(cache_dir: str) -> str:
+    """Return the manifest path inside ``cache_dir``."""
+    return os.path.join(cache_dir, _MANIFEST_NAME)
+
+
+def existing_shard_indices(cache_dir: str) -> set[int]:
+    """Return the shard indices already present in ``cache_dir``."""
+    indices: set[int] = set()
+    if not os.path.isdir(cache_dir):
+        return indices
+    for name in os.listdir(cache_dir):
+        match = _SHARD_RE.match(name)
+        if match is not None:
+            indices.add(int(match.group(1)))
+    return indices
+
+
+def write_manifest(cache_dir: str, manifest: dict[str, Any]) -> str:
+    """Persist the DSpark cache manifest."""
+
+    def _write(tmp_path: str) -> None:
+        with open(tmp_path, "w") as f:
+            json.dump({"format_version": _FORMAT_VERSION, **manifest}, f, indent=2, sort_keys=True)
+
+    return _atomic_write(manifest_path(cache_dir), _write)
+
+
+def read_manifest(cache_dir: str) -> dict[str, Any]:
+    """Load and validate the DSpark cache manifest."""
+    path = manifest_path(cache_dir)
+    if not os.path.exists(path):
+        raise FileNotFoundError(f"DSpark cache manifest not found at {path}. Was the cache fully written?")
+    with open(path) as f:
+        manifest = json.load(f)
+    version = manifest.get("format_version")
+    if version != _FORMAT_VERSION:
+        raise ValueError(
+            f"DSpark cache at {cache_dir} has format_version={version}, expected {_FORMAT_VERSION}. "
+            "Regenerate the cache with the current precompute_dspark."
+        )
+    return manifest
+
+
+def _target_weights_path(cache_dir: str) -> str:
+    return os.path.join(cache_dir, _TARGET_WEIGHTS_NAME)
+
+
+def write_target_weights(cache_dir: str, embed_tokens: torch.nn.Module, lm_head: torch.nn.Module) -> str:
+    """Persist target embedding and lm_head weights for target-free draft initialization."""
+    embed_weight = getattr(embed_tokens, "weight", None)
+    head_weight = getattr(lm_head, "weight", None)
+    if embed_weight is None:
+        raise ValueError("Target input embeddings do not expose a .weight tensor.")
+    if head_weight is None:
+        raise ValueError("Target output embeddings / lm_head do not expose a .weight tensor.")
+    save_file, _ = _load_safetensors()
+    tensors = {
+        "embed_tokens.weight": embed_weight.detach().to(torch.float32).cpu().contiguous(),
+        "lm_head.weight": head_weight.detach().to(torch.float32).cpu().contiguous(),
+    }
+    return _atomic_write(_target_weights_path(cache_dir), lambda tmp: save_file(tensors, tmp))
+
+
+def read_target_weight_modules(cache_dir: str) -> tuple[SimpleNamespace, SimpleNamespace]:
+    """Load target weights and return module-like objects exposing ``.weight``."""
+    _, safe_open = _load_safetensors()
+    path = _target_weights_path(cache_dir)
+    if not os.path.exists(path):
+        raise FileNotFoundError(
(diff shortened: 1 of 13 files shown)


nemo_automodel/recipes/llm/train_dspark.py
```diff
@@ -1099,6 +1204,29 @@ def _log_saved_checkpoint(self, kind: str, epoch: int, step: int) -> None:
         if self.dist_env.is_main and ckpt_cfg is not None and ckpt_cfg.enabled:
             logger.info("Saved %s checkpoint to %s/epoch_%d_step_%d", kind, ckpt_cfg.checkpoint_dir, epoch, step)
 
+    def _forward_batch(self, batch):
+        """Run one batch through live target capture or the offline cache."""
+        batch = {k: v.to(self.device, non_blocking=True) for k, v in batch.items()}
```
khazic, line 1233: The cached branch moves tensors to device but never casts them to `compute_dtype`, and `manifest["dtype"]` is never validated. A cache precomputed with `--dtype fp32` or `fp16` (or a bf16 cache trained on CPU) keeps its stored dtype all the way into the draft's `fc` and crashes with `mat1 and mat2 must have the same dtype` on single-GPU and DDP runs; only FSDP2 survives via `cast_forward_inputs`. Suggest casting the float keys to `self.compute_dtype` here, plus a manifest dtype check.

nemo_automodel/recipes/llm/train_dspark.py
```diff
@@ -333,6 +339,36 @@ def _apply_draft_activation_checkpointing(draft_model: torch.nn.Module, mode: bo
         logger.info("Enabled full activation checkpointing on %d draft layers", len(layers))
 
 
+def _validate_cached_dspark_manifest(
```
khazic, line 309: This validates shapes only, so a cache built from a different checkpoint with the same dims (e.g. two finetunes of the same base, or base vs instruct) passes and the draft silently trains on the wrong model's supervision. `target_model_type` and `num_hidden_layers` are already recorded in the manifest but never compared; checking them here would catch most mixups. The PR description mentions validating "target identity", but that part isn't implemented.

nemo_automodel/recipes/llm/train_dspark.py
```diff
@@ -333,6 +339,36 @@ def _apply_draft_activation_checkpointing(draft_model: torch.nn.Module, mode: bo
...
+            f"but the configured target/layers need {expected_hidden_dim} "
+            f"(hidden_size {hidden_size} x {len(target_layer_ids)} target layers)."
+        )
+    if int(manifest["target_last_hidden_dim"]) != hidden_size:
+        raise ValueError(
+            f"DSpark cache at {cache_dir} has target_last_hidden_dim={manifest['target_last_hidden_dim']}, "
+            f"but the configured target has hidden_size={hidden_size}."
+        )
+    recorded_layer_ids = [int(x) for x in manifest["target_layer_ids"]]
+    if recorded_layer_ids != list(target_layer_ids):
```
khazic, line 377: `manifest["seq_length"]` is also recorded but never checked, and `recipe_cfg.seq_length` is unused on the cached path. A 2048-token cache under a `seq_length: 4096` config trains silently at 2048. Worth a one-line check alongside these.

nemo_automodel/components/datasets/llm/dspark_cache.py
```diff
@@ -0,0 +1,250 @@
...
+from nemo_automodel.shared.import_utils import safe_import_from
+
+_MANIFEST_NAME = "manifest.json"
+_TARGET_WEIGHTS_NAME = "target_weights.safetensors"
+_SHARD_RE = re.compile(r"^shard-(\d{6})\.safetensors$")
+_FORMAT_VERSION = 1
+
+CACHE_KEYS = (
+    "input_ids",
+    "attention_mask",
```
khazic: `attention_mask` is written into every shard and moved H2D every training step, but the cached forward never consumes it: `DSparkTrainerModule.forward` has no `attention_mask` param, and its only online consumer (`generate_batch`) is bypassed in cache mode. It can be dropped from the cached training keys to save disk and a per-step copy.

nemo_automodel/components/datasets/llm/dspark_cache.py
```diff
@@ -0,0 +1,250 @@
...
+    """Persist target embedding and lm_head weights for target-free draft initialization."""
+    embed_weight = getattr(embed_tokens, "weight", None)
+    head_weight = getattr(lm_head, "weight", None)
+    if embed_weight is None:
+        raise ValueError("Target input embeddings do not expose a .weight tensor.")
+    if head_weight is None:
+        raise ValueError("Target output embeddings / lm_head do not expose a .weight tensor.")
+    save_file, _ = _load_safetensors()
+    tensors = {
+        "embed_tokens.weight": embed_weight.detach().to(torch.float32).cpu().contiguous(),
```
khazic: Upcasting the bf16 target weights to fp32 doubles the file size and startup read (roughly +2.4 GB for a 150k-vocab target) with no precision gain, since the draft casts them back to `compute_dtype` anyway. Storing them in the cache dtype seems better.

nemo_automodel/components/speculative/precompute_dspark.py
```diff
@@ -0,0 +1,316 @@
...
+            path = write_shard(args.output_dir, shard_index, merged)
+            logger.info("Wrote %s (%d samples)", path, merged["input_ids"].shape[0])
+        chunks = []
+        buffered = 0
+        shard_index += 1
+
+    with torch.no_grad():
+        for batch in dataloader:
+            batch_size = batch["input_ids"].shape[0]
+            if shard_index in existing:
```
khazic: On `--resume`, every skipped batch still goes through the dataloader, so `ChatDataset` re-renders the chat template and re-tokenizes all completed shards before reaching new work. Since `shuffle=False` gives a deterministic order, skipping directly to the first missing shard's sample index would avoid that.

nemo_automodel/components/datasets/llm/dspark_cache.py
```diff
@@ -0,0 +1,250 @@
...
+    "attention_mask",
+    "loss_mask",
+    "target_hidden_states",
+    "target_last_hidden_states",
+)
+
+DTYPE_MAP = {"bf16": torch.bfloat16, "fp16": torch.float16, "fp32": torch.float32}
+
+
+def _load_safetensors():
```
khazic: Most of this module (`_atomic_write`, shard/manifest helpers, `existing_shard_indices`, the cached dataset/dataloader) is a line-for-line copy of `eagle3_cache.py`, with only `CACHE_KEYS` differing; the `_flush`/resume driver in `precompute_dspark.py` is likewise copied from `precompute_eagle3.py`. Any fix to the on-disk format contract now has to land twice. Consider extracting a shared offline-cache base parameterized by the key set.

nemo_automodel/components/speculative/precompute_dspark.py
```diff
@@ -0,0 +1,316 @@
...
+from nemo_automodel.components.datasets.llm.formatting_utils import (
+    _has_chat_template,
+    _resolve_chat_template,
+)
+from nemo_automodel.components.speculative.dspark.registry import build_target_layer_ids
+from nemo_automodel.components.speculative.dspark.target import HFDSparkTargetModel
+
+logger = logging.getLogger(__name__)
+
+_GEMMA4_MODEL_TYPES = ("gemma4", "gemma4_unified")
```
khazic: These constants and `_read_target_model_type` / `_apply_target_chat_template` duplicate `train_dspark.py`. Precompute and training must agree exactly on model_type detection and chat-template application for the cached loss mask to match training; sharing one copy (e.g. in `dspark/common.py`) removes the drift risk.

khazic:
Once this merges, I will open a follow-up PR adding offline precompute support for very large targets (DeepSeek V4, GLM 5.2, MiniMax M3): the precompute entry point will reuse the recipe's distributed target construction (EP/FSDP sharded loading and the same backend settings), so the offline target forward stays consistent with the online path.

nemo_automodel/components/datasets/llm/dspark_cache.py
```diff
@@ -0,0 +1,172 @@
...
+    embed_weight = getattr(embed_tokens, "weight", None)
+    head_weight = getattr(lm_head, "weight", None)
+    if embed_weight is None:
+        raise ValueError("Target input embeddings do not expose a .weight tensor.")
+    if head_weight is None:
+        raise ValueError("Target output embeddings / lm_head do not expose a .weight tensor.")
+    save_file, _ = load_safetensors(_CACHE_NAME)
+    tensors = {
+        "embed_tokens.weight": embed_weight.detach().to(dtype).cpu().contiguous(),
+        "lm_head.weight": head_weight.detach().to(dtype).cpu().contiguous(),
```
khazic: For tied-embedding targets (e.g. the example `Qwen/Qwen3-0.6B`, which sets `tie_word_embeddings=true`), `get_input_embeddings().weight` and `get_output_embeddings().weight` are the same tensor. On `--device cpu --dtype fp32`, `.to(dtype).cpu().contiguous()` is a no-op chain (already CPU, already fp32, already contiguous), so both dict entries reference the same storage and `safetensors.save_file` raises `Some tensors share memory`, aborting the cache write before any shard is produced. This is exactly the CPU + fp32 cache path that `_validate_cached_dspark_manifest` documents as required for CPU cached training. On GPU it happens to be masked because each `.cpu()` copies. Please force distinct storage, e.g. add `.clone()` to each tensor (or clone one when `head_weight is embed_weight`).

khazic:
LGTM!

HuiyingLi (approved):

PR #2927, opened 2026-07-05
feat(pp): support FusedLinearCrossEntropy under pipeline parallelism

### LINKED ISSUE #2681: FusedLinearCrossEntropy silently no-ops under pipeline parallelism (and doesn't function under PP)
## Summary

`FusedLinearCrossEntropy` (FInCE) **silently falls back to `MaskedCrossEntropy` for every pipeline-parallel run**, so the memory-saving fused loss is never actually used under PP — the last pipeline stage still materializes the full `[batch, seq, vocab]` logits. This affects **all** PP recipes that set `loss_fn: FusedLinearCrossEntropy` (e.g. `examples/llm_pretrain/deepseekv3_pretrain.yaml` at `pp_size: 8`). Beyond the silent fallback, FInCE also does **not currently function under PP even if the fallback is bypassed** — so this issue tracks both (a) stop the silent no-op and (b) make the fused/parallel CE actually work with pipeline parallelism.

Observed (e.g. `llama_3_3_70b_instruct_squad` with `pp_size>1`):
```
WARNING | nemo_automodel.recipes.llm.train_ft | logits_to_keep not found in model.forward. Using MaskedCrossEntropy instead.
```

## Root cause (silent fallback)

`recipes/llm/train_ft.py` (~line 806), run during `setup()`:
```python
if not _supports_logits_to_keep(model) and not isinstance(self.loss_fn, MaskedCrossEntropy):
    logger.warning("logits_to_keep not found in model.forward. Using MaskedCrossEntropy instead.")
    self.loss_fn = MaskedCrossEntropy()

if isinstance(model, AutoPipeline):       # <-- the unwrap happens AFTER the check
    self.model_parts = model.parts
    self.pp = model
```
When PP is enabled, `build_model()` returns an **`AutoPipeline`**, and the capability check runs on that wrapper *before* it is unwrapped to `model_parts`. `_supports_logits_to_keep` → `_get_forward_signature` (`components/utils/model_utils.py`) does `if not callable(getattr(model, "forward", None)): return None` — and `AutoPipeline` (`components/distributed/pipelining/autopipeline.py`) **defines no `forward`** — so it returns `False` unconditionally. FInCE is therefore swapped out for `MaskedCrossEntropy` on every PP run, regardless of whether the underlying model supports `logits_to_keep` (the patched `pipeline_forward_causal_lm` in `pipelining/hf_utils.py` actually does accept it).

## Why FInCE doesn't work under PP even if the check is fixed

Fixing only the check is **necessary but not sufficient**, and would currently crash:

1. **The last PP stage emits projected logits, not hidden states.** `pipeline_forward_causal_lm` (`pipelining/hf_utils.py`) applies `lm_head(hidden_states)` and returns the `[mb, seq, vocab]` logits (and `_precompute_stage_shapes` in `pipelining/functional.py` allocates the last-stage output as `[mb, seq, vocab]`). The PP point-to-point contract requires each stage to emit a tensor, so the full logits are materialized regardless of the loss — which is exactly the memory FInCE is supposed to avoid.
2. **`logits_to_keep` / `output_hidden_states` are never threaded through the schedule.** The PP step calls `self.pp.info.schedule.step(input_ids, target=targets, losses=losses, **batch_filtered)` (`train_ft.py` ~1203) — no `logits_to_keep`, no `output_hidden_states`. (Contrast the non-PP path `out = model(logits_to_keep=1, **batch)`.)
3. **The PP loss path feeds FInCE `hidden_states=None`.** `PipelineCausalLMLoss` (`components/loss/mtp.py`) → `calculate_loss` (`components/loss/utils.py:85-93`): for `FusedLinearCrossEntropy` it pops `hidden_states` and `lm_weight`. The stage output is a bare logits tensor, so `_get_final_hidden_states(output)` returns `None` → FInCE receives `hidden_states=None` and fails.

(FInCE is `cut_cross_entropy`-backed — `components/loss/linear_ce.py` → `from cut_cross_entropy import linear_cross_entropy`.)

## Impact

- Memory-saving fused/parallel CE is silently disabled under PP; the last stage materializes the full `[seq, vocab]` logits (large for big-vocab models, e.g. 128256 for Llama‑3.3‑70B), making the last PP stage the memory-heaviest and limiting batch/seq on memory-tight setups.
- The "fix" is undetectable to users (only a one-line warning) and the claimed loss is not what runs.

## Proposed fixes

**(a) Stop the silent no-op** — run the capability check on the real module, not the wrapper:
```python
if isinstance(model, AutoPipeline):
    self.model_parts = model.parts; self.pp = model
    ...
else:
    self.model_parts = [model]; self.pp = None
loss_probe = self.model_parts[-1] if self.pp is not None else self.model_parts[0]
if not _supports_logits_to_keep(loss_probe) and not isinstance(self.loss_fn, MaskedCrossEntropy):
    # error instead of silently reverting, OR proceed if PP fused-CE is supported
```
At minimum it should **error** rather than silently swap the loss the user asked for.

**(b) Make fused/parallel CE actually work under PP:**
- Have the last stage output **hidden states** (skip the `lm_head` projection) when a fused loss is active, and thread that through `schedule.step` (e.g. an `output_hidden_states`/`logits_to_keep` flag).
- Have `PipelineCausalLMLoss` perform the fused `lm_head` + CE on the last stage's hidden states (give `calculate_loss` the `hidden_states` + `lm_weight`).
- Ensure correctness with **tensor-parallel-sharded vocab** (parallel CE) since these recipes run `tp>1`, so the lm_head/vocab is sharded across TP ranks.

## Repro

Any PP recipe with `loss_fn: FusedLinearCrossEntropy`, e.g. `examples/llm_pretrain/deepseekv3_pretrain.yaml` (`pp_size: 8`) or `examples/llm_finetune/llama3_3/llama_3_3_70b_instruct_squad.yaml` with `pp_size>1` → the warning fires and `MaskedCrossEntropy` runs instead.

Found while fixing the Llama‑3.3‑70B SQuAD H100 OOM (#2673); the working recipe there uses `MaskedCrossEntropy` (which functions correctly under PP), so this is not blocking that PR — but the fused-CE memory savings are unavailable under PP until the above is addressed.

beccohov:
# What does this PR do ?

  Makes `FusedLinearCrossEntropy` actually run under pipeline parallelism instead of
  silently falling back to `MaskedCrossEntropy` on every PP run (#2681).

  # Changelog

  - `recipes/llm/train_ft.py`: run the loss-capability check on the real last-stage
    module, not the `AutoPipeline` wrapper (which defines no `forward`, so the check
    always failed and silently swapped `FusedLinearCrossEntropy` for
    `MaskedCrossEntropy`). Extracted into `_maybe_downgrade_loss_fn`.
  - `recipes/llm/train_ft.py`: gate fused CE under PP to the generic patched forward —
    MoE/custom-forward models (e.g. deepseekv3) fall back to `MaskedCrossEntropy` with a
    clear warning instead of crashing; flag the last stage to emit hidden states when
    fused CE is active.
  - `distributed/pipelining/hf_utils.py`: the last stage skips the `lm_head` projection
    and returns hidden states when flagged; mark the model fused-CE-capable only when the
    generic forward is installed.
  - `distributed/pipelining/functional.py`: size the last-stage output as
    `[mb, seq, hidden]` instead of `[mb, seq, vocab]` on the fused path.
  - `components/loss/mtp.py`: `PipelineCausalLMLoss` treats a bare last-stage tensor as
    hidden states for `FusedLinearCrossEntropy` (running the fused `lm_head`+CE) and raises
    a clear error for the unsupported MTP-tuple + fused-CE combination.
  - Tests: forward emits hidden states; stage output shape switches to hidden; pipeline
    loss routes the bare tensor to hidden states; capability/gate truth table.

  **Pre checks**:

  # Additional Information

  - Related to #2681
  - **Scope:** generic HF causal-LM PP path (covers the Llama-3.3-70B repro). MoE/MTP
    custom-forward models (deepseekv3, nemotron_v3, step3p7) are gated to
    `MaskedCrossEntropy` and left as a follow-up.
  - **Validation:** PP=4 × TP=2 tiny Llama — fused CE trains without the fallback
    warning and produces bit-identical loss/grad-norm to `MaskedCrossEntropy` across steps
    (confirms correctness under TP-sharded vocab).


beccohov:
Hey @akoumpa,
as promised -- here is the PR!

akoumpa:
Hi @beccohov this looks awesome, thank you for making this.

beccohov:
@akoumpa Hey, looks like as a next step I'll need a review 🤗

akoumpa:
Hi @beccohov , thank you for making this! 

I've updated the claude review instructions in main, and i've merged it into your branch; our claude review facility is currently encountering a minor technical issue, so i can't trigger it just yet. If you do have access to claude, please ask it to review according to the repo's guidelines. 

Once we have claude's input, i'll also proceed with reviewing. But from a quick look, I encourage you to include a functional test (e.g., 2 rank + pp=2 with fince to make sure this does not regress).

Thanks again for making this.

beccohov:
Thanks @akoumpa !
I've added the PP test and used Claude for review with the updated instruction. 
Please enable CI again since I committed my changes

akoumpa (approved):

beccohov:
@akoumpa looks like one test was cancelled or failed due to infra. So we need to restart it 🤔

akoumpa:
Hi @beccohov merged main and restarted ci 🤞

beccohov:
Looks like we are ready to merge!
cc @akoumpa

Issue #1027, opened 2026-01-08
(partial) Cudagraph + MoE


akoumpa:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

yaoyu-33:
Proposed direction: treat #1027 as the main tracker for partial MoE CUDA Graph support (related: #551, #678).

Start with Transformer Engine scoped graphs, not full-iteration capture:

```text
Attention   graph
Router      graph
Dispatch    eager (dynamic communication/token counts)
Preprocess  graph once shapes are fixed
Experts     graph later with fixed bucket/capacity
Combine     eager
```

This follows Megatron-Bridge’s safer `transformer_engine` scoped approach. The main gaps on AutoModel `main` are: (1) no stable `cuda_graph_impl/scope/warmup` config, (2) no warmup/capture/replay/cleanup lifecycle, (3) no unified scope discovery across HF/custom models, and (4) missing runtime contracts for stable tensor metadata/storage, FSDP2/DTensor parameter addresses, RNG replay, and dynamic MoE dispatch.

There is substantial prior work on `origin/codex/gptoss-te-moe-graphs-gb200`: `f3a6bed2` adds TE `make_graphed_callables()` scoped boundaries; later commits add explicit FSDP parameter inputs, expert buckets, overflow-to-eager fallback, RNG replay, and full-iteration capture. We should extract/rebase the minimal partial-graph stack first rather than begin from scratch or land the full-iteration/paged-stash stack initially.

Suggested M0 (3–5 engineering days): establish a fixed-shape Qwen3-MoE 30B-A3B BF16 baseline with TE attention/experts and FSDP2+EP; record loss/grad parity, step time, tokens/s, and peak memory. First scope: `attn` only. Explicitly exclude PP, CP, packing, FP8, full activation checkpointing, and dynamic sequence lengths from the initial bring-up.

yaoyu-33:
### AutoModel vs. Megatron-Bridge gap

| Capability | Megatron-Bridge | AutoModel `main` | Gap to close |
|---|---|---|---|
| Configuration | `cuda_graph_impl`, scope/modules, warmup steps | No CUDA Graph configuration | Add a stable, opt-in config contract |
| TE capture | `TECudaGraphHelper` + TE `make_graphed_callables()` | No capture lifecycle | Add warmup, capture, replay, reset, and cleanup |
[...]

PR #2932, opened 2026-07-06
feat(cuda-graphs): capture fixed-capacity MoE execution

yaoyu-33:
# What does this PR do ?

Adds a fixed-capacity `moe` CUDA Graph scope on top of #2917. The captured boundary starts after routing and includes token dispatch, local Transformer Engine `GroupedLinear` experts, and token combine. Attention and MoE scopes can be enabled together with `cuda_graph_modules: [attn, moe]` under FSDP2.

This is part of #1027. It is built on #2917, whose scoped partial-capture implementation credits @hemildesai as co-author.

Stack note: #2943 is the follow-up for shared graph-pool memory. Transactional capture rollback now lives in the #2917 base, so this PR no longer relies on #2943 for failure cleanup.

# Changelog

- Add `cuda_graph_modules: [moe]` and `cuda_graph_moe_capacity_factor` configuration.
- Support EP=1 with local drop-and-pad permutation, fixed TE expert splits, and local unpermutation, including multi-rank FSDP2 jobs.
- Support EP>1 through HybridEP with Megatron-style per-expert drop-and-pad routing.
- Set the static HybridEP receive size to `source_capacity * ep_size * num_local_experts`, avoiding `tokens_per_expert.sum()` and its host synchronization.
- Pass attention and TE expert parameters as explicit graph inputs and guard replay with the existing tensor/control contract checks.
- Unshard the owning FSDP2 transformer block while capturing attention or MoE, then replay with the live materialized parameters inside the block pre-forward lifecycle.
- Keep BF16, activation checkpointing, dynamic shapes, and unsupported dispatcher combinations fail-closed.
- Gate fixed-capacity TE MoE execution on `cuda_graph_modules: [moe]`; setting `cuda_graph_moe_capacity_factor` alone is rejected.
- Avoid CUDA tensor `.tolist()` in the fixed-capacity down-bias path.
- Clarify that HybridEP capacity changes are fail-closed and that capacity limiting may drop routed assignments.

# Design notes

For `N` source tokens, top-k `K`, `E` global experts, and capacity factor `F`, source capacity is:

`ceil(N * K * F / E)`

Every source rank selects exactly that many token slots for every global expert. Real assignments are selected by routing probability; unused slots have zero routing probability. Therefore every local expert receives exactly `source_capacity * ep_size` rows and TE receives constant `m_splits`.

This follows the Megatron Core `moe_pad_expert_input_to_capacity` / HybridEP path. Legacy DeepEP remains rejected: its current dispatch path waits on dynamic communication metadata on the CPU, and padding only after dispatch cannot make capture safe. A DeepEP v2 static-buffer integration would be separate work.

# CUDA experiments

## Combined-scope attribution

Tiny Qwen3-MoE, 2 layers, hidden size 512, 8 experts, top-2, BF16, sequence length 512, local/global batch size 2/4, FSDP2 over 2 H100s. Timing is 8 measured forward/backward/Adam steps after 4 warmup steps.

| Scope | Iteration | Latency vs eager | Throughput vs eager | Peak allocated | Graph stats |
| eager | 27.072 ms | - | - | 0.35 GB | - |
| `attn` | 24.882 ms | -8.1% | +8.8% | 0.53 GB | captured=2, replayed=22, fallback=0 |
| `moe` | 23.702 ms | -12.4% | +14.2% | 0.40 GB | captured=2, replayed=22, fallback=0 |
| `attn, moe` | 20.920 ms | -22.7% | +29.4% | 0.60 GB | captured=4, replayed=44, fallback=0 |

All four modes report the same 12-step loss sequence. This tiny-model run is useful for attribution, but launch overhead is overrepresented.

These numbers compare dynamic eager routing with fixed-capacity graph execution. They therefore include both CUDA Graph launch savings and the change to capacity-limited drop-and-pad work; they are not a pure graph-on/graph-off attribution. Capacity limiting may drop routed assignments when an expert exceeds its configured slots.

## Qwen3-30B-A3B low-load sanity check

`Qwen/Qwen3-30B-A3B`, 48 layers, BF16, sequence length 1024, local/global batch size 1/8, FSDP2 + EP8 HybridEP over 8 H100s, real learned router, 25 forward/backward/Adam steps with 5 warmup steps. The order was eager 1, graph 1, graph 2, eager 2. Setup, TE compilation, and graph capture are excluded from steady-state iteration timing.

| Mode | Run 1 | Run 2 | Mean | Mean MFU | Peak allocated |
| eager | 457.622 ms | 455.914 ms | 456.768 ms | 4.39% | 32.56 GB |
| `attn, moe` | 304.391 ms | 300.639 ms | 302.515 ms | 6.63% | 53.53 GB |

Mean latency decreases by 33.8%; equivalent training throughput increases by 51.0%. Each graph run captured all 96 entries (attention + MoE for 48 layers), replayed 2,304 calls, and had zero fallback. This is a deliberately underloaded launch-overhead result (4.39% eager MFU), not a production-throughput claim.

The memory cost is material: +20.97 GB/GPU in this configuration. Capture is also a one-time startup cost and is not included in the steady-state speedup.

For a controlled replay parity check, the same 30B configuration was run with `lr=0` for 6 steps. Eager and graph produced the same reported loss on every step: `10.3457, 10.3424, 10.3292, 10.3191, 10.3709, 10.3198`. The graph run captured 96 entries, replayed 480 calls, and had zero fallback. With nonzero learning rate, repeated distributed HybridEP/BF16 runs are not bitwise deterministic, so the controlled no-update run is the forward/replay parity check.

## Existing-recipe scaling check

The repository H100 recipe uses local batch 4, sequence length 4096, TE projections/norm/RoPE, DeepEP, and activation checkpointing. It cannot be used as a pure graph-on/graph-off comparison yet: whole-attention graph currently requires torch projections, torch RMSNorm, and unfused RoPE; full MoE graph requires HybridEP and rejects activation checkpointing.

Using the closest currently supported backend on 8 H100-80GB GPUs (`linear: torch`, `rms_norm: torch`, `rope_fusion: false`, TE attention/experts, HybridEP, FSDP2 + EP8), the doubled-batch result is:

| Scope | Local batch x sequence | Iteration | Latency vs eager | Throughput vs eager | Peak allocated | Graph stats |
|---|---:|---:|---:|---:|---:|---:|
| eager | 2 x 1024 | 481.675 ms | - | - | 39.00 GB | - |
| `attn` | 2 x 1024 | 441.681 ms | -8.3% | +9.1% | 45.50 GB | captured=48, replayed=672, fallback=0 |
| `moe` | 2 x 1024 | 405.294 ms | -15.9% | +18.9% | 54.00 GB | captured=48, replayed=672, fallback=0 |
| `attn, moe` | 2 x 1024 | OOM during capture | - | - | 75.43 GiB process usage | failed at layer 45 MoE |

At local batch 2 x sequence 2048, eager runs at 539.834 ms (15.80% MFU, 53.61 GB), but combined capture OOMs earlier at layer 27 attention with 75.37 GiB process usage. At 2 x 1024, combined capture reaches layer 45 before OOM. This identifies cumulative graph-pool memory across 96 per-layer entries as the immediate scaling blocker. Both 48-entry single scopes fit and replay without fallback.

# Validation

- `uv run ruff format .`
- `uv run ruff check --fix .`
- `uv run pytest -q tests/unit_tests/moe tests/unit_tests/recipes/test_partial_cuda_graphs.py`
  - 539 passed, 54 skipped.
- Slurm job 5579217: 2-H100 FSDP2 eager/attention/MoE/combined attribution.
- Slurm job 5579222: 8-H100 Qwen3-30B-A3B symmetric repeated E2E benchmark.
- Slurm job 5579279: 8-H100 Qwen3-30B-A3B controlled `lr=0` replay parity.
- Slurm jobs 5579426 and 5579467: recipe-derived combined-capture memory scaling at 2 x 2048 and 2 x 1024.
- Slurm job 5579620: recipe-derived attention-only and MoE-only attribution at 2 x 1024.

# Known limitations

- Fixed input shape and explicit opt-in only.
- BF16 TE experts only; FP8 is rejected.
- Activation checkpointing and pipeline parallelism are rejected for the full MoE scope.
- Legacy DeepEP is not graph-safe and is rejected; EP>1 currently requires HybridEP.
- Whole-attention graph currently requires torch projections, torch RMSNorm, and unfused RoPE, while the existing H100 recipe uses the TE variants.
- The Qwen3-30B run validates FSDP2 for the transformer blocks and EP8 distribution for experts. Additional FSDP sharding of each rank's local expert weights (`ep_shard > 1`) is not validated.
- CUDA Graph pools increase persistent GPU memory; the 30B combined run used 53.53 GB/GPU versus 32.56 GB/GPU eager.
- Combined attention + MoE capture does not fit at local batch 2 even with sequence length 1024 on H100-80GB; graph-pool sharing/reuse is required before matching existing recipe operating points. Follow-up #2943 addresses this with shared-pool capture and transactional cleanup.

- [x] Read and followed contributor guidelines.
- [x] Added unit coverage for config, capacity/drop-and-pad metadata, static HybridEP sizing, combined scopes, and FSDP2 graph lifecycle.
- [ ] Add user documentation after the API and supported matrix are agreed.

# Additional Information

- Based on #2917
- Related to #1027

### DIFF
--- a/nemo_automodel/components/models/common/utils.py
+++ b/nemo_automodel/components/models/common/utils.py
@@ -14,6 +14,7 @@
 
 import importlib.util
 import logging
+import math
 from contextlib import contextmanager, nullcontext
 from dataclasses import dataclass, field
 from typing import Any, Literal
@@ -160,12 +161,18 @@ class CudaGraphConfig:
     Attributes:
         modules: Per-layer modules to capture with Transformer Engine, following
             Megatron Core's module names. AutoModel supports whole ``attn``,
-            ``moe_router``, and ``moe_preprocess`` scopes, plus the
+            ``moe``, ``moe_router``, and ``moe_preprocess`` scopes, plus the
             AutoModel-specific narrow ``te_dpa`` scope. An empty list disables
-            CUDA graphs.
+            CUDA graphs. The ``moe`` scope captures dispatch, local TE experts,
+            and combine after eager routing.
+        moe_capacity_factor: Drop-and-pad capacity relative to the average
+            number of routed tokens per expert. Required by, and only valid
+            with, the ``moe`` graph scope. Capacity limiting may drop routed
+            expert assignments when an expert is overloaded.
     """
 
-    modules: list[Literal["attn", "te_dpa", "moe_router", "moe_preprocess"]] = field(default_factory=list)
+    modules: list[Literal["attn", "te_dpa", "moe", "moe_router", "moe_preprocess"]] = field(default_factory=list)
+    moe_capacity_factor: float | None = None
 
     def __post_init__(self) -> None:
         """Validate the declarative CUDA-graph configuration."""
@@ -173,7 +180,7 @@ class CudaGraphConfig:
             raise TypeError("cuda_graph.modules must be a list")
         if not all(isinstance(module, str) for module in self.modules):
             raise TypeError("cuda_graph.modules entries must be strings")
-        supported_modules = {"attn", "te_dpa", "moe_router", "moe_preprocess"}
+        supported_modules = {"attn", "te_dpa", "moe", "moe_router", "moe_preprocess"}
         unknown_modules = set(self.modules) - supported_modules
         if unknown_modules:
             raise ValueError(
@@ -186,6 +193,16 @@ class CudaGraphConfig:
             raise ValueError("cuda_graph.modules cannot contain both 'attn' and 'te_dpa'")
         if "moe_preprocess" in self.modules and "moe_router" not in self.modules:
             raise ValueError("'moe_preprocess' in cuda_graph.modules requires 'moe_router'")
+        if self.moe_capacity_factor is not None:
+            if not math.isfinite(self.moe_capacity_factor) or self.moe_capacity_factor <= 0:
+                raise ValueError("cuda_graph.moe_capacity_factor must be positive and finite")
+            if "moe" not in self.modules:
+                raise ValueError("cuda_graph.moe_capacity_factor requires 'moe' in cuda_graph.modules")
+        if "moe" in self.modules:
+            if self.moe_capacity_factor is None:
+                raise ValueError("'moe' in cuda_graph.modules requires cuda_graph.moe_capacity_factor")
+            if any(scope in self.modules for scope in ("moe_router", "moe_preprocess")):
+                raise ValueError("'moe' in cuda_graph.modules cannot be combined with moe_router or moe_preprocess")
 
 
 @dataclass(kw_only=True)
@@ -298,6 +315,16 @@ class BackendConfig:
         if isinstance(self.gate_precision, str):
             self.gate_precision = dtype_from_str(self.gate_precision, default=None)
 
+        graph_modules = self.cuda_graph.modules
+        moe_cuda_graph_enabled = "moe" in graph_modules
+        if moe_cuda_graph_enabled:
+            if self.experts != "te":
+                raise ValueError("'moe' in cuda_graph.modules requires experts='te'")
+            if self.dispatcher not in ("torch", "hybridep"):
+                raise ValueError("'moe' in cuda_graph.modules currently requires dispatcher='torch' or 'hybridep'")
+            if self.te_fp8 is not None:
+                raise ValueError("'moe' in cuda_graph.modules currently supports BF16 TE GroupedLinear only")
+
         # enable_deepep was removed. It is no longer honored; warn (once, on rank 0) if a stale
         # config still sets it so the user migrates to explicit dispatcher/experts. The field is
         # retained only so loading an old config does not crash this kw_only dataclass.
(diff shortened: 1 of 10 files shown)


nemo_automodel/components/models/common/utils.py
```diff
@@ -233,7 +239,10 @@ class BackendConfig:
     # fullgraph can't trace), so it requires attn="sdpa", linear="torch", rms_norm="torch",
     # rope_fusion=False. Default False.
     compile_attn: bool = False
-    cuda_graph_modules: list[Literal["attn", "te_dpa", "moe_router", "moe_preprocess"]] = field(default_factory=list)
+    cuda_graph_modules: list[Literal["attn", "te_dpa", "moe", "moe_router", "moe_preprocess"]] = field(
```
akoumpa: Hi @yaoyu-33 can we make a config that's concerned solely with cudagraphing configuration ala Peft and have it here instead of the individual options?

yaoyu-33: Addressed by grouping the CUDA graph options under `BackendConfig.cuda_graph` using the typed `CudaGraphConfig`, which owns `modules`, `moe_capacity_factor`, paged-stash options added downstream, and their validation.

PR #2939, opened 2026-07-06
fix(bagel): make non-backbone init topology independent

zyzhou5:
# What does this PR do ?

Make initialization of BAGEL-owned modules topology-independent for the `hf_backbones` initialization path.

  Previously, these parameters were initialized after FSDP sharding using rank-specific RNG state, so their values depended on world size and sharding topology. This PR initializes each logical parameter from a dedicated CPU generator seeded with the recipe's global seed, then distributes it into the existing DTensor layout.

  Rank-local runtime RNG remains unchanged. This adds initialization-time work only and introduces no per-step or per-microbatch overhead.

### DIFF
--- a/nemo_automodel/components/models/bagel/hf_backbone_loader.py
+++ b/nemo_automodel/components/models/bagel/hf_backbone_loader.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import gc
 import json
 import logging
+import math
 import pathlib
 from typing import Any, Dict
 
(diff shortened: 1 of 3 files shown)


zyzhou5:
/ok to test [ae0d205](https://github.com/NVIDIA-NeMo/Automodel/pull/2939/commits/ae0d2053cdb49e555773eb604a3da24453dc0b3f)

zyzhou5 (commented):
Left some comments. Overall LGTM.

athitten (approved):
LGTM, thank you @zyzhou5 ! Loss curves in the linear issue AM-636

Issue #2935, opened 2026-07-06
refactor(models): single tie_word_embeddings guard with per-class declared TieSupport


yuhezhang-ai:
## Background

NVIDIA-NeMo/Automodel#2896 completed the `tie_word_embeddings` guard audit from NVIDIA-NeMo/Automodel#2512: every registered model with an `lm_head` now either implements tying explicitly or rejects the unsupported flag value at construction. Coverage today:

* ~29 call sites reject `tie=True` via `reject_unsupported_tied_word_embeddings` (untied-only implementations: deepseek_v3/v32/v4, glm4_moe(+lite/dsa), gpt_oss, qwen3_moe/next/…, mistral4, kimivl, …, plus the four families added in NVIDIA-NeMo/Automodel#2896).
* `gemma4_moe` rejects `tie=False` via the new `reject_unsupported_untied_word_embeddings` (NVIDIA-NeMo/Automodel#2896).
* ~10 families implement tying and honor both values (llama, qwen2, baichuan, mistral3, qwen3_5, ernie4_5, bagel, diffusion_gemma, gemma4_drafter).

The remaining problem is API shape, not behavior: each model constructor must choose **which** of the two reject helpers to call. That per-call-site choice is easy to get wrong for composite configs (the supported value depends on where the controlling flag lives — top-level vs `text_config` vs `thinker_config` vs `llm_config`), and nothing enforces that a newly onboarded model makes the choice at all.

### Layer 1 — per-class declared support + single guard

1. Add a three-state declaration and a single constructor-facing helper in `nemo_automodel/components/checkpoint/utils.py`:

```python
class TieSupport(Enum):
    BOTH = "both"            # implements tie_weights(); both configs validated
    TIED_ONLY = "tied_only"
    UNTIED_ONLY = "untied_only"

def reject_unsupported_tie_word_embeddings(model_cls, config) -> None:
    support = getattr(model_cls, "tie_word_embeddings_support", TieSupport.BOTH)
    requested_tied = get_controlling_tie_word_embeddings(config, model_cls.__name__)
    # raise NotImplementedError on TIED_ONLY vs requested untied, or UNTIED_ONLY vs requested tied
```

2. Each model class declares `tie_word_embeddings_support` as a class attribute (deliberately **separate from** `ModelCapabilities`, which is a parallelism concern). Migration mapping:
   * the ~29 reject-tied call sites → `TieSupport.UNTIED_ONLY` (mechanical),
[...]

PR #2942, opened 2026-07-06
fix(transformers): gate _tie_weights_nemo on tie_word_embeddings flag

### LINKED ISSUE #2941: fix(transformers): _tie_weights_nemo force-ties untied models (ignores tie_word_embeddings)
## Summary

`_tie_weights_nemo` (`nemo_automodel/_transformers/model_init.py:994-1007`) re-ties `lm_head.weight` to the input embedding for any model exposing `_nemo_tied_weights_keys` — **without checking** `config.tie_word_embeddings`. For untied models this silently discards or corrupts trained weights.

Introduced in [#1817](<https://github.com/NVIDIA-NeMo/Automodel/issues/1817>) ("fix: tie weights outside `_init_model`") together with the patched `post_init` in `_transformers/utils.py` that *sets* `_nemo_tied_weights_keys`. [#1817](<https://github.com/NVIDIA-NeMo/Automodel/issues/1817>)'s intent was correct — re-tie **tied** pre-v5 remote-code models (Nemotron-Flash-1B) after checkpoint loading breaks the construction-time alias — but the re-tie fires unconditionally. Pre-dates the NVIDIA-NeMo/Automodel#2512/NVIDIA-NeMo/Automodel#2896 guard work; surfaced during the tie-support audit for NVIDIA-NeMo/Automodel#2935.

## Affected paths

Two ways a model gets `_nemo_tied_weights_keys`:

1. **Static class attribute**: `ernie4_5` dense + MoE (`components/models/ernie4_5/model.py:416,520`). An ERNIE config with `tie_word_embeddings=False` branches correctly in `__init__`, then gets force-tied anyway at the end of `_init_model` (`model_init.py:1040-1041`).
2. **Dynamic (broader)**: the patched `post_init` (`_transformers/utils.py`) sets it for **every pre-v5 remote-code model with list-form** `_tied_weights_keys`. In pre-v5 HF convention, declaring the keys does **not** mean the model is tied — HF's own `tie_weights()` checks the config flag. Many models declare the keys but ship untied checkpoints; all of them are force-tied when loaded through `NeMoAuto*`.

## Observable failure modes

* **Fresh** `from_pretrained` **of an untied model** (silent): HF loads the trained `lm_head.weight`, then `_tie_weights_nemo` aliases it away to `embed_tokens.weight`. No error, no spike — logits simply diverge from the HF reference from the first forward; the model is effectively tied although the checkpoint (and the class it will be served with) is untied.
* **Resume/load of a checkpoint with distinct** `lm_head`**/embedding weights** (loud): after construction the two params share one storage; `load_state_dict`/DCP copies both checkpoint tensors into the same storage — last writer wins, corrupting embeddings and/or head → **loss spike on resume**.

## Proposed fix

Gate the re-tie on the controlling config flag (reusing the canonical resolver from `components/checkpoint/utils.py`):

```python
def _tie_weights_nemo(model):
    if not hasattr(model, "_nemo_tied_weights_keys"):
        return
    if not get_controlling_tie_word_embeddings(model.config, type(model).__name__):
        return  # untied config: the loaded separate lm_head is authoritative
    ...
```

Tied models (Nemotron-Flash) keep the [#1817](<https://github.com/NVIDIA-NeMo/Automodel/issues/1817>) behavior unchanged. Add regression tests for both manifestations: (a) untied list-form remote-code model keeps distinct storage + exact HF logits parity through the `NeMoAuto*` path; (b) save-then-reload of distinct weights round-trips losslessly (no spike); (c) tied model still re-ties (no [#1817](<https://github.com/NVIDIA-NeMo/Automodel/issues/1817>) regression).

## Relationship to NVIDIA-NeMo/Automodel#2935

Independent and complementary: NVIDIA-NeMo/Automodel#2935's constructor guards cover registered classes only. The dynamic path here affects **unregistered** HF/remote-code classes, which no per-class declaration can reach — this gate is the only line of defense for them. (The ERNIE static case additionally becomes unreachable once NVIDIA-NeMo/Automodel#2935 demotes ERNIE to tied-only.)

yuhezhang-ai:
# What does this PR do ?

Fixes #2941: `_tie_weights_nemo` re-tied `lm_head.weight` to the input embeddings for any model exposing
`_nemo_tied_weights_keys`, without consulting the controlling `config.tie_word_embeddings` flag.
`_nemo_tied_weights_keys` records candidate tied keys from pre-v5/list-form `_tied_weights_keys`; it does not prove
that the checkpoint is actually tied.

Impact is real for shipped untied remote-code force-HF recipes. Confirmed affected:

- `examples/llm_finetune/nemotron/nemotron_nano_4b_squad.yaml`
- `examples/llm_finetune/nemotron/nemotron_nano_9b_squad.yaml`
- `examples/llm_finetune/nemotron/nemotron_nano_9b_squad_peft.yaml`

Those recipes load Nemotron Nano 4B/9B through `trust_remote_code: true` + `force_hf: true`. The hub configs set
`tie_word_embeddings=False`, the checkpoints contain separate trained `lm_head.weight` and
`backbone.embeddings.weight`, and the recipe compatibility patch converts remote list-form
`_tied_weights_keys = ["lm_head.weight"]` into `_nemo_tied_weights_keys`. Before this PR, `_tie_weights_nemo` then
forced an alias and discarded the trained head.

The fix gates the re-tie on the canonical `get_controlling_tie_word_embeddings` resolver. Tied models, such as the
Nemotron-Flash case #1817 was written for, keep the existing re-tie behavior unchanged. Models without a `config`
attribute conservatively fall back to re-tying.

## Before / after

**Real recipe-path source load: `nvidia/NVIDIA-Nemotron-Nano-9B-v2`.**

Using the same recipe condition that exposes the bug (`apply_cache_compatibility_patches()` + `trust_remote_code=True`
+ `force_hf=True`), compare a raw HF source load against a NeMoAuto source load on a seeded 128-token prompt:

| branch | `lm_head` aliased? | max abs diff vs HF | mean abs diff vs HF | cosine |
|---|---:|---:|---:|---:|
| main | `true` | `22.5625` | `2.9893` | `-0.0224` |
| this PR | `false` | `0.25` | `0.0141` | about `1.000` |

This is the silent manifestation: the model loads successfully, but logits diverge from the raw HF reference immediately
on main because the trained untied `lm_head` is aliased away.

**Unit/mechanism coverage.**

The regression tests fail on main and pass on this branch:

- `test_untied_config_keeps_separate_lm_head`
- `test_untied_state_dict_roundtrip_is_lossless`
- `test_tied_config_reties`
- `test_missing_config_still_reties`

Targeted result on this branch: `4 passed`.

**Resume smoke note.**

A real 9B FSDP2 DCP recipe resume smoke was also run. This PR produced a checkpoint at step 3, then main and this PR
both resumed one step from that same checkpoint. Both resumed step 4 at `loss=0.466808`.

That result means the DCP checkpoint resume path was correct in this smoke. It does not cover the bug fixed here: on
main, the model is already wrong immediately after the initial NeMoAuto source load, before any training checkpoint is
saved or resumed. This is also why the current checkpoint robustness test can miss the issue: it captures reference
logits from the already-loaded NeMo model, then validates checkpoint reloads against that reference. We should add a
separate source-load parity check that compares raw HF source load vs NeMoAuto source load before training starts.

# Changelog

- `_tie_weights_nemo` now returns early when the controlling `tie_word_embeddings` flag is `False`, preserving
  separately loaded `lm_head.weight` for untied models.
- Added `TestTieWeightsNemoConfigGate` regression tests: untied preservation, tied re-tie, missing-config fallback, and
  lossless construct-then-load round-trip.

**Pre checks**:
- [x] Make sure you read and followed the [Contributor
      Guide](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md).
  - `tests/unit_tests/_transformers/test_model_init.py::TestTieWeightsNemoConfigGate`
  - No public API change.
- [x] Does this PR introduce a breaking change?
  - No. This restores the checkpoint-config contract for untied models and preserves tied behavior.

# Additional Information
* Related to #2935 but intentionally kept as a minimal correctness fix instead of refactoring the full tie-weight
  pipeline.
* Introduced by the #1817-era workaround for pre-v5/list-form `_tied_weights_keys`. That workaround is still needed for
  tied remote-code models; this PR only adds the missing config gate for untied checkpoints.
* Existing checkpoint robustness coverage does not exercise this exact source-load failure. It captures reference logits
  from the already-constructed NeMo model, then compares checkpoint reloads against that reference. If the source load is
  already corrupted, the corrupted model can become the baseline. We should add an opt-in raw-HF-source vs
  NeMoAuto-source parity check for remote-code/force-HF recipes.

### DIFF
--- a/nemo_automodel/_transformers/model_init.py
+++ b/nemo_automodel/_transformers/model_init.py
@@ -995,6 +995,15 @@ def _tie_weights_nemo(model):
     if not hasattr(model, "_nemo_tied_weights_keys"):
         return
 
+    # Only re-tie when the controlling config flag actually requests tied
+    # embeddings. ``_nemo_tied_weights_keys`` names the *candidate* tied keys
+    # (pre-v5 list-form ``_tied_weights_keys`` semantics); it does not mean the
+    # model is tied. Re-tying an untied model here would alias away the trained
+    # ``lm_head.weight`` that ``from_pretrained`` just loaded (see #2941).
+    config = getattr(model, "config", None)
+    if config is not None and not checkpoint_utils.get_controlling_tie_word_embeddings(config, type(model).__name__):
+        return
+
     def get_module_by_fqn(model, fqn):
         from functools import reduce
 
--- a/tests/unit_tests/_transformers/test_model_init.py
+++ b/tests/unit_tests/_transformers/test_model_init.py
@@ -703,3 +703,85 @@ class TestTryGetRemoteCodeModelCls:
         cfg.auto_map = {"AutoModelForCausalLM": "modeling.MyModel"}
         result = _try_get_remote_code_model_cls(cfg, "/some/path", "AutoModelForCausalLM", {})
         assert result is None
+
+
+class TestTieWeightsNemoConfigGate:
+    """_tie_weights_nemo must honor the controlling tie_word_embeddings flag (#2941).
+
+    ``_nemo_tied_weights_keys`` names the candidate tied keys (pre-v5 list-form
+    ``_tied_weights_keys`` semantics); it does not mean the model is tied.
+    Re-tying an untied model aliases away the trained ``lm_head.weight`` that
+    ``from_pretrained`` just loaded.
+    """
+
+    @staticmethod
+    def _make_model(tie: bool | None) -> nn.Module:
+        from transformers import PretrainedConfig
+
+        class _TinyModel(nn.Module):
+            _nemo_tied_weights_keys = {"lm_head.weight": "model.embed_tokens.weight"}
+
+            def __init__(self):
+                super().__init__()
+                self.model = nn.Module()
+                self.model.embed_tokens = nn.Embedding(10, 4)
+                self.lm_head = nn.Linear(4, 10, bias=False)
+
+        model = _TinyModel()
+        if tie is not None:
+            model.config = PretrainedConfig(tie_word_embeddings=tie)
+        return model
+
+    def test_untied_config_keeps_separate_lm_head(self):
+        """tie_word_embeddings=False: the loaded lm_head must not be aliased away."""
+        from nemo_automodel._transformers.model_init import _tie_weights_nemo
+
+        model = self._make_model(tie=False)
+        lm_head_before = model.lm_head.weight.detach().clone()
+
+        _tie_weights_nemo(model)
+
+        assert model.lm_head.weight is not model.model.embed_tokens.weight
+        assert model.lm_head.weight.data_ptr() != model.model.embed_tokens.weight.data_ptr()
+        torch.testing.assert_close(model.lm_head.weight, lm_head_before)
+
+    def test_tied_config_reties(self):
+        """tie_word_embeddings=True: keep the #1817 re-tie behavior."""
+        from nemo_automodel._transformers.model_init import _tie_weights_nemo
+
+        model = self._make_model(tie=True)
+        _tie_weights_nemo(model)
+
+        assert model.lm_head.weight is model.model.embed_tokens.weight
+
+    def test_missing_config_still_reties(self):
+        """No config attribute: fall back to the conservative #1817 re-tie."""
+        from nemo_automodel._transformers.model_init import _tie_weights_nemo
+
+        model = self._make_model(tie=None)
+        _tie_weights_nemo(model)
+
+        assert model.lm_head.weight is model.model.embed_tokens.weight
+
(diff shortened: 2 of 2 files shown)


yuhezhang-ai:
/ok to test [9199d1e](https://github.com/NVIDIA-NeMo/Automodel/pull/2942/commits/9199d1e54294845ad3a95f943b463b2886e94aca)

akoumpa (approved):

Issue #2948, opened 2026-07-07
DSpark offline hidden-state generation is single-node only; does not scale to very large targets


khazic:
## Summary

The DSpark offline path (target hidden-state generation) is currently implemented for a **single node only**. It works fine for small/medium targets, but it does not scale to very large targets whose weights cannot be held on one box.

## Details

The offline generation loads and runs the full target model on a single node to capture its intermediate and final hidden states. For very large targets, the model does not fit at load time on a single machine, so the offline path is effectively unusable at that scale.

This limitation is already visible in the training recipe: the only current workaround for a large target on one node is to load a reduced number of layers, and a draft trained against a reduced target is not usable. See `_resolve_reduced_target_layers` in `nemo_automodel/recipes/llm/train_dspark.py`, whose docstring notes the full DeepSeek-V4-Flash target OOMs at load on a single 8x80GB box.

## Plan

I will add a **distributed (multi-node)** implementation of the DSpark offline generation soon, so the target can be sharded across nodes and hidden states can be produced for very large models. Filing this to track the work.

PR #2953, opened 2026-07-07
feat(speculative): add distributed DSpark offline precompute

khazic:
# What does this PR do ?

Adds a distributed (multi-node) DSpark offline precompute, so target supervision can be generated offline for targets too large to load on a single node (DeepSeek-V4-Flash, GLM-5.2). Implements #2948.

The single-process `precompute_dspark` explicitly rejects these model types because the full target OOMs at load on one 8x80GB box. The new entry point loads the target frozen through the same expert-parallel / FSDP2 distributed path the online training recipe already uses (routed experts shard across ranks, FP8 base weights dequantize on load), runs it once over the dataset, and writes the identical DSpark cache that `train_dspark.py` consumes via `recipe_args.cached_target_path`.

Launch mirrors multi-node training:

```bash
torchrun --nnodes=4 --node-rank=0 --nproc_per_node=8 \
  --master-addr=<NODE0_IP> --master-port=29500 \
  -m nemo_automodel.recipes.llm.precompute_dspark_dist \
  -c examples/speculative/dspark/deepseek_v4_flash_dspark_precompute.yaml
```

# Changelog

- Extract the DeepSeek V4 / GLM-5.2 frozen-target builders out of `TrainDSparkRecipe` into `recipes/llm/_dspark_target_build.py` so the training recipe and the precompute share one implementation (no behavior change to online training).
- Add `partition_cache_shards` to `offline_cache.py`: contiguous, shard-aligned per-rank sample blocks. Every shard is owned by exactly one rank, so ranks write global-indexed shards straight into a shared output directory with no post-hoc merge.
- Add `write_cache_shards_distributed`: per-rank write loop that pads its forward-step count to the global maximum with dummy batches, keeping the EP/FSDP collectives in lockstep; shares the shard accumulator with the single-process writer. Writes are idempotent (atomic replace).
- Add `nemo_automodel/recipes/llm/precompute_dspark_dist.py`: config-driven (reuses the training YAML's `distributed:` block for the target mesh), torchrun-launched; rank zero gathers the DTensor-sharded `embed_tokens` / `lm_head` to full tensors and writes the target weights and manifest. Small text targets (Qwen3, Gemma4) are also accepted and run data-parallel replicated.
- Unify the cache schema in `dspark_cache.py` (`compute_batch_cache`, `build_cache_manifest`, `manifest_mismatch_fields`) so the single-process and distributed producers emit identical caches by construction; `precompute_dspark` now points V4/GLM users at the distributed entry point instead of a generic rejection.
- Add example precompute configs for DeepSeek-V4-Flash and GLM-5.2 and document the two-phase offline flow (precompute, then cached training) in the speculative README.
- MiniMax M3 stays out of scope: the DSpark cache schema is text-only, so the multimodal target keeps the online path.

**Pre checks**:

# Additional Information

Related issue: #2948 (closes it).

Validation done (CPU, no GPU access on my dev box):

- `pytest tests/unit_tests/speculative/ tests/unit_tests/recipes/llm/test_precompute_dspark_dist.py tests/unit_tests/recipes/llm/test_train_dspark_helpers.py` passed: `665 passed, 13 skipped` (skips are the GPU-only FlexAttention tests).
- `pytest tests/functional_tests/speculative/test_precompute_dspark_dist_equivalence.py` passed: runs a real tiny random Qwen3 target on two spawned gloo ranks (real all-reduce step sync and barriers, uneven per-rank steps to force the dummy-batch lockstep padding path) and asserts the distributed cache is bit-identical (`rtol=0, atol=0`) to the single-process `write_cache_shards` reference, sample by sample.
- `ruff format --check` / `ruff check` and `lint-imports` pass on all touched files.

Validation still needed (no multi-node GPU access here):

- Real sharded-target run: DeepSeek-V4-Flash or GLM-5.2 with the example precompute config on 2+ nodes (or the single-node `target_num_hidden_layers` reduced smoke), then a short `train_dspark` run consuming the produced cache via `recipe_args.cached_target_path`. The EP/FSDP forward with hidden-state hooks is exactly the code path online training already exercises, but the dummy-batch lockstep padding with real NCCL collectives deserves a live check.


khazic:
@HuiyingLi Could you help validate the GPU/multi-node part of this PR when you get a chance? The CPU-side correctness is covered (a functional test asserts the 2-rank distributed cache is bit-identical to the single-process reference), but I have no multi-node GPU access, so two things still need a live check: (1) a real sharded-target precompute with the DeepSeek-V4-Flash or GLM-5.2 example config (a single-node reduced-layer smoke via target_num_hidden_layers also works), and (2) a short train_dspark run consuming the produced cache through recipe_args.cached_target_path. Happy to iterate on anything that comes up. Thanks!

HuiyingLi:
Validated at head `7dcc9fdc87dbe55adb8cc2bdea04a36fa0be297c` — the requested GPU/multi-node validation passes.

- Ran a full 43-layer DeepSeek-V4-Flash distributed precompute on 2 nodes / 16 × 80 GB GPUs with FSDP2 + EP16 (no reduced-layer override), using 64 samples at sequence length 128 and cache shard size 5.
- The uneven-work path was exercised with real NCCL/DeepEP collectives: rank 12 ran 4 real + 1 dummy forward, while ranks 13–15 ran 0 real + 5 dummy forwards. All ranks stayed in lockstep with no collective timeout or desynchronization.
- Precompute produced 13 contiguous cache shards plus the gathered target embedding/lm-head weights. Validation confirmed all expected keys, shapes and dtypes, finite tensors, 64 samples, and 1,472 supervised tokens.
- Ran a short cached `train_dspark` phase on 1 node / 8 GPUs using `recipe_args.cached_target_path`. The log confirms the target model was not loaded; all 8 training steps completed with finite losses. PyTorch peak memory was 5.45 GiB/GPU (versus a 72.88 GiB observed peak during full-target precompute).
- Slurm job completed successfully in 7m34s with exit code 0.
- CPU/regression checks also passed: the broader speculative suite reported 664 passed / 11 skipped, the focused PR set (including the two-rank Gloo equivalence test) reported 98 passed, and Ruff/import-linter passed.

Result: **PASS**. I found no validation blocker.

examples/speculative/dspark/deepseek_v4_flash_dspark_precompute.yaml
```diff
@@ -0,0 +1,81 @@
+# Distributed DSpark offline precompute for a DeepSeek-V4-Flash target.
```
HuiyingLi, line 1: **[P1] Make the command-only configs pass example-YAML validation**

`validate-nightly-recipes` currently fails for this file and `glm_5.2_dspark_precompute.yaml` because both lack a top-level `recipe` target. Since these configs are launched through `python -m nemo_automodel.recipes.llm.precompute_dspark_dist`, please either exempt these exact command-only configs in `tools/lint_example_yamls.py` or introduce a genuine precompute recipe target. Adding `TrainDSparkRecipe` merely to satisfy the linter would misrepresent the entry point.

nemo_automodel/components/datasets/llm/dspark_cache.py
```diff
@@ -62,6 +62,60 @@
...
+    hidden_size = int(target_text_config.hidden_size)
+    return {
+        "target_model": target_model,
+        "target_model_type": target_model_type,
+        "target_vocab_size": int(target_text_config.vocab_size),
+        "hidden_size": hidden_size,
+        "num_hidden_layers": int(target_text_config.num_hidden_layers),
+        "seq_length": seq_length,
+        "dtype": dtype,
+        "num_samples": num_samples,
```
HuiyingLi, line 123: **[P2] Protect reruns from silently mixing cache inputs**

This manifest has no identity for the dataset/split, shuffle seed, effective chat template, masking settings, or resolved target revision. A same-length different input therefore passes `_ensure_output_dir_compatible`; because the distributed rerun replaces shards individually, an interrupted rerun leaves a contiguous, readable mixture of old and new supervision. I reproduced that state with sample IDs `[100, 101, 2, 3]`, and `CachedDSparkDataset` accepted it. Please include a stable input/target signature and/or publish each run through a staged generation plus completion marker so a partial overwrite cannot be treated as a valid cache.

HuiyingLi (commented):
I found two actionable issues. The distributed forward/cache equivalence and live multi-node GPU path otherwise look sound.

examples/speculative/dspark/deepseek_v4_flash_dspark_precompute.yaml
khazic: Fixed in 92a07ba10: both command-only configs are now exempted in tools/lint_example_yamls.py (EXCLUDED_RECIPE_FILES, with a comment noting they are launched via python -m precompute_dspark_dist), and a linter unit test pins the exemption. Went with the exemption rather than a recipe target to keep the entry point honest, as you suggested.

nemo_automodel/components/datasets/llm/dspark_cache.py
khazic: Fixed in 92a07ba10, both halves: the manifest now records the run input identity (train_data_path, train_split, shuffle_seed, mask_reasoning_content, sha256 of the effective chat template), so your [100, 101, 2, 3] scenario fails the compatibility check up front; and producers publish staged (manifest lands with complete=false before the first shard, flipped to true only after the last), so consumers reject an interrupted rerun while a same-config rerun into an incomplete dir is still allowed. Legacy caches without the field stay readable. Unit tests cover each rejection path. One deliberate scope cut: the complete marker lives in the DSpark layer for now; pushing it into the shared offline_cache layer (EAGLE-3 has the same window, partially covered by its shard-count check) would be a follow-up.

khazic:
Fixed the L0_Unit_Tests_CPU/GPU failures in 103a28ea0. Root cause: some other test earlier in the same pytest session leaks an initialized `torch.distributed` default process group (never destroys it), so `test_make_sync_max_steps_identity_without_process_group`'s "no process group" precondition is already false by the time it runs, and the next test's own `init_process_group()` then hits "trying to initialize the default process group twice". Both tests pass in isolation; only fail as part of the full suite, which is why local `pytest tests/unit_tests/recipes/llm/test_precompute_dspark_dist.py` never caught it. Guarded both with `pytest.skip(...)` when a group is already live, the same idiom already used elsewhere in this test tree (`test_train_eagle1_grad_accum.py`, `test_eagle3_token_cache.py`) rather than force-destroying a group this test does not own.

Separately, `gb200_L2_MoE` also failed on this run, but that is an unrelated GPU-topology flake ("No NVLink connection between GPU 0 and GPU 1") in `test_te_down_bias_parity.py`, a MoE test this PR does not touch.

khazic:
Correction to my last comment: the gb200_L2_MoE/NVLink note does not apply to this PR — I mixed it up with a different PR's CI run. Disregard that paragraph; the only real failure on this PR was the process-group leak fixed in 103a28ea0.

HuiyingLi (approved):

PR #2954, opened 2026-07-07
feat(kd): support separate student and teacher meshes

### LINKED ISSUE #2950: Support separate mesh for KD
**Is your feature request related to a problem? Please describe.**
For teacher and student models, we want to be able to support different parallelisms and, optionally, meshes hosted on disjointed ranks.

**Describe the solution you'd like**
See above

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

akoumpa:
# What does this PR do ?

Adds opt-in knowledge distillation with the student and teacher on disjoint device meshes, allowing each model to use an independent DP/TP/CP/PP layout.

# Changelog

- Add `separate_meshes`, `distributed`, and `teacher_distributed` configuration for KD recipes.
- Build rank-aware student and teacher meshes and validate that their explicit world sizes exactly cover the global world.
- Bridge nested batches and logits between disjoint rank sets, including unequal DP replica counts, TP full-vocabulary logits, CP gather/shard, and PP microbatch capture.
- Use model-specific process groups for checkpoint, signal, scheduler, dataloader serialization, and control collectives.
- Preserve the existing shared-mesh behavior by default.
- Add four-GPU Llama 3.2 example YAMLs for teacher TP2, CP2, and PP2 layouts under [`examples/llm_kd/llama3_2/`](https://github.com/NVIDIA-NeMo/Automodel/tree/akoumpa/separate_mesh_kd/examples/llm_kd/llama3_2).
- Add four-GPU Qwen3.5 VLM example YAMLs for teacher DP2, TP2, and CP2 layouts under [`examples/vlm_kd/qwen3_5/`](https://github.com/NVIDIA-NeMo/Automodel/tree/akoumpa/separate_mesh_kd/examples/vlm_kd/qwen3_5).
- Add unit tests, a 4-rank CPU functional test, a 4-GPU fixture, and [the complete TP2/CP2/PP2 LLM loss table](https://github.com/NVIDIA-NeMo/Automodel/blob/akoumpa/separate_mesh_kd/tests/functional_tests/llm_pretrain_and_kd/kd_separate_mesh_results.md).

# Review findings addressed

Ran an uncapped review against the latest repository review prompt and contributor skills after merging current `main`. Fixed 10 actionable findings:

1. Made synchronous/asynchronous checkpointing, addons, PEFT saves, consolidation, and barriers model-subgroup safe.
2. Made serialized dataset construction process-group aware instead of bypassing it with a boolean trap.
3. Corrected the Llama context-parallel capability declaration used by the new CP example.
4. Pinned the VLM teacher DP/TP/CP examples to the SDPA backend required by their native CP path.
5. Rejected unsupported separate-mesh expert parallelism early with a clear configuration error.
6. Consolidated duplicated LLM/VLM transport and TP vocabulary-shard helpers into `KDMeshBridge`.
7. Added distributed SIGTERM handling to teacher-only worker ranks.
8. Moved the private PyTorch CP unshard integration behind the distributed CP component boundary.
9. Documented tensor shapes, layouts, ownership, and microbatch contracts for new KD/CP helpers.
10. Added real four-process subgroup routing/checkpoint coverage and PP teacher-microbatch ordering coverage.

**Pre checks**:

Validation:

- `ruff` lint and format checks pass for all changed Python files.
- 146 focused KD/VLM/distributed/training/example tests pass.
- 204 checkpoint tests pass (`11 skipped`).
- The latest focused KD utility run passes (`11 passed`), including a real four-process Gloo test with disjoint student DP2 and teacher TP2 meshes plus subgroup-local async checkpoint initialization.
- The standalone 4-rank CPU bridge test passes with finite KD loss and gradients.
- cw-dfw 4-GPU LLM tests passed for teacher TP2 (10 steps), CP2 (1 step), and PP2 (1 step); every loss was finite and matched `0.5 * CE + 0.5 * KD` within `1e-5`.
- All six separate-mesh example YAMLs pass the example linter and load through the CLI config parser.
- `git diff --check` passes.

# Additional Information

LLM separate meshes support DP, TP, CP, and PP. VLM separate meshes support DP, TP, and CP; VLM PP remains unsupported, matching the existing VLM pipeline-parallel limitation.

### DIFF
--- a/examples/llm_kd/llama3_2/llama3_2_1b_kd_separate_mesh_teacher_cp2.yaml
+++ b/examples/llm_kd/llama3_2/llama3_2_1b_kd_separate_mesh_teacher_cp2.yaml
@@ -1,0 +1,102 @@
+# Knowledge distillation with disjoint 2-GPU student and teacher meshes.
+# The student uses data parallelism and the teacher uses context parallelism.
+#
+# To run on four GPUs:
+#   automodel examples/llm_kd/llama3_2/llama3_2_1b_kd_separate_mesh_teacher_cp2.yaml --nproc-per-node 4
+
+recipe: KnowledgeDistillationRecipeForNextTokenPrediction
+
+step_scheduler:
+  global_batch_size: 32
+  local_batch_size: 1
+  ckpt_every_steps: 200
+  val_every_steps: 100
+  num_epochs: 2
+
+dist_env:
+  backend: nccl
+  timeout_minutes: 10
+
+rng:
+  _target_: nemo_automodel.components.training.rng.StatefulRNG
+  seed: 1111
+  ranked: true
+
+# Student on global ranks [0, 1].
+model:
+  _target_: nemo_automodel.NeMoAutoModelForCausalLM.from_pretrained
+  pretrained_model_name_or_path: meta-llama/Llama-3.2-1B
+  torch_dtype: bf16
+  attn_implementation: sdpa
+
+# Teacher on global ranks [2, 3].
+teacher_model:
+  _target_: nemo_automodel.NeMoAutoModelForCausalLM.from_pretrained
+  pretrained_model_name_or_path: meta-llama/Llama-3.2-3B
+  torch_dtype: bf16
+  attn_implementation: sdpa
+
+separate_meshes: true
+distributed:
+  strategy: fsdp2
+  dp_size: 2
+  tp_size: 1
+  cp_size: 1
+  pp_size: 1
+  sequence_parallel: false
+
+teacher_distributed:
+  strategy: fsdp2
+  dp_size: 1
+  tp_size: 1
+  cp_size: 2
+  pp_size: 1
+  sequence_parallel: false
+
+checkpoint:
+  enabled: true
+  checkpoint_dir: checkpoints/llama3_2_1b_kd_separate_mesh_teacher_cp2/
+  model_save_format: safetensors
+  save_consolidated: false
+
+loss_fn:
+  _target_: nemo_automodel.components.loss.masked_ce.MaskedCrossEntropy
+
+kd_ratio: 0.5
+kd_loss_fn:
+  _target_: nemo_automodel.components.loss.kd_loss.KDLoss
+  ignore_index: -100
+  temperature: 1.0
+  fp32_upcast: true
+
+optimizer:
+  _target_: torch.optim.Adam
+  betas: [0.9, 0.999]
+  eps: 1e-8
+  lr: 1.0e-5
+  weight_decay: 0
+
+dataset:
+  _target_: nemo_automodel.components.datasets.llm.squad.make_squad_dataset
+  dataset_name: rajpurkar/squad
+  split: train
+  seq_length: 2048
+  padding: max_length
+  truncation: true
+
+dataloader:
+  _target_: torchdata.stateful_dataloader.StatefulDataLoader
+  collate_fn: nemo_automodel.components.datasets.utils.default_collater
+  shuffle: false
+
+validation_dataset:
+  _target_: nemo_automodel.components.datasets.llm.squad.make_squad_dataset
+  dataset_name: rajpurkar/squad
+  split: validation
+  seq_length: 2048
+  padding: max_length
+  truncation: true
+
+validation_dataloader:
+  _target_: torchdata.stateful_dataloader.StatefulDataLoader
+  collate_fn: nemo_automodel.components.datasets.utils.default_collater
--- a/examples/llm_kd/llama3_2/llama3_2_1b_kd_separate_mesh_teacher_cp2.yaml
+++ b/examples/llm_kd/llama3_2/llama3_2_1b_kd_separate_mesh_teacher_cp2.yaml
@@ -26,14 +26,14 @@ rng:
 model:
   _target_: nemo_automodel.NeMoAutoModelForCausalLM.from_pretrained
   pretrained_model_name_or_path: meta-llama/Llama-3.2-1B
-  torch_dtype: bf16
+  torch_dtype: bfloat16
   attn_implementation: sdpa
 
 # Teacher on global ranks [2, 3].
 teacher_model:
   _target_: nemo_automodel.NeMoAutoModelForCausalLM.from_pretrained
   pretrained_model_name_or_path: meta-llama/Llama-3.2-3B
-  torch_dtype: bf16
+  torch_dtype: bfloat16
   attn_implementation: sdpa
 
 separate_meshes: true
@@ -70,11 +70,15 @@ kd_loss_fn:
   fp32_upcast: true
 
 optimizer:
-  _target_: torch.optim.Adam
+  _target_: transformer_engine.pytorch.optimizers.fused_adam.FusedAdam
   betas: [0.9, 0.999]
   eps: 1e-8
   lr: 1.0e-5
   weight_decay: 0
+  adam_w_mode: true
+  bias_correction: true
+  master_weights: true
+  master_weight_dtype: torch.float32
 
 dataset:
   _target_: nemo_automodel.components.datasets.llm.squad.make_squad_dataset
--- a/examples/llm_kd/llama3_2/llama3_2_1b_kd_separate_mesh_teacher_pp2.yaml
+++ b/examples/llm_kd/llama3_2/llama3_2_1b_kd_separate_mesh_teacher_pp2.yaml
@@ -1,0 +1,100 @@
+# Knowledge distillation with disjoint 2-GPU student and teacher meshes.
+# The student uses data parallelism and the teacher uses pipeline parallelism.
+#
+# To run on four GPUs:
+#   automodel examples/llm_kd/llama3_2/llama3_2_1b_kd_separate_mesh_teacher_pp2.yaml --nproc-per-node 4
+
+recipe: KnowledgeDistillationRecipeForNextTokenPrediction
+
+step_scheduler:
+  global_batch_size: 32
+  local_batch_size: 1
+  ckpt_every_steps: 200
+  val_every_steps: 100
+  num_epochs: 2
+
+dist_env:
+  backend: nccl
+  timeout_minutes: 10
+
+rng:
+  _target_: nemo_automodel.components.training.rng.StatefulRNG
+  seed: 1111
+  ranked: true
+
+# Student on global ranks [0, 1].
+model:
+  _target_: nemo_automodel.NeMoAutoModelForCausalLM.from_pretrained
+  pretrained_model_name_or_path: meta-llama/Llama-3.2-1B
+  torch_dtype: bf16
+  attn_implementation: sdpa
+
+# Teacher on global ranks [2, 3].
+teacher_model:
+  _target_: nemo_automodel.NeMoAutoModelForCausalLM.from_pretrained
+  pretrained_model_name_or_path: meta-llama/Llama-3.2-3B
+  torch_dtype: bf16
+  attn_implementation: sdpa
+
+separate_meshes: true
+distributed:
+  strategy: fsdp2
+  dp_size: 2
+  tp_size: 1
+  cp_size: 1
+  pp_size: 1
+  sequence_parallel: false
+
+teacher_distributed:
+  strategy: fsdp2
+  dp_size: 1
+  tp_size: 1
+  cp_size: 1
+  pp_size: 2
+  sequence_parallel: false
+  pipeline:
+    pp_schedule: 1f1b
+    pp_microbatch_size: 1
+    scale_grads_in_schedule: false
+
+checkpoint:
+  enabled: true
+  checkpoint_dir: checkpoints/llama3_2_1b_kd_separate_mesh_teacher_pp2/
+  model_save_format: safetensors
+  save_consolidated: false
+
+loss_fn:
+  _target_: nemo_automodel.components.loss.masked_ce.MaskedCrossEntropy
+
+kd_ratio: 0.5
+kd_loss_fn:
+  _target_: nemo_automodel.components.loss.kd_loss.KDLoss
+  ignore_index: -100
+  temperature: 1.0
+  fp32_upcast: true
+
+optimizer:
+  _target_: torch.optim.Adam
+  betas: [0.9, 0.999]
+  eps: 1e-8
+  lr: 1.0e-5
+  weight_decay: 0
+
+dataset:
+  _target_: nemo_automodel.components.datasets.llm.squad.make_squad_dataset
+  dataset_name: rajpurkar/squad
+  split: train
+
+dataloader:
+  _target_: torchdata.stateful_dataloader.StatefulDataLoader
+  collate_fn: nemo_automodel.components.datasets.utils.default_collater
+  shuffle: false
+
+validation_dataset:
+  _target_: nemo_automodel.components.datasets.llm.squad.make_squad_dataset
+  dataset_name: rajpurkar/squad
+  split: validation
+
+validation_dataloader:
+  _target_: torchdata.stateful_dataloader.StatefulDataLoader
+  collate_fn: nemo_automodel.components.datasets.utils.default_collater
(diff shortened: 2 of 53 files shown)


akoumpa:
Uncapped review completed against the latest in-repo review guidelines after merging current `main`.

- Reviewed the complete PR diff without the workflow's seven-finding cap.
- Found and fixed 10 actionable issues in `d9f20702`.
- Revalidated formatting, focused KD/VLM/distributed coverage, the checkpoint suite, example configs, and real four-process CPU routing/checkpoint behavior.

akoumpa:
Follow-up uncapped review and bf16 parity validation completed on the latest `main`-merged branch.

- Fixed 6 additional actionable findings beyond the earlier 10-finding uncapped review: subset EP mesh construction, model-local download synchronization, explicit-group barrier reuse, Gemma4 compatibility in the requested image, OOM-free HF key enumeration, and reproducible parity fixtures/reporting.
- Ran with `nemo_auto_nightly_15_may_2026.sqsh`, the Automodel worktree mounted at `/workspace`, and bf16 student/teacher weights (fp32 only for configured loss reductions).
- Separate layouts keep the student on exactly one node and place the teacher on 3 nodes for TP2, 3 nodes for PP3, and 4 nodes for EP8/ES4.
- All shared-vs-separate comparisons pass at `rtol=1%`, `atol=0.01`; maximum relative differences are 0.1261% total loss and 0.2530% KD loss. EP8 KD is identical at logged float precision.
- The KD metric's absolute scale is expected because the current `KDLoss` reports teacher-to-student cross-entropy, including teacher entropy, rather than a zero-based KL value.
- Local validation: 214 passed, 1 deselected; Ruff lint/format and config loading pass; the final comparator replay passes against the completed cw-dfw traces.

Evidence:
- [Detailed cw-dfw results](https://github.com/NVIDIA-NeMo/Automodel/blob/akoumpa/separate_mesh_kd/tests/functional_tests/llm_pretrain_and_kd/kd_sep_mesh_cw_dfw_results.md)
- [W&B project](https://wandb.ai/Nemo-automodel/kd_sep_mesh)
- [W&B parity summary](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/xgjsx2y0)

akoumpa:
Final zero-based KL and numerics review completed on `a154eb19` after merging current `main` and applying the latest trusted review policy without the seven-finding cap.

Root cause and fixes:
- `KDLoss` documented forward KL but logged soft-target cross-entropy, including teacher entropy. Normal, chunked, and TP paths now compute `KL(P_teacher || P_student)`; student gradients are unchanged and covered at temperatures 1.0/2.5.
- The Gemma fixture used the smaller model's chat template and sliding windows could omit the teacher generation prefix. The 31B/26B teacher then expected a missing channel token, inflating true KL to ~13.44. The fixture now uses the teacher generation prompt, masks prompt labels, and repeats the complete transcript.
- Corrected MoE custom-vs-HF true KL is `1.83363` vs `1.82491` (0.48%), clearing the custom backend as the cause.
- The uncapped follow-up review also fixed model-local remote-code cache synchronization, stale soft-CE evidence, an unreproducible MoE baseline, parity checks that omitted grad norm/KD settings, tensor contracts required by the latest policy, and missing real TP/fallback tests.

Final BF16 cw-dfw matrix (`nemo_auto_nightly_15_may_2026.sqsh`, mounted code, repo-pinned Transformers 5.12.1 overlay):

| Layout | Student | Teacher | CE | Zero-based KL | Grad norm | W&B |
|---|---|---|---:|---:|---:|---|
| Shared dense | DP8, 1 node | DP8, 1 node | 10.4323130 | 2.4816835 | 824 | [eekmgoh9](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/eekmgoh9) |
| Separate TP2 | DP8, 1 node | DP12 x TP2, 3 nodes | 10.4323130 | 2.4399059 | 824 | [q09df5b1](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/q09df5b1) |
| Separate PP3 | DP8, 1 node | DP8 x PP3, 3 nodes | 10.4323130 | 2.4647844 | 820 | [djpndi6m](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/djpndi6m) |
| Shared MoE | DP8, 1 node | HF DP8, 1 node | 10.4813356 | 1.8249131 | 812 | [4lr58vi9](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/4lr58vi9) |
| Separate EP8 | DP8, 1 node | DP32 x EP8 x ES4, 4 nodes | 10.4813356 | 1.8336279 | 820 | [xu5gzzf3](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/xu5gzzf3) |

The strengthened comparator passes at `rtol=1%`, `atol=0.05`: CE is exact, max total-loss diff is 0.3235%, and max grad-norm diff is 0.9852%. TP2's KL diff is 0.04178 absolute / 1.6834% relative; a clean restart reproduced `2.4399059` exactly, so this is deterministic BF16 reduction order, not drift. [Parity summary](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/rcaa6vbb) · [full report](https://github.com/NVIDIA-NeMo/Automodel/blob/akoumpa/separate_mesh_kd/tests/functional_tests/llm_pretrain_and_kd/kd_sep_mesh_cw_dfw_results.md)

Validation: Ruff lint/format clean; 59 focused tests passed, including real 2/4/10-process paths; the final comparator replay passed against all completed traces. No cw-dfw jobs remain.

akoumpa:
SFT-completion KD numerics refresh after replacing the short raw continuation fixture with YAML-owned prompt/completion data.

Pushed commit: `26fc2abd` (`test(kd): use sft completion data`)

Setup:
- BF16, `kd_ratio=0.5`, `temperature=1.0`, LR=0, one training step
- `nemo_auto_nightly_15_may_2026.sqsh`, Automodel branch mounted as code
- W&B project: https://wandb.ai/Nemo-automodel/kd_sep_mesh
- The dataset now masks the chat/generation prompt and supervises the assistant completion only, so CE is in the expected range instead of the earlier ~10.x synthetic-continuation values.

| Layout | Teacher path/topology | Slurm | CE | zero-based KL | Total | Grad norm | W&B |
| --- | --- | ---: | ---: | ---: | ---: | ---: | --- |
| Shared dense DP8 | 31B HF teacher, 1 node | 13694730 | 3.1751444 | 2.0904799 | 2.6328123 | 612 | [qjx5ktp0](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/qjx5ktp0) |
| Separate dense TP2 | student DP8, teacher DP12 x TP2, 3 nodes | 13695345 | 3.1751444 | 2.1105356 | 2.6428401 | 620 | [z1r45bfh](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/z1r45bfh) |
| Separate dense PP3 | student DP8, teacher DP8 x PP3, 3 nodes | 13695347 | 3.1751444 | 2.0890758 | 2.6321104 | 620 | [j75ljl6m](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/j75ljl6m) |
| Shared MoE DP8 | 26B A4B HF teacher, 1 node | 13694975 | 3.1778483 | 2.3157218 | 2.7467852 | 600 | [6rp1w3a1](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/6rp1w3a1) |
| Separate MoE EP8 | student DP8, custom teacher DP32 x EP8, 4 nodes | 13695346 | 3.1778483 | 2.3832476 | 2.7805481 | 612 | [oov0fshg](https://wandb.ai/Nemo-automodel/kd_sep_mesh/runs/oov0fshg) |

Dense parity against shared dense:
- TP2: CE exact; KL +0.0200558 abs / +0.96%; total +0.0100279 abs / +0.38%; grad +8 / +1.31%.
- PP3: CE exact; KL -0.0014040 abs / -0.07%; total -0.0007019 abs / -0.03%; grad +8 / +1.31%.

MoE status:
- HF shared vs custom EP8 has exact CE, but KL is +0.0675259 abs / +2.92% and total is +0.0337629 abs / +1.23%.
- A shared custom MoE DP8 baseline is not valid right now: it hits expert-weight shape mismatch (`[128, 2816, 704]` checkpoint tensor vs `[16, 2816, 704]` runtime tensor), so expert weights are being sharded incorrectly without EP.
- A shared custom MoE DP1 baseline was also tried to isolate HF-vs-custom backend effects, but failed with optimizer-state OOM while allocating Adam state after loading the teacher (`13695960`).

Validation on the pushed code:
- `uv run --no-sync ruff format --check tests/functional_tests/llm_pretrain_and_kd/kd_separate_mesh_test_utils.py tests/unit_tests/recipes/test_kd_utils.py`
- `uv run --no-sync ruff check tests/functional_tests/llm_pretrain_and_kd/kd_separate_mesh_test_utils.py tests/unit_tests/recipes/test_kd_utils.py`
- `uv run --no-sync pytest -q tests/unit_tests/recipes/test_kd_utils.py` -> 15 passed
- `git diff --check`

yuhezhang-ai:
Reviewed the full PR — core `KDMeshBridge` myself plus a parallel deep pass over the checkpoint, distributed/CP-mesh, recipe/loss, and model/example subsystems. This is a large, genuinely hard distributed feature and it's **well-engineered and thoroughly validated**: the two highest-risk pieces (rank-offset mesh construction and CP unshard reversal) are correct against torch internals, the checkpoint subgroup fix (`_new_gloo_process_group(use_local_synchronization=True)`) is the right one, and the wave-routing is deadlock-free on the happy path (orphan teacher replicas on the partial last wave recompute-and-drop with no output route). Approving in principle; the findings below are mostly latent / edge-case rather than happy-path blockers.

**Ranked findings:**

**1. [Confirmed correctness — latent] Vocab-shard split mismatch in `match_student_vocab_shard` (`kd_utils.py:382`).** It slices the teacher's full logits with `torch.tensor_split(teacher, tp_size, dim=-1)`, but the student's TP-sharded logits follow DTensor `Shard` = `torch.chunk` semantics. These disagree when `vocab_size % tp_size != 0`. Verified empirically (repo torch): for `V=10, tp=4`, `tensor_split` gives rank2=`{6,7}`/rank3=`{8,9}` while DTensor `Shard` gives rank2=`{6,7,8}`/rank3=`{9}` — so KD compares misaligned vocab columns and produces wrong gradients **silently**. The Llama (128256) / Qwen examples have even splits so tests pass, but the helper is general. Suggest deriving the slice from the DTensor's own local shape/offset instead of re-splitting.

**2. [Medium — hang risk] Per-batch control-group collective in validation assumes equal batch counts.** `_run_validation_epoch` calls `_get_separate_teacher_logits` → `broadcast_command` + wave loop **per validation batch** over `control_group`. The base recipe tolerates unequal per-DP-replica shard lengths (one allreduce at epoch end); this path does not. If the val sampler doesn't pad to equal length across student DP replicas (e.g. 100 samples / 3 replicas), replicas issue different broadcast counts → hang until NCCL timeout. Training is safe (StepScheduler equalizes steps); validation isn't. Worth confirming the val sampler pads per replica, or gating the teacher call on a synchronized batch count.

**3. [Medium — robustness] No clean teardown on the error path.** `_run_teacher_worker` installs `DistributedSignalHandler(group=teacher_group)` but never calls `signals_received()`, so the teacher stops only on the student's `STOP_TEACHER`, broadcast in the student's `finally`. If a student rank raises mid-step (e.g. OOM in backward) while the teacher is blocked in a route broadcast, that `STOP` is out of position → global hang until timeout instead of clean shutdown. Consider polling the teacher's SIGTERM handler between waves.

**4. [Low-Med — behavioral regression] `kd_loss` drops the `-inf` student-logit mask.** Old code masked `torch.isinf(s_logits)` before summing; the new `_forward_kl_from_log_probs` masks only `teacher_prob > 0`. For a position with finite teacher prob but `-inf` student logit (models that hard-mask blocked/padding vocab), `log_ratio = teacher_log_prob - (-inf) = +inf` → `prob·(+inf) = NaN`, where the old path yielded 0. Finite logits (Llama/Qwen) unaffected. Also note the *reported* `kd_loss` shifts from cross-entropy `H(P,Q)` to zero-based `KL` — gradients identical, but logged magnitudes are no longer comparable to pre-PR runs; worth a changelog line.

**5. [Low-Med] Diffusers index rename uses global rank 0, not group rank (`checkpointing.py:546`).** `_maybe_rename_index_for_diffusers` is guarded by `get_rank() == 0` (global) while every other save-path guard was converted to `get_rank(group=self.process_group)`. If the saving model's subgroup excludes global rank 0, the diffusers index is never renamed → silently broken consolidated checkpoint (no hang). Should be group-relative. The `is_rank_0()` offline-consolidation-script guard shares the same assumption (cosmetic).

**Minor / scope:**
- `pp_kd_loss_fn` silently depends on `loss_fn` `reduction="sum"` (CE `num_label_tokens=None`, KD `num_batch_labels=1`); `reduction="mean"` would corrupt the blend with no guard.
- `utils.py:71`: subgroup `barrier(group=...)` drops the `monitored_barrier` timeout the world path has — a stuck subset rank hangs silently.
- `gemma4_moe/model.py` (+78/−21) mask refactor is correct and well-tested but out of scope for this feature (transformers-compat hardening, not exercised by KD) — arguably its own PR.
- VLM examples name `Qwen/Qwen3.5-4B`/`-9B` without a `-VL` suffix despite the `ImageTextToText` recipe + `freeze_vision_tower`/`freeze_audio_tower` — worth confirming these resolve to the multimodal checkpoints.

**Test coverage** is strong for happy paths and subgroup routing; the gaps map to the findings above (no test for `vocab % tp != 0`, unequal-length validation, mid-step failure teardown, or `-inf` student logits).

Overall: solid, well-validated work. None of these block the demonstrated configs, but #1–#3 fail silently or hang rather than erroring clearly, so they're worth addressing before use beyond the even-vocab / padded-validation happy path.

yuhezhang-ai:
@akoumpa  my agent post some comments above.

But my main concern is about the dtype precision:
the LLM examples pin `torch_dtype: bf16` on both models; the VLM dp2 example pins neither. Since the optimizer is plain Adam/AdamW (no separate fp32 master), a bf16 student trains with bf16 master weights, and small updates swamp against bf16's 7-bit mantissa, which is exactly this regime (lr=1e-5, smooth KD gradients). 

I would suggest making all six configs explicit and consistent: teacher bf16 (frozen, forward-only, logits already fp32-upcast) and student fp32 (opt into fp32 masters), unless you're deliberately accepting lower-res updates.

akoumpa:
@yuhezhang-ai addressed the optimizer precision concern in current head `407062f0997b5f9f82c6a7ce860c8953cec4167b`.

What changed:
- Switched all PR KD YAMLs that train the student from `torch.optim.Adam`/`AdamW` to `transformer_engine.pytorch.optimizers.fused_adam.FusedAdam`.
- Set `master_weights: true` and `master_weight_dtype: torch.float32` in each of those optimizer blocks, so the bf16 student storage path has explicit fp32 optimizer master weights.
- Made the VLM KD configs explicit about model dtype (`model.torch_dtype: bfloat16`, `teacher_model.torch_dtype: bfloat16`) instead of relying on defaults.
- Normalized the LLM KD examples from the `bf16` alias to `bfloat16` for consistency.
- Updated the functional KD YAMLs used by the PR parity/smoke runs the same way, including the tiny separate-mesh smoke config.
- Added a unit guard that scans the nine KD YAMLs touched by the PR and asserts `FusedAdam + master_weights + master_weight_dtype=torch.float32` plus explicit bf16 student/teacher storage.

Validation:
- `uv run --no-sync ruff format --check tests/unit_tests/recipes/test_kd_utils.py`
- `uv run --no-sync ruff check tests/unit_tests/recipes/test_kd_utils.py`
- YAML contract scan over the nine KD configs
- `uv run --no-sync pytest -q tests/unit_tests/recipes/test_kd_utils.py` -> 16 passed
- `git diff --check`

yuhezhang-ai (dismissed):

yuhezhang-ai (approved):

yuhezhang-ai:
One non-blocking follow-up remains: KDMeshBridge.match_student_vocab_shard() uses torch.tensor_split, which disagrees with DTensor Shard partitioning when vocabulary size is not divisible by TP size. Current examples are unaffected, but future uneven-vocabulary student-TP configurations could fail or select the wrong slice.

PR #2957, opened 2026-07-07
feat(dspark): log acceptance and confidence metrics during training

kashif:
## What

DSpark training only logged the loss and its three terms, so during a run there was no signal for how well the draft was actually doing. This adds the acceptance diagnostics to the training logs.

## Changes

- Compute per-batch `accept_rate` (mean per-token acceptance probability), `tau` (expected accepted block length), and confidence-head calibration `abs_error`/`bias`/`cumprod_bias` in the DSpark loss.
- Surface them on `DSparkStepMetrics` and log them next to the existing losses (wandb, metric logger, and the step log line).
- The new metrics ride the existing DP all-reduce window in the recipe, so the logged values are correct across data-parallel ranks and grad-accum micro-batches.
- Diagnostics are detached and computed under `no_grad`, so they add no backward cost and do not change the trained weights.

## Testing

`uv run pytest tests/unit_tests/speculative/ -k dspark` (86 passed), plus new range checks for `accept_rate` in [0, 1] and `tau >= 1`.


kashif:
@HuiyingLi thanks, let me align the tests so they match the same convention. Sorry about that!

nemo_automodel/components/speculative/dspark/loss.py
```diff
@@ -258,6 +329,16 @@ def compute_dspark_loss(
             "ce_loss": local_ce_loss.detach(),
             "l1_loss": local_l1_loss.detach(),
             "confidence_loss": local_confidence_loss.detach(),
+            "accept_rate": loss_terms["accept_rate_num"] / (loss_terms["accept_rate_den"] + 1e-6),
```
khazic: This scalar is a mean of per-micro-batch local ratios, while `accept_rate@k` in the recipe is a ratio of DP-reduced sums, so the two disagree whenever micro-batches have different valid-token counts. The exact global value is already available for free as `sum(pos_num) / sum(pos_den)` after the reduce; deriving it there would also let you drop `accept_rate_num`/`accept_rate_den`.

nemo_automodel/components/speculative/dspark/loss.py
```diff
@@ -258,6 +329,16 @@ def compute_dspark_loss(
...
+            "tau": loss_terms["tau_num"] / (loss_terms["tau_den"] + 1e-6),
```
khazic: When a micro-batch has no valid blocks on some rank, this yields `tau = 0 / 1e-6 = 0`, which then drags the window mean below 1 even though tau is >= 1 by definition. Reducing `tau_num`/`tau_den` across the window (like the per-position sums) instead of averaging local ratios would remove that bias.

nemo_automodel/recipes/llm/train_dspark.py
```diff
@@ -1313,28 +1358,36 @@ def run_train_validation_loop(self):
...
-                                    step=self.runtime.global_step,
-                                )
+                                wandb_metrics = {
+                                    "train/loss": avg["loss"],
+                                    "train/ce_loss": avg["ce_loss"],
+                                    "train/tv_loss": avg["l1_loss"],
+                                    "train/confidence_loss": avg["confidence_loss"],
+                                    "train/accept_rate": avg["accept_rate"],
+                                    "train/tau": avg["tau"],
+                                    "train/confidence_abs_error": avg["confidence_abs_error"],
```
khazic: When the confidence head is disabled these log a constant 0, and the same happens to `accept_rate`/`tau` when `l1_loss_alpha=0` with no confidence head (the TV distance is never computed). A flat zero curve in wandb reads like a collapsed acceptance rate; consider gating these keys on the corresponding switches.

nemo_automodel/recipes/llm/train_dspark.py
```diff
@@ -1285,23 +1305,64 @@ def run_train_validation_loop(self):
...
+                                dtype=torch.float32,
+                            )
+                            reduced = self._dp_allreduce(
+                                torch.cat([scalars, running_accept_pos_num, running_accept_pos_den])
+                            )
+                            n_scalars = scalars.numel()
+                            w = reduced[:n_scalars].tolist()
+                            pos_num = reduced[n_scalars : n_scalars + self.block_size]
+                            pos_den = reduced[n_scalars + self.block_size :]
+                            count = max(1.0, w[10])
```
khazic, line 1338: This pack/slice arithmetic (hard-coded `w[10]`, `w[4]/w[5]`, `w[6..8]/w[9]`, and the two `block_size` slices) has no unit test, and it is exactly the spot that silently breaks the next time someone inserts a metric into `scalars`. A small test that feeds known window sums through the reduce/unpack and checks the resulting `avg` dict would lock it down.

HuiyingLi (approved):

Issue #2958, opened 2026-07-07
[Tracking] Speculative decoding: implemented features and remaining gaps


khazic:
# [Tracking] Speculative decoding: implemented features and remaining gaps

Snapshot of `main` as of 2026-07-07, updated 2026-07-22. One place to see what the speculative-decoding subsystem (`nemo_automodel/components/speculative/`, the `train_*` recipes, `examples/speculative/`) already ships, which PR added each piece, what is in flight, and what is not implemented yet.

_Update 2026-07-10: #2953, #2959, #2966, and #2976 merged since the original snapshot._
_Update 2026-07-11: fp8 draft training and LoRA/PEFT drafts (#2963, merged 2026-07-08) and draft-side context parallelism (#2918) are now landed, so they move to "Implemented". Sequence-packing coverage for the remaining drafts is now in flight (#3002, #3003, #3004, #3005, #3023). The "Not implemented" list below is corrected accordingly._
_Update 2026-07-14: sequence packing for DeepSeek MLA EAGLE-3, EAGLE-1/2, DFlash, DSpark Qwen3, Domino, and JetSpec (#3002, #3003, #3004, #3005, #3023), DSpark training-time acceptance metrics (#2957), and the variable-prefix DFlash plus EAGLE-3 LK objectives (#3032) are now merged._
_Update 2026-07-14: a design RFC for the disaggregated streaming target backend is now open at #3062 (see the "Not implemented yet" item below)._
_Update 2026-07-22: Gemma4 EAGLE-3 target support and its example configs (#3071, #3073, #3077, #3079), unified DFlash / Domino / JetSpec validation metrics (#3072), periodic real accept-length eval during training (#3037), the on-policy regeneration loop (#3042), and the fp8 draft convergence smoke (#3081) are merged, so they move to "Implemented" and their entries are removed from the gap list. The first VLM speculative training PRs (#3166, #3173) and the streaming data-plane series (#3084, #3085, #3094) are now in flight. The remaining gaps are re-ranked below._

### Algorithms

| Algorithm | Recipe | Key PRs |
|---|---|---|
| EAGLE-1 / EAGLE-2 | `train_eagle1.py` / `train_eagle2.py` | #2263, #2264, feature-noise augmentation #2470, TP #2829, sequence packing #3003 |
[...]

PR #2963, opened 2026-07-08
feat(speculative): add fp8 draft training and LoRA draft adaptation

khazic:
## What does this PR do?

Implements the "fp8 draft training and LoRA/PEFT drafts" item of #2958.

### FP8 draft training (all spec-decode recipes)

- All draft-training recipes (EAGLE-1/2, EAGLE-3 / P-EAGLE, DFlash / Domino / JetSpec, DSpark) now accept the same top-level `fp8:` YAML block as the SFT recipes, reusing `components/quantization/fp8.py` (`FP8Config`, `build_fp8_config`, `apply_fp8_to_model`).
- A shared `apply_draft_fp8` helper in `recipes/llm/_spec_train_utils.py` swaps the draft's `nn.Linear` layers to torchao `Float8Linear` in place, before the DDP / FSDP2 wrap, so draft forward and backward GEMMs run in fp8 on SM89+ GPUs. `emulate: true` allows numerics testing on pre-Hopper GPUs.
- DSpark's FSDP2 path supports `enable_fsdp_float8_all_gather` with a per-optimizer-step `precompute_float8_dynamic_scale_for_fsdp` hook, mirroring `train_ft.py`.
- The frozen target model is never converted (fp8-quantized target checkpoints are already handled by dequant-on-load in the DSpark V4 / GLM path).

### LoRA draft adaptation (EAGLE-3)

- The EAGLE-3 recipe accepts the SFT recipes' `peft:` block (`PeftConfig`), reusing `components/_peft/lora.py`. The base draft is frozen, only `lora_A`/`lora_B` train, and the checkpointer flips to the standard adapter-only PEFT path (`is_peft=True`, consolidated export disabled).
- New `recipe_args.draft_weights_path` warm-starts the draft from a consolidated safetensors export (single file, flat directory, or `model.safetensors.index.json` sharded layout). This is what makes adapter training meaningful; it also works without `peft:` for full-FT continued training.
- Unsupported combinations fail fast instead of silently misconfiguring: `peft:` on DFlash / Domino / JetSpec / DSpark / EAGLE-1/2 raises (their drafts carry trainable non-LoRA modules, e.g. Domino's `prefix_gru` and DSpark's Markov / confidence heads, which the LoRA freeze would silently lock), and `peft:` with `parallel_drafting` or `fp8:` raises in EAGLE-3.

### Docs and examples

- `examples/speculative/README.md`: new "FP8 draft training" and "LoRA draft adaptation" sections plus config-reference rows.
- New example configs: `eagle3/qwen3_eagle3_fp8.yaml`, `eagle3/qwen3_eagle3_lora.yaml`.

## Validation

- 21 new unit tests in `tests/unit_tests/recipes/llm/test_spec_draft_fp8_lora.py` plus two checkpointer tests in `test_eagle_build_checkpointer.py`, all CPU-only.
- Full `tests/unit_tests/recipes/llm/` and `tests/unit_tests/speculative/` suites pass locally.
- Not yet validated on fp8-capable hardware: my GPUs are A100 (SM80), which torchao float8 rejects without `emulate`. See the comment below asking for an H100 smoke run.


khazic:
@HuiyingLi Could you (or someone with H100 access) help validate the fp8 path end to end? My GPUs are A100 (SM80), which torchao float8 rejects without `emulate`, so I could only cover the fp8 wiring with unit tests plus `emulate: true` numerics.

A short smoke run with `examples/speculative/eagle3/qwen3_eagle3_fp8.yaml` (a few dozen steps on any Qwen3 target) would confirm: (1) the Float8Linear swap logs `Successfully converted model to FP8` and reports the converted module count, (2) loss tracks the bf16 baseline over the first steps, and (3) tokens/s improves vs the same config with `fp8.enabled: false`. If you can also flip `enable_fsdp_float8_all_gather: true` on a DSpark FSDP2 run, that covers the all-gather + per-step scale precompute path too. Happy to adjust anything that comes up.

HuiyingLi:
Validated the native FP8 EAGLE-3 path at head `63c319043c56582ea4c992e3fc468cc46ad1189a` on H100 80 GB GPUs, using `Qwen/Qwen3-8B`, 4K packed sequences, 7 TTT steps, and two matched FP8/BF16 seed pairs.

The A/B controls were identical within each seed (model initialization, data order, optimizer, batch/step count); only `fp8.enabled` changed. The first 10 steps were excluded from throughput statistics.

- Native conversion passes: both FP8 runs logged `Successfully converted model to FP8`; 8/9 draft linears became `Float8Linear` (the configured `lm_head` exclusion remained BF16).
- Convergence passes:
  - seed 42: loss correlation `0.9999969`, relative loss MAE `0.0297%`, final loss BF16/FP8 `6.46545 / 6.46646`
  - seed 43: loss correlation `0.9999820`, relative loss MAE `0.0766%`, final loss BF16/FP8 `5.89483 / 5.89338`
- Throughput does **not** pass:
  - seed 42: BF16 `8,854 tok/s`, FP8 `6,679 tok/s` -> `0.754x`
  - seed 43: BF16 `8,806 tok/s`, FP8 `6,698 tok/s` -> `0.761x`
  - replicated geometric-mean FP8/BF16 throughput: **`0.757x`** (FP8 is about **24.3% slower**).

[W&B comparison summary with loss/throughput curves and all four runs](https://wandb.ai/Nemo-automodel/automodel-pr2963-fp8/runs/y1v7tjqi)

Result: the FP8 wiring and convergence criteria pass, but the requested speedup criterion fails under the PR's example settings.

khazic:
@HuiyingLi Thanks for the thorough A/B validation, and for catching the throughput regression. That result is consistent with how torchao float8 behaves in eager mode: the tensorwise path adds per-GEMM amax reduction, scale, and cast ops that are memory-bound, and they only pay off once inductor fuses them into the GEMM prologue. The EAGLE-3 draft was running fully eager, so the fp8 runs paid the cast overhead without ever getting the fused-GEMM benefit.

Pushed 2773a7ef9 to address this: all spec-decode recipes now accept the SFT recipes' top-level `compile:` block. The draft is compiled in place with `nn.Module.compile()` after the fp8 swap (in-place so the draft's object identity and checkpoint keys are unchanged), and `examples/speculative/eagle3/qwen3_eagle3_fp8.yaml` now ships with `compile.enabled: true`. The README documents that `fp8:` should be paired with `compile:`.

Could you rerun the same A/B at the new head with one change: both arms with `compile.enabled: true`, toggling only `fp8.enabled`? That isolates the fp8 contribution on top of the compiled baseline, which is the configuration the example now recommends. First-step compile time should be excluded along with your existing 10-step warmup exclusion. If fp8+compile still does not beat bf16+compile on these GEMM shapes (the draft is a single 4096-wide layer, which is near the float8 break-even point), I would keep the feature but document it as beneficial for larger drafts (for example DSpark drafts on V4/GLM-scale targets), and I am happy to run that DSpark measurement as a follow-up.

HuiyingLi:
Reran the requested compiled FP8/BF16 A/B at head `2773a7ef981fab56f7f1eaeb43bc73a6f2d241d0` on H100 80 GB GPUs, using `Qwen/Qwen3-8B`, 4K packed sequences, 7 TTT steps, and matched seeds 42/43. Both arms used `compile.enabled: true`; only `fp8.enabled` changed. Steps 1–10, including lazy compilation, were excluded from throughput statistics.

- Setup passes: all four runs installed the compiled draft call path. Both FP8 runs logged successful native conversion; 8/9 draft linears became `Float8Linear`, with the configured `lm_head` exclusion remaining BF16. Paired parameter fingerprints matched exactly.
- Convergence passes:
  - seed 42: correlation `0.9999969`, relative loss MAE `0.0300%`, final loss BF16/FP8 `6.46558 / 6.46640`
  - seed 43: correlation `0.9999879`, relative loss MAE `0.0663%`, final loss BF16/FP8 `5.89464 / 5.89308`
- Throughput passes:
  - seed 42: BF16 `11,852 tok/s`, FP8 `12,073 tok/s` -> `1.019x`
  - seed 43: BF16 `11,727 tok/s`, FP8 `12,142 tok/s` -> `1.035x`
  - replicated geometric-mean FP8/BF16 throughput: **`1.027x`** (FP8 is approximately **2.7% faster**)
- The focused compile/FP8 unit tests passed: `48 passed`.

[W&B comparison summary with loss/throughput curves and all four runs](https://wandb.ai/Nemo-automodel/automodel-pr2963-fp8/runs/7tzk3ltt)

Result: **PASS** for the requested convergence and speed criteria. The compiled path reverses the prior eager-mode regression (`0.757x`), although the resulting speedup is modest.

khazic:
Thanks for the fast rerun. That settles it: compiled fp8 clears both criteria (1.027x geomean over compiled bf16, convergence identical), and the eager regression is confirmed as a compile issue rather than an fp8 one. Pushed a17b06b9f recording the measured expectations in the README (eager fp8 0.76x, compiled fp8 1.03x, compile alone ~1.34x over eager bf16) plus the note that the fp8 gain scales with draft GEMM size, so the single 4096-wide EAGLE-3 layer is the floor and larger DSpark drafts should see more. From my side the PR is ready for review.

nemo_automodel/recipes/llm/train_eagle3.py
```diff
@@ -830,14 +893,18 @@ def _build_checkpointer(self, target_path: str) -> None:
...
         ckpt_kwargs = dict(
             enabled=True,
             checkpoint_dir=default_dir,
             model_save_format="safetensors",
             model_repo_id=str(target_path),
             model_cache_dir=hf_constants.HF_HUB_CACHE,
-            save_consolidated=True,
-            is_peft=False,
+            save_consolidated=not is_peft,
+            is_peft=is_peft,
```
HuiyingLi, line 991: [P1] Please pass the PEFT config through the custom save path. Setting `is_peft=True` installs `PeftAddon`, but `TrainEagle3Recipe.save_checkpoint()` later calls `checkpointer.save_model()` without `peft_config=self.peft_config`. The addon then dereferences `None` while building adapter metadata (`peft_config.dim`), so every checkpoint-enabled LoRA run fails at its final save. Please pass the config and add a test that performs an actual adapter save.

nemo_automodel/recipes/llm/train_eagle3.py
```diff
@@ -180,6 +184,54 @@ def _validate_peagle_gates(backend: str, cached_target_path, packed_sequence_siz
...
+    writes). Loaded with ``strict=False`` because the vocab-mapping buffers
+    (``d2t``/``t2d``) are regenerated by ``set_vocab_mapping`` each run and
+    need not be present.
+    """
+    state_dict = _load_hf_safetensors_checkpoint(path)
+    if not state_dict:
+        raise FileNotFoundError(f"draft_weights_path={path!r} contains no .safetensors files")
+    if state_dict.keys().isdisjoint(draft_model.state_dict().keys()):
+        raise ValueError(f"draft_weights_path={path!r} shares no keys with the draft model; wrong checkpoint?")
+    missing, unexpected = draft_model.load_state_dict(state_dict, strict=False)
```
HuiyingLi, line 223: [P1] This load silently desynchronizes the draft vocabulary during domain adaptation. Consolidated EAGLE checkpoints contain persistent `d2t`/`t2d` buffers, and this call runs after `set_vocab_mapping(selected_token_ids)`, so it overwrites the newly selected mapping with the base draft's old mapping. The trainer constructed below still receives the new `selected_token_ids`; for LoRA, the frozen `lm_head` cannot reconcile that mismatch. Please recover/validate the base checkpoint mapping and propagate it to the trainer/target mapping, or fail explicitly when the current mapping differs.

examples/speculative/eagle3/qwen3_eagle3_lora.yaml
```diff
@@ -0,0 +1,51 @@
...
+  freeze_embeddings: true
+  trust_remote_code: false
+  shuffle_seed: 42
+  log_every_steps: 10
+  max_grad_norm: 1.0
+  packed_sequence_size: 4096
+  # LoRA adapts an EXISTING draft; point this at the consolidated safetensors
+  # export of a previously trained draft (the `model.safetensors` directory a
+  # full-FT run writes). Without it the adapters would sit on random weights.
+  draft_weights_path: ./outputs/eagle3_qwen3_8b_perfectblend/checkpoints/final/consolidated
```
HuiyingLi: [P2] This path is not produced by the EAGLE checkpoint layout. Full-FT checkpoints are written under `checkpoints/epoch_*_step_*/model/consolidated`, with `checkpoints/LATEST` updated to the newest checkpoint; there is no `checkpoints/final/consolidated`. Please use `checkpoints/LATEST/model/consolidated` (or an explicit epoch directory) so the documented warm-start workflow resolves.

HuiyingLi (changes requested):
The FP8/compile path looks sound, but the new LoRA path has two correctness blockers in checkpointing and vocabulary mapping. I also found one broken path in the LoRA example; details are inline.

nemo_automodel/recipes/llm/train_eagle3.py
khazic: Fixed in a43a57e73: save_checkpoint now passes peft_config=self.peft_config through to save_model. Added the regression test (asserts the kwarg is forwarded, and None without LoRA) plus an end-to-end test that saves a LoRA-patched draft through the real Checkpointer and asserts adapter_model.safetensors contains only lora_ keys.

nemo_automodel/recipes/llm/train_eagle3.py
khazic: Fixed in a43a57e73, via the fail-explicitly option: _load_draft_weights now compares the checkpoint d2t/t2d buffers against the mapping the run just selected (including one-sided presence and shape) and raises with a pointer to selected_token_ids_path before touching any weights. Propagating the checkpoint mapping outward instead would silently retarget the dataloader, so I kept the mapping source of truth with the run config. Three tests cover match / mismatch / one-sided.

examples/speculative/eagle3/qwen3_eagle3_lora.yaml
khazic: Fixed in a43a57e73: the example now points at checkpoints/LATEST/model/consolidated and the comment documents the epoch_*_step_* layout plus the same-mapping requirement.

khazic:
Follow-up to the review fixes: I ran a deeper self-review over the whole branch and closed six more confirmed gaps in 929dc2f4e before asking for another look. Highlights: the final checkpoint of a LoRA run now exports the merged serve-ready draft to model/consolidated (no external merge needed); adapter_config.json records the warm-start draft as its base instead of the target; peft now rejects freeze_embeddings: false instead of silently overriding it; _load_draft_weights hard-errors on missing keys; CompileConfig.options previously crashed the new in-place compile path and silently disabled compilation in compile_model, now routed through torch.compile options correctly; the LoRA example sets selected_token_ids_path so the documented flow actually passes the new vocab-mapping guard. Also reverted ~30 unrelated lint-churn test files that had slipped into an earlier commit. Full details in the commit message; ready for re-review.

nemo_automodel/recipes/llm/train_eagle3.py
```diff
@@ -180,6 +184,123 @@ def _validate_peagle_gates(backend: str, cached_target_path, packed_sequence_siz
...
+    """
+    state = {k: v.detach().cpu() for k, v in draft_model.state_dict().items() if ".lora_" not in k and "lora_" != k[:5]}
+    for module_name, module in draft_model.named_modules():
+        if not (hasattr(module, "lora_A") and hasattr(module, "lora_B")):
+            continue
+        key = f"{module_name}.weight" if module_name else "weight"
+        delta = (module.lora_B.weight.detach().cpu().float() @ module.lora_A.weight.detach().cpu().float()) * float(
+            module.scale
+        )
+        state[key] = (state[key].float() + delta).to(draft_model.state_dict()[key].dtype)
```
HuiyingLi: Thanks for adding the final merged export—this makes the serving workflow much more convenient. One supported PEFT option still needs handling here: with `use_dora=True`, the trained effective weight is row-scaled by `lora_magnitude / ||W + scale * B @ A||`, whereas this code merges only `W + scale * B @ A` and drops `lora_magnitude`. The resulting consolidated checkpoint is therefore not numerically equivalent to the trained model (a small parity check showed a maximum output difference of about 1.25). Could we either apply the DoRA magnitude scaling during merge and add a forward-parity test, or explicitly reject DoRA for this recipe until that merge is supported?

examples/speculative/eagle3/qwen3_eagle3_lora.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # export of a previously trained draft. A full-FT run writes checkpoints
+  # under `checkpoints/epoch_*_step_*/model/consolidated`, with
+  # `checkpoints/LATEST` symlinked to the newest one.
+  draft_weights_path: ./outputs/eagle3_qwen3_8b_perfectblend/checkpoints/LATEST/model/consolidated
+  # REQUIRED with draft_weights_path + compressed vocab: reuse the BASE run's
+  # cached token mapping. Set the same path in the base run's config so its
+  # mapping is cached here; without it this run rescans the (different)
+  # adaptation dataset, derives a different mapping, and the warm start fails
+  # fast at load (the frozen lm_head rows are tied to the base mapping).
+  selected_token_ids_path: ./outputs/eagle3_qwen3_8b_perfectblend/draft_token_map.pt
```
HuiyingLi, line 38: The explicit shared cache path is a helpful way to document the mapping requirement. Could we also add this same `selected_token_ids_path` to the referenced `qwen3_eagle3_perfectblend.yaml` base config? At present that base run never creates `draft_token_map.pt`; running the two examples as provided leaves this file absent, so the adaptation run rebuilds a map from the domain dataset and then likely fails the mismatch guard. Declaring the path in both configs would make the documented base-to-adaptation workflow work as written.

HuiyingLi (commented):
Thanks for the comprehensive follow-up and self-review. The earlier findings are addressed; I found two remaining edge cases in the newly added workflow, noted inline.

nemo_automodel/recipes/llm/train_eagle3.py
khazic: Fixed in bc80f4fd8, via the merge-support option: the export now applies the magnitude row-rescaling W_eff = (lora_magnitude / ||W + scale * B @ A||_row) * (W + scale * B @ A), matching LinearLoRA DoRA at eval time, and a parametrized forward-parity test asserts the merged weight reproduces the live adapter forward for both plain LoRA and DoRA.

examples/speculative/eagle3/qwen3_eagle3_lora.yaml
khazic: Fixed in bc80f4fd8: the base perfectblend config now sets the same selected_token_ids_path, so the documented pair runs as written (base run caches draft_token_map.pt, the adaptation run consumes it).

HuiyingLi (approved):

PR #2966, opened 2026-07-08
feat(speculative): add multi-dataset acceptance-length benchmark sweep

khazic:
## What does this PR do?

Implements the "multi-dataset acceptance benchmark suite" item of #2958.

### Problem

`bench_sglang.py` / `bench_vllm.py` each measure acceptance length / speedup against ONE dataset per invocation. The same draft's acceptance rate varies sharply by task type (conversational data tends to accept far more tokens than math or code), so getting a draft's acceptance profile across task types meant manually invoking each script once per dataset and collating the numbers by hand.

### `bench_sweep.py`

Drives either engine (`--engine sglang|vllm`) through several datasets in one pass and reports a per-dataset table plus a completed-weighted aggregate.

- **Default dataset suite**: the four benchmarks the EAGLE / EAGLE-2 papers report acceptance/speedup on -- MT-Bench (multi-turn chat), HumanEval (code), GSM8K (math), Alpaca (single-turn instruction-following). None ship a chat-messages column, so each is read via the new `--prompt-column` path (see below).
- `--datasets-config <path.yaml>` swaps in a custom dataset list (see `examples/speculative/bench_sweep/spec_bench_datasets.yaml`, which mirrors the built-in default and doubles as a template); `--datasets <name...>` narrows to a subset.
- **Failure isolation**: one dataset failing to load or benchmark (bad HF id, a transient issue for that request, ...) is reported as an error row and does not abort the sweep. A bad *shared* flag (e.g. `--num-prompts 0`) is validated once up front instead, and raises immediately rather than being reported as N duplicate per-dataset failures.
- **SGLang caveat**: `avg_spec_accept_length` is a server-cumulative running average with no reset/delta API, so sweeping more than one dataset against one live SGLang server contaminates every dataset after the first with prior traffic. A warning is logged when this applies (`--engine vllm`'s counter-diffing approach has no such caveat). Documented in the module docstring and README.

### Supporting changes (to avoid duplicating logic)

- `bench_common._load_prompts` gains a `--prompt-column` path (a raw text field wrapped into a single-turn user message) alongside the existing `--messages-column` path, since none of the four default datasets are chat-messages-shaped. Also available directly on `bench_sglang.py`/`bench_vllm.py` for a single custom dataset, not just through the sweep. `bench_common._report_summary` factors out the print-JSON-and-optionally-write-file tail the two engine scripts shared.
- `bench_sglang.py` / `bench_vllm.py`: split `_run` into `_run_summary` (pure compute, returns the metrics dict or `None`) + `_run` (unchanged external behavior), so the sweep can drive one dataset at a time without the printing / `--output-json` side effects. Verified byte-for-byte behavior preserved against the existing test suites (all 57 pre-existing tests pass unchanged).
- `tools/lint_example_yamls.py`: exempt the new dataset-list example config (not a recipe, consumed via `--datasets-config`).

## Validation

- 611 unit tests pass in `tests/unit_tests/speculative/` (94 new: `test_bench_sweep.py` + `test_bench_common.py`'s new `--prompt-column` coverage), plus the example-YAML linter suite.
- No network / GPU required for any of this -- everything is tested via the same hermetic mocking pattern the existing `bench_sglang`/`bench_vllm` tests use (`_run_summary` monkeypatched per engine, no live server).
- Not yet validated against a real SGLang/vLLM server end to end. Happy to have someone with server access run a smoke sweep if useful, but the orchestration logic itself has no server-specific surface beyond what `bench_sglang`/`bench_vllm` already exercise (unchanged) in production.

### DIFF
--- a/nemo_automodel/components/speculative/bench_sweep.py
+++ b/nemo_automodel/components/speculative/bench_sweep.py
@@ -83,6 +83,11 @@ class DatasetSpec:
     Exactly one of ``messages_column`` (an existing OpenAI-messages list) or
     ``prompt_column`` (a raw text field, wrapped into a single-turn user
     message) must be set -- see ``bench_common._load_prompts``.
+
+    ``prompt_context_column`` is an optional second raw-text field appended to
+    ``prompt_column`` (separated by a blank line) when it is non-empty, for
+    datasets whose task context lives in a separate column (e.g. Alpaca's
+    ``input``). It is only valid alongside ``prompt_column``.
     """
 
     name: str
@@ -91,6 +96,7 @@ class DatasetSpec:
     dataset_name: str | None = None
     messages_column: str | None = None
     prompt_column: str | None = None
+    prompt_context_column: str | None = None
     max_new_tokens: int | None = None
 
     def __post_init__(self):
@@ -99,12 +105,19 @@ class DatasetSpec:
                 f"dataset {self.name!r}: exactly one of messages_column / prompt_column must be set "
                 f"(got messages_column={self.messages_column!r}, prompt_column={self.prompt_column!r})"
             )
+        if self.prompt_context_column and not self.prompt_column:
+            raise ValueError(
+                f"dataset {self.name!r}: prompt_context_column={self.prompt_context_column!r} requires "
+                "prompt_column to be set (it is appended to the prompt_column text)."
+            )
 
 
-# The classic EAGLE / EAGLE-2 paper suite: multi-turn chat (MT-Bench), code
-# (HumanEval), math (GSM8K), and single-turn instruction-following (Alpaca).
+# The classic EAGLE / EAGLE-2 paper suite: chat (MT-Bench, first turn), code
+# (HumanEval), math (GSM8K), and instruction-following (Alpaca).
 DEFAULT_DATASET_PRESETS: tuple[DatasetSpec, ...] = (
-    DatasetSpec(name="mt_bench", input_data="HuggingFaceH4/mt_bench_prompts", split="train", prompt_column="turns"),
+    # HuggingFaceH4/mt_bench_prompts stores the two-turn sequence under ``prompt``
+    # (a list); the first turn is used (see bench_common._extract_prompt_text).
+    DatasetSpec(name="mt_bench", input_data="HuggingFaceH4/mt_bench_prompts", split="train", prompt_column="prompt"),
     DatasetSpec(
         name="humaneval",
         input_data="openai/openai_humaneval",
@@ -113,7 +126,16 @@ DEFAULT_DATASET_PRESETS: tuple[DatasetSpec, ...] = (
         max_new_tokens=512,
     ),
     DatasetSpec(name="gsm8k", input_data="openai/gsm8k", dataset_name="main", split="test", prompt_column="question"),
-    DatasetSpec(name="alpaca", input_data="tatsu-lab/alpaca", split="train", prompt_column="instruction"),
+    # Alpaca's task context lives in a separate ``input`` field (non-empty for
+    # ~40% of rows); append it so those prompts are not truncated to the bare
+    # instruction. The reference ``output`` is never included.
+    DatasetSpec(
+        name="alpaca",
+        input_data="tatsu-lab/alpaca",
+        split="train",
+        prompt_column="instruction",
+        prompt_context_column="input",
+    ),
 )
 
 _ENGINE_MODULES = {"sglang": bench_sglang, "vllm": bench_vllm}
@@ -176,6 +198,7 @@ def _dataset_args(base_args: argparse.Namespace, spec: DatasetSpec) -> argparse.
     args.dataset_name = spec.dataset_name
     args.messages_column = spec.messages_column
     args.prompt_column = spec.prompt_column
+    args.prompt_context_column = spec.prompt_context_column
     if spec.max_new_tokens is not None:
         args.max_new_tokens = spec.max_new_tokens
     return args
@@ -215,6 +238,13 @@ async def _run_sweep(args: argparse.Namespace, specs: list[DatasetSpec]) -> list
             continue
         if summary is None:
             results.append({"dataset": spec.name, "error": "no usable prompts loaded"})
+            continue
+        # _run_workload swallows per-request HTTP errors, so an engine whose every
+        # request failed still returns a summary with completed=0 rather than raising.
+        # Treat that as an error row so it is excluded from num_datasets_ok and the
+        # sweep's exit status reflects that no benchmark actually ran.
+        if summary.get("completed", 0) < 1:
+            results.append({"dataset": spec.name, "error": "all requests failed (0 completed)"})
             continue
         results.append({"dataset": spec.name, **summary})
     return results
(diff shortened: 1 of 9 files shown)


nemo_automodel/components/speculative/bench_sweep.py
```diff
@@ -0,0 +1,345 @@
...
+            raise ValueError(
+                f"dataset {self.name!r}: exactly one of messages_column / prompt_column must be set "
+                f"(got messages_column={self.messages_column!r}, prompt_column={self.prompt_column!r})"
+            )
+
+
+# The classic EAGLE / EAGLE-2 paper suite: multi-turn chat (MT-Bench), code
+# (HumanEval), math (GSM8K), and single-turn instruction-following (Alpaca).
+DEFAULT_DATASET_PRESETS: tuple[DatasetSpec, ...] = (
+    DatasetSpec(name="mt_bench", input_data="HuggingFaceH4/mt_bench_prompts", split="train", prompt_column="turns"),
```
HuiyingLi: Thank you for including MT-Bench in the default suite—this is a very useful addition. I noticed one schema mismatch that prevents this entry from running as shipped: the current [Hugging Face dataset](https://huggingface.co/datasets/HuggingFaceH4/mt_bench_prompts) exposes the two-turn sequence under `prompt`, rather than `turns`. With this preset, `_load_prompts` returns zero prompts and the default sweep records MT-Bench as an error row. Could we update both this preset and the example YAML to use `prompt_column="prompt"`, and add a representative schema test? Since `_extract_prompt_text` currently keeps only the first item, it would also be helpful either to describe this as a first-turn MT-Bench workload or to preserve the full two-turn flow so the multi-turn description remains accurate.

nemo_automodel/components/speculative/bench_sweep.py
```diff
@@ -0,0 +1,345 @@
...
+    DatasetSpec(
+        name="humaneval",
+        input_data="openai/openai_humaneval",
+        split="test",
+        prompt_column="prompt",
+        max_new_tokens=512,
+    ),
+    DatasetSpec(name="gsm8k", input_data="openai/gsm8k", dataset_name="main", split="test", prompt_column="question"),
+    DatasetSpec(name="alpaca", input_data="tatsu-lab/alpaca", split="train", prompt_column="instruction"),
```
HuiyingLi: The Alpaca preset is a helpful inclusion. One concern is that selecting only `instruction` omits the `input` field, which carries required task context for many rows; the [dataset card](https://huggingface.co/datasets/tatsu-lab/alpaca) notes that roughly 40% of examples include it. In live validation, 23 of the first 64 rows had non-empty input—for example, “Identify the odd one out” loses “Twitter, Instagram, Telegram.” That leaves a substantial portion of the benchmark prompts incomplete and can make the acceptance statistics misleading. Could we add a small composition/template path that combines `instruction` and non-empty `input` without including the reference output, along with a regression test for both forms?

nemo_automodel/components/speculative/bench_sweep.py
```diff
@@ -0,0 +1,345 @@
...
+        try:
+            summary = await engine._run_summary(dataset_args)
+        except Exception as exc:  # noqa: BLE001 -- one bad dataset must not abort the sweep
+            logger.error("[%s] failed: %s", spec.name, exc)
+            results.append({"dataset": spec.name, "error": str(exc)})
+            continue
+        if summary is None:
+            results.append({"dataset": spec.name, "error": "no usable prompts loaded"})
+            continue
+        results.append({"dataset": spec.name, **summary})
```
HuiyingLi, line 249: The per-dataset isolation here is thoughtfully designed. There is one failure path where the sweep currently reports success: `_run_workload` intentionally catches request-level HTTP errors, so when every request fails, the engine returns a summary with `completed=0` instead of raising or returning `None`. This row is then counted in `num_datasets_ok`, and a sweep where every request failed exits with status `0`. Could we treat summaries with zero completed requests as error rows (or have the engine return `None` in that case) and add an all-requests-failed regression test? That would keep both the aggregate and process exit status aligned with the actual benchmark outcome.

nemo_automodel/components/speculative/bench_sweep.py
khazic: Good catch, fixed. The preset and example YAML now use `prompt_column: prompt` (the real column on HuggingFaceH4/mt_bench_prompts); `_extract_prompt_text` reduces the two-turn list to its first turn, so this is a first-turn MT-Bench workload now, and I've updated the comments/docstring to say so. Added a schema test that the presets carry the right columns.

nemo_automodel/components/speculative/bench_sweep.py
khazic: Agreed. Added `DatasetSpec.prompt_context_column`: `_load_prompts` appends a non-empty secondary field (Alpaca's `input`) to the instruction, blank-line separated, and never reads the reference `output`. The alpaca preset and example YAML now set it. Added regression tests for both forms (input present -> joined, input empty/absent -> bare instruction).

nemo_automodel/components/speculative/bench_sweep.py
khazic: Fixed. `_run_sweep` now records a summary with `completed < 1` as an error row ("all requests failed (0 completed)"), so it drops out of `num_datasets_ok` and the exit status reflects that nothing actually ran. Added an all-requests-failed regression test.

nemo_automodel/components/speculative/bench_sweep.py
```diff
@@ -0,0 +1,375 @@
...
+and collating the output by hand.
+
+Default dataset suite -- the four benchmarks the EAGLE / EAGLE-2 papers report
+acceptance / speedup numbers on: MT-Bench (multi-turn chat), HumanEval (code),
+GSM8K (math), and Alpaca (single-turn instruction-following). None of these
+ship a chat-messages column the way ``bench_sglang``/``bench_vllm``'s
+``--messages-column`` expects, so each reads a raw text field instead
+(``bench_common``'s ``--prompt-column`` path: the field is wrapped into a
+fresh single-turn user message; a list value, e.g. MT-Bench's multi-turn
+``turns`` column, uses its first entry).
```
HuiyingLi: Thank you for addressing all three findings so quickly. I rechecked the updated presets and failure handling, and those fixes look good. I noticed just one small documentation follow-up: this module docstring still describes MT-Bench as “multi-turn chat” and refers to a “turns” column, while the current dataset uses the “prompt” column and this sweep intentionally benchmarks only the first turn. The README repeats the “multi-turn chat” wording. Would you mind updating both places to say “MT-Bench (first turn)” and replacing “turns” with “prompt”? Everything else looks good from my re-review—thank you again!

khazic: Done, thanks. Updated both the module docstring and examples/speculative/README.md to say "MT-Bench (first turn)" and to describe the two-turn `prompt` column (dropping the stale "multi-turn chat" / "turns" wording). Pushed in d2908e0c4.

khazic: Everything else looks good from your re-review

HuiyingLi (approved):

PR #2974, opened 2026-07-09
fix: chunk variable-length all-gather backward

mohamed-alansary:
## Summary

This PR reduces peak memory in the expert-parallel variable-length all-gather backward path.

Previously, `_AllGatherConcatVarlenFn.backward` materialized a contiguous copy of the full concatenated gradient, all-reduced it, and then narrowed out the local rank slice. For long-context expert-parallel batches, that full `[sum(tokens_per_rank), hidden]` temporary can be large.

This change all-reduces one source-rank slice at a time, further split into bounded token chunks, and only copies the reduced chunks that belong to the local rank into `grad_local`.

## Details

- Adds a private `_ALL_GATHER_CONCAT_VARLEN_BACKWARD_CHUNK_TOKENS = 4096` implementation constant.
- Replaces the full-gradient `contiguous()` + `all_reduce()` path with chunked `all_reduce()` calls.
- Uses `.clone()` for each chunk so the incoming `grad_output` is not mutated by the in-place collective.
- Preserves the existing autograd contract and return shape.

The chunk size is kept as a private implementation constant instead of a public `MoEConfig` field because this is intended to be a transparent memory-safety improvement; 4096 token chunks bound the temporary all-reduce buffer while keeping collective launch overhead amortized.

## Tests

Added `test_all_gather_concat_varlen_backward_reduces_chunks`, which verifies:

- local ranks with non-empty and empty local slices
- the full gradient shape is never all-reduced
- chunk shapes are reduced as expected
- `grad_output` remains unchanged

Ran:

```bash
python -m pytest tests/unit_tests/moe/test_experts.py -k all_gather_concat_varlen_backward_reduces_chunks -v
```

Result:
```text
3 passed
```


hemildesai:
Hi @mohamed-alansary, thanks a lot for your contribution. Can you post a before/after convergence or finetuning run with memory usage statistics?

mohamed-alansary:
Thanks @hemildesai . I do not yet have a clean before/after convergence or finetuning run for this isolated PR, so I do not want to overclaim.

The motivation came from a long-context Gemma4 MoE LoRA-GRPO run where policy training reached `loss.backward()` and OOMed with a ~10.65 GiB allocation in the variable-length EP all-gather backward path. That matches the previous `grad_output.contiguous()` full concatenated-gradient clone in `_AllGatherConcatVarlenFn.backward`.

This PR keeps the same all-reduce + local-slice backward semantics, but reduces source-rank token chunks instead of reducing the full concatenated gradient at once. If you need it, I can also run a before/after finetuning smoke with memory stats around the policy backward step, including peak allocated/reserved memory and relevant training metrics.

PR #2979, opened 2026-07-09
perf: chunk MoE per-expert loop by routed tokens

mohamed-alansary:
## Summary

This PR reduces peak transient memory in the MoE per-expert loop path by chunking routed-token processing for each local expert.

Previously, when many tokens routed to the same local expert, `_forward_loop` materialized the full per-expert intermediate tensors at once:

```python
x_idx = x.gather(dim=0, index=idx_b)
gate_and_up_out = x_idx @ gate_and_up_proj
activated = self.expert_activation_grouped(gate_and_up_out, w)
expert_out = activated @ down_proj
```

For long-context MoE batches, a hot expert can receive many routed tokens, making these intermediates large enough to dominate peak memory.

This change processes routed tokens for each expert in bounded chunks, while preserving the same math and keeping the existing fp32 accumulator behavior unchanged.

## Details

- Adds a private `_MOE_LOOP_EXPERT_CHUNK_TOKENS = 4096` implementation constant.
- Chunks the per-expert loop along the routed-token dimension.
- Keeps per-expert parameter views outside the chunk loop:
  - `gate_and_up_proj`
  - `down_proj`
  - optional gate/up bias
  - optional down-projection bias
- Keeps heavy intermediates chunk-local:
  - `x_idx`
  - `gate_and_up_out`
  - `activated`
  - `expert_out`
- Accumulates each chunk into the existing fp32 `y` buffer with `scatter_add_`.
- Does not add a public config knob or environment-variable switch.

The implementation is semantically equivalent to the previous unchunked loop: each chunk computes the same routed-token contribution and scatters it into the same destination accumulator.

## Chunk Size

The chunk size is intentionally a private module constant rather than a user-facing config option.

`4096` routed tokens keeps the largest per-chunk intermediates to tens of MB for common MoE hidden/intermediate dimensions, while still amortizing kernel launch overhead. Smaller chunk sizes can add unnecessary launch overhead, while much larger chunks reduce the memory-bounding benefit.

Maintainers can tune or promote this later if production models surface a need for configurability.

## Numerical Behavior

The accumulator remains fp32, matching the previous implementation.

Chunking changes the order in which duplicate destination rows are accumulated via `scatter_add_`, so tiny fp32 associativity differences are possible. The added regression test compares chunked and effectively unchunked execution at tight tolerance and checks both forward output and gradients.

## Tests

Added `test_loop_path_expert_chunking_matches_unchunked`, which verifies:

- chunked output matches effectively unchunked output
- input gradients match
- gate/up projection gradients match
- down projection gradients match
- gate/up bias gradients match when expert bias is enabled
- down-projection bias gradients match when expert bias is enabled
- both `expert_bias=False` and `expert_bias=True`
- both `chunk_size=1` and `chunk_size=2`

The test intentionally routes most tokens to one expert so the chunking path is exercised across multiple chunk boundaries.

Ran:

```bash
python -m pytest tests/unit_tests/moe/test_experts.py -k expert_chunking_matches_unchunked -v
```

Result:

```text
4 passed, 95 deselected
```


akoumpa:
Hi @mohamed-alansary , thanks for making this, i was wondering if by any chance you had this tested and can comment on any perf gains / losses with this change.

Thank you.

mohamed-alansary:
Thanks @akoumpa . I have not run a clean end-to-end throughput benchmark for this PR yet, so I do not want to claim a throughput gain.

This change is primarily memory-motivated. It came from long-context MoE debugging where a heavily routed expert could make the loop path materialize a very large per-expert routed-token slice; the observed failure was at the `_forward_loop` `x.gather` step with a ~9.81 GiB allocation. Chunking the per-expert path bounds those transient tensors by processing routed tokens in smaller chunks.

So I would expect this to be a memory/throughput tradeoff: for experts with <=4096 routed tokens, behavior should be effectively unchanged; for hot experts above that threshold, peak memory is bounded, but throughput may be lower due to extra loop iterations and smaller matmuls.

The added test covers correctness by comparing chunked vs effectively unchunked execution for outputs and gradients, including expert-bias cases. I can run a specific throughput benchmark as a follow-up if there is a preferred MoE perf test/config you would like me to use.

PR #2982, opened 2026-07-09
fix(attention): initialize FlexAttention masks

akoumpa:
# What does this PR do ?

Initialize FlexAttention block masks from the current query/key geometry before the no-sink fast path reads the cache, fixing DeepSeekV3 pretraining failures caused by the undeclared `mask_key` attribute.

# Changelog

- Build and cache causal or sliding-window masks for both sink and no-sink attention paths.
- Key cached masks by window, batch size, query heads, sequence geometry, and device.
- Add a CPU regression test covering causal SDPA parity, cache reuse, and batch-geometry changes.

**Pre checks**:

# Validation

- `uv run pytest tests/unit_tests/attention` (`27 passed, 9 skipped`)
- `uv run ty check nemo_automodel/components/attention/flex_attention.py`
- Compiled CUDA DeepSeekV3 MLA forward/backward smoke test with finite loss and gradients
- [8-GPU DeepSeekV3 FlexAttention pretrain: 50/50 steps passed](https://gitlab-master.nvidia.com/dl/jet/ci/-/jobs/358889776)
- [Passing NeMo Bench pipeline](https://gitlab-master.nvidia.com/dl/gwe/nemo_bench/-/pipelines/57404623)

# Additional Information

- Related to [AM-371](https://linear.app/nvidia/issue/AM-371/nemo-bench-deepseek-v3-pretrain-flexattentionforward-fails-on-missing)

### DIFF
--- a/nemo_automodel/components/attention/flex_attention.py
+++ b/nemo_automodel/components/attention/flex_attention.py
@@ -22,10 +22,9 @@ from torch.nn.attention.flex_attention import (
     flex_attention,
 )
 
-# FlexAttention mask type. For each mask type, we initialize it at most once per
-# batch. To record what it is initialized, FLEX_ATTN_MASK_T is used as the key to
-# track the initialized mask.
-FLEX_ATTN_MASK_T = tuple[str, int | None] | tuple[int, int, int]
+# Cache masks by window, tensor geometry, and device so a BlockMask is reused only
+# for compatible attention inputs.
+FLEX_ATTN_MASK_T = tuple[int, int, int, int, int, torch.device]
 
 
 # Adapted from https://github.com/pytorch/torchtitan/pull/1559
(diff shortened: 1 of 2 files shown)


yuhezhang-ai:
Approving — this correctly fixes the never-assigned `mask_key` crash by building/caching the mask inline ahead of the sink branch, and the CPU regression test covers exactly the path that was broken. 

Two optional, non-blocking thoughts:

**1. Cache key over-specifies dimensions.** The key is `(sliding_window, batch_size, num_q_heads, q_len, kv_len, device)` and the mask is built with those exact `B`/`H` dims — but the causal / sliding-window `mask_mod`s never reference `b` or `h`, so the mask is identical across batch sizes and head counts. As written, a new (identical) `BlockMask` is built whenever batch size changes (e.g. a last partial batch, or eval). Building broadcastable masks with `B=1, H=1` and keying on `(sliding_window, q_len, kv_len, device)` would be equally correct and cheaper (fewer builds, less memory). Minor: `FLEX_ATTN_MASK_T` types element 0 as `int`, but the code still guards `sliding_window is not None` — worth making the annotation and the check agree.

**2. Out of scope, just confirming:** the sink path (`flex_attn(..., return_lse=True)`) omits `scale`, while the no-sink path passes `scale=scale`. If a caller ever passes a custom `scale` alongside `sink_weights`, it would silently fall back to the kernel default `1/sqrt(D)`. Looks fine for current callers (GPT-OSS uses the default), but flagging in case it should be threaded through.

yuhezhang-ai (approved):

Issue #2985, opened 2026-07-09
[Tracking] Large-scale MoE-VLM training improvements


yuhezhang-ai:
## Summary

We have been running large-scale MoE-VLM training (Qwen3.5-VL-scale, EP/TP/CP/FSDP2 mixes up to 128 GPUs) on top of Automodel and accumulated a set of fixes and features that are mature enough to upstream. This issue tracks the themed PR series; each PR ships with a CI-runnable (CPU / 1-GPU) test as its proof, with large-scale runs cited as supporting evidence.

Grouping is by coherent subsystem (one area, one reviewer, shared tests) rather than many atomic PRs, and risky/large changes are kept standalone so they cannot block the safe wins.

## PR queue

| # | Themed PR | Contents | Proof | Status | PR |
| -- | -- | -- | -- | -- | -- |
| 1 | MoE TP/EP correctness | TP-safe MoE parallelization; EP+TP gradient correctness; TE grouped-expert zero-grad handling | 264 unit tests (passing), validated in large-scale campaign | merged | NVIDIA-NeMo/Automodel#2995 |
| 2 | megatron-fsdp compatibility | compat fixes for megatron-fsdp 0.5 API (flattened DP-CP, TP DTensor local shape, API drift) | 149 unit tests (passing) | merged | NVIDIA-NeMo/Automodel#2986 |
| 3 | Vision-tower AC on the EP-MoE path | follow-up to NVIDIA-NeMo/Automodel#2840: the MoE AC path (`ep_size>1`) does not cover the vision tower; also a math-SDPA recompute `context_fn` for full-block vision AC | AC-scoping + recompute-parity unit tests (181 passing incl. regressions) | merged | NVIDIA-NeMo/Automodel#2993  <br>  <br>a relevant PR landed too: NVIDIA-NeMo/Automodel#3025 |
| 4 | CP vision sharding (frame-level) | frame-level context-parallel sharding of vision inputs; docs + 40 unit tests. Sharding scope by design: frozen ViT splits frames across CP × TP; trainable ViT splits within CP only (TP ranks recompute the ViT) to keep TP-replica gradients correct; pure-TP (cp=1) sharding not enabled | unit tests (40, passing), loss parity + throughput data from validation runs | merged | NVIDIA-NeMo/Automodel#2990 |
[...]

PR #2986, opened 2026-07-09
fix(distributed): Megatron-FSDP 0.5.0 compatibility

yuhezhang-ai:
# Summary

Makes the Megatron-FSDP code path work with the 0.5.0 API and fixes correctness/integration gaps that surface at scale.

Per review feedback, AutoModel no longer installs the official PyPI build while the required fixes are unreleased. The dependency now resolves from an immutable commit in a Megatron-LM fork:

* Fork: [yuhezhang-ai/Megatron-LM](<https://github.com/yuhezhang-ai/Megatron-LM>)
* Pinned commit: [`920e1d247ed0b3b79cb9eb941c3af087869b1b43`](<https://github.com/yuhezhang-ai/Megatron-LM/commit/920e1d247ed0b3b79cb9eb941c3af087869b1b43>)
* Source base: Megatron-LM `core_v0.18.0`, whose `megatron-fsdp` package metadata is `0.5.0`
* Package subdirectory: `megatron/core/distributed/fsdp/src`

The full commit SHA is pinned in `[tool.uv.sources]` and both lockfiles. The dependency constraint remains `megatron-fsdp==0.5.0` because AutoModel intentionally supports that API only. Once the fixes are available in an official release, the source override can be removed and the version pin updated.

There is no runtime monkeypatch or source rewriting in AutoModel.

## Main changes

1. **Pin the patched Megatron-FSDP source**

   The fork commit carries focused fixes for three published 0.5.0 defects:
   * NVIDIA/Megatron-LM#5786: `make_fsdp_dtensor` used a DTensor's global trailing shape when viewing a rank-local flat buffer. The fix derives the shape from `to_local_if_dtensor(orig_param)`.
   * NVIDIA/Megatron-LM#5787: `ParamAndGradBuffer.update_main_grads` repeated the same global-shape view for cached main gradients. It now uses the rank-local shape.
   * NVIDIA/Megatron-LM#5789: `MegatronFSDP.forward` called `self.module.forward(...)` directly and bypassed root-module hooks. It now invokes `self.module(...)`.
2. **Resolve flattened** `dp_cp` **meshes**

   `DeviceMesh._flatten()` records flattened dimensions in the root mesh's flatten mapping rather than `mesh_dim_names`. `MegatronFSDPManager.parallelize` now resolves `dp_cp` through the existing `get_flat_mesh` helper, while preserving support for meshes with a literal `dp_cp` dimension.
3. **Map precision controls onto the 0.5.0 API**

   Megatron-FSDP replaced `grad_reduce_in_fp32`, `preserve_fp32_weights`, and `check_for_nan_in_grad` with a `MixedPrecisionPolicy` and a separate per-parameter NaN reporter.

   `_megatron_fsdp_compat_kwargs` translates AutoModel's config onto the 0.5.0 `fully_shard` signature. Any older or unknown signature fails loudly with an error naming the required `megatron-fsdp==0.5.0` API instead of guessing a translation.

   `MegatronFSDPConfig.report_nan_in_param_grad` is an explicit opt-in, defaulting to `False`. Because 0.5.0 has no equivalent buffer-level NaN check, a truthy `check_for_nan_in_grad` is dropped with a one-time warning that points users to the precise reporter.
4. **Derive Megatron-FSDP wrapping units from the model**

   When `megatron_fsdp_unit_modules` is unset, AutoModel derives the wrapping classes from the model's `_no_split_modules`. It walks the instantiated model and resolves the actual classes, so both NeMo-custom and Hugging Face backends work. Resolution fails loudly when no matching modules exist.

   This replaces the previous hard-coded Llama FQN, which matched zero modules for NeMo-custom models and caused `total_fsdp_module=0` followed by `ZeroDivisionError`.
5. **Preserve Megatron-FSDP parameter state across reload/re-tie**

   Megatron-FSDP stores load-bearing state such as `_megatron_fsdp_model`, `_is_shared`, and `orig_param` on each `Parameter`. AutoModel's post-wrap reload and lm-head re-tie can replace those objects.

   Following the maintainer guidance in NVIDIA/Megatron-LM#5790, AutoModel snapshots the distributed parameter attributes before replacement and restores them afterward by parameter name. This unblocks optimizer sharding and the first backward pass on the pure-DP path.
6. **Complete recipe integration**
   * The retrieval bi-encoder recipe registers its optimizer with the wrapped Megatron-FSDP model via `shard_optimizers_for_megatron_fsdp`.
   * `dp_size == 1` configurations keep the TP-only fallback and skip Megatron-FSDP wrapping.
   * The two shipped Llama 3.2 Megatron-FSDP configs are re-enabled in release CI.

# Motivation

Training with the published Megatron-FSDP 0.5.0 source was broken or silently incorrect in several ways:

* The legacy precision kwargs raise `TypeError` against the 0.5.0 API.
* On `MeshContext`-built meshes with CP, the shard group can silently exclude CP.
* Rowwise TP parameters can deadlock during DTensor construction because global trailing dimensions are applied to rank-local storage.
* Cached main-gradient views repeat the same shape error after the first optimizer step.
* Calling the wrapped module's `.forward()` directly bypasses root hooks.
* Parameter replacement during AutoModel reload/re-tie drops Megatron-FSDP ownership state.
* The old hard-coded unit-module FQN can produce zero FSDP units for NeMo-custom models.
* The retrieval recipe did not register its optimizer with Megatron-FSDP.

# What's included

* `pyproject.toml`, `uv.lock`, and `docker/common/uv-pytorch.lock`
  * Keep the `megatron-fsdp==0.5.0` API constraint.
  * Resolve it from the exact fork commit and package subdirectory.
* `nemo_automodel/components/distributed/megatron_fsdp.py`
  * Resolve flattened `dp_cp`.
  * Preserve and restore Megatron-FSDP parameter attributes.
  * Plumb `report_nan_in_param_grad`.
* `nemo_automodel/components/distributed/config.py`
  * Add `report_nan_in_param_grad=False`.
  * Default `megatron_fsdp_unit_modules` to `None` for model-owned derivation.
* `nemo_automodel/components/distributed/parallelizer.py`
  * Map precision settings to the 0.5.0 API.
  * Derive wrapping classes from `_no_split_modules`.
  * Keep the `dp_size == 1` TP-only fallback.
* `nemo_automodel/recipes/retrieval/train_bi_encoder.py`
  * Register the optimizer with the wrapped model.
* Focused unit coverage for mesh selection, configuration plumbing, wrapping-unit derivation, parameter-state preservation, precision translation, strict signature handling, warnings, and the DP=1 fallback.

# Validation

## Local and CPU tests

* Megatron-LM fork regression tests: **3 passed**
  * Rank-local shape in `make_fsdp_dtensor`
  * Rank-local cached-main-grad view
  * Root forward-hook dispatch
* AutoModel `TestMegatronFSDPStrategyParallelize`: **29 passed**
* Ruff formatting and lint: passed
* `uv lock --check`: passed
* Both lockfiles resolve the full fork SHA.

## GPU and CI evidence

* The pure-DP `llm_finetune` path was validated end to end on 4×H100 at `tp=1, dp=4`: three finite training steps, decreasing loss/grad norm, and checkpoint save.
* The two shipped Llama 3.2 Megatron-FSDP release jobs passed on 8×H100 before the dependency-source change:
  * `llama3_2_1b_squad_megatron_fsdp`
  * `llama3_2_1b_hellaswag_megatron_fsdp`
  * Green pipeline: [58695995](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/58695995>)
* A fresh exact-job rerun pinned to this PR's fork-source commit is passed too:
  * Pipeline: [58965370](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/58965370>)
  * Scope: `llm / release`, EOS, x86-only AutoModel build
* After carrying the [#5976](<https://github.com/NVIDIA-NeMo/Automodel/issues/5976>) optimizer fix (fork commit `455389c`), the same two Megatron-FSDP release jobs pass on the fixed pin: [59252936](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/59252936>).
* These jobs verify the Megatron-FSDP path trains end to end. Gradient-norm parity with FSDP2/DDP is confirmed on Qwen3-0.6B (8×H100) after carrying the [#5976](<https://github.com/NVIDIA-NeMo/Automodel/issues/5976>) optimizer fix — mfsdp ZeRO-2/ZeRO-3 match FSDP2 and DDP within ≤0.55% at every step (see the resolved gap below).

# Known upstream gaps

* \*\*Megatron-FSDP optimizer under-update (\*\*NVIDIA/Megatron-LM#5974) — fixed. `fully_shard_optimizer`'s zero-gradient pre-initialization step (which materializes optimizer state for DCP checkpoint loading) advanced Adam's step counter, so the first real training update used the wrong bias correction and applied a materially smaller step (\~1.34×). This is fixed upstream in NVIDIA/Megatron-LM#5976 (make the pre-init inert and reset the per-group and per-parameter step counters), carried on the pinned fork (commit `455389c`). Grad-norm parity with FSDP2/DDP is restored — verified on Qwen3-0.6B (8×H100): mfsdp ZeRO-2/ZeRO-3 within ≤0.55% of FSDP2 at every step (vs +42% at step 1 before), and the minimal-repro update ratio drops from 1.35× to 1.01×. The fork override is dropped and the pin moved to an official Megatron-FSDP release once [#5976](<https://github.com/NVIDIA-NeMo/Automodel/issues/5976>) merges and ships (tracked in NVIDIA-NeMo/Automodel#3191).
* NVIDIA/Megatron-LM#5788 (a suspected temporary all-gather bucket lifetime race) turned out **not to be a defect** and has been retracted. `AllGatherPipeline.all_gather_params` issues `all_gather_stream.wait_stream(torch.cuda.current_stream())` before bucket storage is allocated, and the CUDA caching allocator only recycles a block for allocations on the allocating stream, so bucket reuse is already ordered after the consuming compute kernels. Nothing is patched or pending for it here.
* Tensor Parallelism together with Megatron-FSDP (`tp>1, dp>1`) remains out of scope. A vocab-embedding parameter can be `Shard(dim=0)` while its autograd gradient is `Replicate`, causing `_grad_acc` to copy a full-vocab gradient into rank-local `main_grad`. This reproduces on natively untied models and is tracked upstream in NVIDIA/Megatron-LM#5790 / draft NVIDIA/Megatron-LM#5836.
* The supported path in this PR is pure DP with `tp_size=1`.

---


hemildesai (commented):
Can we just upgrade Megatron FSDP to 0.5.0 and drop support for the older versions instead of having to maintain both?

yuhezhang-ai:
/ok to test [368212f](https://github.com/NVIDIA-NeMo/Automodel/pull/2986/commits/368212f22b85f1a176076a56a1ca7df44c8d945b)

yuhezhang-ai:
Context from @zhiqi-li (patch author): groot isn't using the Megatron-FSDP strategy right now — Muon and scaling runs are all on FSDP2 — so for their needs this could be deferred.

Thinking about AM more broadly, I believe it's worth landing. Two things are bundled: 1) 0.5.0 API compatibility (today, megatron_fsdp against the published wheel either raises at startup or silently drops CP), and 2) a fail-closed runtime layer fixing four real bugs in the 0.5.0 wheel itself — a rowwise-TP view deadlock, a stale master-grad view, an allocator free without record_stream, and a MegatronFSDP.forward hook bypass. The last two hit any sharded run, so any external megatron_fsdp customer is affected with no sign the fault is in the wheel, not AM.

@hemildesai, thanks for the earlier note; you were right, so I dropped the legacy path: it's now 0.5.0-only and fails loudly on anything else.

What I plan to work on before this could merge: (1) real 4-GPU tp=2, dp=2 validation — the current tests are CPU/mock and the existing TP2 test is dp=1, so it doesn't exercise these fixes; (2) upstream megatron-fsdp issues filed + a removal tracker for the runtime layer; (3) the NaN-check default change flagged as a breaking behavior change.
 
 @akoumpa  Given no active internal consumer plus that remaining work, I'd like your call on the path: shall we land this PR soon, or park for later?

yuhezhang-ai:
@akoumpa Regarding the megatron FSDP2 style API: yes, 0.5.0's fully_shard is the FSDP2-style entry point and it does simplify the API side — our shim is now just a thin config→policy mapping. But it doesn't help the four wheel bug fixes: those are bugs in megatron-fsdp's sharding internals that run regardless of the entry point, and 0.5.0 (the latest) still has all four.

Status: the API-compat + four wheel patches are done and unit-tested, and the rowwise-TP init fix is validated load-bearing on 4 GPUs. But end-to-end dp>1 can't be fully validated yet. It's blocked by additional upstream fragility in 0.5.0, which I've filed as NVIDIA/Megatron-LM#5786–#5790. Given there's also no active internal consumer, we may park this until those upstream fixes land, at which point the runtime patch layer can be deleted.

akoumpa:
Hi @yuhezhang-ai, thanks a lot for taking this on,

rgd versions, we don't need to maintain compatibility with older megatronfsdp versions. let's use 0.5.0's fully_shard-compatible API and drop / simplify any code that we had before to support the older mfsdp API. 

> Status: the API-compat + four wheel patches are done and unit-tested, and the rowwise-TP init fix is validated load-bearing on 4 GPUs. But end-to-end dp>1 can't be fully validated yet. It's blocked by additional upstream fragility in 0.5.0, which I've filed as https://github.com/NVIDIA/Megatron-LM/issues/5786–#5790. Given there's also no active internal consumer, we may park this until those upstream fixes land, at which point the runtime patch layer can be deleted. 

if TP is not supported, I think it's not the end of the world. I would prioritize ensuring 0.5.0 composes well with PP / EP, and make sure losses match and there's a perf gain. For TP, if the solution takes long, we can revisit in follow up PR.

yuhezhang-ai:
Hi @akoumpa 
I've investigated more and here's where things stand.

**What works today (this PR):** the non-TP DP sharding path — ZeRO-1/2/3 — is validated end-to-end (single- and multi-node, losses matching). Perf is model-dependent: for small/medium **dense** LLMs, ZeRO-2 is ~5% faster than FSDP2 (fair no-reshard baseline) and matches DDP speed at ~3 GB less memory — a genuine win. For the vision retriever nemotron vl I tested, it's the *weakest* option (no speed or memory edge over FSDP2 once precision is matched, and TE-FusedAdam can't run under its flat-buffer sharding).

**What's not here — and isn't a regression:**
- **PP/EP: never integrated** — the manager only builds a dp/tp/cp mesh, so there's no pipeline/expert path at all. Greenfield, a net-new build whenever there's real demand — not something this PR breaks.
- **TP+DP: wraps and shards, but the backward pass is blocked upstream.** At tp=2/dp=2 the model wraps on all ranks (#5786), the TP plan applies, and the optimizer shards — but the first backward fails on a vocab-embedding grad-shape mismatch; it's a general tp>1 issue that belongs upstream. 

**The 5 upstream bugs (all filed on Megatron-LM):**

| # | Issue | Status |
|---|---|---|
| #5786 | rowwise-TP global-shape view (init deadlock) | **patched downstream** (temporary shim) |
| #5787 | cached main-grad global-shape view | **patched downstream** (shim) |
| #5789 | forward bypasses root module hooks | **patched downstream** (shim) |
| #5788 | bucket-free missing record_stream | **needs upstream** — our record_stream workaround regressed peak memory, so deferred to the maintainer's event-based fix |
| #5790 | per-param state / vocab-embed grad routing under TP | non-TP half **patched downstream**; the tp>1 backward grad-shape case **needs upstream** |

The three downstream patches live in a fail-closed compat layer that's SHA-locked to the 0.5.0 wheel and designed to be deleted once a fixed release ships.

**My take:** I'd lean toward landing this as the enabling foundation (the wheel bug-fixes + a working, loss-matched non-TP DP path with a real small perf win for dense LLMs), but I'm equally fine parking it to wait for Megatron V2 and the upstream fixes. What do you think?

yuhezhang-ai:
/ok to test [8d19b2f](https://github.com/NVIDIA-NeMo/Automodel/pull/2986/commits/8d19b2f993a761e01f34c27c8475daa6a143909a)

yuhezhang-ai:
Hi @akoumpa , small follow-up, I made a bit further fix, and it now clears the CI errors on the remaining two Megatron-FSDP configs (pure-DP, non-TP), tracked in [AM-741](https://linear.app/nvidia/issue/AM-741/optimizer-not-registered-with-distributed-parameters-before), green run on real Llama-3.2-1B here:

https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/58695995

So it may be further worth landing. See my earlier comment on the PR for the full status/details.

nemo_automodel/components/distributed/megatron_fsdp_compat.py
```diff
@@ -0,0 +1,961 @@
...
+from importlib import metadata as importlib_metadata
+from pathlib import Path
+from types import CodeType, ModuleType, SimpleNamespace
+from typing import Any
+
+import torch.distributed as dist
+
+logger = logging.getLogger(__name__)
+
+_MEGATRON_FSDP_TP_DTENSOR_VERSION = "0.5.0"
```
akoumpa: @yuhezhang-ai is this file necessary? I thought we discussed dropping support for previous megatron fsdp versions

yuhezhang-ai: Yes, necessary, and it's not for old versions. This file is SHA-256-pinned to 0.5.0 only: it verifies the wheel's exact source before rewriting it and fails loud on any other version. What it does is patch three real bugs in the published 0.5.0 wheel itself — make_fsdp_dtensor (#5786, rowwise-TP deadlock), update_main_grads (#5787), and MegatronFSDP.forward bypassing root hooks (#5789) — which otherwise break dp>1 training. The green CI run on real Llama-3.2-1B depends on these.

The "dropping support for previous versions" we discussed was the pre-0.5.0 API translation (the legacy precision-boolean passthrough), that's already removed from parallelizer.py. This compat module is purely bug-patches for 0.5.0 and is designed to be deleted once a fixed megatron-fsdp release ships (the three bugs are filed upstream).

akoumpa: Thanks @yuhezhang-ai , i misunderstood.

in that case, my recommendation is to avoid using the official package until the fix has been released. Until then, my recommendation is to fork megatron-lm under your account, create a branch with the fixes and use a pinned commit from that branch, the pinned commit is a hard requirement. Once the fix has been released we can update the pyproject. 

Please let me know if that's ok with you.

yuhezhang-ai:
/ok to test [1a5aea3](https://github.com/NVIDIA-NeMo/Automodel/pull/2986/commits/1a5aea3f98fe4407429c1011c34886f16866078e)

yuhezhang-ai:
/ok to test [5df5a86](https://github.com/NVIDIA-NeMo/Automodel/pull/2986/commits/5df5a86e95ca1cd0ebb442749c36e67d933d1f16)

pyproject.toml
```diff
@@ -271,6 +271,7 @@ override-dependencies = [
 prerelease = "allow"
 
 [tool.uv.sources]
+megatron-fsdp = { git = "https://github.com/yuhezhang-ai/Megatron-LM.git", rev = "920e1d247ed0b3b79cb9eb941c3af087869b1b43", subdirectory = "megatron/core/distributed/fsdp/src" }
```
akoumpa: Thanks a lot for the pinned version @yuhezhang-ai

akoumpa (dismissed):
Thanks @yuhezhang-ai , I think the current PR can be merged;

One exit criteria i want to confirm is Yuhe, have you run fsdp2 vs mfsdp and compare gradient norms? I recall previous mfsdp had incorrect gradient scaling, and i want to make sure we cover this and don't run into it again. I don't expect this version to have it, but better safe than sorry. 

However, i feel we should follow up on:
- quantifying perf gain over fsdp2 
- i recall from discussions with mfsdp devs that mfsdp supports unified buffer with multiple precisions within it, therefore making per-dtype wrapping unnecessary, i think we don't do that right now . 
- precision plugin, in particular, i don't know if semantics are 100% same as fspd2's
- composition with other parallelisms, in particular, ep, tp, pp -- I think we should validate these before claiming support for mfsdp.


Finally, one more question, i recall mfsdp having fp8 / cudagraph support, have you tried these by any chance?

Thanks again @yuhezhang-ai !

yuhezhang-ai:
Thanks Alex! Answers below.

**Exit criterion — fsdp2 vs mfsdp grad norms:** ran it, and it caught a real one. Same model/seed/config: step-0 gradients are identical across FSDP2/DDP/mfsdp, but after the first optimizer step mfsdp diverges **+42%** in grad-norm while FSDP2 and DDP agree to <0.5%. Ruled out noise, gradient scaling (grads *are* correctly averaged), grad-clip, overlap, and fp32-master precision — it's the **optimizer update path**: mfsdp under-applies the step (~1.3–2× smaller update on the same gradient), both ZeRO-2 and ZeRO-3. So it's the behavior you remembered (technically the update, not the reduction). Minimal repro filed upstream: **NVIDIA/Megatron-LM#5974**.

**Proposal:** merge as **opt-in / experimental** with this documented as a known issue (PR body + tracking issue #3191), guidance to use FSDP2 for correctness-critical runs, and a pin bump to the fixed release once #5974 lands. If you'd rather hold until the upstream fix, I'm fine with that too — your call.

On your follow-ups (all tracked in #3191):

- **Perf over FSDP2 — model-dependent, so no blanket claim:** on Qwen3-0.6B (small dense LLM), 8×H100, mfsdp ZeRO-3 was ~5.6% faster than FSDP2 at matched sharding (ZeRO-2 ~7.7%, a lighter strategy FSDP2 can't do). But on the vision-heavy Llama-Nemotron VL retriever (8×H100), mfsdp was the *slowest* sharded option — no win over FSDP2 — because its per-unit-collective overhead isn't hidden when the vision forward is compute-light. So the speedup is compute-density-dependent, not universal, and provisional regardless until #5974 is fixed (a faster-but-under-updating optimizer isn't a real win yet).
- **Unified multi-precision buffer** (avoid per-dtype wrapping): agreed, we don't do that today — good follow-up.
- **Precision-plugin semantics vs FSDP2:** not fully verified, and #5974 is itself an optimizer-semantics divergence, so I wouldn't claim parity yet.
- **EP/TP/PP composition:** agree we shouldn't claim support. TP+DP is blocked upstream (vocab-embedding grad, NVIDIA/Megatron-LM#5790); EP/PP aren't integrated. This PR's supported path is pure DP, `tp_size=1`.
- **fp8 / CUDA-graph:** not tried yet — added to the follow-up list.

akoumpa:
Thanks @yuhezhang-ai , 

I think we can label this as eng-preview in the release notes, it's ok for this first release.

yuhezhang-ai:
@akoumpa Update: the upstream fix for the optimizer step issue landed as a PR NVIDIA/Megatron-LM#5976 ("Fixes #5974"). I verified it on our real run (Qwen3-0.6B, 8×H100): with #5976 applied, mfsdp ZeRO-2/ZeRO-3 now match FSDP2 and DDP grad-norm within ≤0.55% at every step (was +42% at step 1), losses match, and the minimal repro's update ratio drops from 1.35× → 1.01×. So the exit criterion passes.

I've carried #5976 on the pinned fork (455389c) and bumped the pin, CI is running on the fixed commit. I think this is good to merge now after CI. Can you help approve again?

yuhezhang-ai:
/ok to test [9677499](https://github.com/NVIDIA-NeMo/Automodel/pull/2986/commits/9677499909c9049c59240a2523a204347a4866f9)

yaoyu-33 (approved):

akoumpa (approved):

pyproject.toml
```diff
@@ -277,6 +277,7 @@ override-dependencies = [
 prerelease = "allow"
 
 [tool.uv.sources]
+megatron-fsdp = { git = "https://github.com/yuhezhang-ai/Megatron-LM.git", rev = "455389c480af6b3acdca74c7830c68b3274eb083", subdirectory = "megatron/core/distributed/fsdp/src" }
```
akoumpa, line 280: Hi @yuhezhang-ai , i know this PR uses the established pattern , but IIUC when someone tries pip install (not uv) then it will install 0.5.0 not your revision, just something to keep in mind for when we move to pip completely.

ko3n1g (approved):

PR #1711, opened 2026-04-07
feat: FSDP2 w weight prefetching and async TP optimization

ZhiyuLi-Nvidia:
# What does this PR do ?

Perf Improvement:
* llama3-70b peft MFU [31%](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/mqd71y3t/logs?nw=nwuserzhiyul) from previous [23%](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/c2xixp8u/logs?nw=nwuserzhiyul) 
* llama3-70b peft 2 node MFU [29%](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/wy501akk/logs?nw=85l7f0alfxb) from previous [19%](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/5gszgzsy/logs) 
* qwen2.5-32b peft MFU [30%](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/466n3aei/logs?nw=85l7f0alfxb) from previous [25%](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/i3kybjt3/logs)
* qwen2.5-32b peft 2 nodes MFU [28%](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/50zrfw28/logs?nw=85l7f0alfxb) from previous [25%](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/i3kybjt3/logs)
* llama3-70b full weight tuning MFU [39%](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/fxzrgwd8/logs?nw=85l7f0alfxb) from previous [36%](https://wandb.ai/nvidia/automodel-dev-zhiyul/runs/3k27x0e7)

**FSDP2 throughput optimizations + removal of combined projection modules.**

### FSDP2 Optimizations (`FSDP2Config` new flags)
- **Weight prefetching**: configurable forward/backward prefetch depth
- **Async tensor parallel** (`enable_async_tensor_parallel`): overlaps ReduceScatter with compute via `_micro_pipeline_tp`; requires `sequence_parallel=True`
- **Per-layer compile** (`enable_compile`): applies `torch.compile` after checkpoint load to avoid `_orig_mod` key mismatches
- **`patch_is_packed_sequence`**: removes per-layer CPU-GPU sync (`aten::is_nonzero`) and enables static shapes for compile

### Compile Compatibility
[...]

PR #2987, opened 2026-07-09
feat(distributed): shape TPLinear/LinearLoRA graphs for async-TP fusion

yuhezhang-ai:
# Summary

Make async tensor parallelism (`enable_async_tensor_parallel`, added in [#1711](<https://github.com/NVIDIA-NeMo/Automodel/issues/1711>)) actually fuse. When Dynamo is tracing with `torch._inductor.config._micro_pipeline_tp` set, `TPLinear.forward` and `LinearLoRA.forward` now emit the native `F.linear` graph that Inductor's async-TP pass can pattern-match, instead of the `torch.bmm` form. Every other compile and eager path is unchanged.

# Motivation

[#1711](<https://github.com/NVIDIA-NeMo/Automodel/issues/1711>) added the `enable_async_tensor_parallel` knob, which sets `torch._inductor.config._micro_pipeline_tp = True` and enables symmetric memory for the TP group so Inductor can emit `fused_all_gather_matmul` / `fused_matmul_reduce_scatter`.

However, the fusion never fires for layers routed through `TPLinear` (and `LinearLoRA`'s base projection). Inductor's micro-pipeline-TP pass recognizes the reshape-mm-reshape pattern that `F.linear` lowers to for 3-D input; it does not recognize `torch.bmm`. `TPLinear` deliberately uses `bmm` under compile because `F.linear`'s `aten.view` cannot flatten a sharded DTensor dimension (sequence parallelism) and recurses in AOT-autograd backward tracing. The result today: async-TP is "enabled" but the collectives are never overlapped for these layers.

This PR threads the needle: async-TP tracing gets the fusable `F.linear` graph (with the bias add kept outside `F.linear`, so the collective sees the matmul as its direct producer/consumer), while all other compile paths keep the DTensor-safe `bmm` fallback.

# What's included

* `nemo_automodel/shared/tp_linear.py` (new): lower-level graph-shaping policy shared by distributed TP and PEFT. It provides `tp_linear_forward()`, the async-TP gate, and the native linear graph with bias outside `F.linear`. The dispatch normalizes negative DTensor shard dimensions, so the standard row-parallel `Shard(-1)` layout is correctly treated as last-dimension sharding and remains eligible for fusion.
* `nemo_automodel/components/distributed/parallel_styles.py`: `TPLinear.forward` delegates to the shared dispatch; docstring updated.
* `nemo_automodel/components/_peft/lora.py`: `LinearLoRA.forward` delegates its base projection to the same shared dispatch; 2-D inputs traced under compile take `F.linear` instead of crashing in `torch.bmm`.
* CPU unit tests:
  * `tests/unit_tests/distributed/test_tp_linear.py`: gate is False in eager even with the flag set / False under compile without the flag / True with both; forward and backward numerical equivalence of `_async_tp_linear` vs `F.linear` (2-D/3-D, with/without bias); compiled-graph shape assertions via a graph-capturing backend — with the flag the Dynamo graph contains `F.linear` and no `torch.bmm` (`fullgraph=True`), without it the `bmm` path is preserved; eager behavior unaffected by the flag; `Shard(1)` keeps the safe `bmm` fallback while row-parallel `Shard(-1)` takes the fusable linear path.
  * `tests/unit_tests/_peft/test_lora.py`: LinearLoRA numerics under async-TP shaping match eager; 2-D input under compile tracing matches eager; DTensor regressions cover both `Shard(1)` fallback and `Shard(-1)` fusion eligibility across the base, LoRA-A, and LoRA-B projections.

# Validation

* Focused unit suite on commit `5acd9cfab`: **43 passed** on an H100 compute node (`test_tp_linear.py` + `test_lora.py`).
* Ruff formatting/lint and the repository import-layer contract passed for all changed files.
* Manual two-GPU H100 cluster validation (not a CI test) on commit `66a1cb4ac`:
  * compiled a tiny two-rank TP MLP with real CUDA DTensors, sequence-parallel input, symmetric memory, and Inductor micro-pipeline TP enabled;
  * matched the unsharded reference output within BF16 tolerance;
  * inspected the post-grad Inductor graph and confirmed both `symm_mem.fused_all_gather_matmul.default` and `symm_mem.fused_matmul_reduce_scatter.default` were emitted.
* Validated in internal large-scale runs (up to 128 GPUs) with `enable_async_tensor_parallel: true` + `enable_compile: true`, where the fused all-gather-matmul / matmul-reduce-scatter kernels fire on TP linears.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


yuhezhang-ai:
/ok to test [66a1cb4](https://github.com/NVIDIA-NeMo/Automodel/pull/2987/commits/66a1cb4ac554f0843290a6f036a5a2fba472bb97)

nemo_automodel/shared/tp_linear.py
```diff
@@ -0,0 +1,123 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Linear graph-shaping helpers shared by dense and PEFT tensor parallelism.
+
+Inductor's async tensor-parallel pass (``torch._inductor.config._micro_pipeline_tp``)
```
akoumpa, line 17: Hi @yuhezhang-ai is there a way not to have this file in the middle of components? Admittedly I'm not 100% sure where it should or whether we should break it. Please let me know what you think.

yuhezhang-ai: Good point. This module is shared by the distributed TPLinear path and PEFT’s LinearLoRA, which is why I initially kept it at the components/ root rather than making either component depend on the other. I agree that the root placement is awkward, though. I'll move it to `nemo_automodel/shared/tp_linear.py`; it only depends on PyTorch/DTensor and represents a lower-level graph-shaping primitive shared by both components. I don’t think splitting it is necessary because the gate, fusable linear implementation, and fallback dispatch form one cohesive policy.

yuhezhang-ai:
/ok to test [5acd9cfab](https://github.com/NVIDIA-NeMo/Automodel/pull/2987/commits/5acd9cfab634b97dfbfd4cdcf5f49ce4541dbb77)

nemo_automodel/shared/tp_linear.py
akoumpa: I'm wondering whether we could move this into the parallelizer 🤔

yuhezhang-ai: I considered that, but `parallelizer.py` already imports `parallel_styles.py`, where `TPLinear` lives. Moving the helper into `parallelizer.py` would make `parallel_styles.py -> parallelizer.py -> parallel_styles.py` circular. `LinearLoRA` would also need to depend on the full distributed parallelization orchestrator for a forward-time tensor primitive. That’s why I put the small dependency-free implementation in `shared/`.

akoumpa: I still think it's not ideal. 

but other than that, have you validated the expected results (e.g., comm + comp overlap)?

yuhezhang-ai: Yes, I validated this with a focused two-H100 distributed probe using real CUDA DTensors, confirming numerical parity and that both expected fused comm+compute kernels execute. 

But I’ll also run an async-TP on/off A/B with a real training config to quantify the end-to-end performance impact, and post the result later.

yuhezhang-ai: Follow-up A/B is complete: on a real Llama 3.1 8B full-SFT recipe with TP=2 on two H100s (5 warmup + 10 measured steps, same-node off/on), async TP improved median throughput from 9,969 to 10,650 tokens/s (+6.8%), and the generated Inductor code used the micro-pipelined fused matmul+reduce-scatter path.

I also checked LoRA in two reversed-order pairs; it was effectively flat because that graph only reached fused all-gather+matmul, so the end-to-end gain is workload-dependent.

yaoyu-33 (approved):

akoumpa (approved):

PR #2937, opened 2026-07-06
refactor(distributed): unify CP input prep and dispatch across models

HuiyingLi:
# What does this PR do ?

Unifies Automodel's context-parallelism plumbing behind one contract and one dispatch: **every CP backend is a** `ContextParallelismSharder`, every recipe forward goes through a single `prepare_cp_forward` call, and the resolved sharder is returned to the caller — making it the public downstream CP interface proposed in NVIDIA-NeMo/Automodel#2861. Model-specific CP logic lives in model directories; the framework builds sharders for the paths it owns (generic torch `context_parallel`, TE/THD, MagiAttention). Closes NVIDIA-NeMo/Automodel#2879; delivers the contract for NVIDIA-NeMo/Automodel#2861.

# Changelog

Reviewable sequentially; later commits also delete the transitional compatibility earlier ones introduced, so the final tree carries no shims. By theme:

**The** `ContextParallelismSharder` **contract** (`components/distributed/cp_sharder.py`) — a dataclass of two required callables plus shard-time-captured facts, replacing the private batch keys (`_cp_make_batch_fn`, `_cp_metadata_*`, `_cp_full_logits_grad_touch`):

* `shard_batch(cp_mesh, tp_mesh, batch, *, loss_mask=None, padding_token_id=0) -> (ctx_factory, batch)`
* `local_token_global_indices(cp_mesh, padded_seq_len, device) -> LongTensor` — the global position of every local token, the universal layout coordinate system. `shard_token_tensor` / `gather_token_tensor` (differentiable all-gather + reorder) are synthesized from it. Closed-form for contiguous/round-robin; data-dependent layouts (THD `cu_seqlens` partitioning, magi's dispatch solver) install the partition their `shard_batch` just computed.

The sharder carries no backend tag — nothing may branch on which backend produced it (a `layout` diagnostic string existed transiently and was removed as consumer-less).

[...]

PR #2990, opened 2026-07-09
feat(distributed): frame-level context-parallel vision-tower sharding

yuhezhang-ai:
## Summary

Adds frame-level context-parallel (CP) vision-tower sharding and wires it into the VLM recipe plus two dense production model paths:

* `nemo_automodel.components.distributed.cp_vision_shard` partitions independent image/video frames across the CP group, runs one local vision slice per rank, and differentiably all-gathers `pooler_output` plus optional `deepstack_features` in original order.
* The VLM recipe publishes the CP-only process group around each model forward (and its backward during training), matching the merged sharder contract where vision encoding and embedding happen inside the model forward.
* Dense `Qwen3_5ForConditionalGeneration` uses the helper for in-forward image and video encoding.
* Dense `Qwen3VLForConditionalGeneration` adds a Hugging Face-compatible registered model path. Mixed image/video batches use one vision forward. Its model-owned hook is metadata/sharder-only; embedding, vision encoding, visual-mask construction, DeepStack alignment, and differentiable sequence sharding run inside `forward` per microbatch.
* Typed YAML policy under `distributed`, runnable CP2 examples (Qwen3.5-4B and Qwen3-VL-8B), model coverage documentation, CPU composition/parity tests, and two-rank Gloo collective tests are included.

Part of NVIDIA-NeMo/Automodel#2985.

## Scope and planned follow-ups

| Stage | Scope |
| -- | -- |
| **This PR** | Shared vision-sharding primitive, typed recipe configuration, and production CP integration for **dense Qwen3.5 VLM** and **dense Qwen3-VL**, with examples, documentation, focused forward/backward parity, real-collective coverage, and exact-config GPU smoke tests. This establishes the reusable correctness contract; it is not a large-scale performance qualification. |
| **Stacked MoE follow-up** | Consume this primitive to enable **Qwen3.5-MoE** CP training, including its model-specific in-forward multimodal path and composed **EP x CP** trainable-vision gradient and collective-ordering validation. |
| **Large-scale/final integration after the stacked upstream work lands** | Exercise representative large-scale FSDP2 training across dense and MoE models, especially trainable vision towers, packed image/video workloads, and composed EP x CP. Revalidate end-to-end loss, gradient, memory, and throughput there; this is series-level qualification, not additional scope for this primitive PR. |
| **Separate, non-blocking work** | Tensor-parallel integration is lower priority and is not required for this PR or the MoE CP follow-up. Qwen2.5-VL is not being added because its AutoModel surface is being deprecated. Qwen3-VL-MoE CP integration is not currently planned and can be added later if an active use case requires it. |

The primitive is model-agnostic only for vision towers matching the independent-frame `visual(pixel_values, grid_thw=..., return_dict=True)` contract. This is not blanket support for every model with "Qwen" in its name: each additional VLM family requires a model-owned pre-embed integration and tensor-contract coverage.

## Configuration

Vision sharding is declarative, lives under the multimodal distributed policy, and is disabled by default:

```yaml
distributed:
  cp_size: 2
  multimodal:
    vision:
      frame_sharding:
        enabled: true
        mesh_dims: [cp]
        min_tokens: 2048
        cost_alpha: auto  # infer 3 * vision_hidden_size
```

`cost_alpha: auto` is the default. `null` remains a backward-compatible alias, while a non-negative integer is an exact experiment override (`0` restores the pure-quadratic `p**2` partition cost). Only `mesh_dims: [cp]` is currently supported.

Use `--distributed.multimodal.vision.frame_sharding.enabled false` for a recorded A/B baseline. There is no environment-variable control path.

## Motivation

VLM CP embeds media before entering the causal CP attention context. Without this helper, every CP rank runs the full non-causal vision tower over all media, causing `cp_size`-fold redundant ViT compute and full-media vision activations on every rank.

Qwen3.5/Qwen3-VL-style towers build vision attention boundaries per frame. Their frames can therefore be processed independently and concatenated back in entry/frame order, up to normal floating-point numerical tolerance.

## Design

* **Frame-level partitioning:** each `(t, h, w)` entry expands into `t` independent frame units. Contiguous cuts preserve gather order and allow a single video to span CP ranks.
* **Cost-aware balance:** frame cost is `p * (p + alpha)`, where `p = h * w`. By default `alpha` is inferred as `3 * vision_hidden_size`; callers may provide an exact non-negative override. This is the model-aware default.
* **Differentiable gather:** variable-length local outputs are padded, all-gathered, and sliced. Backward uses reduce-scatter SUM, routing gradients to the ranks that computed each frame.
* **Uniform collective ordering:** every rank makes the same distribute/pad/replicate choice. When frames are fewer than ranks, minimal dummy frames keep the vision-forward/collective sequence uniform and their gathered tail is removed before the loss.
* **Collective-safe validation:** rank-local output validation reaches group consensus before later collectives, so one invalid rank cannot leave peers blocked.
* **Small-work fallback:** workloads below `min_tokens` remain replicated to avoid collective overhead.

### Sharding-group scope

* **Trainable vision tower:** CP-only group. This is gradient-correct when the vision tower is replicated over non-CP axes.
* **Frozen vision tower:** the lower-level helper also permits a wider CP x TP group when explicitly declared, but the built-in recipe currently publishes CP-only for both frozen and trainable towers.
* **Pure TP (**`cp_size == 1`**):** vision sharding is inactive.

## Validation

**Validation boundary:** the H100 runs below are production-path integration smokes with frozen vision towers. Trainable-tower backward correctness is covered here by differentiable-gather gradient parity and real two-rank collectives. Large-scale NCCL/FSDP2 training with trainable vision towers, and its representative performance characterization, is intentionally assigned to the stacked/final integration work above.

* Real small Qwen3.5 and Qwen3-VL vision towers preserve pooled output ordering across entry and single-video frame partitions; Qwen3-VL coverage also checks every DeepStack output.
* Dense Qwen3-VL keeps non-CP Hugging Face parity and full image pre-embed/DeepStack/logit parity, including tied-weight behavior.
* The Qwen3-VL in-forward CP path checks mask/sequence contracts, head-tail ordering, padding, and gradient coverage.
* Two-rank Gloo tests exercise the actual all-gather/reduce-scatter path, gradient parity, dummy padding, and divergent-rank validation without deadlock.
* Local review runs: 313 selected CPU tests passed (3 expected skips), 4/4 real-collective Gloo tests passed, Ruff and import-linter passed, and 292 MDX files validated. On the final nested-config head, an additional 69 distributed/Qwen3.5 tests, 108 recipe/parser tests, and 8 Qwen3-VL in-forward CP tests passed; Ruff formatting/checks are clean. On the final rebased head (`b0c095cd3`), 295 focused tests passed (3 expected skips), and all 4 real-collective Gloo tests passed again.
* **Exact post-**NVIDIA-NeMo/Automodel#2763 **namespace release CI (head** `0f14f0c14`**):** both official 1x8 H100 recipes installed the exact PR head, logged the production vision-tower `SHARDED` path over CP2, and completed with finite losses and gradients. Qwen3-VL-8B completed its six-step dataset epoch; Qwen3.5-4B completed all 50 configured steps. NeMo CI: [parent pipeline](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/59833709>), [AutoModel child](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/59837059>), [VLM leaf](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/59837277>), [Qwen3-VL-8B](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/375597262>), [Qwen3.5-4B](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/375597263>).
* **Exact Qwen3-VL-8B recipe A/B** (`ca0ae15f4`, code-equivalent to the final head; 1x8 H100, FSDP2 CP2/DP4, frozen vision tower, RDR, one six-step epoch): both variants consumed the same label-token sequence. Max/mean absolute loss delta was 0.0058/0.00225. Excluding compile step 0, observed mean TPS was 4.3% higher with sharding and peak memory was 41.70 GiB for both. Because this is only five steady steps over a small, sample-varying workload, the TPS observation is smoke evidence, not a benchmark claim. W&B: [replicated](<https://wandb.ai/Nemo-automodel/yuhez_workspace/runs/gbc52yjo>), [sharded](<https://wandb.ai/Nemo-automodel/yuhez_workspace/runs/c4pioop7>).
* **Exact Qwen3.5-4B sharded recipe smoke** (`ca0ae15f4`; 1x8 H100, FSDP2 CP2/DP4, frozen vision tower, MedPix-VQA): completed 10/10 steps with finite loss and gradients; peak logged memory was 28.77 GiB. This verifies that the newly added 4B example, model path, and CP vision-shard collective execute together; it is not an A/B performance result. W&B: [sharded](<https://wandb.ai/Nemo-automodel/yuhez_workspace/runs/ln6bttgk>).
* **Longer exact-head Qwen3.5-0.8B A/B** (`83a085a12`, 1x8 H100, FSDP2 CP2/DP4, frozen vision tower, MedPix-VQA, 30 steps): max/mean absolute loss delta was 0.0031/0.00098. Excluding compile step 0, sharding improved mean TPS by 9.5% over steps 1-29; the stricter steps 5-29 window improved by 6.7% (paired geometric mean +8.2%). Overall memory was unchanged for this small, language-model-dominated one-image workload (7.24 GiB median for both; 8.24 vs 8.25 GiB peak). W&B: [replicated](<https://wandb.ai/Nemo-automodel/yuhez_workspace/runs/iw3mwnjw>), [sharded](<https://wandb.ai/Nemo-automodel/yuhez_workspace/runs/02wuuk8e>).

## Update: shared multimodal policy namespace

[#2763](<https://github.com/NVIDIA-NeMo/Automodel/pull/2763>) has merged, and this PR now extends its typed `distributed.multimodal` namespace through `MultimodalDistributedConfig`. The policy lives at `distributed.multimodal.vision.frame_sharding`, aligned with [#3219](<https://github.com/NVIDIA-NeMo/Automodel/issues/3219>). Its `mesh_dims: [cp]` field makes the selected sharding axis explicit; `[cp]` is the only currently supported value and unsupported dimensions are rejected. This is a configuration/plumbing alignment only; the frame partitioning, collective, model-capability gate, and validation behavior described above are unchanged.


yuhezhang-ai:
Superseded by the corrected exact-head `/ok to test` command below (SHA transcription error).

athitten:
@yuhezhang-ai there's a CP refactor PR from @HuiyingLi #2937 , might want to check if this PR needs to align with the refactored design, not sure if it impacts. Just adding as an FYI

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
+---
+title: "Context-Parallel Vision Sharding"
+description: "Distribute the VLM vision tower across context-parallel ranks at frame granularity, removing the per-rank replicated ViT forward."
+---
+
+## Overview
+
+Vision-language models run the vision tower (ViT) in a **pre-embed** step *outside* the
+context-parallel (CP) attention context: the model embeds all images/videos, scatters the
+vision features into `inputs_embeds`, and only *then* shards the sequence across CP ranks.
```
jgerh:
```suggestion
Vision-language models run the vision tower (ViT) in a pre-embed step outside the
context-parallel (CP) attention context. The model embeds all images and videos, scatters the
vision features into `inputs_embeds`, and then shards the sequence across CP ranks.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+Vision attention is non-causal and incompatible with CP ring attention, so this step cannot
+reuse the sequence shard.
+
+Without vision sharding, that pre-embed `visual(...)` call runs on the **full, un-sharded set
+of images on every CP rank**: with `cp_size=N` the vision tower does **N× redundant compute**
+and holds **O(all-images)** vision activations on each rank.
```
jgerh:
```suggestion
Without vision sharding, the pre-embed `visual(...)` call runs on the full, unsharded set
of images on every CP rank. With `cp_size=N`, the vision tower performs N times the necessary
compute and holds `O(all-images)` vision activations on each rank.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+`nemo_automodel.components.distributed.cp_vision_shard` removes this redundancy by
+partitioning the visual input across the CP group at **frame granularity**, running the
+vision tower on each rank's slice only, and all-gathering the per-frame embeddings back in
+order. The downstream scatter into `inputs_embeds` and the sequence shard consume
+numerically equivalent (`allclose`) embeds through the unchanged code path.
+
+## Design
+
+### Key property: vision frames are independent
```
jgerh:
```suggestion
### Vision Frames Are Independent
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+Without vision sharding, that pre-embed `visual(...)` call runs on the **full, un-sharded set
+of images on every CP rank**: with `cp_size=N` the vision tower does **N× redundant compute**
+and holds **O(all-images)** vision activations on each rank.
+
+`nemo_automodel.components.distributed.cp_vision_shard` removes this redundancy by
+partitioning the visual input across the CP group at **frame granularity**, running the
+vision tower on each rank's slice only, and all-gathering the per-frame embeddings back in
+order. The downstream scatter into `inputs_embeds` and the sequence shard consume
+numerically equivalent (`allclose`) embeds through the unchanged code path.
```
jgerh:
```suggestion
numerically equivalent (`allclose`) embeddings through the unchanged code path.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+## Design
+
+### Key property: vision frames are independent
+
+Qwen3.5/Qwen3-VL-style vision towers have **zero cross-frame coupling**:
```
jgerh:
```suggestion
Qwen3.5- and Qwen3-VL-style vision towers have no cross-frame coupling:
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+- attention is **per-frame** — `cu_seqlens` is built from `grid_thw` so each frame's `h*w`
+  patches attend only within that frame;
+- both the rotary and the learned position embeddings are **spatial-only, repeated
+  identically per frame** (temporal reasoning happens later, in the LLM via mRoPE);
+- the patch embedding is per temporal-patch.
```
jgerh:
```suggestion
- Attention is per frame. `cu_seqlens` is built from `grid_thw`, so each frame's `h*w`
  patches attend only within that frame.
- The rotary and learned position embeddings are spatial only and repeat identically for
  each frame. Temporal reasoning happens later in the LLM through mRoPE.
- The patch embedding is per temporal patch.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+So, with all CP ranks holding identical FSDP-all-gathered vision weights:
```
jgerh:
```suggestion
With all CP ranks holding identical fully sharded data parallel (FSDP) all-gathered vision
weights, the following relationships apply:
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+```
+visual((t,h,w)) == visual(t × (1,h,w))    and    visual(all_frames) == concat_r visual(frames_of_rank_r)
+```
+
+hold up to **numerical precision** (`allclose`). Every `(t,h,w)` entry is expanded into `t` per-frame units, the *frames*
+are partitioned contiguously across the CP group, each rank runs `visual` on its slice, and a
```
jgerh:
```suggestion
These relationships hold within numerical precision (`allclose`). Every `(t,h,w)` entry
expands into `t` per-frame units. The frames are partitioned contiguously across the CP group,
each rank runs `visual` on its slice, and a
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+differentiable all-gather reassembles the embeddings in the original order. Sharding at frame
+granularity (rather than whole-entry) lets a **single large video** — or a pack with fewer
+videos than ranks — shard across ranks, and balances finer. Images are `t == 1`, i.e. one
+unit each.
```
jgerh:
```suggestion
granularity instead of by whole entry allows a single large video, or a pack with fewer
videos than ranks, to be sharded across ranks and provides finer-grained balancing. Images
have `t == 1`, which means that each image is one unit.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+### Flow (per CP rank `r`)
```
jgerh:
```suggestion
### Flow for Each CP Rank `r`
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+```
```
jgerh:
```suggestion
The following flow shows how each CP rank processes the vision input:

```text
```

docs/guides/cp-vision-frame-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+             pad with (world - num_frame_units) minimal (1,sms,sms) zero dummy frames
+             cuts = one frame unit per rank                           [every rank runs ONE frame]
+         else:
+             cuts = contiguous, attention-cost-balanced partition of FRAMES, >=1 frame/rank
+         local  = self.visual(pixel_rows[r], coalesce(frames_of_rank_r))
+         full   = all-gather(local) over the CP group, reassembled in frame order
+         full   = full[:real_token_count]                             # drop the dummy tail (pad path)
+  # unchanged from here: masked_scatter into inputs_embeds, sequence shard
+```
+
```
jgerh, line 52:
```suggestion
The following table describes the primary design concerns and their resolutions:

```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+| **Gradient parity** | Every rank reassembles the full embeds, then keeps only its sequence shard. A frame computed by rank `c` lands in shard-owner rank `s`; reduce-scatter-SUM routes `s`'s gradient back to `c` (summing the straddling case). The total vision-parameter gradient is numerically equivalent (`allclose`) to the replicated baseline's. |
+| **FSDP / no deadlock** | The distribute / pad / replicate choice is a pure function of `(num_frame_units, world, enabled)` and thus identical on all ranks; every rank makes exactly one `visual` call and the same number of vision-shard collectives. Fewer frame units than ranks **pads with minimal dummy frames** (one unit per rank) so no rank is starved; the dummy embeddings are sliced off before the loss and contribute exactly zero gradient. The all-gather inside `maybe_distribute_visual` is the synchronization boundary — do **not** add an extra post-visual CP barrier. |
+| **Small workloads** | Batches whose visual token count is below `distributed.cp_vision_sharding.min_tokens` (default `2048`) use the replicated path: they have little memory pressure but would still pay the all-gather synchronization cost. |
```
jgerh:
```suggestion
| Load Balance | A contiguous partition of per-frame units is balanced by `p*(p+alpha)`, where `p` is the per-frame attention sequence size (`h*w`), with at least one frame per rank. The quadratic term captures per-frame attention. `alpha*p` adds a model-scaled proxy for the linear per-patch work, such as QKV and MLP projections, that pure `p**2` ignores. As a result, packs that mix large image frames with many small video frames no longer place most small frames on a few ranks. By default, `alpha = 3 * vision_hidden_size` is discovered from the model (Qwen `visual.config.hidden_size`, nested `vision_config`, Nemotron `vit_hidden_size`, and RADIO `embed_dim`, without class-name checks). Unknown towers fall back to `alpha=0`, which uses the legacy pure-quadratic calculation. The contiguous partition allows the gathered blocks to concatenate in their original frame order without reshuffling. The partition is deterministic, so every rank computes the same cuts. |
| Differentiable All-Gather | Per-rank `[max_tokens, H]` blocks are padded to the global maximum and all-gathered. The forward pass uses all-gather and concatenation, and the backward pass uses reduce-scatter SUM. The blocks are then sliced back to their actual token counts. |
| Gradient Parity | Every rank reassembles the full embeddings, then keeps only its sequence shard. A frame computed by rank `c` lands in shard-owner rank `s`. Reduce-scatter SUM routes the gradient from `s` back to `c` and sums gradients when a frame spans shards. The total vision-parameter gradient is numerically equivalent (`allclose`) to the replicated baseline gradient. |
| FSDP Without Deadlock | The distribute, pad, or replicate choice is a pure function of `(num_frame_units, world, enabled)` and is therefore identical on all ranks. Every rank makes exactly one `visual` call and the same number of vision-shard collectives. When there are fewer frame units than ranks, the operation pads the input with minimal dummy frames, one unit per rank, so every rank receives work. The dummy embeddings are sliced off before the loss and contribute exactly zero gradient. The all-gather inside `maybe_distribute_visual` is the synchronization boundary. Do not add an extra post-visual CP barrier. |
| Small Workloads | Batches with a visual token count below `distributed.cp_vision_sharding.min_tokens` (default `2048`) use the replicated path. These batches have little memory pressure but would still incur the all-gather synchronization cost. |
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+`maybe_distribute_visual` also all-gathers an optional `deepstack_features` list
+(Qwen3-VL-style towers); `last_hidden_state` is intentionally left as the local shard since
+`embed_multimodal` only reads `pooler_output` / `deepstack_features`.
```
jgerh:
```suggestion
`maybe_distribute_visual` also all-gathers an optional `deepstack_features` list for
Qwen3-VL-style towers. `last_hidden_state` intentionally remains the local shard because
`embed_multimodal` reads only `pooler_output` and `deepstack_features`.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+### Sharding scope: frozen vs. trainable vision tower
```
jgerh:
```suggestion
### Choose a Sharding Scope for a Frozen or Trainable Vision Tower
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+The caller **declares** the scope of the process group it publishes via
+`set_cp_vision_group(group, config=..., spans_only_cp=...)`, and the declaration is **enforced**:
+`maybe_distribute_visual` raises a `ValueError` — on every rank, before any collective —
+when the vision tower has trainable parameters but the published group was not declared
```
jgerh:
```suggestion
The caller declares the scope of the process group that it publishes by using
`set_cp_vision_group(group, config=..., spans_only_cp=...)`, and the function enforces the
declaration. On every rank, `maybe_distribute_visual` raises a `ValueError` before any
collective when the vision tower has trainable parameters but the published group is not
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+CP-only (`spans_only_cp=False`).
+
+| Vision tower | Helper-supported sharding group | Why |
```
jgerh:
```suggestion
| Vision Tower | Helper-Supported Sharding Group | Why |
```

docs/guides/cp-vision-frame-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+`embed_multimodal` only reads `pooler_output` / `deepstack_features`.
+
+### Sharding scope: frozen vs. trainable vision tower
+
+The caller **declares** the scope of the process group it publishes via
+`set_cp_vision_group(group, config=..., spans_only_cp=...)`, and the declaration is **enforced**:
+`maybe_distribute_visual` raises a `ValueError` — on every rank, before any collective —
+when the vision tower has trainable parameters but the published group was not declared
+CP-only (`spans_only_cp=False`).
+
```
jgerh, line 73:
```suggestion

The supported sharding group depends on whether the vision tower is frozen or trainable:

```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+|---|---|---|
+| **Frozen** | Full **CP × TP** rank set (`spans_only_cp=False`) | Frame independence + replicated weights make the forward numerically equivalent (`allclose`) for *any* rank group with identical weights, and `requires_grad=False` means no backward ever runs through the gather — zero gradient risk. |
+| **Trainable** | **CP group only** (`spans_only_cp=True`, the default) | The vision tower is replicated (not tensor-parallel) across TP ranks. Gathering frames across CP × TP would make the all-gather's reduce-scatter(SUM) backward accumulate the vision gradient **tp-fold**: each TP-replicated sequence shard backprops into the single rank that computed the frame. TP ranks therefore redundantly compute the ViT — and `maybe_distribute_visual` **raises** if a trainable tower is published with `spans_only_cp=False`. |
+| **Pure TP** (`cp_size == 1`) | Not enabled | With no CP there is no CP group to publish: either no group is set, or the published group has size 1 — `maybe_distribute_visual` falls back to the replicated path whenever the group size is ≤ 1. |
```
jgerh:
```suggestion
| Frozen | Full CP × TP rank set (`spans_only_cp=False`) | Frame independence and replicated weights make the forward pass numerically equivalent (`allclose`) for any rank group with identical weights. Because `requires_grad=False`, no backward pass runs through the gather, so there is no gradient risk. |
| Trainable | CP group only (`spans_only_cp=True`, the default) | The vision tower is replicated, not tensor parallel, across TP ranks. Gathering frames across CP × TP would cause the all-gather reduce-scatter SUM backward pass to accumulate the vision gradient `tp` times. Each TP-replicated sequence shard propagates its gradient into the single rank that computed the frame. TP ranks therefore compute the ViT redundantly, and `maybe_distribute_visual` raises an error if a trainable tower is published with `spans_only_cp=False`. |
| Pure TP (`cp_size == 1`) | Not enabled | Without CP, there is no CP group to publish. Either no group is set, or the published group has size `1`. `maybe_distribute_visual` falls back to the replicated path whenever the group size is less than or equal to `1`. |
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+Extending the trainable case to CP × TP would require gathering over CP × TP while reducing
+gradients over CP only (or scaling by `1/tp`), plus a gradient-parity test — until then the
+CP-only group is correct by construction.
```
jgerh:
```suggestion
gradients over CP only, or scaling by `1/tp`, and adding a gradient-parity test. Until then,
the CP-only group is correct by construction.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+The built-in dense Qwen3.5 and Qwen3-VL integrations publish the **CP-only group for both
+frozen and trainable towers**. A wider frozen-tower group is supported by the helper, but a
+caller must publish that group explicitly; the recipe does not derive or publish CP × TP.
```
jgerh:
```suggestion
The built-in dense Qwen3.5 and Qwen3-VL integrations publish the CP-only group for both
frozen and trainable towers. The helper supports a wider frozen-tower group, but a
caller must publish that group explicitly. The recipe does not derive or publish CP × TP.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+    enabled: true
+    min_tokens: 2048
+    cost_alpha: auto  # infers 3 * vision_hidden_size
+```
+
+The policy is disabled by default, so existing CP recipes retain replicated vision behavior.
+Enable it explicitly in the recipe. For an A/B baseline, use the recorded CLI override
+`--distributed.cp_vision_sharding.enabled false` rather than an environment variable.
+
+Runnable CP2 examples are available for
```
jgerh:
```suggestion
Runnable CP size 2 examples are available for
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+Programmatic callers pass the same typed policy while publishing the runtime-only process
+group:
+
+```python
+from nemo_automodel.components.distributed import cp_vision_shard
+
+config = cp_vision_shard.CpVisionShardingConfig(enabled=True)
+# Trainable vision tower: publish the CP-only group (spans_only_cp defaults to True).
+token = cp_vision_shard.set_cp_vision_group(cp_group, config=config)
+# Frozen vision tower only: a wider CP x TP group may be published instead --
```
jgerh:
```suggestion
# Frozen vision tower only: a wider CP x TP group can be published instead.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+# token = cp_vision_shard.set_cp_vision_group(cp_tp_group, config=config, spans_only_cp=False)
+try:
+    # pre-embed forward; embed_multimodal routes self.visual(...) through
+    # cp_vision_shard.maybe_distribute_visual(self.visual, pixel_values, grid_thw)
+    ...
+finally:
+    cp_vision_shard.reset_cp_vision_group(token)
+```
+
+Configuration fields:
```
jgerh:
```suggestion
The following configuration fields control CP vision sharding:
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+| Field | Default | Effect |
+|---|---|---|
+| `distributed.cp_vision_sharding.enabled` | `false` | Enables CP vision sharding. `false` preserves the replicated ViT path. |
+| `distributed.cp_vision_sharding.min_tokens` | `2048` | Visual-token threshold below which the replicated path is used. |
+| `distributed.cp_vision_sharding.cost_alpha` | `auto` | Linear term `alpha` in the `p*(p+alpha)` partition cost. `auto` discovers `3 * vision_hidden_size` from the model; `null` remains a backward-compatible alias. A non-negative integer is an exact override (`0` restores the legacy pure-quadratic `p**2` partition). Partition-only knob — forward/gradients are identical for any cuts. |
```
jgerh:
```suggestion
| `distributed.cp_vision_sharding.min_tokens` | `2048` | Sets the visual token threshold below which the replicated path is used. |
| `distributed.cp_vision_sharding.cost_alpha` | `auto` | Sets the linear term `alpha` in the `p*(p+alpha)` partition cost. `auto` discovers `3 * vision_hidden_size` from the model. `null` remains a backward-compatible alias. A nonnegative integer is an exact override. A value of `0` restores the legacy pure-quadratic `p**2` partition. This field affects partitioning only. The forward pass and gradients are identical for any cuts. |
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+  backward; end-to-end `maybe_distribute_visual` (including deepstack, video entries, and a
+  single video split across ranks at frame granularity); every fallback (including `None`
+  media inputs); the trainable-tower sharding-scope guard; and the pad path end-to-end
+  through the production code, asserting `sum_r grad_r == replicate grad` and that every
+  padded (dummy) rank contributes exactly zero gradient.
+- **Qwen3.5 composition tests**
+  (`tests/unit_tests/distributed/test_cp_vision_shard_qwen3_5.py`, CPU-only): the real,
+  small `Qwen3_5VisionModel` produces the same pooled output for a replicated forward and
+  concatenated contiguous frame slices, including the single-video frame-split case.
+- **Qwen3-VL composition and model tests**
```
jgerh:
```suggestion
Validation includes the following tests and runs:

- Unit tests (`tests/unit_tests/distributed/test_cp_vision_shard.py`, CPU only) validate
  `visual(full) == concat_r visual(slice_r)` for the forward pass and vision-parameter
  gradient at world sizes 2–4. They also validate partition balance, coverage, and order.
  Other tests cover variable-token all-gather forward and backward passes and end-to-end
  `maybe_distribute_visual` behavior. The end-to-end tests include DeepStack, video entries,
  a single video split across ranks at frame granularity, every fallback (including `None`
  media inputs), the trainable-tower sharding-scope guard, and the pad path through the
  production code. The pad-path test asserts `sum_r grad_r == replicate grad`. It also
  verifies that every padded dummy rank contributes exactly zero gradient.
- Qwen3.5 composition tests
  (`tests/unit_tests/distributed/test_cp_vision_shard_qwen3_5.py`, CPU only) verify that the
  small `Qwen3_5VisionModel` produces the same pooled output for a replicated forward pass
  and concatenated contiguous frame slices, including the single-video frame-split case.
- Qwen3-VL composition and model tests
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+  (`tests/unit_tests/distributed/test_cp_vision_shard_qwen3_vl.py` and
+  `tests/unit_tests/models/qwen3_vl/test_qwen3_vl_cp_preembed.py`, CPU-only): the real
```
jgerh:
```suggestion
  `tests/unit_tests/models/qwen3_vl/test_qwen3_vl_cp_preembed.py`, CPU only) verify that the
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+  Qwen3-VL tower preserves both pooled and DeepStack outputs across image-entry and
+  single-video frame partitions; the production model preserves non-CP Hugging Face parity,
+  uses one mixed image/video vision forward, and shards DeepStack tensors with gradient
```
jgerh:
```suggestion
  single-video frame partitions. The production model preserves non-CP Hugging Face parity,
  uses one mixed image and video vision forward, and shards DeepStack tensors with gradient
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+  coverage in the same token order as the CP sequence.
+- **Real-collective tests** (`tests/unit_tests/distributed/test_cp_vision_shard_gloo.py`,
+  2-rank gloo on CPU): `maybe_distribute_visual` forward + backward through the actual
+  `all_gather` / `reduce_scatter(SUM)` collectives — each rank backprops its sequence shard
+  and the group-summed vision-parameter gradients must equal the single-process replicated
+  reference, for both the balanced-partition path (with deepstack) and the dummy-pad path
+  (padded rank receives exactly zero gradient). A real Qwen3.5 tower test additionally
+  validates forward parity through the production collective path.
+- **Internal validation runs** (Qwen3.5-class VLM): with `cp_size=2`, loss parity with the
```
jgerh:
```suggestion
- Real-collective tests (`tests/unit_tests/distributed/test_cp_vision_shard_gloo.py`,
  two-rank Gloo on CPU) validate the `maybe_distribute_visual` forward and backward passes
  through the `all_gather` and `reduce_scatter(SUM)` collectives. Each rank propagates
  gradients from its sequence shard, and the group-summed vision-parameter gradients must
  equal the single-process replicated reference. The tests cover the balanced-partition
  path with DeepStack and the dummy-pad path, where the padded rank receives exactly zero
  gradient. A Qwen3.5 tower test also validates forward parity through the production
  collective path.
- Internal validation runs with a Qwen3.5-class VLM and `cp_size=2` show loss parity with the
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+  replicated baseline (step-0 loss bit-identical, later steps within bf16 noise), lower peak
+  memory, and **about +19% tokens/sec per GPU**; validated with both a frozen and a trained
+  vision tower using the CP-only group.
```
jgerh:
```suggestion
  memory, and approximately 19% more tokens per second per GPU. The runs covered both a
  frozen vision tower and a trained vision tower using the CP-only group.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+Because the memory saving depends on activation checkpointing (a checkpointed vision tower
+retains only block-boundary activations), the primary win is **compute**: activation
+checkpointing runs the vision forward twice and sharding divides both by ~`cp_size`.
```
jgerh:
```suggestion
Because memory savings depend on activation checkpointing, a checkpointed vision tower
retains only block-boundary activations. The primary benefit is reduced compute. Activation
checkpointing runs the vision forward pass twice, and sharding divides both passes by
approximately `cp_size`.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+## Limitations
+
+<Note>
+**The built-in dense Qwen3.5 and Qwen3-VL callers shard within the CP group only**, whether
```
jgerh:
```suggestion
The built-in dense Qwen3.5 and Qwen3-VL callers shard within the CP group only, whether
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+the tower is frozen or trainable. This is always gradient-correct. The lower-level helper
+permits a wider CP × TP group only for a frozen tower, but the current recipe does not
+publish that wider group.
+</Note>
```
jgerh:
```suggestion

</Note>
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+- **Memory saving is modest under activation checkpointing.** Retained vision activations
+  are already small when the tower is checkpointed; the throughput win is the larger,
+  consistent benefit.
+- **Gathered embeds are full on every rank.** The final per-frame embeddings (small compared
+  to the ViT's internal activations) are reconstructed in full before the existing sequence
+  shard; sharding them too is a separate, more invasive effort.
+- **Model wiring required outside dense Qwen3.5 and dense Qwen3-VL.** The helper is
+  model-agnostic, but each additional VLM family's
+  `embed_multimodal` must route its `visual(...)` call through `maybe_distribute_visual`
```
jgerh:
```suggestion
The following limitations apply:

- Memory savings are modest under activation checkpointing. Retained vision activations
  are already small when the tower is checkpointed. The throughput improvement is the
  larger and more consistent benefit.
- Gathered embeddings are complete on every rank. The final per-frame embeddings are small
  compared to the ViT's internal activations, but they are reconstructed in full before the
  existing sequence shard. Sharding these embeddings is a separate, more invasive effort.
- Model wiring is required outside dense Qwen3.5 and dense Qwen3-VL. The helper is
  model-agnostic, but each additional VLM family's `embed_multimodal` function must route
  its `visual(...)` call through `maybe_distribute_visual`. The recipe must also publish the
  CP group around the pre-embed forward pass.
```

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+- attention is **per-frame** — `cu_seqlens` is built from `grid_thw` so each frame's `h*w`
+  patches attend only within that frame;
+- both the rotary and the learned position embeddings are **spatial-only, repeated
+  identically per frame** (temporal reasoning happens later, in the LLM via mRoPE);
+- the patch embedding is per temporal-patch.
+
+So, with all CP ranks holding identical FSDP-all-gathered vision weights:
+
+```
```
jgerh:
```suggestion
```text
```

jgerh (commented):
Completed tech pubs review of docs/guides/cp-vision-sharding.mdx and provided copyedits and suggested text revisions.

jgerh (commented):
Completed tech pubs review of docs/guides/cp-vision-sharding.mdx and provided copyedits and suggested text revisions.

yuhezhang-ai:
/ok to test [53f6326](https://github.com/NVIDIA-NeMo/Automodel/pull/2990/commits/53f6326a72ae477f9a2d02c29dc8ddced3e45d79)

yuhezhang-ai:
/ok to test [b0c095c](https://github.com/NVIDIA-NeMo/Automodel/pull/2990/commits/b0c095cd3d6cf432dd71dc4ad09fd55d12c1ea29)

akoumpa (dismissed):

examples/vlm_finetune/qwen3/qwen3_vl_8b_cp2_vision_shard.yaml
```diff
@@ -0,0 +1,99 @@
...
+  dp_size: none
+  tp_size: 1
+  cp_size: 2
+  pp_size: 1
+  ep_size: 1
+  sequence_parallel: false
+  activation_checkpointing: true
+  cp_vision_sharding:
+    enabled: true
+    min_tokens: 0
```
akoumpa: I'm wondering if that's the best way to represent this, given that for example, we may want to be able to support hybrid parallelism, like tp + cp for the encoder and cp + pp for the rest of the model. WDYT @yuhezhang-ai ? because also we may have audio / video encoders etc

yuhezhang-ai:
/ok to test [b4a7857](https://github.com/NVIDIA-NeMo/Automodel/pull/2990/commits/b4a7857f16a3c2a91c17bfadb00b6e9a3f4bb718)

examples/vlm_finetune/qwen3/qwen3_vl_8b_cp2_vision_shard.yaml
yuhezhang-ai: Until we have the hybrid parallelism, the current cp based vision frame sharding is still a good feature, and I feel live under `distributed` is reasonable?
It is also only applicable to vision tower where the frames/images are encoded independently. Encoders with cross-frame coupling, and audio encoders, would need separate model-specific handling.  So cp_vision_sharding is narrower than a general encoder-parallelism.

yuhezhang-ai:
/ok to test [72b5e5e](https://github.com/NVIDIA-NeMo/Automodel/pull/2990/commits/72b5e5eb60573e8ae3aec255f698ca0b4dde3c48)

examples/vlm_finetune/qwen3/qwen3_vl_8b_cp2_vision_shard.yaml
yuhezhang-ai: btw I added `supports_cp_vision_sharding` to ModelCapabilities, defaulting to false. Only dense Qwen3.5 and Qwen3-VL opt in here, and unsupported models now fail clearly when the option is enabled. We can design a separate per-submodule topology once hybrid encoder parallelism is implemented.

docs/guides/cp-vision-sharding.mdx
```diff
@@ -0,0 +1,116 @@
...
+Context parallelism (CP) splits a model's language sequence across multiple GPUs. In a
+vision-language model, however, the vision tower normally processes every image and video
+on every CP rank before the language sequence is split. This repeats the same vision
+computation on each rank.
+
+Context-parallel vision sharding distributes independent image and video frames across the
+CP group. Each rank processes only its assigned frames, and the resulting visual embeddings
+are gathered in their original order before the existing language-sequence shard. This
+reduces duplicated vision-tower computation while preserving the model's input ordering and
+gradients.
```
yaoyu-33: Could we clarify the terminology here? This does not apply context-parallel attention inside the vision tower. It shards independent vision frames over the existing CP ranks, gathers their embeddings, and then applies ordinary sequence CP to the multimodal sequence.

yuhezhang-ai:
/ok to test [f5bc718](https://github.com/NVIDIA-NeMo/Automodel/pull/2990/commits/f5bc7180b31111702d7dc4e596fd0837bcf0b180)

docs/guides/cp-vision-sharding.mdx
yuhezhang-ai: Thanks! addressed, and I also rename the feature to `cp_vision_frame_sharding`, hoping it's clearer.

yuhezhang-ai:
/ok to test [0f14f0c](https://github.com/NVIDIA-NeMo/Automodel/pull/2990/commits/0f14f0c1486b78b282865aed121ef71a819e3904)

yuhezhang-ai:
@HuiyingLi  can you help review this PR?

akoumpa (approved):

PR #2840, opened 2026-06-29
fix(distributed): fix and scope vlm activation checkpointing

yuhezhang-ai:
## Overview

This PR fixes inconsistent activation checkpointing behavior for VLMs and adds an explicit `activation_checkpointing_scope` option for users who want to tune the memory/speed tradeoff in the generic FSDP2/DDP path.

The default remains `activation_checkpointing_scope: all`, so users do not need to change configs. With `activation_checkpointing: true`, the generic FSDP2/DDP path checkpoints all extracted trainable language and multimodal layer groups by default. Fully frozen extracted layers are skipped automatically, while LoRA-adapted layers are still checkpointed because the adapter parameters remain trainable.

## Influence Scope

The practical behavior change is concentrated in retrieval-wrapped VLMs and generic FSDP2/DDP VLM configs where the vision path is trainable, including full vision finetuning and vision-side LoRA/adapters. Those cases can now have vision layers discovered and checkpointed under the default `all` scope, which can reduce activation memory but may reduce speed.

Most existing normal VLM finetune examples that enable activation checkpointing freeze the vision tower. For those configs, the generic FSDP2/DDP path now explicitly skips frozen vision layers, so this PR is mostly a unification and overhead-avoidance change rather than a correctness change to their current training behavior.

The VLM example audit is:

- `examples/vlm_finetune/qwen3_5/qwen3_6_27b.yaml` had `freeze_vision_tower: false`, but Huiying confirmed the intent is to freeze the vision tower. This PR corrects the config to `freeze_vision_tower: true` and removes the now-redundant `activation_checkpointing_scope: language` override.
- The dense Qwen3.5 examples `qwen3_5_4b.yaml`, `qwen3_5_9b.yaml`, and `qwen3_5_4b_neat_packing.yaml` keep the vision tower trainable, but they do not enable activation checkpointing, so this PR does not change their current runtime behavior.
- Other normal VLM finetune examples found with `activation_checkpointing: true` set `freeze_vision_tower: true`, so the frozen-layer skip keeps them from recomputing frozen vision towers.
- BAGEL pretrain/SFT keeps vision trainable with activation checkpointing enabled, but BAGEL continues to use its existing model-specific full-layer AC path, which this PR preserves.
- EP MoE VLM benchmark configs use the MoE parallelizer AC path, which checkpoints text/MoE decoder blocks and does not currently use `activation_checkpointing_scope` to include VLM vision towers. With that path, this PR is not expected to materially affect their performance.

[...]

PR #2993, opened 2026-07-09
fix(moe): checkpoint trainable vision towers on the expert-parallel path

yuhezhang-ai:
## Summary

Follow-up to #2840. That PR fixed and scoped VLM activation checkpointing on the generic FSDP2/DDP path, and its docs explicitly noted the gap it left open: expert-parallel MoE configs (`ep_size > 1`) use the MoE parallelizer's own activation checkpointing path, which only checkpoints the text/MoE decoder blocks and never touches the vision tower. For MoE VLMs (e.g. `Qwen3VLMoeForConditionalGeneration`) with a trainable vision tower, every ViT activation is kept, which dominates activation memory on media-heavy batches.

This PR closes that gap with full config parity to the generic path and no new knobs:

1. The MoE `apply_ac` now also checkpoints trainable multimodal tower blocks (vision, and audio once a model registers an audio layer group), in both full and selective modes, reusing the per-model layer-group mapping introduced by #2840.
2. `distributed.activation_checkpointing_scope` is now respected on the expert-parallel path — the same field with the same semantics as the generic FSDP2/DDP path.
3. Recompute safety under ambient SDPA backend forcing is handled automatically: the tower checkpoint wrappers snapshot the SDPA backend set at checkpoint-forward time and restore it for the backward-time recompute.

## Design

**Tower wrapping (same as the generic path).** Trainable multimodal tower blocks get the exact same per-submodule wrapping as the generic FSDP2/DDP path — attention, MLP, and norms wrapped via the existing `apply_submodule_checkpointing` (no new wrapping logic, no divergence between the two paths). This includes towers configured with flash attention: replaying SDPA/flash attention inside checkpoint backward is safe and is what the generic path (and the wider ecosystem) already does.

**Automatic SDPA backend snapshot/restore (no knobs).** Checkpoint replay of SDPA faults if and only if the effective SDPA backend set differs between the forward and the backward-time recompute — e.g. when ambient backend forcing (an `sdpa_kernel` pin or module-level backend toggling) is active at forward time but does not span the recompute; the mismatch surfaces as a deterministic `CheckpointError: Recomputed values ... have different metadata`. The new `sdpa_backend_snapshot_context_fn` (a `context_fn` for non-reentrant `checkpoint_wrapper`, attached to the tower attention/MLP wrappers) snapshots the ambient backend-enabled set (`flash`, `mem_efficient`, `cudnn`, `math` — exactly what `torch.nn.attention.sdpa_kernel` toggles) at checkpoint-region entry and re-pins it for the recompute. Parity holds by construction: a no-op in clean environments, correct under one-sided ambient forcing. (Known caveat: forcing toggled *inside* the region between attention calls is not captured — the snapshot is taken once at region entry.) RNG state is preserved (dropout/DropPath towers redraw the forward's mask on recompute) and the wrapper's default determinism check stays on.

**Scope wiring.** `activation_checkpointing_scope` is threaded from the strategy config through `instantiate_infrastructure` → `parallelize_model` → `apply_ac`, following the existing `activation_checkpointing` threading. The scope decides WHICH groups participate (`all` → decoder blocks plus any trainable multimodal towers, matching the generic path's scope filter; `language` → decoder blocks only; `vision`/`audio`/`multimodal` → only the selected trainable tower blocks, skipping decoder AC entirely — including selective mode); the existing `activation_checkpointing: bool|str` value decides HOW the selected decoder blocks are checkpointed. Scopes irrelevant to the model resolve through the same layer-groups mechanism: groups the model does not expose are simply absent.

## What's included

- `nemo_automodel/components/moe/parallelizer.py`
  - `apply_ac` / `parallelize_model`: new `activation_checkpointing_scope: str | list[str] | tuple[str, ...] = "all"` parameter (normalized via the shared `normalize_activation_checkpointing_scope`); decoder wrapping is gated on the scope, and tower-only scopes skip the decoder path entirely (including its `hidden_size`/`num_experts` derivation).
  - `_apply_multimodal_tower_ac`: scope-driven group selection over the #2840 layer-group mapping (`all` selects vision + audio; `multimodal` expands to both); per-submodule wrapping with the snapshot `context_fn` on attention/MLP wrappers; a cheap trainable-tower detection (vision and audio attribute names) keeps the heavy transformers-aware dense parallelizer import off text-only MoE call paths; skips fully frozen towers (consistent with the generic path).
- `nemo_automodel/components/distributed/activation_checkpointing.py`
  - `sdpa_backend_snapshot_context_fn`: snapshots the ambient SDPA backend-enabled set at checkpoint-region entry and returns `(nullcontext(), sdpa_kernel([captured...]))`.
  - `apply_submodule_checkpointing` / `_wrap_first_existing_attr`: optional keyword-only `context_fn` applied to the attention and MLP wrappers (norm wrappers stay plain — they dispatch no SDPA). Existing callers are unchanged.
- `nemo_automodel/_transformers/infrastructure.py`: threads the strategy config's normalized `activation_checkpointing_scope` into the MoE `parallelize_fn` partial, next to the existing `activation_checkpointing` threading.
- `nemo_automodel/components/distributed/parallelizer.py`: `get_model_layer_groups`, a typed public accessor over the per-model layer-group mapping, so the MoE parallelizer no longer imports the private `_extract_model_layer_groups` across components.
- `docs/guides/gradient-checkpointing.mdx`: the MoE-path note now documents that the scope is respected (same field, same semantics as the generic path) for trainable multimodal towers, with no additional configuration needed.
- Tests:
  - `tests/unit_tests/moe/test_parallelizer_vision_ac.py` (CPU): on a tiny `Qwen3VLMoeForConditionalGeneration`-shaped model, scope parity on the EP path — default/`all` wraps decoder + ViT attention/MLP submodules (full and selective modes, FA-configured towers included), `language` wraps the decoder only, `vision`/`multimodal` wrap only the tower and skip decoder AC entirely (works without `hidden_size`/`num_experts`), `audio` wraps nothing on a vision-only model; frozen towers are skipped; text-only models unchanged.
  - `tests/unit_tests/distributed/test_activation_checkpointing.py`: the snapshot `context_fn`'s recompute context restores exactly the captured flag set (pure flag assertions, CPU-safe); a checkpointed real-SDPA block recomputes under the forward-time backend set after a one-sided ambient pin (flags recorded inside the wrapped attention, CPU-safe); without the `context_fn` the recompute observably diverges; a 1-GPU test proves the failure mode end-to-end — a one-sided ambient pin faults plain checkpoint replay with `CheckpointError` and passes with the snapshot `context_fn`, with gradient parity against an unwrapped reference; a dropout-bearing block keeps gradient parity across recompute (RNG-state preservation).
  - `tests/unit_tests/_transformers/test_infrastructure.py`: `instantiate_infrastructure` threads the normalized scope into the MoE `parallelize_fn` partial.

## Why there is no MATH pin / FA fallback / opt-in spec

Controlled checkpoint-replay experiments across two torch/cuDNN generations (and matching transformers majors) established:

- Unpinned SDPA replay under activation checkpointing is deterministic and safe — including flash attention. The earlier claim that "replaying FlashAttention inside checkpoint backward is not safe" was wrong (this repo already checkpoint-wraps FA-backed attention elsewhere, e.g. the retriever models).
- Replay divergence was reproducible only when ambient SDPA backend forcing applied at forward time did not span the backward-time recompute; the one production fault motivating this work occurred under an expert-parallel MoE stack with module-level backend forcing. Dense/CP recipes are unaffected (their training context spans `loss.backward()`).
- Since replay faults iff the forward/recompute backend sets differ, snapshotting the forward-time set and restoring it at recompute is correct by construction and costs nothing in clean environments — so it needs no configuration surface at all. No `vision_checkpoint_spec`, no pinned backend list, no environment variables.

## Proof / validation

- Unit tests above assert the externally observable contract: scope parity between the EP-MoE and generic paths, and — on GPU — the deterministic `CheckpointError` under one-sided ambient forcing without the snapshot `context_fn` vs. clean replay with gradient parity with it.
- Replay-safety and backend-divergence behavior validated across two torch/cuDNN generations; replay divergence only observed under ambient backend forcing that does not span the recompute.
- Validated in internal large-scale MoE-VLM training runs (up to 128 GPUs) with trainable vision towers under expert parallelism.

Part of #2985.


yuhezhang-ai:
a relevant issue found and fixed: https://github.com/NVIDIA-NeMo/Automodel/pull/3025

yuhezhang-ai:
/ok to test [d1357bd](https://github.com/NVIDIA-NeMo/Automodel/pull/2993/commits/d1357bdd5fecbe608f64f06afce1a0397d7ad17c)

yuhezhang-ai:
/ok to test [19b19a2](https://github.com/NVIDIA-NeMo/Automodel/pull/2993/commits/19b19a2ea2b43c2d48d43cf8b8945eee71afd0fd)

nemo_automodel/_transformers/infrastructure.py
```diff
@@ -347,6 +347,10 @@ def instantiate_infrastructure(
         parallelize_fn = partial(
             parallelize_model,
             activation_checkpointing=activation_checkpointing,
+            # The AC scope lives on the strategy config (normalized in its
+            # __post_init__); thread it through so expert-parallel configs keep
+            # scope parity with the generic FSDP2/DDP path.
+            activation_checkpointing_scope=getattr(distributed_config, "activation_checkpointing_scope", "all"),
```
akoumpa, line 353: 🤔 @yuhezhang-ai i'm wondering if we need getattr here or could do without?

yuhezhang-ai: The getattr is intentional and I'd keep it: `MegatronFSDPConfig` deliberately doesn't define `activation_checkpointing_scope` because MegatronFSDP doesn't support activation checkpointing at all (`megatron_fsdp.py` logs "not yet supported… Skipping"). Since this branch runs for any strategy when ep_size > 1, the getattr(..., "all") lets the expert-parallel parallelizer — which does honor scope — default cleanly when the strategy config is MegatronFSDP, without adding a scope field to a config that wouldn't act on it. It also mirrors the `getattr(distributed_config, "activation_checkpointing", False)` two lines up.

akoumpa (approved):

PR #2039, opened 2026-04-24
feat: DeepSeek V4 Flash support

khazic:
## Status — DeepSeek V4 Flash full finetuning is now working ✅

End-to-end **full finetuning** of DeepSeek V4 Flash trains correctly on this branch. The forward-numerics gap that remained earlier in this PR is closed — five independent layer-parity bugs were uncovered by per-tensor dump bisection vs the DeepSeek inference reference (`dsv4flash/inference/model.py`) and all five are now fixed.

The loss curve at the bottom of this PR is from a real **43-layer full-finetune run** with the full attention zoo (SWA + CSA + HCA) live — not the 4-layer smoke.

Many thanks to **@HuiyingLi** for the strong support landing this — the layer-parity work, and fixes to RoPE / HC / CSA / PP stage handling are what got this over the finish line.

## Bugs fixed (post-bringup)

1. **RoPE convention** — released DSV4-Flash uses INTERLEAVED-pair RoPE (`view_as_complex` on `(2k, 2k+1)` packing); HF transformers PR 45616 reused Llama-style `rotate_half` (pairs `(d, d+rd/2)`), which is the wrong dim-to-frequency mapping for these weights. Fixed via `_apply_partial_rope_interleaved`. *Effect:* `kv_post_rope` cos 0.866 → 0.999 after one block.
2. **Hyper-Connections `post` / `comb` formulas** — `post` needs `2 * sigmoid(...)` (no `+eps`); the trained weights expect range `(0, 2)`, not `(eps, 1+eps)`. `comb` needs `softmax(dim=-1) + eps` followed by col-norm-first Sinkhorn (`iters - 1` alternating row/col passes); HF PR 45616's sigmoid+sinkhorn variant converges to a different distribution.
3. **Dual RoPE base on main attention** — V4 has two RoPE bases (`theta=10000` for `compress_ratio==0`, `theta=160000` with YaRN for `compress_ratio!=0`), and the compress-rope must apply to the **main** attention Q/KV on `compress_ratio>0` layers, not just the compressor sub-module. Reference: `dsv4flash/inference/model.py:476-501` builds `freqs_cis` with `compress_rope_theta` whenever `compress_ratio != 0`.
[...]

PR #2994, opened 2026-07-09
ci(review): strengthen model review invariants

akoumpa:
## Summary

- require checkpoint-free model paths to initialize every active parameter, persistent buffer, and routing table before first use
- require model-specific logic to live under `nemo_automodel/components/models/<model>/` and classify violations elsewhere as critical findings
- standardize tensor-layout documentation on forms such as `hidden_states: Tensor of shape [batch, sequence, hidden].`, allow conventional dimension names without definitions, and require a compliant rewrite example in review findings
- require focused no-checkpoint state coverage and a finite first forward/backward on a representative backend or topology

## Why

PR #2991 exposed a gap in checkpoint-free initialization review: DeepSeek-V4 random-init training left HyperConnection tensors uninitialized and left the HashGate route table all zeros, which violated DeepEP routing invariants and produced a NaN gradient norm on the first backward pass.

The review prompt also lacked a strict ownership gate for model-specific behavior and prescribed symbolic tensor layouts without a single reusable documentation format. These additions make model boundaries and tensor contracts concrete and actionable.

## Developer impact

Model-specific architecture, initialization, weight-layout, and parallelization policy must stay with the owning model package. Tensor docstrings use a consistent semantic shape form, while review findings for unclear layouts must show an acceptable rewrite. Model changes that add parameters, buffers, routing tables, or random-init paths must demonstrate valid checkpoint-free initialization.

## Validation

- `git diff --check -- .github/workflows/claude-review.yml`
- parsed `.github/workflows/claude-review.yml` with `yaml.safe_load`
- asserted that the rendered prompt contains the model ownership path, critical severity, common tensor-layout example, and required review rewrite example

Follow-up to #2991 and NVBug 6317402.

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -107,6 +107,16 @@ jobs:
           - ambiguous public signatures: boolean-trap parameters, multiple
             same-typed positional runtime arguments that should be keyword-only,
             or long anonymous tuple returns that need a named typed result.
+        - Model-specific ownership is mandatory. Model-specific logic may live
+          only under `nemo_automodel/components/models/<model>/`. Treat logic as
+          model-specific when it names or branches on a model or family, hard-codes
+          its module/parameter paths or weight layout, or implements its
+          architecture, initialization, or parallelization policy. If changed code
+          adds or expands such logic anywhere else, including another directory
+          under `nemo_automodel/components/` or a recipe, report a critical finding
+          and require relocation to the owning model package. Shared components may
+          expose generic typed contracts and invoke model-owned hooks, but must not
+          encode model identities or model-specific policy.
         - Tensor contract docstrings. For every new function or method, or any
           function/method whose signature or tensor-handling body is materially
           changed, that accepts a tensor input, require its docstring to document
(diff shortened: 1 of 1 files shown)


akoumpa:
### Blind review-prompt validation

I ran the updated `.github/workflows/claude-review.yml` prompt in a fresh subagent without providing the known bug root cause or fix. The subagent reviewed the complete 26-file, non-`uv.lock` DeepSeek-V4 onboarding diff (`17ed5796^..17ed5796`, PR #2039) and inspected the corresponding code on current `origin/main` (`1197b628`).

The prompt produced two **High** findings:

1. **Checkpoint-free initialization leaves backbone and MTP HyperConnection parameters uninitialized.**
   - `DeepseekV4HyperConnection` and `DeepseekV4HyperHead` allocate parameters with `torch.empty`.
   - The production initialization path does not initialize those tensors in the backbone, final head, or MTP blocks.
   - A small CPU probe that constructed the model without checkpoint loading and invoked the production initializer reported:

     ```text
     attn_hc.fn finite=False
     logits finite=False
     loss=nan
     gradients finite=False
     ```

2. **Checkpoint-free HashGate routing initializes every top-k slot to expert 0.**
   - `tid2eid` is created as zeros and reset to zeros by `init_weights()`.
   - With `topk=2`, the production initializer probe returned:

     ```text
     [[0, 0], [0, 0], [0, 0]]
     ```

   - This violates the downstream distinct-experts-per-token routing requirement used by DeepEP. The existing test checked only the route dtype, not range or uniqueness.

**Result:** the new checkpoint-free initialization guideline flags both root causes fixed by #2991: active `torch.empty` state that is never initialized and an all-zero routing placeholder that violates the dispatcher's structural contract.

This was a faithful local subagent simulation of the review prompt, not a hosted GitHub Actions invocation.

HuiyingLi (approved):

PR #2996, opened 2026-07-09
fix: bound chunked CE memory and preserve packing masks

yuhezhang-ai:
## Summary

This PR keeps two independently testable fixes from the Amber upstream work:

1. **Bound the saved-activation memory of the existing** `ChunkedCrossEntropy`**.**  <br>The sum-reduced path now saves only original-dtype logits and recomputes each fp32 softmax chunk during backward. This fixes the gap in the existing implementation, whose forward was chunked but whose autograd graph retained all fp32 chunk activations until backward.
2. **Preserve indexed packing masks for Flash Attention 2.**  <br>Transformers 5.x mask preprocessing treats 2D attention masks as binary padding masks and can coerce AutoModel's integer per-document masks to bool, erasing packed-document boundaries. `configure_packing()` now preserves indexed masks through a generic FA2 compatibility shim.

The previously proposed `non_moe` / `non_moe_no_attn` activation-checkpointing modes and the duplicate `MaskedCrossEntropy(chunk_size=...)` API have been removed.

## Implementation

* `ChunkedCrossEntropy(reduction="sum")` uses a private custom autograd function:
  * forward upcasts only one `[chunk_len, vocab]` slice at a time;
  * backward recomputes fp32 softmax per chunk;
  * only original-dtype logits and integer/bool metadata are saved;
  * logits are never mutated;
  * unsupported higher-order gradients fail explicitly.
* Existing `chunk_len` remains the only chunking configuration surface.
* `compile=False` now correctly bypasses `torch.compile` on the legacy non-sum path.
* The packing compatibility shim is limited to FA2 and actual indexed masks:
  * a constant `[1, 1, 1, 1]` probe validates the installed private Transformers early-exit contract at configuration time;
  * the complete probed return tuple is preserved instead of guessing its arity or trailing fields;
  * ordinary integer 0/1 padding masks still delegate to Transformers;
  * an incompatible private API raises before the remaining packing globals are installed, preventing silent cross-document attention.

## Validation

* `331 passed, 2 skipped` across focused loss, packing, distributed, MoE, and recipe regression tests.
* `27 passed, 1 skipped` across the final packing compatibility and adjacent FlashAttention packing tests.
* `11 passed, 1 skipped` in the final chunked-CE test file after cleanup.
* `ruff format .`
* `ruff check --fix .`


nemo_automodel/components/loss/masked_ce.py
```diff
@@ -19,10 +19,125 @@
...
 
+def _validate_chunk_size(chunk_size: int) -> int:
+    """Validate that ``chunk_size`` is a positive integer and return it as ``int``."""
+    chunk_size = int(chunk_size)
+    if chunk_size <= 0:
+        raise ValueError(f"chunk_size must be greater than zero; got {chunk_size}.")
+    return chunk_size
+
+
+class _ChunkedMaskedCESum(torch.autograd.Function):
```
akoumpa: Hi @yuhezhang-ai  we already have a chunked ce, maybe you want to update it instead?

https://github.com/NVIDIA-NeMo/Automodel/blob/e2d117f24fea5ecdbc35c1788b2ae22b0bc8b233/nemo_automodel/components/loss/chunked_ce.py

yuhezhang-ai:
/ok to test [0efbb7d](https://github.com/NVIDIA-NeMo/Automodel/pull/2996/commits/0efbb7dc60a3f57f7631f13a77d86e7fe937bfb1)

akoumpa (approved):

PR #2997, opened 2026-07-09
fix(optim): drop zero-numel local DTensor shards before TE FusedAdam

yuhezhang-ai:
## Summary

TransformerEngine FusedAdam's `multi_tensor_apply` kernel has no empty-tensor
guard: feeding it a zero-numel parameter faults with a CUDA "misaligned
address" / "illegal memory access" at the first optimizer step. FSDP2 produces
exactly such parameters as a matter of course — this PR filters them out of
the param list / param groups before TE ever sees them, on both TE FusedAdam
construction paths:

- the typed `FusedAdamConfig._build_optimizer` (registry name `fused_adam`),
- the YAML `_target_: transformer_engine.pytorch.optimizers.fused_adam.FusedAdam`
  factory escape hatch (`OptimizerFromFactoryConfig`), which is what every
  example recipe currently uses.

The filter is exact, not an approximation: a parameter whose local shard is
empty has **all** of its elements resident on other ranks, whose optimizers
update them. This rank simply has nothing to do for that tensor, so dropping
it from the local optimizer changes no update anywhere. A rank-local warning
is logged with the number of dropped shards.

## Motivation: zero-numel shards from small dense towers on wide meshes

FSDP2 shards every parameter tensor independently along dim-0 across the
shard group. Any tensor whose dim-0 is smaller than the shard-group size
leaves **zero-numel local shards on the tail ranks**. LLM-only training never
hits this (LM tensor dim-0 >= hidden size >> shard group), which is why the
bug hides until a VLM run unfreezes a small dense vision tower on a wide
mesh: ViT biases, LayerNorm weights, class/position embeddings, and conv
patch kernels all have small dim-0. At (say) 128-way sharding, every such
tensor is empty on most ranks, and the first `optimizer.step()` kills the job
deterministically on the tail rank(s) with an unhelpful CUDA fault deep in
`multi_tensor_apply.cuh`. Freezing the tower "fixes" it, which makes the
failure look spurious.

The same applies to any small trainable module (adapters, small heads,
retrieval towers) sharded over a wide mesh — the guard is not VLM-specific,
which is why it is named after what it actually does (dropping zero-numel
local DTensor shards) rather than after VLM.

## What's included

- `nemo_automodel/components/optim/optimizer.py`
  - `_local_numel(param)` — rank-local element count (`to_local().numel()`
    for DTensor, `numel()` otherwise).
  - `_drop_empty_local_shards(params)` — filters flat param lists and
    param-group dicts; drops groups that become empty; preserves group
    options; warns with the dropped count.
  - `_is_te_fused_adam(factory)` — identity-based detection of TE's
    `FusedAdam` for the factory escape hatch. Import-free: it only consults
    `sys.modules`, so the optional TE dependency is never imported by the
    check itself (if TE was never imported, the factory cannot be TE's class).
  - Wired into `FusedAdamConfig._build_optimizer` and into
    `OptimizerFromFactoryConfig.build` / `build_from_param_groups` (TE
    FusedAdam factories only — other factories, e.g. `torch.optim`, are
    untouched; torch handles empty tensors fine).
- `tests/unit_tests/optim/test_fused_adam_empty_local_shards.py`
  - CPU tests (no TE required — TE is stubbed via `sys.modules`, so the tests
    observe exactly which parameters reach the TE constructor):
    - flat-list filtering + warning on `FusedAdamConfig`;
    - param-group filtering with group options preserved and empty groups
      dropped (`build_from_param_groups`);
    - a real CPU `DTensor` (single-rank gloo mesh) with **non-empty global
      shape but empty local shard** proving the check uses local, not global,
      numel;
    - factory-escape-hatch filtering for a TE-FusedAdam factory, and a
      negative test that non-TE factories (`torch.optim.SGD`) are not
      filtered.
  - 1-GPU test (skipped without CUDA, `importorskip("transformer_engine")`):
    a real TE FusedAdam step succeeds with a zero-numel parameter present,
    and the surviving parameter is actually updated and finite.

## Proof

- The new unit tests are CPU/1-GPU runnable in CI (`pytest
  tests/unit_tests/optim/test_fused_adam_empty_local_shards.py -q`): they
  assert, on both construction paths, that zero-numel local shards never
  reach the TE constructor (including a real CPU `DTensor` whose global shape
  is non-empty but whose local shard is empty), that group options survive
  filtering, that non-TE factories are untouched, and (1 GPU) that a real TE
  FusedAdam step over the filtered params updates the surviving parameter.
- The single-GPU distillation of the fault is simply a zero-numel parameter
  in TE FusedAdam's param list — exactly the local shard FSDP2 leaves on tail
  ranks. On the TE version in the affected production stack this faults
  deterministically inside `multi_tensor_apply` (`CUDA_LAUNCH_BLOCKING=1`
  pins it in `multi_tensor_apply.cuh`); on TE 2.11 nightly the same input no
  longer faults (see Notes), which is precisely why Automodel should not
  depend on any particular TE version's tolerance of empty tensors.
- At scale, multi-hundred-GPU VLM runs with an unfrozen dense vision tower
  fail at the first optimizer step without this guard and train normally with
  it; frozen-tower runs are immune, consistent with the dim-0 analysis.

## Relationship to TransformerEngine

This is an Automodel-side fix and is intentionally **independent of any
TE-side fix**: skipping locally-empty shards in the optimizer is exact,
avoids pointless kernel launches for tensors this rank does not own, and
protects every TE version users may run (TE's `multi_tensor_apply` chunking
has no *explicit* empty-tensor guard on `main`; empty tensors are only
skipped as an arithmetic side effect of contributing zero chunks).

The fault is confirmed live on a current TE build and root-caused —
**NVIDIA/TransformerEngine#3202**: `multi_tensor_apply`'s "struct full" flush
is only checked inside the per-chunk loop, so a zero-numel tensor landing in
the last `TensorListMetadata` slot skips the flush and subsequent tensors
write their sizes/addresses out of bounds within the kernel-arg struct. With
enough trailing tensors this crashes (CUDA illegal memory access / misaligned
address — reproduced deterministically in 4-GPU and 128-GPU FSDP2 trainings
and in a single-GPU standalone script); with fewer, the metadata corruption
is **silent** and produces wrong optimizer updates with no error. Whether a
given run crashes, silently corrupts, or works depends only on the
numel-sequence-dependent slot position of the empty shard. That is why this
guard rejects the input at the optimizer boundary rather than relying on TE's
implicit handling: it prevents both failure modes on every TE version, and
dropping the empty local shards is mathematically exact. This PR does not
block on the TE-side fix.

Part of #2985.

### DIFF
--- a/nemo_automodel/components/optim/optimizer.py
+++ b/nemo_automodel/components/optim/optimizer.py
@@ -38,6 +38,7 @@ Megatron-FSDP sharding).  Subclasses only implement the small
 
 from __future__ import annotations
 
+import functools
 import importlib
 import inspect
 import logging
@@ -569,22 +570,47 @@ def _drop_empty_local_shards(params: list[Any]) -> list[Any]:
             dim-0-sharded local shard may be empty on this rank.
 
     Returns:
-        ``params`` with zero-numel local shards removed.  Param groups whose
-        parameter list becomes empty are dropped; the remaining groups keep
+        ``params`` with zero-numel local shards removed.  Param groups keep
         their other options unchanged.
+
+    Raises:
+        ValueError: If every parameter of the flat list — or of any single
+            param group — is locally empty on this rank.  Neither outcome has
+            a safe representation: dropping a whole group makes
+            ``optimizer.param_groups`` rank-asymmetric (LR/WD schedulers
+            address groups positionally, e.g. ``param_groups[0]``, so ranks
+            would silently schedule different lr/wd for shards other ranks
+            own), while keeping an empty group breaks torch DCP's flattened
+            optimizer-state load, which indexes the first param of each group.
     """
+    remedy = (
+        "This happens when FSDP2 shards parameters whose dim-0 is smaller than the shard group "
+        "(e.g. tiny biases/norm weights of a small module on a wide mesh). Merge such parameters "
+        "into a group that keeps at least one non-empty local shard on every rank, or shrink the "
+        "sharding mesh so every rank owns at least one element."
+    )
     params = list(params)
     dropped = 0
     if params and isinstance(params[0], dict):
         filtered: list[Any] = []
-        for group in params:
+        for idx, group in enumerate(params):
             kept = [p for p in group["params"] if _local_numel(p) > 0]
             dropped += len(group["params"]) - len(kept)
-            if kept:
-                filtered.append({**group, "params": kept})
+            if not kept:
+                raise ValueError(
+                    f"Every parameter in optimizer param group {idx} has a zero-numel local shard on "
+                    f"this rank; TE FusedAdam cannot hold them (multi_tensor_apply faults on empty "
+                    f"tensors) and the group can be neither dropped nor kept empty. {remedy}"
+                )
+            filtered.append({**group, "params": kept})
     else:
         filtered = [p for p in params if _local_numel(p) > 0]
         dropped = len(params) - len(filtered)
+        if params and not filtered:
+            raise ValueError(
+                f"Every trainable parameter of this model part has a zero-numel local shard on this "
+                f"rank, leaving TE FusedAdam with an empty parameter list. {remedy}"
+            )
     if dropped:
         logger.warning(
             "Dropped %d parameter(s) with zero-numel local shards from TE FusedAdam on this rank: "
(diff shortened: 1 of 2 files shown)


yuhezhang-ai:
hi @akoumpa , this one is ready too, can you please help review

hemildesai (approved):

PR #2998, opened 2026-07-09
refactor(models): single tie_word_embeddings guard via TieSupport

### LINKED ISSUE #2935: refactor(models): single tie_word_embeddings guard with per-class declared TieSupport

Achyuthan-S:
# What does this PR do ?

Replaces the two-helper `tie_word_embeddings` guard API with a single per-class  <br>declaration + one constructor guard, per the design agreed in NVIDIA-NeMo/Automodel#2935.

Each registered model that builds an `lm_head` declares a `TieSupport` policy  <br>(`BOTH` / `TIED_ONLY` / `UNTIED_ONLY`) as a class attribute; a single  <br>`reject_unsupported_tie_word_embeddings(model_cls, config)` reads that policy and  <br>the controlling flag (`get_controlling_tie_word_embeddings`) and rejects an  <br>unsupported request at construction, so the per-call-site choice between two  <br>helpers — easy to get wrong for composite configs — no longer exists.

Opening as **Draft**: the layers land as separate commits (agreed in NVIDIA-NeMo/Automodel#2935).

# Changelog

**Landed (commit 1 — mechanical, no behavior change):**

* Add `TieSupport` enum + `reject_unsupported_tie_word_embeddings(model_cls, config)`  <br>in `checkpoint/utils.py`; remove `reject_unsupported_tied_word_embeddings` and  <br>`reject_unsupported_untied_word_embeddings`.
* Migrate the 30 existing guard call sites 1:1 — 29 `UNTIED_ONLY`, gemma4_moe  <br>`TIED_ONLY` — via a per-class `tie_word_embeddings_support` declaration.
* Port the checkpoint-utils and qwen3_moe guard tests to the new guard; add  <br>`BOTH` / undeclared-default coverage.

**Remaining (subsequent commits in this PR):**

- [ ] Declarations per the audit: demote baichuan / mistral3 / bagel →  <br>`UNTIED_ONLY`; add ernie4_5 / diffusion_gemma / gemma4_drafter →  <br>`TIED_ONLY`; promote llama / qwen2 / qwen3_5 → `BOTH`.
- [ ] bagel: read the controlling nested (`text_config`/`llm_config`) value; wire  <br>the guard into `build_bagel_from_hf_backbones` (registry bypass).
- [ ] Registry-iterating test with an explicit exempt list (retrieval encoders,  <br>VLM containers, …), one comment per exemption.
- [ ] Demotion test debt: flip ernie4_5 untied tests → rejection; baichuan  <br>tie=True rejection + drop the unconditional `_tied_weights_keys`; add qwen2  <br>untied test; add qwen3_5 tied+untied tests.
- [ ] BOTH gaps: qwen2 `tie_weights()` override; qwen3_5 post-swap re-tie verify.
- [ ] llama micro-cleanup: align the adapter's  <br>`getattr(config, "tie_word_embeddings", True)` with the `False` default in  <br>`__init__`/`tie_weights`.
- [ ] Layer 2: `NeMoAuto*.from_pretrained` symmetric flip guard (raise on any  <br>mismatch between the checkpoint's raw `tie_word_embeddings` and the  <br>requested value, both directions); reword the PP validator at  <br>`pipelining/hf_utils.py:742`.
- [ ] Docs: onboarding skill §2.3 + checklist require the declaration, and  <br>instruct per-model `from_pretrained` implementers to call the shared flip  <br>check.

**Pre checks**:

# Additional Information

* Closes NVIDIA-NeMo/Automodel#2935
* Builds on merged NVIDIA-NeMo/Automodel#2896 (the two-helper API this replaces) and NVIDIA-NeMo/Automodel#2942 (the  <br>`_tie_weights_nemo` gate the ernie4_5 demotion lines up with).

cc @yuhezhang-ai


yuhezhang-ai:
Thanks for already tracking the Qwen3.5 post-swap re-tie item. One concrete data point: Qwen2 and `Qwen3_5ForCausalLM` already alias/separate correctly in a tiny tied/untied construction probe. The remaining failure I see is specifically `Qwen3_5ForConditionalGeneration` with `tie_word_embeddings=True`: after replacing `self.model.language_model` with `Qwen3_5DenseTextBackbone`, `lm_head.weight` is no longer aliased to `model.language_model.embed_tokens.weight`. Calling `tie_weights()` after the swap fixes it.

The supported dense Qwen3.5 VLM configs are mixed by size (`0.8B`/`2B`/`4B` tied, `9B`/`27B`/`Qwen3.6-27B` untied), so I think this class should land as `TieSupport.BOTH` with tiny tied/untied alias tests.

yuhezhang-ai:
For the final `BOTH` declarations, can we keep the model-onboarding contract consistent: explicit `_tied_weights_keys`, explicit `tie_weights()` using the actual local FQNs, and tiny tied/untied alias tests? Qwen2 currently behaves correctly via the HF base implementation, so this is not fixing a known runtime bug, but making the declared BOTH support robust and self-documenting.

Could you also include a small VLM audit table before marking this ready? The risky pattern is HF parent init tying weights, followed by replacing `language_model` / `lm_head`. Gemma4 already re-ties after the swap; Qwen3.5 dense VLM still needs that. Most other supported custom VLM examples look untied-only, but Qwen3.5 dense VLM and Mistral3FP8VLM are mixed/BOTH.

Achyuthan-S:
Thanks — agreed on both, and I ran the swap/re-tie audit across the custom VLMs.

Confirmed your Qwen3.5 finding: `Qwen3_5ForConditionalGeneration` swaps in
`Qwen3_5DenseTextBackbone` (model.py:845) with no re-tie, so tie=True orphans the
alias; the `tie_weights()` at :670 belongs to the `ForCausalLM` sibling. I'll land
this class as BOTH with a `tie_weights()` that re-ties after the swap, plus tiny
tied/untied alias tests.

VLM audit table (the risky pattern = HF parent ties on init, then language_model/
lm_head is replaced):

| Class | HF parent ties | Replaces LM/head | Re-ties after | Verdict |
|---|---|---|---|---|
| Mistral3FP8VLMForConditionalGeneration | yes | no swap | n/a (HF tie holds) | BOTH |
| Qwen3_5ForConditionalGeneration | yes | yes (:845) | NO → adding | BOTH |
| Qwen3_5ForCausalLM | no | yes + own head | yes (:670) | BOTH |
| Gemma4ForConditionalGeneration | yes | yes (:1021) | yes (:1050) | TIED_ONLY |
| Qwen3VLMoeForConditionalGeneration | yes | yes + own head | no | UNTIED_ONLY |
| Qwen3_5MoeForConditionalGeneration | yes | yes + own head | no | UNTIED_ONLY |
| MiniMaxM3 / KimiVL / KimiK25VL / LLaVAOneVision / NemotronOmni / NemotronParse / Step3p7 | no | own separate head | n/a | UNTIED_ONLY |
| Qwen2_5Omni / Qwen3OmniMoe thinkers | HF thinker | no re-tie | n/a | UNTIED_ONLY |

Every UNTIED_ONLY VLM is safe because the constructor guard rejects tie=True
before the swap can orphan the head; gemma4 already re-ties; mistral3_vlm doesn't
swap so HF's tie holds.

On the contract: I'll add explicit `_tied_weights_keys` + `tie_weights()` (local
FQNs) + tied/untied tests to qwen2, Qwen3_5ForCausalLM, and
Qwen3_5ForConditionalGeneration. For Mistral3FP8VLM — since it doesn't swap and
uses HF's inherited tie — do you want an explicit local override there too, or is
the inherited contract + the alias tests I added sufficient?

yuhezhang-ai:
For `Mistral3FP8VLMForConditionalGeneration`, I think the inherited HF `tie_weights()` plus the tied/untied alias tests are sufficient. The explicit local override is important for our custom classes that own or swap the embedding/head path, like Qwen2 and Qwen3.5, especially `Qwen3_5ForConditionalGeneration` after the language-model replacement.

Since Mistral3FP8VLM does not swap `language_model`/`lm_head`, a local override would just duplicate HF’s contract and may be more brittle than useful. The alias tests are the right proof here.
Thanks!

yuhezhang-ai:
btw, after the tasks are done, on unit test side, Can we make the TieSupport declaration mandatory via the registry-iterating test? The helper currently defaults undeclared classes to BOTH, so review/ci test is the only thing catching a missing declaration. I think each registered class that owns/builds an lm_head should explicitly declare TieSupport, with an explicit commented exemption list for classes that do not own the language head / are not CausalLM-style.

Achyuthan-S:
Sounds good — keeping Mistral3FP8VLM on HF's inherited tie_weights + the alias
tests, no local override.

And agreed on making the declaration mandatory: the registry-iterating test will
require every MODEL_ARCH_MAPPING class that owns/builds an lm_head to declare
TieSupport, with an explicit commented exemption list for the non-CausalLM /
no-head classes (retrieval encoders, VLM containers, …). I'll add that as its own
commit so the undeclared→BOTH default can never silently hide a missing verdict.

yuhezhang-ai:
@Achyuthan-S Thanks for the latest update. Since this PR includes correctness fixes we’d like to land soon (Mistral3FP8VLM BOTH and Qwen3.5 post-swap re-tie), do you think you’ll have time to finish the remaining checklist items shortly?

The main remaining pieces seem to be Layer 2 `from_pretrained` flip guard / PP validator wording, Bagel handling, the registry TieSupport declaration test, the remaining TIED_ONLY declarations/tests, and docs. If you’re busy, I’m happy to help take over so we can get the correctness fixes merged faster.

Achyuthan-S:
Thanks @yuhezhang-ai ! No need to take over — I've pushed the rest of the checklist:

- Remaining verdicts: ernie4_5 (dense+MoE) / diffusion_gemma / gemma4_drafter →
  TIED_ONLY (ernie untied tests flipped to rejection); bagel → UNTIED_ONLY (guard
  in __init__ covers the build_bagel_from_hf_backbones bypass; resolver already
  reads the nested text_config). The registry enumeration also surfaced mistral4's
  Mistral3ForConditionalGeneration (head lives in the Mistral4 backbone) — declared
  UNTIED_ONLY, guard reads config.text_config since the composite Mistral3Config
  top-level defaults tie=True.
- Registry test: every MODEL_ARCH_MAPPING lm_head class must declare TieSupport,
  with a commented exempt list (retrieval encoders) + a stale-entry check.
- Layer 2: reject_tie_word_embeddings_flip in NeMoAuto*.from_pretrained (symmetric,
  no materialization, conservative — re-reads the raw config, only fires on a real
  flip); PP validator reworded.
- Onboarding §2.3 + checklist updated (declaration + the flip check for the ~33
  from_pretrained classmethods that bypass the bridge).

479 green across the tie suite. Two flagged for review: diffusion_gemma/gemma4_drafter
construction is fork-gated so I leaned on the registry + guard unit tests, and the
from_pretrained flip integration is the piece most worth a close look. Ready for review.

nemo_automodel/_transformers/auto_model.py
```diff
@@ -254,6 +255,37 @@ def _maybe_dequantize_fp8_for_peft(hf_native_quant_cfg, peft_config, pretrained_
...
+    :func:`reject_tie_word_embeddings_flip`. Conservative by design: only runs for a
+    string checkpoint path and silently returns if the raw config cannot be re-read, so
+    it never blocks a load except on a genuine flip.
+
+    Args:
+        pretrained_model_name_or_path: The from_pretrained source (only str paths checked).
+        hf_config: The resolved config with user overrides applied (the requested value).
+        kwargs: The from_pretrained kwargs (hub-locating keys are reused for the raw load).
+    """
+    if not isinstance(pretrained_model_name_or_path, str):
```
yuhezhang-ai, line 277: This skips the new flip guard for local checkpoints passed as `pathlib.Path`, even though `from_pretrained` accepts `str | os.PathLike`. In that case a user can still flip `tie_word_embeddings` silently. Can we normalize path-like inputs with `os.fspath()` and run the same raw-config check for both `str` and `os.PathLike` sources?

Achyuthan-S: Oh yes you are right , good catch , thank you — fixed  . Normalized path-like sources with os.fspath() so the raw-config flip check runs for both str and os.PathLike, and added pathlib.Path coverage (rejected flip, matching value, and an assertion that the Path is normalized to str before the raw read).

yuhezhang-ai:
/ok to test [09cee09](https://github.com/NVIDIA-NeMo/Automodel/pull/2998/commits/09cee0972ca73cb4a3d9f4013716d071c009b2f1)

yuhezhang-ai:
/ok to test [484420c](https://github.com/NVIDIA-NeMo/Automodel/pull/2998/commits/484420ce484bb88903cae800884ad4b5a8140517)

yuhezhang-ai:
Thanks @Achyuthan-S for driving this through and for the thorough model audit and follow-up fixes. I only added a small maintainer follow-up to harden and simplify the model-onboarding skill guidance, make the explicit tie-weight contract consistent, and clean up the helper ownership so import-linter is satisfied. Focused local checks are green (import-linter, Ruff, and 74 unit tests). Triggering CI on the latest head now.

yuhezhang-ai:
/ok to test [4b0683d](https://github.com/NVIDIA-NeMo/Automodel/pull/2998/commits/4b0683d66c4c273a82cd60b486b1ea8e81ae9ff2)

yuhezhang-ai:
Hi @akoumpa , could you help run `/nvskills-ci` here? My request was skipped because I don’t have admin permission.

Also, if you want to take a look, this PR makes tied-embedding handling more robust by requiring models with an lm_head to explicitly declare their tie support and define tie_weights() when they own the embedding/head relationship. It also reduces reliance on ambiguous nested versus top-level config fields, particularly for VLMs.

Scoped CI:
[Ministral3 AM-657 coverage](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/57981639): full fine-tuning and PEFT passed; the later allowed checkpoint-robustness failure is an unrelated offline-tokenizer issue.
[Qwen3.5 and Mistral4 VLM coverage](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/57981666): both jobs passed.

yuhezhang-ai:
/ok to test [b965726](https://github.com/NVIDIA-NeMo/Automodel/pull/2998/commits/b965726a152809059e3ecbf3d6019fc496effc5e)

Achyuthan-S:
Thanks for pushing this over the line, @yuhezhang-ai — the runtime tie-support wiring and the extra tie_weights() hooks are nice hardening on top of the guard. CI results look great (AM-657 FT+PEFT, Qwen3.5, Mistral4 VLM all green). Happy with the refinements — nothing further from me; ready whenever the code-owner review lands.

yuhezhang-ai (approved):

PR #2999, opened 2026-07-09
fix(distributed): support packed CP for Llama, Qwen2, and Qwen3

### LINKED ISSUE #1473: packing + CP
Hi. I would like to ask if the packing + CP is already planned as it is not possible [yet](https://github.com/NVIDIA-NeMo/Automodel/blob/bb0c0bfb08798daec2d56e0b271ddacd4d43afed/nemo_automodel/_transformers/kernel_patches.py#L188).

**Is your feature request related to a problem? Please describe.**
Packing is a way of how to test or fill data for long context training. It seems that long context training with limited number of nodes is not possible yet.

**Describe the solution you'd like**
It would be amazing to combine packing process defined here with CP to enable long context training.

**Describe alternatives you've considered**
I tried combinations without CP but was not successful. The `MegatronPretraining` under the dataset can create long-context samples but it seems not supported by the CP>1.

**Additional context**

Thank you very much! I did lots of experiments with NeMo2 and the separation into Automodel for pre-training and SFT is great idea.

Ondřej

akoumpa:
# What does this PR do ?

Adds packed THD context parallelism for the dense Llama, Qwen2, and Qwen3 implementations.

Dense models previously kept the HuggingFace BSHD attention path even when packed batches were converted to THD, and their Transformer Engine attention modules were not bound to the context-parallel process group. This caused packed document boundaries and CP metadata to be ignored. Dense Qwen3 additionally had no custom TE path that preserved its per-head Q/K RMSNorm.

# Changelog

* Add Transformer Engine THD attention paths for dense Llama, Qwen2, and Qwen3 while preserving their existing BSHD behavior.
* Preserve Qwen3 per-head Q/K RMSNorm and verify its BSHD logits and parameter gradients against HuggingFace.
* Configure dense Transformer Engine attention modules with the CP process group and communication mode.
* Support fused RoPE for packed THD tensors under context parallelism.
* Mark all three model families as CP-capable and preserve Qwen3's optimized TP plan.
* Add runnable packed CP=2 configs for Llama 3.2 1B, Qwen2.5 0.5B, and Qwen3 0.6B.
* Add CPU packed-versus-per-document forward/backward parity tests and a two-layer, two-GPU THD CP=2 functional test.

**Pre checks**:

- [X] Read and followed the contributor guidelines.
- [X] Added focused unit and functional coverage.
- [X] Added runnable Llama, Qwen2, and Qwen3 example configs.

Validation:

* `195 passed` in the targeted CPU unit suite.
* Dense Qwen3 BSHD logits and every parameter gradient match the HuggingFace reference; its state-dict round-trip is exact.
* `ruff format --check .` and `ruff check .`.
* `lint-imports`: 524 files and 1187 dependencies analyzed; component contract kept.
* All 438 example YAMLs validated, including the required new-recipe `ci:` fields.
* Two-GPU Slurm job `13644623` passed with two decoder layers and CP=2:
  * Llama: logits max diff `0.001953`, normalized layer-gradient max diff `0.000000`.
  * Qwen2: logits max diff `0.001953`, normalized layer-gradient max diff `0.000000`.
  * Qwen3: logits max diff `0.001953`, normalized layer-gradient max diff `0.000057`.

# Additional Information

* Linear: NVIDIA-NeMo/Automodel#1473
* Related to NVIDIA-NeMo/Automodel#1473.


nemo_automodel/components/models/qwen3/model.py
```diff
@@ -0,0 +1,484 @@
...
+        if backend.attn == "te":
+            # Ordinary BSHD inputs retain the HuggingFace attention interface.
+            self._te_thd_only = True
+            self.attn_module, self.attn_func = initialize_attn_module_and_func(
+                attn_impl="te",
+                num_attention_heads=config.num_attention_heads,
+                num_qk_channels=self.head_dim,
+                num_v_channels=self.head_dim,
+                softmax_scale=self.scaling,
+                num_gqa_groups=config.num_key_value_heads,
```
yuhezhang-ai, line 77: [P1] Configure TE attention for tensor parallelism before allowing packed TP+CP

This `DotProductAttention` is created with the global head counts and the Transformer Engine default `tp_size=1`. The optimized TP plan later column-shards Q/K/V, so each TP rank passes local K/V heads while this module still expects the global `num_gqa_groups`; Transformer Engine rejects that mismatch at the first packed attention call. Please construct or configure the DPA with the resolved TP group and size for Llama, Qwen2, and Qwen3, or reject the TP+packed-CP composition until it is supported, and add a TP2×CP2 test.

nemo_automodel/components/models/qwen2/model.py
```diff
@@ -100,24 +118,80 @@ def forward(
...
             cos, sin = position_embeddings[:2]
             query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)
 
+        if is_thd:
+            if past_key_values is not None:
+                raise ValueError("Packed THD attention does not support past_key_values.")
+            attn_module = getattr(self, "attn_module", None)
+            if attn_module is None:
+                raise ValueError("Packed THD attention requires backend.attn='te'.")
+            window_size = (-1, 0) if self.sliding_window is None else (self.sliding_window, 0)
```
yuhezhang-ai: [P1] Convert the HF sliding-window length to the TE inclusive radius

The HF causal sliding-window contract keeps exactly `N` tokens (`kv_idx > q_idx - N`), whereas Transformer Engine defines `window_size=(left, right)` with inclusive bounds. Passing `(self.sliding_window, 0)` therefore allows `N+1` tokens. Please use `(self.sliding_window - 1, 0)` here and in Qwen3, and add packed sliding-attention parity coverage; the current GPU test exercises full attention only.

nemo_automodel/components/models/llama/model.py
```diff
@@ -104,6 +110,18 @@ def __init__(
...
+            # BSHD continues through the HuggingFace attention interface; this
+            # module is selected only for packed THD batches.
+            self._te_thd_only = True
+            self.attn_module, self.attn_func = initialize_attn_module_and_func(
+                attn_impl="te",
+                num_attention_heads=config.num_attention_heads,
+                num_qk_channels=self.head_dim,
+                num_v_channels=self.head_dim,
+                softmax_scale=self.scaling,
+                num_gqa_groups=config.num_key_value_heads,
```
yuhezhang-ai, line 135: [P2] Preserve configured attention dropout on the THD path

Transformer Engine `DotProductAttention` defaults `attention_dropout` to `0.0`. This constructor omits it even though the model stores `config.attention_dropout`, so packed THD training silently disables nonzero attention dropout while the BSHD path honors it. Please pass `attention_dropout=config.attention_dropout` in the Llama, Qwen2, and Qwen3 TE constructors and cover a nonzero setting.

yuhezhang-ai (commented):
I found three correctness gaps in the packed Transformer Engine path. The existing coverage stays green because it exercises TP=1, full attention, and `attention_dropout=0.0`.

nemo_automodel/components/models/llama/model.py
```diff
@@ -326,32 +423,53 @@ def forward(
...
-            past_key_values=past_key_values,
-            position_ids=position_ids,
-        )
+        causal_mask = None
+        if not is_thd:
+            causal_mask = create_causal_mask(
+                config=self.config,
+                inputs_embeds=inputs_embeds,
+                attention_mask=attention_mask,
+                cache_position=cache_position,
```
yuhezhang-ai: Transformers 5.12.1 no longer accepts `cache_position` in `create_causal_mask`, so the ordinary BSHD forward raises `TypeError` here (and in the equivalent Qwen2/Qwen3 calls). Please just remove `cache_position` from these mask calls; older versions do not need it either.

nemo_automodel/_transformers/infrastructure.py
```diff
@@ -747,20 +759,31 @@ def apply_model_infrastructure(
     # and clobber the model-owned ring (the original double-apply bug). Non-TE MoE
     # is not excluded by the _uses_te_attention check, so gate on ep_size: only
     # dense (non-MoE) models need this pass.
-    if mesh.cp_size > 1 and mesh.ep_size <= 1 and not _uses_te_attention(model):
+    if mesh.cp_size > 1 and mesh.ep_size <= 1:
```
yuhezhang-ai: Packed TP without CP still skips Transformer Engine TP setup because this whole block requires `cp_size > 1`. With `TP=2, CP=1`, Q/K/V are sharded but `DotProductAttention` remains configured with `tp_size=1`. The new TP2×CP2 test enters this block, so it cannot catch the TP-only case. Please configure TP independently of CP and add a packed TP2/CP1 test.

yuhezhang-ai (dismissed):

PR #3001, opened 2026-07-10
fix(vlm): load Tulu-3 directly from HF Hub instead of local meta JSON

athitten:
## Summary

The Tulu-3 VLM SFT recipes loaded data through `make_meta_dataset`, pointed at a **local** `tulu3_train_meta.json` (→ `tulu3_train.jsonl`) that only exists after a one-time data-prep dump. That file is absent in Nemo-CI, so the runs died with `FileNotFoundError`.

This PR adds `make_tulu3_dataset`, which pulls `allenai/tulu-3-sft-mixture` straight from the HF Hub — no meta-JSON / JSONL prep step. It delegates per-row conversion to the **same** helper the meta-JSON path used (`_convert_sharegpt_to_conversation`), so the resulting data composition is identical to dumping the split to JSONL and loading it via `make_meta_dataset`: no turn cap, `system` turns dropped, every row kept in split order, text-only conversations. The dataset stays Arrow-backed (`.map`) so the full split isn't copied into a Python list.

Fixes [AM-628](https://linear.app/nvidia/issue/AM-628/missing-tulu3-train-metajson-file-leads-to-filenotfounderror).

## Affects

* `examples/vlm_finetune/gemma4/gemma4_31b_tulu3_text_cp8_16k.yaml`
* `examples/vlm_finetune/gemma4/gemma4_e4b_tulu3_text_cp16_64k.yaml`
* `examples/vlm_finetune/minimax_m3/minimax_m3_vl_sft_tulu3_text_cp8_16k.yaml`

All three now set `dataset._target_: ...make_tulu3_dataset` with `path_or_dataset: allenai/tulu-3-sft-mixture`.

## Validation

* Added `TestMakeTulu3Dataset` unit tests (happy path, `system`-dropped parity with the meta path, no turn cap, all-rows-kept-in-order, extra-kwargs ignored).
* MiniMax M3: before (local meta JSON, dotted green) vs after (`make_tulu3_dataset`, solid green) loss curves overlay closely, confirming identical composition.

Screenshot 2026-07-09 at 10 49 43 PM

* Gemma4 31b (local meta JSON and make_tulu3_dataset matching very closely):

Screenshot 2026-07-10 at 8 30 39 AM

* Gemma4 e4b (local meta JSON and make_tulu3_dataset matching very closely):

Screenshot 2026-07-10 at 8 31 47 AM

* `ruff format` / `ruff check` clean.

## Test plan

- [X] Validated all  the three recipes run without `FileNotFoundError`
- [X] `pytest tests/unit_tests/datasets/vlm/test_datasets.py -k TestMakeTulu3Dataset`

### DIFF
--- a/examples/vlm_finetune/gemma4/gemma4_31b_tulu3_text_cp8_16k.yaml
+++ b/examples/vlm_finetune/gemma4/gemma4_31b_tulu3_text_cp8_16k.yaml
@@ -66,8 +66,9 @@ loss_fn:
   _target_: nemo_automodel.components.loss.masked_ce.MaskedCrossEntropy
 
 dataset:
-  _target_: nemo_automodel.components.datasets.vlm.datasets.make_meta_dataset
-  path_or_dataset: logs/gemma4_tulu3_seqpack_cp/data/tulu3_train_meta.json
+  # Pull Tulu-3 straight from the HF Hub (no meta-JSON / JSONL data-prep step).
+  _target_: nemo_automodel.components.datasets.vlm.datasets.make_tulu3_dataset
+  path_or_dataset: allenai/tulu-3-sft-mixture
   split: train
   truncate: false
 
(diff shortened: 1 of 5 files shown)


examples/vlm_finetune/gemma4/gemma4_31b_tulu3_text_cp8_16k.yaml
```diff
@@ -66,8 +66,9 @@ loss_fn:
   _target_: nemo_automodel.components.loss.masked_ce.MaskedCrossEntropy
 
 dataset:
-  _target_: nemo_automodel.components.datasets.vlm.datasets.make_meta_dataset
-  path_or_dataset: logs/gemma4_tulu3_seqpack_cp/data/tulu3_train_meta.json
+  # Pull Tulu-3 straight from the HF Hub (no meta-JSON / JSONL data-prep step).
```
HuiyingLi, line 69: Thanks for capturing this

HuiyingLi (approved):

PR #3003, opened 2026-07-10
feat(speculative): support sequence packing in the EAGLE-1/2 draft

khazic:
## What

Extends EAGLE-3 sequence packing to the EAGLE-1 / EAGLE-2 draft. The EAGLE-1/2 recipe already builds its dataloader with `build_eagle3_dataloader` but never passed `packed_sequence_size`, so packing was unreachable. This wires it end to end, mirroring the EAGLE-3 packing path:

- **Draft** (`draft_llama_v12.py`): `forward` accepts `position_ids` / `seq_lens`. Under packing it builds the shared `[B, 1, T, T]` block-causal additive mask (`build_block_causal_additive_mask`) instead of the plain causal + padding mask, and uses the per-document `position_ids` for RoPE so each document is rotated from its own origin. Fails loud if `seq_lens` is passed without `position_ids`.
- **Target wrapper** (`target_v12.py`): `generate_batch` accepts `position_ids` / `seq_lens` / `doc_remaining`; under packing the target runs with the block-causal mask (SDPA/eager) or FlashAttention varlen (batch size 1), so its hidden states do not leak across documents, and it carries the packing metadata onto `EagleTargetBatch`.
- **Trainer** (`core_v12.py`): threads `position_ids` / `seq_lens` to the draft and gates supervision with `doc_remaining > 0`. EAGLE-1/2 is single-step, so this is the reduction of the EAGLE-3 TTT gate (`step_idx < doc_remaining`) at `step_idx == 0`: it drops each document's last real token, whose target, under the wrapper's global left-shift, belongs to the next document.
- **Recipe** (`train_eagle1.py`): reads `packed_sequence_size` into both dataloaders, splats the packing metadata into `generate_batch` and the trainer, and validates incompatible configs at setup (`_validate_packing_gates`): packing with context parallelism (`cp_size>1`, which EAGLE-1/2 has no sequence-sharding path for), and a FlashAttention target with `micro_batch_size>1`.

The packed dataset (`build_packed_eagle3_dataset` / `_pack_collate`) is reused unchanged; only the recipe needed to enable it.

## Correctness verification

CPU unit tests (16 pass in the packing file, no regressions in the EAGLE-1/2 suite), including:

- **Golden parity**: packing N documents into one row produces the same loss and every draft gradient (tight fp32 tolerance) as N padded single-document rows, run through the real `HFEagleTargetModel` + `EagleTrainerModule`.
- **Document isolation**: perturbing only document B's inputs leaves document A's draft output, and the target's input hidden states, bit-identical.
- **Single-doc equivalence**, and **fail-loud guards** (missing `position_ids`; FlashAttention target with batch > 1; `cp_size>1` with packing).

GPU recipe smoke (1x A100-80GB, bf16), the full recipe with a real Qwen3-4B target and `packed_sequence_size` enabled:

<details>
<summary>Training log</summary>

```
epoch=0 step=1  loss=2.8629 acc=0.0000 lr=0.0001
epoch=0 step=2  loss=2.2494 acc=0.2143 lr=9.9808e-05
epoch=0 step=3  loss=2.0738 acc=0.3643 lr=9.92338e-05
epoch=0 step=5  loss=1.8130 acc=0.6294 lr=9.69613e-05
epoch=0 step=8  loss=1.6369 acc=0.7857 lr=9.09108e-05
epoch=0 step=12 loss=1.4594 acc=0.7471 lr=7.86894e-05
epoch=0 step=18 loss=1.3350 acc=0.7929 lr=5.5e-05
epoch=0 step=24 loss=1.2639 acc=0.8176 lr=3.13106e-05
epoch=0 step=30 loss=1.2525 acc=0.8357 lr=1.47177e-05
epoch=0 step=35 loss=1.2533 acc=0.8235 lr=1e-05
Finished epoch 1/1 completed_steps=35
```

</details>

Loss falls 2.86 -> 1.25 and next-token accuracy rises 0.0 -> ~0.82 with packing on, exercising the full path (packed dataloader -> target block-causal -> draft block-causal + `doc_remaining` gating).

## Scope

EAGLE-1/2 draft only, per the incremental plan (the DeepSeek MLA EAGLE-3 draft is a separate PR). Extending packing to DFlash / DSpark and resolving packing with context parallelism remain follow-ups.

### DIFF
--- a/nemo_automodel/components/speculative/eagle/core_v12.py
+++ b/nemo_automodel/components/speculative/eagle/core_v12.py
@@ -94,18 +94,14 @@ class EagleTrainerModule(nn.Module):
     ) -> EagleStepMetrics:
         """Run one EAGLE-1 / EAGLE-2 training step.
 
-        All per-token tensors are ``[B, T]`` (``input_ids`` / ``loss_mask`` /
-        ``attention_mask`` / ``position_ids`` / ``doc_remaining``) except the
-        ``[B, T, H]`` hidden states and ``[B, T, V]`` ``target_logits``.
-
-        Packing (``seq_lens`` ``[B, max_docs]`` long, per-document lengths summing
-        to ``T``) makes the draft's attention document-level block-causal and its
-        RoPE per-document via ``position_ids``. ``doc_remaining`` ``[B, T]`` (real
-        tokens after each slot within its document) gates supervision: a
-        document's last real token (``doc_remaining == 0``) predicts the *next*
-        document's first token through the wrapper's global left-shift, so it is
-        dropped. This is the single-step reduction of the EAGLE-3 TTT gate
-        (``step_idx < doc_remaining`` at ``step_idx == 0``).
+        Per-token tensors are ``[B, T]`` (``position_ids`` / ``doc_remaining``
+        included) except the ``[B, T, H]`` hidden states and ``[B, T, V]``
+        ``target_logits``; ``seq_lens`` is ``[B, max_docs]``. When packing is on,
+        ``position_ids`` / ``seq_lens`` make the draft block-causal and per-document,
+        and ``doc_remaining`` (real tokens after each slot within its document) gates
+        supervision: a document's last real token (``doc_remaining == 0``) is dropped
+        because the wrapper's global left-shift makes its target the next document's
+        first token.
         """
         if self.training and self.feature_noise > 0:
             # EAGLE feature-noise augmentation (see __init__): add
--- a/nemo_automodel/components/speculative/eagle/core_v12.py
+++ b/nemo_automodel/components/speculative/eagle/core_v12.py
@@ -88,8 +88,25 @@ class EagleTrainerModule(nn.Module):
         input_hidden_states: torch.Tensor,
         target_hidden_states: torch.Tensor,
         target_logits: torch.Tensor,
+        position_ids: torch.Tensor | None = None,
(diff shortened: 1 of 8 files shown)


HuiyingLi:
I found one packing-path issue that looks worth fixing before merge.

`nemo_automodel/recipes/llm/train_eagle1.py` now rejects packed FlashAttention targets when `micro_batch_size > 1`, and the error tells users to load the target with `attn_implementation='sdpa'`. But `TrainEagle1Recipe` does not currently forward any `recipe_args.target_attn_implementation` value into `NeMoAutoModelForCausalLM.from_pretrained`; `target_kwargs` only includes `trust_remote_code`, `torch_dtype`, `force_hf`, and optional `distributed_setup`.

In environments where the default target attention is `flash_attention_2`, that makes `packed_sequence_size > 0` with `micro_batch_size > 1` effectively unusable from YAML, even though the gate says SDPA is the supported workaround.

Could you add the same passthrough used by the neighboring recipes (`train_eagle3.py`, `train_dflash.py`, `train_dspark.py`)? Roughly:

```python
target_attn_implementation = recipe_cfg.get("target_attn_implementation", None)
if target_attn_implementation is not None:
    target_kwargs["attn_implementation"] = target_attn_implementation
```

A small unit test that the kwarg is forwarded would also pin the behavior.

HuiyingLi (approved):

PR #3022, opened 2026-07-10
test(ci): accept SHA-pinned claude-review template refs in the policy test

khazic:
## What

`test_review_workflow_keeps_main_caller_wiring` asserts the exact `uses:` ref of the claude-review caller workflow (`_claude_review.yml@v1.7.0`). #3013 repinned the template to a full commit SHA (`7d857ec #v1.8.3`) without updating the assertion, so the test now fails on `main` and on every PR whose CI merges with current `main` (first seen as an `aea1fdf` mismatch during the v1.8.2 window, e.g. the `L0_Unit_Tests_CPU` failure on #3002).

This keeps the security intent of the test (the callee repo and workflow path must stay exactly `NVIDIA-NeMo/FW-CI-templates/.github/workflows/_claude_review.yml`) while allowing the ref to move between release tags (`vX.Y.Z`) and full 40-hex commit SHAs, so routine repins of the template no longer break `main`.

## Verification

`tests/unit_tests/test_claude_review_policy.py` passes against the current SHA-pinned workflow (84 passed), and the ref regex still accepts the previous tag form (`v1.7.0`). A `uses:` pointing at any other repo, path, or a non-tag/non-SHA ref still fails the test.

### DIFF
--- a/tests/unit_tests/test_claude_review_policy.py
+++ b/tests/unit_tests/test_claude_review_policy.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import re
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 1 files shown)


HuiyingLi (approved):

akoumpa (approved):

akoumpa:
I've manually confirmed this fix; FMing

PR #3024, opened 2026-07-10
fix(datasets): reject non-prefix multiturn mask templates

khazic:
# Problem

The fallback answer-only mask locates each assistant turn with tokenized conversation-prefix lengths:

```text
start = len(render(messages[:assistant_index]))
end = len(render(messages[:assistant_index + 1]))
```

This is correct only when every standalone prefix is also an exact token prefix of the fully rendered conversation.

Some chat templates rewrite earlier turns after later messages are appended. Qwen3 is one concrete example: an assistant turn rendered by itself can contain its `<think>` block, but the same turn is rendered without that block once a later user message exists.

```text
render([user_1, assistant_1])
    = user_1 + think_1 + answer_1

render([user_1, assistant_1, user_2, assistant_2])
    = user_1 + answer_1 + user_2 + think_2 + answer_2
```

The old code still used the longer standalone-prefix length as the end of `assistant_1`. The resulting mask extended past the real answer boundary and could silently supervise tokens from `user_2`, tool messages, or later turns. Training continued without an error, but with incorrect labels.

The same unstable-prefix assumption was also used to locate `reasoning_content`. A rewritten prefix could therefore mask the wrong tokens, while simply skipping every rewritten prefix could leave rendered reasoning tokens supervised.

# Fix

- Validate each tokenized prefix against the full-conversation tokenization before using its length as an assistant boundary.
- Require an exact token prefix by default.
- Preserve the existing boundary tolerance only for multi-token prefixes ending in the tokenizer's known EOS token.
- Raise a descriptive `ValueError` when assistant spans cannot be located safely. The error directs users to a chat template with `{% generation %}` blocks instead of producing corrupted labels.
- Locate each reasoning span by comparing two full-conversation renders, with only the target message's `reasoning_content` cleared.
- Map reasoning spans from the untruncated render into left-truncated and right-truncated training windows.
- Reject truncation results that are not a contiguous prefix or suffix of the untruncated render.

For Qwen3-style history rewriting, earlier reasoning that is absent from the final render produces no mask span, while reasoning that remains in the final render is still masked correctly.

# Testing

- `127 passed` across `test_formatting_utils.py`, `test_tokenizer_apply_functions.py`, `test_agent_chat.py`, and agent recipe wiring tests.
- Added focused CPU coverage for Qwen3-style history rewriting, preserved reasoning under unrelated header rewriting, EOS boundary handling, left truncation, right truncation, and invalid noncontiguous truncation.
- `ruff format --check`
- `ruff check`
- `git diff --check`

- [x] Read and followed the contributor guidelines
- [x] Added focused CPU unit tests
- [x] Updated affected documentation and error messages

### DIFF
--- a/nemo_automodel/components/datasets/llm/formatting_utils.py
+++ b/nemo_automodel/components/datasets/llm/formatting_utils.py
@@ -129,6 +129,23 @@ def _maybe_shift_mask_for_left_padding(
     return [0] * pad_len + mask[: len(mask) - pad_len]
 
 
+def _is_consistent_render_prefix(prefix_ids: "list[int]", reference_ids: "list[int]") -> bool:
+    """Check that a prefix render matches the start of the full-conversation render.
+
+    Locating spans by prefix lengths is only correct when
+    ``render(messages[:k])`` reproduces the first tokens of ``render(messages)``.
+    The final prefix token is excluded from the comparison: many templates close
+    the rendered text with a token (an EOS or a trailing newline) that is
+    replaced or merged once the next message is appended, and that one-token
+    boundary fuzziness is already inherent to prefix-length arithmetic.
+    """
+    if not prefix_ids:
+        return True
+    if len(prefix_ids) > len(reference_ids):
+        return False
+    return prefix_ids[:-1] == reference_ids[: len(prefix_ids) - 1]
+
+
 def _build_multiturn_assistant_mask(
     tokenizer: "PreTrainedTokenizer",
     formatted_text: List[Dict[str, Any]],
@@ -137,7 +154,7 @@ def _build_multiturn_assistant_mask(
     tools: Optional[List[Dict]] = None,
     truncation: Union[str, bool] = "do_not_truncate",
     seq_length: Optional[int] = None,
-    full_length: Optional[int] = None,
+    unpadded_full_ids: "list[int] | None" = None,
 ) -> List[int]:
     """Build a fallback loss mask that supervises every assistant turn.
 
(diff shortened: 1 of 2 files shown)


nemo_automodel/components/datasets/llm/formatting_utils.py
```diff
@@ -129,6 +129,23 @@ def _maybe_shift_mask_for_left_padding(
...
+    The final prefix token is excluded from the comparison: many templates close
+    the rendered text with a token (an EOS or a trailing newline) that is
+    replaced or merged once the next message is appended, and that one-token
+    boundary fuzziness is already inherent to prefix-length arithmetic.
+    """
+    if not prefix_ids:
+        return True
+    if len(prefix_ids) > len(reference_ids):
+        return False
+    return prefix_ids[:-1] == reference_ids[: len(prefix_ids) - 1]
```
khazic: Excluding the final token leaves the original silent-corruption path open. Any one-token prefix is accepted, and a template that replaces its closing token with the next turn's first token is accepted even though `len(prefix_ids)` is one past the true boundary. Please tolerate only a known removable terminator, such as `eos_token_id`, or require an exact prefix match.

nemo_automodel/components/datasets/llm/formatting_utils.py
```diff
@@ -249,6 +308,19 @@ def _build_reasoning_mask(
             truncation=truncation,
             seq_length=seq_length,
         )
+        if not (
```
khazic: Prefix divergence does not imply that `reasoning_content` disappeared. A template can rewrite an earlier header while still rendering this turn's reasoning, and this branch then leaves those reasoning tokens supervised despite `mask_reasoning_content=True`. Render the full conversation again with only this message's `reasoning_content` cleared and diff those two full renders. This also handles Qwen3 historical turns without relying on prefix stability.

nemo_automodel/components/datasets/llm/formatting_utils.py
khazic: Fixed in `4de4600d0`: prefix comparison is exact by default, and only a multi-token prefix ending in the tokenizer's EOS may use the existing boundary tolerance.

nemo_automodel/components/datasets/llm/formatting_utils.py
khazic: Fixed in `4de4600d0`: reasoning spans now compare two full-conversation renders, with explicit mapping for left and right truncation.

hemildesai:
/ok to test [4de4600](https://github.com/NVIDIA-NeMo/Automodel/pull/3024/commits/4de4600d095afe6fdbf263dc5d5914cf7eb82c96)

HuiyingLi (approved):
TYSM

PR #3025, opened 2026-07-10
fix(distributed): resolve VLM AC layer-group paths structurally, not by version

yuhezhang-ai:
## Summary

`_get_model_layer_group_specs()` in
`nemo_automodel/components/distributed/parallelizer.py` listed stale module
paths for several VLM families. On the repo pin (`transformers==5.12.1`) those
FQNs no longer resolve; `_extract_model_layer_groups()` skips FQNs that do not
resolve and returned `{}` (or a partial mapping) **without any diagnostic**,
which turned activation checkpointing and per-layer FSDP sharding into a
**silent no-op** for the affected models: training runs fine, produces correct
losses, and simply OOMs earlier / recomputes nothing — nobody sees an error.

This PR fixes the resolution **structurally**: each layer group lists every
known historical location of its container, and extraction takes the first
candidate that resolves. No `transformers.__version__` gating anywhere —
including the pre-existing Gemma3 gates, which were also removed.

### Why structural, not version-gated (the wrong-boundary finding)

The VLM module-tree standardization is **not** a v4→v5 boundary. Verified by
meta-instantiating tiny configs per version:

| transformers | Qwen2-VL tree | Gemma3 / Llava tree |
|---|---|---|
| 4.51.3 | `model.layers` + `visual.blocks` | `language_model.model.layers` + `vision_tower.vision_model.encoder.layers` |
| 4.57.1 | `model.language_model.layers` + `model.visual.blocks` | `model.language_model.layers` + `model.vision_tower.vision_model.encoder.layers` |
| 5.8.1 / 5.12.1 | `model.language_model.layers` + `model.visual.blocks` | `model.language_model.layers` + `model.vision_tower.encoder.layers` (flattened tower) |

So the restructure landed **mid-4.x** (the 4.52 standardization), and v5 only
added the CLIP/SigLIP tower flattening plus removal of the deprecated
top-level alias properties (`language_model`, `visual`, `vision_tower`) that
standardized 4.x still carries. Any `_is_transformers_v5_or_higher()` gate is
therefore wrong-boundaried:

- A v4-branch spec of `language_model.layers` / `visual.blocks` picks paths
  that are **wrong for 4.52–4.x** (they only "work" via deprecation aliases,
  and `language_model.layers` was *never* a real registered path for
  pre-standardization Gemma3/Llava, whose text tower sits at
  `language_model.model.layers`).
- The **pre-existing Gemma3 gates on main** (both in
  `_get_model_layer_group_specs()` and in `get_hf_tp_shard_plan()`) had the
  same wrong boundary: on 4.57.1 they select the "v4" branch even though the
  tree is already standardized.

Version pinning of module paths is the wrong tool entirely; the module tree
itself is the source of truth.

### Resolution semantics

Candidate FQNs within a group are alternative locations of the same logical
container across releases; `_extract_model_layer_groups()` now takes the
**first candidate that resolves** (canonical path listed first). This matches
the existing Mistral3 multi-candidate behavior (where exactly one candidate
resolves per version) while making the alias-carrying 4.x releases safe: with
the previous collect-all semantics, a deprecated alias like top-level `visual`
on 4.52–4.x would resolve *in addition to* `model.visual` and double-count —
and double-wrap — every vision block.

## What's included

1. **Structural multi-candidate specs** (`parallelizer.py`)
   - All `_is_transformers_v5_or_higher()` conditionals removed from
     `_get_model_layer_group_specs()`. Each family's spec is a flat candidate
     list covering every empirically verified tree shape:
     - Qwen2-VL / Qwen2.5-VL: language `["model.language_model.layers",
       "model.layers"]`, vision `["model.visual.blocks", "visual.blocks"]`.
     - Gemma3 and Llava/LlavaNext/LlavaNextVideo/LlavaOnevision (same
       lineage): language `["model.language_model.layers",
       "language_model.model.layers"]`, vision
       `["model.vision_tower.vision_model.encoder.layers",
       "model.vision_tower.encoder.layers",
       "vision_tower.vision_model.encoder.layers"]`.
   - `_extract_model_layer_groups()` resolves first-match-wins per group (see
     above).
   - The pre-existing version gate in `get_hf_tp_shard_plan()`'s Gemma3 branch
     is replaced by a structural check on **registered child modules**
     (`named_children()`, not `hasattr`, because standardized 4.x keeps the
     deprecated `language_model` alias property): top-level `language_model`
     child → `language_model` prefix, else `model` prefix. On 4.57.1 the old
     gate picked the `language_model` prefix, producing plan FQNs that don't
     exist as registered module paths.
   - `_is_transformers_v5_or_higher()` now has **zero callers** in the repo;
     the helper is left in place.
2. **Silent no-op made loud** (`parallelizer.py`, from the first commit)
   - When a model class *is* in the mapping but none of its spec FQNs resolve,
     `_extract_model_layer_groups` emits a `logger.warning` naming the model
     class and the per-group FQNs that failed to resolve. It does not raise:
     partial trees (e.g. pipeline-parallel stages) may legitimately lack some
     groups, and downstream scoping handles empty results.
3. **`use_cache` disabled on sub-configs too** (`activation_checkpointing.py`,
   second commit, unchanged)
   - `detect_kv_sharing_and_maybe_disable_cache` used to set `use_cache=False`
     only on the composite VLM config. The standardized text model reads
     `text_config.use_cache` (still `True`), so a `DynamicCache` stayed alive
     under checkpointing and the checkpointed text self-attention appended K/V
     both at forward and at recompute, failing backward with a
     `CheckpointError` metadata mismatch (saved seq-len `S` vs recomputed
     `2S`). This was previously masked for Qwen2-VL by bug 1 (no layers were
     checkpointed at all). The function now also disables `use_cache` on every
     declared sub-config (`type(config).sub_configs`, falling back to
     `text_config`) that exposes the attribute; KV-shared models are left
     untouched.

## Tests

All version monkeypatching is gone from the AC-extraction tests; they are now
purely structural (tiny stub trees on the real HF classes via `__new__`, no
downloads, no version checks):

- `tests/unit_tests/distributed/test_parallelizer.py`
  - Qwen2-VL + Qwen2.5-VL: both historical shapes (pre-standardization
    `model.layers`+`visual.blocks`, standardized nested tree) extract both
    groups; parametrized.
  - Gemma3 and all four Llava classes: all three historical shapes
    (pre-standardization, standardized 4.x with nested tower, v5 flattened
    tower) extract both groups; parametrized.
  - Alias regression test: a standardized tree that *also* exposes the
    deprecated top-level `visual`/`language_model` paths (as on 4.52–4.x)
    yields each layer exactly once.
  - Total-mismatch case asserts extraction returns `{}` **and** warns with the
    class name and unresolved FQNs (kept).
  - `get_hf_tp_shard_plan` Gemma3: both tree shapes map to the structurally
    correct plan prefix (`language_model` vs `model`).
- `tests/unit_tests/distributed/test_activation_checkpointing.py` (unchanged)
  - composite config: cache disabled on composite **and** `text_config`;
    sub-config without `use_cache` does not grow a phantom field.
  - legacy config without `sub_configs` falls back to `text_config`.
  - KV-shared model keeps `use_cache=True` everywhere.

Test evidence — the same two files pass unmodified against **three**
transformers versions (login node, torch 2.10, `PYTHONPATH` prepend for the
non-default versions), demonstrating version-independence:

```
transformers 5.8.1  : 122 passed
transformers 5.12.1 : 122 passed  (repo pin)
transformers 4.57.1 : 122 passed
```

(4.51.3 was used only for meta-instantiation shape verification; the repo
itself does not import against transformers that old.)


yuhezhang-ai:
/ok to test [31e823b](https://github.com/NVIDIA-NeMo/Automodel/pull/3025/commits/31e823bd0071989b859164ea91668a2052691833)

akoumpa (approved):
LGTM on the current code;

Note, @yuhezhang-ai, we've discussed offline we want to refactor this and move to a per-model side-bar that relies less on named attribute probing (not at all if possible), so as to increase maintainability.

PR #3028, opened 2026-07-11
fix(models): temporarily disable TE fused RoPE globally

### LINKED ISSUE #3027: Temporarily disable TE fused RoPE globally (breaks HF/vLLM logprob parity in some models)
### Summary
TE's fused RoPE kernel is being **temporarily disabled globally** across all model families. The fused kernel computes cos/sin in **fp32 in-kernel**, while HF and vLLM rotate with **bf16** cos/sin tables. This mismatch breaks logprob parity against HF reference models and against vLLM rollouts (RL), and is not confined to a single family.

### Why / impact
- Confirmed to cause divergence in **Qwen3-MoE** (logprob divergence vs vLLM; see related items below).
- The same fp32-vs-bf16 RoPE discrepancy is a latent risk for **every family** that can use the fused path — so we are turning it off everywhere rather than family-by-family, until a proper parity-preserving fix lands.
- **Trade-off:** fused RoPE is a throughput optimization; disabling it will slightly increase attention-block cost. This is an intentional, temporary correctness-over-speed choice.

### What changed
Force `rope_fusion = False` in `BackendConfig.__post_init__` (`nemo_automodel/components/models/common/utils.py`) — the single chokepoint every `BackendConfig` passes through (including the `dataclasses.replace()` in `resolve_rope_fusion`). This overrides:
1. the legacy `None -> (HAVE_TE and cuda) -> True` default,
2. any explicit `rope_fusion: true` set in a recipe/config,
3. both the resolver-based families and the families that read `backend.rope_fusion` directly (`llama`, `qwen2`, `deepseek_v3`, `kimi_k25_vl`, `kimivl`, `mistral4`).

A one-time rank-0 warning is logged when an explicit `True` is overridden.

### Scope
- All model families, all recipes, training + inference/rollout paths.
- No config changes required from users; the toggle is code-level and centralized.

### Exit criteria (to re-enable)
- Root-cause fix so the fused RoPE kernel is numerically consistent with bf16 HF/vLLM RoPE (or the reference paths are aligned to the fused kernel), verified by logprob-parity tests against HF and vLLM.
- Remove the override block in `BackendConfig.__post_init__` to restore per-family behavior.

### Related
- Qwen3-MoE vLLM logprob divergence (root cause: TE fused RoPE fp32 vs bf16 cos/sin).
- AM-656 (KL divergence between original and HF-loaded model).

HuiyingLi:
## What

Force `rope_fusion=False` for every `BackendConfig` in `__post_init__`, disabling TE fused RoPE globally.

## Why

The fused RoPE kernel computes cos/sin in fp32 in-kernel while HF/vLLM rotate with bf16 tables, breaking logprob parity in some models (e.g. Qwen3-MoE). Temporary correctness-over-speed workaround until a parity fix lands — see NVIDIA-NeMo/Automodel#3027.

## Notes

* Single chokepoint: overrides both the default and any explicit `rope_fusion: true` in a recipe/config; logs a one-time rank-0 warning when overriding an explicit `True`.
* To re-enable, remove the block in `BackendConfig.__post_init__`.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

### DIFF
--- a/nemo_automodel/components/models/common/utils.py
+++ b/nemo_automodel/components/models/common/utils.py
@@ -233,8 +233,7 @@ class BackendConfig:
         # in some models. See #3027. This is the one chokepoint every BackendConfig passes
         # through, so it also overrides an explicit rope_fusion=True from a recipe/config.
         if self.rope_fusion:
-            if not torch.distributed.is_initialized() or torch.distributed.get_rank() == 0:
-                logger.warning("rope_fusion is temporarily force-disabled globally (see #3027).")
+            logger.warning("rope_fusion is temporarily force-disabled globally (see #3027).")
         self.rope_fusion = False
 
         # Normalize te_fp8: dict -> TEFp8Config, None stays None
--- a/nemo_automodel/components/models/common/utils.py
+++ b/nemo_automodel/components/models/common/utils.py
@@ -228,6 +228,15 @@ class BackendConfig:
     compile_attn: bool = False
 
     def __post_init__(self):
+        # TEMPORARY: force TE fused RoPE off globally. The fused kernel computes cos/sin
+        # in fp32 in-kernel while HF/vLLM rotate with bf16 tables, breaking logprob parity
+        # in some models. See #3027. This is the one chokepoint every BackendConfig passes
+        # through, so it also overrides an explicit rope_fusion=True from a recipe/config.
+        if self.rope_fusion:
+            if not torch.distributed.is_initialized() or torch.distributed.get_rank() == 0:
+                logger.warning("rope_fusion is temporarily force-disabled globally (see #3027).")
+        self.rope_fusion = False
+
         # Normalize te_fp8: dict -> TEFp8Config, None stays None
         if isinstance(self.te_fp8, dict):
             self.te_fp8 = TEFp8Config(**self.te_fp8)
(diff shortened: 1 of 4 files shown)


nemo_automodel/components/models/common/utils.py
```diff
@@ -228,6 +228,15 @@ class BackendConfig:
     compile_attn: bool = False
 
     def __post_init__(self):
+        # TEMPORARY: force TE fused RoPE off globally. The fused kernel computes cos/sin
+        # in fp32 in-kernel while HF/vLLM rotate with bf16 tables, breaking logprob parity
+        # in some models. See #3027. This is the one chokepoint every BackendConfig passes
+        # through, so it also overrides an explicit rope_fusion=True from a recipe/config.
+        if self.rope_fusion:
+            if not torch.distributed.is_initialized() or torch.distributed.get_rank() == 0:
```
akoumpa: @HuiyingLi logger is already torch.dist aware :)

nemo_automodel/components/models/common/utils.py
```diff
@@ -228,6 +228,15 @@ class BackendConfig:
...
+                logger.warning("rope_fusion is temporarily force-disabled globally (see #3027).")
```
akoumpa:
```suggestion
            logger.warning("rope_fusion is temporarily force-disabled globally (see #3027).")
```

akoumpa (dismissed):

nemo_automodel/components/models/common/utils.py
HuiyingLi: Indeed! Thanks!

HuiyingLi:
Thanks @akoumpa — suggestion applied in 958cedc8e3eb26fea5280c389f89c7d447aba2ff. /ok to test 958cedc8e3eb26fea5280c389f89c7d447aba2ff

nemo_automodel/components/models/common/utils.py
HuiyingLi: Thanks for the commit!

akoumpa (approved):

PR #3030, opened 2026-07-11
chore: CI proxy for #3024

HuiyingLi:
CI-only proxy PR. Do not merge / do not review. Points at the exact head commit of #3024 (3cc3401f6dc16daadcd471ec264b5d26dd06e199) so internal CI runs under the internal-contributor queue; results post back to #3024 via the shared SHA. Source of truth: #3024. Close once CI completes.


HuiyingLi:
Closing this CI-only proxy because the source PR #3024 already has a successful CICD NeMo run on the same head SHA. The proxy run was still queued on gb200 and is no longer needed.

PR #3037, opened 2026-07-11
feat(speculative): add periodic real accept-length eval

khazic:
## Summary

1. Add an optional decode_eval recipe block that periodically snapshots the EAGLE-3 draft and measures real engine accept length on a reserved GPU.
2. Run vLLM and the benchmark in an isolated worker process while training continues, then log completed results as train/tau_real.
3. Harden worker lifecycle, resume behavior, port selection, snapshot cleanup, and optional-eval failure containment.

## Testing

1. ruff check and ruff format check passed on all changed Python files.
2. 123 related unit tests passed, including decode eval, serve and bench integration, EAGLE-3 cleanup, tau metrics, and offline cache coverage.

### DIFF
--- a/examples/speculative/eagle3/qwen3_eagle3_perfectblend.yaml
+++ b/examples/speculative/eagle3/qwen3_eagle3_perfectblend.yaml
@@ -18,6 +18,12 @@ recipe_args:
   num_workers: 0
   num_epochs: 1
   ttt_steps: 7
+  # decode_eval:                   # periodic REAL accept-length eval in a vLLM server (train/tau_real)
+  #   every_steps: 500             # launch cadence (optimizer steps); checked at log points
+  #   cuda_visible_devices: "7"    # REQUIRED: GPU reserved for the eval engine (exclude it from torchrun)
+  #   num_prompts: 32              # fixed prompt set drawn from input_data (default: train_data_path)
+  #   max_new_tokens: 256
+  #   port: 8199
   draft_vocab_size: 32000
   freeze_embeddings: true
   trust_remote_code: false
--- a/examples/speculative/eagle3/qwen3_eagle3_perfectblend.yaml
+++ b/examples/speculative/eagle3/qwen3_eagle3_perfectblend.yaml
@@ -23,7 +23,7 @@ recipe_args:
   #   cuda_visible_devices: "7"    # REQUIRED: GPU reserved for the eval engine (exclude it from torchrun)
   #   num_prompts: 32              # fixed prompt set drawn from input_data (default: train_data_path)
   #   max_new_tokens: 256
-  #   port: 8199
+  #   port: 0                    # auto-select a free loopback port
   draft_vocab_size: 32000
   freeze_embeddings: true
   trust_remote_code: false
--- a/nemo_automodel/components/speculative/decode_eval.py
+++ b/nemo_automodel/components/speculative/decode_eval.py
@@ -42,12 +42,16 @@ from __future__ import annotations
 import json
 import logging
 import os
+import shutil
+import signal
+import socket
 import subprocess
 import sys
 import time
 import urllib.request
 from argparse import Namespace
 from dataclasses import asdict, dataclass, fields
+from typing import Any
 
 logger = logging.getLogger(__name__)
 
@@ -90,6 +94,7 @@ class DecodeEvalConfig:
     target_model: str
     input_data: str
     output_dir: str
+    num_speculative_tokens: int
     num_prompts: int = 32
     concurrency: int = 8
     max_new_tokens: int = 256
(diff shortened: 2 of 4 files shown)


khazic:
## Correctness review of the decode-eval lifecycle

I traced the full launch → reap → collect → shutdown state machine against the failure modes an out-of-process periodic eval can hit (orphaned engines, resume replay, port collisions, worker crashes, teardown races). Summary: the logic is correct, and every path that can fail is contained so it never takes the training run down. Details below for reviewers.

### What it does

An optional `recipe_args.decode_eval` block turns the trainer's simulated `tau_sim` into a real one. On a step cadence, rank 0 snapshots the draft to the consolidated layout and launches a detached worker subprocess on a reserved GPU; training never blocks. The worker reuses the already-merged `serve_vllm` conversion to boot a real vLLM speculative server, drives a fixed prompt set through the existing `bench_vllm` runner, and writes the engine's real `accept_length` to a JSON sidecar the trainer picks up as `train/tau_real`. The trainer process never imports vllm.

### Lifecycle guarantees I verified

- **At most one eval in flight.** `maybe_launch` reaps the previous worker, then skips if one is still running; the cadence bucket only advances after `Popen` succeeds, so a launch that raised mid-snapshot retries at the next log point instead of silently burning the boundary (`test_failed_launch_retries_same_bucket_and_cleans_snapshot`).
- **Snapshot D2H copy is rank-0 only and safe.** `self.draft_model` is the unwrapped module (DDP wraps `trainer_module`, not the draft), so `state_dict()`/`config` on rank 0 hold the full replicated weights. The assumption is documented in `export_draft_snapshot`. The whole hook sits inside the existing `if self.dist_env.is_main:` log block, so it never enters a collective path and cannot desync ranks.
- **Resume does not replay history.** The runner's constructor marks every already-present `result.json` as collected and drops its snapshot, so restarting from a checkpoint with the same `output_dir` neither re-logs stale `tau_real` points nor lets a relaunched step report a stale file (`test_runner_resume_ignores_old_results_and_removes_snapshot`).
- **Worker failures are visible, not silent.** `_reap_finished_worker` logs a nonzero exit with the worker-log path and cleans the snapshot; a crashed eval surfaces in the training log instead of just making `tau_real` stop appearing (`test_failed_worker_is_reported_and_snapshot_removed`).
- **No orphaned engines on teardown.** The worker is a session leader (`start_new_session=True`) and the vLLM server is in its group, so `shutdown` sends `SIGTERM` to the group and escalates to `SIGKILL` on the group (not just the worker PID) after 30s, taking the engine down with it (`test_shutdown_escalates_to_process_group_kill`). Teardown collects with `launch=False` so it never starts a fresh eval on the way out.
- **Disk does not grow without bound.** Each snapshot's heavyweight `model/` dir is removed as soon as its result is collected (or its worker fails, or at shutdown), so only the JSON sidecars persist.
- **Config plumbing is strict.** Unknown `decode_eval` keys raise (typos fail loudly rather than silently keeping a default), a missing reserved GPU raises, and `input_data` falls back to `train_data_path` via `.get()` so the offline cached-target path (which has no `train_data_path`) is not broken by this feature being merely present.
- **Result read/write is race-free.** The worker writes atomically via `os.replace`; `collect()` only ever sees the fully-written file or none, and the step in the JSON is validated against the directory step.

### Evidence

- `pytest tests/unit_tests/speculative/test_decode_eval.py` (21 passed) plus the adjacent EAGLE-3 suites (`test_train_eagle3_flush`, `test_train_eagle3_tau`, `test_eagle3`: 70 passed, 2 skipped).
- `ruff check`, `ruff format --check`, and `ty check` clean on the changed files.

### Design caveats worth calling out

- **The reserved GPU is operator-managed.** `cuda_visible_devices` must name a card the training world does not use, and the torchrun `--nproc-per-node` has to be shrunk to match. Nothing inside the training process validates this (it can't reliably map torchrun's device placement), so a misconfiguration means the eval engine competes for a training GPU. This is documented in the config docstring; flagging it as an operational contract rather than a code issue.
- **Free-port selection has the usual TOCTOU window.** `_resolve_worker_port` binds a probe socket, closes it, and vLLM binds the port afterward (the window spans the `resolve_draft_artifacts` export). If another process grabs the port in between, vLLM fails to bind, `_wait_for_server` observes the dead engine and raises, the worker exits nonzero and is reported, and the next cadence retries. So the race is self-healing and visible rather than silent; on a reserved single-eval GPU it is very unlikely to trigger at all.

### Still owed

Unit tests mock the subprocess and engine, so the snapshot → `serve_vllm` conversion → real vLLM boot → `bench_vllm` measurement path is exercised only through its own existing coverage, not end to end. A real single-node GPU smoke run (train EAGLE-3 a few hundred steps with `decode_eval` enabled on a spare card, confirm `train/tau_real` lands and climbs) is the remaining validation; I will attach the sanitized log once it is done.

HuiyingLi (approved):

PR #3046, opened 2026-07-13
feat(optim): support per-parameter-group learning rate

### LINKED ISSUE #2961: Support per parameter group learning rate

khazic:
Closes #2961.

## What

Adds per-parameter-group learning rate: a subset of parameters, matched by name, can be given its own optimizer parameter group with an `lr` / weight-decay multiplier. This mirrors Megatron-LM's per-group `lr_mult` scaling (e.g. a reduced learning rate for MoE router parameters).

```yaml
optimizer:
  _target_: torch.optim.AdamW
  lr: 2.0e-5
  weight_decay: 0.01
  param_group_overrides:
    - pattern: "router"     # Python regex matched against each parameter name (re.search)
      lr_mult: 0.1          # this group trains at 0.1 x the scheduled learning rate
      wd_mult: 1.0          # optional; defaults to 1.0
```

## How

Most of the machinery already existed, so this is a small, surgical change:

- The LR scheduler (`OptimizerParamScheduler.step`) already applies a per-group `lr_mult` / `wd_mult` and honors per-group `max_lr` / `min_lr`. No scheduler change.
- New `ParamGroupOverride` dataclass (`pattern`, `lr_mult`, `wd_mult`) and a `param_group_overrides` field on `OptimizerConfig`.
- `OptimizerConfig.build` now iterates `named_parameters()` and, when overrides are set, partitions parameters into groups: each parameter joins the **first** override whose `pattern` matches; unmatched parameters form the default group. Each override group carries `lr_mult` / `wd_mult` (read by the scheduler) plus an initial `lr` / `weight_decay` pre-scaled by the multiplier so the override holds even before the first scheduler step.
- Works for the standard torch optimizers on **both** config paths: the typed configs (`_target_: adamw`) and the common factory path (`_target_: torch.optim.AdamW`, where overrides arrive inside `kwargs`). Dion-family configs do their own grouping and warn if `param_group_overrides` is set.
- The default (unmatched) group is placed first so `LRSchedulerConfig.build`'s base-LR detection (`param_groups[0]["lr"]`) reads the unscaled learning rate.
- A pattern that matches nothing is skipped with a warning (no empty param group).

Default is no overrides, i.e. a single param group, so existing configs are unchanged.

## Tests

`tests/unit_tests/optim/test_optim_config.py` (`TestParamGroupOverrides`):

- dict-to-`ParamGroupOverride` coercion and invalid-type rejection
- no overrides -> single group (unchanged)
- grouping by pattern on both the typed and factory paths, including pre-scaled initial `lr` / `weight_decay` and default-group-first ordering
- non-matching pattern -> dropped with a warning
- first-matching-override-wins
- `param_group_overrides` stripped from the torch constructor kwargs
- end-to-end: `OptimizerParamScheduler.step` applies `lr_mult` / `wd_mult` to the right group

All `tests/unit_tests/optim/` pass (138 passed, 6 skipped) and `ruff format` / `ruff check` are clean on the changed files.

Docs: the finetune guide documents the new `param_group_overrides` option.


khazic:
### GPU validation (FSDP2 / DTensor)

Beyond the CPU unit tests, I validated the feature on a real GPU (single A100-80GB) under **FSDP2**, so the optimizer builds its parameter groups over sharded **DTensor** parameters and the LR scheduler drives real `optimizer.step()` calls. Setup: a small Llama config (4 layers), `fully_shard` on every layer + root, `build_optimizer_config("torch.optim.AdamW", {..., param_group_overrides:[{pattern:"mlp", lr_mult:0.1, wd_mult:0.0}]})`, an `OptimizerParamScheduler` (5-step warmup + cosine decay), then 15 real forward/backward/step iterations.

What the log shows:

- `params=39 DTensor_sharded=39` — every parameter is a sharded DTensor (FSDP2 active).
- Two param groups are built: `default` (27 params, `lr_mult=1.0`) and `mlp-override` (12 params, `lr_mult=0.1`, `wd_mult=0.0`).
- At construction both groups store the **unscaled** base `lr=1.0e-3` (the fix that keeps `LRSchedulerConfig` base-LR detection correct); the scheduler applies the multiplier.
- `ratio(mlp/default) == 0.1000` at **every** step through warmup and cosine decay, and the `mlp` group's weight decay is `0.0` (`wd_mult=0.0`). Training runs without error.

<details>
<summary>GPU run log (single A800-80GB, FSDP2, torch 2.12)</summary>

```
[env] torch=2.12.0+cu130 world_size=1 device=NVIDIA A800-SXM4-80GB
[fsdp2] params=39 DTensor_sharded=39
[optim] AdamW param_groups=2
  group0 [default] nparams=27 lr=1.000e-03 wd=0.100 lr_mult=1.0 wd_mult=1.0
  group1 [mlp-override] nparams=12 lr=1.000e-03 wd=0.100 lr_mult=0.1 wd_mult=0.0
[train] step  loss     lr_default   lr_mlp       ratio(mlp/default)
          1   6.9145  2.8000e-04  2.8000e-05  0.1000
          2   6.9890  4.6000e-04  4.6000e-05  0.1000
          3   7.0031  6.4000e-04  6.4000e-05  0.1000
          4   7.0311  8.2000e-04  8.2000e-05  0.1000
          5   6.8784  1.0000e-03  1.0000e-04  0.1000
          6   6.9665  9.9645e-04  9.9645e-05  0.1000
          7   6.9487  9.8586e-04  9.8586e-05  0.1000
          8   7.0130  9.6840e-04  9.6840e-05  0.1000
          9   7.0438  9.4434e-04  9.4434e-05  0.1000
         10   6.9365  9.1406e-04  9.1406e-05  0.1000
         11   6.9645  8.7804e-04  8.7804e-05  0.1000
         12   7.0120  8.3684e-04  8.3684e-05  0.1000
         13   6.9454  7.9112e-04  7.9112e-05  0.1000
         14   6.9799  7.4160e-04  7.4160e-05  0.1000
         15   6.9813  6.8906e-04  6.8906e-05  0.1000
[check] mlp group weight_decay=0.0000 (wd_mult=0.0 -> expect 0.0)
[check] ratio(mlp/default) == lr_mult(0.1) every step -> per-group LR works under FSDP2
[done] OK
```
</details>

(The bf16-Adam precision warning in the raw log is the pre-existing generic advisory, unrelated to this change.)

docs/guides/llm/finetune.mdx
jgerh:
```suggestion
Pretrained language models are general-purpose: they know a lot about language but nothing about your particular domain, terminology, or task. Fine-tuning bridges that gap. You fine-tune the model on your own examples so it produces answers that are accurate and relevant for your use case, without the cost of training a model from scratch. The result is a model optimized for your data that you can evaluate, publish, and deploy. This guide walks you through that process end-to-end with NeMo AutoModel (from installation through training, evaluation, and deployment) using [Meta Llama 3.2 1B](https://huggingface.co/meta-llama/Llama-3.2-1B) and the [SQuAD v1.1](https://huggingface.co/datasets/rajpurkar/squad) dataset as a running example.
```

jgerh:
```suggestion
Docker containers are ephemeral, and files written inside the container are lost when it stops. The `-v` flag in the `docker run` command above bind-mounts a local `checkpoints/` directory into the container so that saved checkpoints persist across runs. For more details, see [Save Checkpoints When Using Docker](/development/checkpointing#save-checkpoints-when-using-docker).
```

jgerh:
```suggestion
Training is configured through a [YAML](https://en.wikipedia.org/wiki/YAML) configuration file with three required sections (model, dataset, and step_scheduler) plus an optional **peft** section. The following sections explain each one. For the complete copy-pastable file, see [Full Config YAML](#full-config-yaml).
```

jgerh:
```suggestion
To discover which parameters a section accepts, look up the Python signature of its `_target_`. For example, `torch.optim.Adam` accepts `lr`, `betas`, `eps`, and `weight_decay`. These are the keys you can set in the YAML.
```

jgerh:
```suggestion
1. **Resolves `_target_`**: Imports the Python path and obtains the callable (class or function).
2. **Calls it**: Passes every other key in the section as a keyword argument.
```

jgerh:
```suggestion
**The `recipe` key.** Every config file includes a top-level `recipe` key that tells the CLI *which recipe class* to run. You can write it as a **short name** or as a **fully-qualified Python path**, and both resolve to the same class:
```

jgerh:
```suggestion
The short name form is a convenience. The CLI scans all recipe modules under `nemo_automodel.recipes` and matches the bare class name. If you invoke the recipe script directly with `torchrun` instead of the `automodel` CLI, the `recipe` key is not required because the script itself *is* the recipe.
```

jgerh:
```suggestion
**Not every section uses `_target_`.** Some sections like `step_scheduler`, `distributed`, and `checkpoint` are plain key-value groups consumed directly by the recipe. They control training schedule, parallelism strategy, and checkpoint behavior without instantiating a Python object.
```

jgerh:
```suggestion
This guide uses **SQuAD v1.1** as a running example. Swap the dataset by changing `_target_` and the dataset arguments. See [Integrate Your Own Text Dataset](/datasets/text-dataset) and [Dataset Overview: LLM, VLM, and Retrieval Datasets](/datasets/overview).
```

jgerh:
```suggestion
### Optional PEFT
```

jgerh:
```suggestion
Including a `peft:` section enables LoRA fine-tuning. Remove it entirely to run SFT instead. See [Switch Between SFT and PEFT](#switch-between-sft-and-peft).
```

jgerh:
```suggestion
#### QLoRA for Quantized Low-Rank Adaptation
```

jgerh:
```suggestion
To enable QLoRA, add a `quantization:` section alongside the `peft:` section in your configuration. Note two differences from the standard PEFT configuration previously described. `target_modules` uses the broader `"*_proj"` pattern to apply LoRA to all projection layers (wider coverage compensates for precision loss from 4-bit weights), and `dim` is increased from 8 to 16 for additional adapter capacity.
```

jgerh:
```suggestion
Unlike the sections above, `step_scheduler` has **no `_target_`**. It is not instantiated into a Python object. Instead, the recipe reads its keys directly to control the training loop (how many epochs to run, when to checkpoint, and when to validate). This is typical of sections that configure *behavior* rather than *components*.
```

jgerh:
```suggestion
All other settings (distributed strategy, optimizer, checkpointing, and logging) use sensible defaults. See the [Full Configuration Reference](#full-configuration-reference) to customize them.
```

jgerh, line 277:
```suggestion

</Note>
```

jgerh:
```suggestion
The `--nproc-per-node=8` flag specifies the number of GPUs per node. Adjust as needed. For a single GPU, omit the `--nproc-per-node` option.
```

jgerh:
```suggestion
Alternatively, you can invoke the recipe [script](https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/recipes/llm/train_ft.py) directly using [torchrun](https://docs.pytorch.org/docs/stable/elastic/run.html) as shown below.
```

jgerh:
```suggestion
Each log line reports the current loss, gradient norm, peak GPU memory, and tokens per second (TPS). Small fluctuations between steps (for example, 0.2057 to 0.2557) are normal. Look at the overall downward trend rather than individual values.
```

jgerh:
```suggestion
Checkpoints are saved as Hugging Face-compatible safetensors. By default, SFT writes sharded model weights plus a generated `model/consolidate.sh` helper. Run the helper after training to create `model/consolidated/` for Transformers, vLLM, lm-eval-harness, and other Hugging Face ecosystem tools. You can also set `save_consolidated: final` to export consolidated HF weights only for the final checkpoint. Use `save_consolidated: every` (or legacy `true`) only when you intentionally want inline HF export at every checkpoint save. **PEFT checkpoints** contain only the adapter weights (megabytes instead of gigabytes) and are saved directly under `model/`. They do not use `model/consolidate.sh`. At inference time you must load the original base model and apply the adapter on top. This distinction affects every downstream step (inference, publishing, and deployment).
```

jgerh:
```suggestion
### Validation Loss During Training
```

jgerh:
```suggestion
A decreasing validation loss across checkpoints indicates the model is learning. If validation loss plateaus or increases while training loss continues to drop, the model might be overfitting, so consider stopping earlier or reducing the learning rate.
```

jgerh:
```suggestion
### Post-Training Evaluation Using lm-eval-harness
```

jgerh:
```suggestion
For task-specific benchmarks (such as MMLU, GSM8K, and HellaSwag accuracy), use [lm-eval-harness](https://github.com/EleutherAI/lm-evaluation-harness) with the fine-tuned checkpoint. For SFT runs with `save_consolidated: false`, run `bash checkpoints/epoch_0_step_20/model/consolidate.sh` before pointing evaluation at `model/consolidated/`:
```

jgerh:
```suggestion
Ensure vLLM is installed (`pip install vllm`, or use an environment that includes it).
```

jgerh:
```suggestion
PEFT adapter serving uses the `vLLMHFExporter` class, which is provided by the `nemo` package (a separate dependency from `nemo-automodel`).
```

jgerh:
```suggestion
<Info>
Install both packages before proceeding:

```bash
pip install nemo vllm
```
```

Comment (jgerh) [inline docs/guides/llm/finetune.mdx]:
```suggestion
# Controls epoch count, batch sizes, and how often to checkpoint and validate.
# No _target_ (these are plain values read directly by the recipe).
```

jgerh:
```suggestion
# No _target_ (consumed directly by the recipe).
```

jgerh:
```suggestion
# No _target_ (consumed directly by the recipe).
```

jgerh:
```suggestion
                          # Increase for very long contexts (e.g., 32k+ tokens).
```

jgerh:
```suggestion
# cache) a pretrained Hugging Face model and wraps it with NeMo distributed-training
# support. Any from_pretrained kwarg is accepted (such as cache_dir and torch_dtype).
```

jgerh:
```suggestion
# No _target_ (plain key-value group consumed by the recipe).
```

docs/guides/llm/finetune.mdx
```diff
@@ -798,6 +798,23 @@ optimizer:
...
+  weight_decay: 0.01
+  param_group_overrides:
+    - pattern: "router"     # Python regex matched against each parameter name (re.search)
+      lr_mult: 0.1          # this group trains at 0.1 x the scheduled learning rate
+      wd_mult: 1.0          # (optional) multiplier for weight decay; defaults to 1.0
+```
+
+Each parameter joins the group of the **first** override whose `pattern` matches its name; parameters matching no pattern keep the base learning rate. A pattern that matches nothing is skipped with a warning. This works for the standard `torch.optim` optimizers; Dion-family optimizers do their own parameter grouping and ignore `param_group_overrides`.
+
 ## Distributed Training: TP, PP, CP, and EP
```
jgerh:
```suggestion
## Distributed Training with TP, PP, CP, and EP
```

docs/guides/llm/finetune.mdx
jgerh:
```suggestion
| **Data Parallelism (DP)** | `dp_size` | Replicates the model on each group of GPUs; each replica trains on a different data batch. | Default. Scales batch size linearly with GPU count. |
| **Tensor Parallelism (TP)** | `tp_size` | Splits individual weight matrices (attention, MLP) across GPUs within a node. | Model is too large to fit on a single GPU, or you want to reduce per-GPU memory at the cost of extra communication. |
| **Pipeline Parallelism (PP)** | `pp_size` | Assigns different *layers* (stages) to different GPUs and pipelines micro-batches through them. | Very deep models that do not fit even with TP, or multi-node training where TP's all-reduce is too expensive across nodes. |
| **Context Parallelism (CP)** | `cp_size` | Splits the input *sequence* across GPUs so each GPU processes a portion of the context. | Very long sequences that exceed single-GPU memory. |
| **Expert Parallelism (EP)** | `ep_size` | Distributes MoE experts across GPUs so each GPU holds a subset of experts. | Mixture-of-Experts models only. |
```

jgerh:
```suggestion
EP does not appear in this formula. Experts are distributed across the DP×CP rank groups, with the constraint that `(dp_size × cp_size)` must be divisible by `ep_size`.
```

jgerh:
```suggestion
#### Data Parallelism as the Default
```

jgerh:
```suggestion
TP splits weight matrices across GPUs within a single node. Set `tp_size` to the number of GPUs you want to shard over (typically 2, 4, or 8, and should divide evenly into the number of attention heads):
```

jgerh:
```suggestion
CP splits the sequence across GPUs, which is useful for very long contexts that exceed single-GPU memory. Set `cp_size` to the target split factor:
```

jgerh:
```suggestion
When `cp_size > 1`, fused RoPE is automatically disabled. Some models also require the Transformer Engine (TE) attention backend for CP with packed sequences. The framework will raise an error with instructions if this applies.
```

jgerh:
```suggestion
#### Expert Parallelism for MoE Models
```

jgerh:
```suggestion
EP only applies to Mixture-of-Experts models (such as Qwen3-MoE, Mixtral, and DeepSeek-V3). For dense models, leave `ep_size` at `1` or omit it.
```

jgerh:
```suggestion
- **`(dp_size × cp_size) % ep_size == 0`**, where EP shares the DP×CP groups.
- **TP within a node, PP across nodes** is the typical layout. TP requires fast NVLink bandwidth, whereas PP tolerates higher latency.
- **Start simple.** Use DP-only first. Add TP if the model does not fit on one GPU. Add PP for very large models. Add CP for long sequences. Add EP only for MoE architectures.
```

jgerh:
```suggestion
- [Integrate Your Own Text Dataset](/datasets/text-dataset) to swap the SQuAD example for your own data.
- [Recipes and End-to-End Examples](/recipes-e2e-examples/overview) to browse the full set of recipes available in NeMo AutoModel. See also the [`examples/llm_finetune/`](https://github.com/NVIDIA-NeMo/Automodel/tree/main/examples/llm_finetune) directory for ready-to-run configurations.
- [Dataset Overview: LLM, VLM, and Retrieval Datasets](/datasets/overview) to see all supported dataset types across LLM, VLM, and retrieval tasks.
- [Knowledge Distillation](/recipes-e2e-examples/knowledge-distillation) to distill a fine-tuned model into a smaller one.
```

jgerh:
```suggestion
  num_epochs: 1     # Will train over the dataset for one epoch.
```

jgerh:
```suggestion
# _target_ → torch.optim.Adam: any torch.optim class can be used here (such as
# AdamW and SGD). All remaining keys become kwargs to the constructor.
```

jgerh (commented):
Completed tech pubs review of docs/guides/llm/finetune.mdx and provided copyedits and suggested text revisions.

jgerh (commented):
Completed tech pubs review of docs/guides/llm/finetune.mdx and provided copyedits and suggested text revisions.

khazic:
Thanks @jgerh, applied all your tech-pubs copyedits in df2fd47.

HuiyingLi:
Thank you @khazic  Do you have test/examples on TEFusedAdam optimizer? Thank you!

khazic:
@HuiyingLi Yes, TE FusedAdam is supported. Per-group overrides flow through both optimizer paths:

- **Typed `FusedAdamConfig`**: `build()` hands TE FusedAdam the per-group dicts. `FusedAdam` subclasses `torch.optim.Optimizer`, so the extra `lr_mult` / `wd_mult` keys live on each group and `OptimizerParamScheduler.step()` applies them every step, exactly as it does for the torch optimizers.
- **Factory escape hatch (`_target_: transformer_engine.pytorch.optimizers.FusedAdam`)**: same grouping, and the TE-specific `_drop_empty_local_shards` guard already handles the grouped shape (it filters zero-numel local shards per group and preserves each group's `lr_mult` / `wd_mult`).

The CPU unit tests parametrize the grouping over `torch.optim.Adam` / `AdamW`, since TE FusedAdam needs a GPU plus TE and can't run in the CPU suite. I'll add a TE FusedAdam validation: an FSDP2 GPU run with `param_group_overrides` on TE FusedAdam (mirroring the AdamW log above), plus a CPU test for the grouped `_drop_empty_local_shards` path. Posting the log shortly.

khazic:
Here is the TE FusedAdam validation you asked about, @HuiyingLi. Same setup as the AdamW run above (small Llama, 4 layers, `fully_shard` on every layer plus root, `OptimizerParamScheduler` with 5-step warmup plus cosine decay), but the optimizer is **TransformerEngine `FusedAdam`** (fp32 master weights), built with `param_group_overrides=[{pattern:"mlp", lr_mult:0.1, wd_mult:0.0}]`.

What the log confirms:

- **Both construction paths** build the two groups: the typed `FusedAdamConfig`, and the factory `_target_: transformer_engine.pytorch.optimizers.FusedAdam` escape hatch (the latter runs through the TE-specific `_drop_empty_local_shards` guard, which handles the grouped shape).
- `params=39 DTensor_sharded=39`: every parameter is a sharded DTensor, so FSDP2 is active and the optimizer builds its groups over DTensor params.
- Group `default` (27 params, `lr_mult=1.0`) and `mlp-override` (12 params, `lr_mult=0.1`, `wd_mult=0.0`). Both store the unscaled base `lr=1e-3` at construction; the scheduler applies the multiplier.
- `ratio(mlp/default) == 0.1000` at every step through warmup and cosine decay, and the `mlp` group's weight decay stays `0.0` (`wd_mult=0.0`). 15 real forward/backward/step iterations run without error.

<details>
<summary>GPU run log (single A800-80GB, FSDP2, TE FusedAdam, torch 2.12)</summary>

```
[env] torch=2.12.0+cu130 world_size=1 device=NVIDIA A800-SXM4-80GB
[fsdp2] params=39 DTensor_sharded=39
[optim] transformer_engine.pytorch.optimizers.fused_adam.FusedAdam (TE FusedAdam) param_groups=2
  group0 [default] nparams=27 lr=1.000e-03 wd=0.100 lr_mult=1.0 wd_mult=1.0
  group1 [mlp-override] nparams=12 lr=1.000e-03 wd=0.100 lr_mult=0.1 wd_mult=0.0
[optim-factory] TE FusedAdam via _target_ param_groups=2 mlp_lr_mult=0.1 (TE _drop_empty_local_shards guard active)
[train] step  loss     lr_default   lr_mlp       ratio(mlp/default)
          1  6.9766  2.0000e-04  2.0000e-05  0.1000
          2  7.0200  4.0000e-04  4.0000e-05  0.1000
          3  7.0028  6.0000e-04  6.0000e-05  0.1000
          4  6.9319  8.0000e-04  8.0000e-05  0.1000
          5  6.9497  1.0000e-03  1.0000e-04  0.1000
          6  6.9606  9.9975e-04  9.9975e-05  0.1000
          7  6.9933  9.9902e-04  9.9902e-05  0.1000
          8  6.9723  9.9779e-04  9.9779e-05  0.1000
          9  6.9189  9.9607e-04  9.9607e-05  0.1000
         10  6.9461  9.9386e-04  9.9386e-05  0.1000
         11  6.9704  9.9117e-04  9.9117e-05  0.1000
         12  6.9665  9.8800e-04  9.8800e-05  0.1000
         13  6.9878  9.8434e-04  9.8434e-05  0.1000
         14  6.9700  9.8022e-04  9.8022e-05  0.1000
         15  6.9855  9.7562e-04  9.7562e-05  0.1000
[check] mlp group weight_decay=0.0000 (wd_mult=0.0 -> expect 0.0)
[check] ratio(mlp/default) == lr_mult(0.1) every step -> per-group LR works with TE FusedAdam under FSDP2
[done] OK
```
</details>

khazic:
Pushed the CPU unit test in 0d7199b: `TestDropEmptyLocalShards` covers the zero-numel-shard guard on both the flat param list and the grouped `param_group_overrides` shape (empty local shards dropped, per-group `lr_mult` / `wd_mult` preserved, fully-empty group or list raises). Local run: `4 passed`, full `test_optim_config.py` `52 passed, 6 skipped`.

HuiyingLi (approved):

PR #3049, opened 2026-07-13
refactor(dspark): pack the log window in a helper

kashif:
Follow up to #2957. The DSpark trainer packed its window metrics into one tensor for a single all-reduce and then unpacked them by hard coded index (w[10], w[4]/w[5], the block_size slices). That arithmetic had no test, and inserting a metric into the packed list would have silently misaligned everything after it.

Moves the packing, the unpacking, and the accumulation into a small _DSparkMetricWindow dataclass. Pack and unpack both walk the same name tuple, so the two cannot drift apart. Behaviour is unchanged: losses still divide by the micro batch count, the acceptance diagnostics still divide their summed numerator by their summed denominator once, and a diagnostic with a zero denominator is still omitted.

Adds unit tests for the pack layout, the window averaging, the ratio of sums across two ranks, the omission of unmeasured diagnostics, and reset.


khazic:
I will review it

khazic:
Looks good to me. @huiying, could you please start CI?

HuiyingLi:
Hi @kashif  could you please resolve the conflict? Thanks!

HuiyingLi (approved):

HuiyingLi (approved):

PR #3051, opened 2026-07-13
fix(recipe): add generation-marked chat templates

khazic:
# What does this PR do ?

Configure the Qwen3 MoE THD and OLMo SQuAD finetuning recipes to use generation-marked chat templates for both training and validation datasets.

The native Qwen3 Thinking chat template rewrites earlier assistant turns when later user turns are present. The OLMo template can also render a user-only prefix differently from the same user turn in a complete conversation, depending on the Transformers version. Prefix-based answer-only mask construction therefore rejects these samples. Generation-marked templates allow the tokenizer to return exact assistant masks directly.

# Changelog

- Add the Qwen3 MoE chat template to the training and validation dataset configuration.
- Add a generation-marked OLMo template without changing rendered training text.
- Allow the SQuAD dataset builder to accept a chat template string or path.
- Configure both OLMo SQuAD recipes to use the new template.
- Add CPU regression tests for the Qwen3 MoE and OLMo recipe configurations and the SQuAD template override.

**Pre checks**:

# Additional Information

- Reproduced with `Qwen/Qwen3-30B-A3B-Thinking-2507` and a two-turn conversation.
- The native template raises the non-prefix answer-only mask error.
- Reproduced the OLMo failure with a tokenizer that rewrites a user-only prefix.
- The configured generation-marked templates produce valid assistant masks.
- The OLMo template renders the same training text as the model's native template.


HuiyingLi (approved):
TYSM!

Issue #3062, opened 2026-07-14
[RFC] Disaggregated streaming backend for speculative-decoding draft training


khazic:
# [RFC] Disaggregated streaming backend for speculative-decoding draft training

## Ownership

This feature is assigned to @kashif (Kashif Rasul).

## Summary

Introduce a streaming, producer/consumer feature pipeline for draft-model
(EAGLE / DFlash / DSpark) training so that target-model feature generation and
draft-model training run as independently-paced pools connected by a
lightweight reference queue plus a pluggable feature store, instead of the
current synchronous request/response remote backend.

The first producer backend keeps using AutoModel's own target-model forward
(the existing `Eagle3TargetBackend` path); no external inference engine is
required for the initial landing. This is a design proposal only. Implementation
is split into an independently-reviewable PR chain (see "Phased plan").

## Motivation

Today the only train/inference disaggregation path is the EAGLE-3 remote backend
(`nemo_automodel/components/speculative/eagle/remote/`). It cleanly separates the
target GPUs from the draft GPUs over HTTP (control) + NCCL (tensors), with
prefetch and multiple remote URLs. But the trainer and the target server are
still bound in a tight request/response relationship:

- The trainer decides when to request a batch and blocks on the matching reply;
  each server serializes to one in-flight generate (`_generate_lock`,
  `remote/server.py`), and the client keeps one in-flight request per server to
  preserve NCCL recv ordering (`train_eagle3.py` prefetch loop, around line 775).
- There is no global queue, no producer/consumer backpressure, and no durable,
  resumable feature references. If a target server errors mid-step, it directly
  stalls the training step.
- Target and draft throughput must roughly match. Scaling the target side means
  changing `remote_urls`, and the trainer prefetch depth is capped to the number
  of remote servers (`_resolve_prefetch_depth`, `train_eagle3.py`).
- The remote backend exists for EAGLE-3 only. DFlash and DSpark are
  colocated-only (`train_dflash.py`, `train_dspark.py` load the target
  in-process), so their large-target story is worse.

For very large MoE targets (100B+ class, tens of GPUs, multi-node, long runs),
we want target inference and draft training to overlap continuously, scale
independently, absorb producer jitter, and tolerate a single producer failing
without killing the run. That requires turning the current "run one remote
forward on demand" into an asynchronous feature-production system.

Non-goal for small setups: 8B / 30B single-node runs where target and draft
throughput are close are well served by the existing colocated or remote
[...]

PR #3084, opened 2026-07-15
feat(streaming): add data-plane contracts and local feature store

kashif:
## What does this PR do ?

PR 1 of NVIDIA-NeMo/Automodel#3062. Adds the tensor-free control-plane contracts and an in-process feature store that the next three PRs will plug producers, consumers, and remote transports into. No trainer wiring yet, behavior-neutral.

## Changelog

* New `nemo_automodel/components/speculative/streaming/` package:
  * `SampleRef` + `FeatureSpec`: frozen, tensor-free dataclasses for the control plane. Each ref carries `feature_keys`, per-feature `feature_specs`, `algorithm`, `schema_version`, residency, and version counters.
  * `assert_no_tensors`: structural (not nominal) guard against tensors, numpy arrays, and duck-typed tensor-likes anywhere in the dataclass / dict / list tree. Run inside `SampleRef.__post_init__` so a misbehaving producer fails fast.
  * `FeatureAlgorithm` enum (`eagle3`, `dflash`, `dspark`) with a per-algorithm required-features check.
  * `FeatureStore` ABC + `StoreHandle` + `StoreHealth`: six-method data-plane contract with consume-once refcounting, ints-only health snapshot for backpressure, and a URI partition so cross-store refs are rejected at materialization.
  * `LocalFeatureStore`: in-process implementation with sample-count and byte caps (bytes as the hard backstop per RFC Q2) and high / low-watermark hysteresis.
  * `SampleRefQueue` + `Lease` + `VisibilityTimeout`: lease / ack / fail queue with visibility-timeout reclaim and watermark-based backpressure driven by `StoreHealth` ints only.
* 60 unit tests under `tests/unit_tests/speculative/streaming/`:
  * refs: invariants, frozen, algorithm-requirements, structural `assert_no_tensors` walk, tensor-free construction guard.
  * local_store: round-trip, detached copies, multi-handle refcount, residency caps, watermark transitions, lifecycle.
  * queue: FIFO acquire/ack, fail redelivery, visibility-timeout reclaim, pause/resume callback hysteresis, duplicate-id rejection.
* `ruff format` clean (10 / 10 files), `ruff check` clean.
* Wider sweep: `825 passed, 12 skipped` across `tests/unit_tests/speculative/`. Zero regressions.

## Additional Information

* Out of scope (deferred to PR 2 / 3 / 4 per the RFC): async producer, `SharedDirFeatureStore`, `NcclFeatureStore`, recipe wiring, the EAGLE-3 schema registry.


nemo_automodel/components/speculative/streaming/queue.py
```diff
@@ -0,0 +1,357 @@
...
+            RuntimeError: if the queue is closed while the producer is
+                blocked, so a producer does not silently swallow a
+                shutdown signal.
+        """
+        while True:
+            with self._put_cv:
+                if self._closed:
+                    raise RuntimeError("SampleRefQueue is closed while put was waiting on backpressure")
+                health = self._store.health()
+                if not health.high_watermark_hit:
```
khazic: `put_blocks_until_below` resumes as soon as `not health.high_watermark_hit` (resident < high), so `low_watermark_hit` is never read. The pause-at-high / resume-at-low hysteresis this module documents does not actually hold: a producer sitting near the high watermark flaps one-in/one-out on every consumer release. Resume should gate on `low_watermark_hit`, otherwise the low watermark is dead.

nemo_automodel/components/speculative/streaming/queue.py
```diff
@@ -0,0 +1,357 @@
...
+        *,
+        visibility_timeout: VisibilityTimeout | None = None,
+        high_watermark_bytes: int | None = None,
+        low_watermark_bytes: int | None = None,
+        on_pause: Callable[[StoreHealth], None] | None = None,
+        on_resume: Callable[[StoreHealth], None] | None = None,
+    ) -> None:
+        self._store = store
+        self._vt = visibility_timeout or VisibilityTimeout()
+        self._high_bytes = high_watermark_bytes
```
khazic, line 177: `high_watermark_bytes` / `low_watermark_bytes` are stored in `self._high_bytes` / `self._low_bytes` but never read anywhere; backpressure comes entirely from the store's own `high_watermark_hit`. So these ctor args (and the docstring's `0.75 / 0.25 * capacity` defaults) silently do nothing. Either wire them into the resume decision or drop them to avoid a misleading API.

nemo_automodel/components/speculative/streaming/queue.py
```diff
@@ -0,0 +1,357 @@
...
+    def ack(self, lease: Lease) -> None:
+        """Mark a leased ref as successfully consumed and free its queue slot.
+
+        Does NOT touch the store -- the consumer's :meth:`FeatureStore.get`
+        return value carries a :class:`~nemo_automodel.components.speculative.streaming.store.StoreHandle`
+        that the consumer must hand to :meth:`FeatureStore.release` to drop
+        the tensors. The queue's responsibility ends at "lease no longer held".
+        """
+        with self._lock:
+            active = self._outstanding.pop(lease.ref.sample_id, None)
```
khazic: `_sample_counters` gets a `setdefault(sample_id, 0)` on every `put` but is never removed on a terminal `ack` (nor cleared in `close`), so it grows one entry per unique `sample_id` for the life of the run. On a long streaming job that is an unbounded leak. Pop `lease.ref.sample_id` here once the lease is acked for good.

nemo_automodel/components/speculative/streaming/queue.py
```diff
@@ -0,0 +1,357 @@
...
+        so the queue can reclaim the slot and the store can drop the
+        sample.
+        """
+        with self._put_cv:
+            if self._closed and not self._pending and not self._outstanding:
+                return None
+            if not self._pending:
+                self._put_cv.wait(timeout=poll_interval)
+                if not self._pending:
+                    return None
```
khazic, line 366: `acquire` returns `None` on a transient empty poll here, but also on closed-and-drained (line 240), and the docstring says `None` means shutdown. A consumer cannot tell "nothing right now, retry" from "stop": breaking on `None` exits early whenever the producer is momentarily behind, and there is no public `closed` getter to disambiguate. Consider a distinct close signal so the two cases are separable.

kashif:
All four review findings fixed in commit 1b575a37 on this branch (PR 2 rebased onto it):

1. **Hysteresis** (`put_blocks_until_below`): resume now gates on `low_watermark_hit`, pause still gates on `high_watermark_hit`, in the band the producer's existing state is preserved. A producer at resident=high-1 now stays paused until the consumer drains past low.

2. **Ctor args wired in**: `_should_pause` / `_should_resume` read the queue's `high_watermark_bytes` / `low_watermark_bytes` first (resident >= high_bytes to pause, <= low_bytes to resume) and fall back to the store's health() booleans when ctor args are None. Misconfigured pairs (low >= high, non-positive) raise `ValueError` at construction. Docstring's stale `0.75 / 0.25 * capacity` defaults are gone.

3. **`_sample_counters` leak**: popped in `ack()` once the lease is removed from outstanding. `fail` and `reclaim_expired` re-set the counter on re-enqueue so the redelivery bookkeeping stays consistent. One test pins this; another pins `fail` -> `ack` terminal behavior.

4. **`is_closed` property**: `acquire()` docstring now distinguishes the two `None` cases and points at `is_closed`. New test pins both branches.

The pre-existing `test_queue_pause_and_resume_callbacks_fire_on_watermark_transitions` test relied on the old broken behavior (only released `p1`, leaving resident at ~32 KiB which equals low) -- updated it to release both warm and p1 so resident drops strictly below low. 18/18 queue tests pass, 66/66 streaming tests pass, 831 passed / 12 skipped wider sweep.

nemo_automodel/components/speculative/streaming/queue.py
```diff
@@ -0,0 +1,437 @@
...
+    def ack(self, lease: Lease) -> None:
+        """Mark a leased ref as successfully consumed and free its queue slot.
+
+        Does NOT touch the store -- the consumer's :meth:`FeatureStore.get`
+        return value carries a :class:`~nemo_automodel.components.speculative.streaming.store.StoreHandle`
+        that the consumer must hand to :meth:`FeatureStore.release` to drop
+        the tensors. The queue's responsibility ends at "lease no longer held".
+        """
+        with self._lock:
+            active = self._outstanding.pop(lease.ref.sample_id, None)
```
khazic: A late ACK for an expired lease removes whichever lease currently has the same `sample_id`. If the ref has already been reclaimed and re-leased, this ACK silently discards the new consumer's active lease. Verify the active lease identity or use a unique lease token before removing it. The same issue applies to `fail`.

nemo_automodel/components/speculative/streaming/queue.py
```diff
@@ -0,0 +1,437 @@
...
+                            health.capacity_bytes,
+                        )
+                        self._producer_paused = True
+                        if self._on_pause is not None:
+                            try:
+                                self._on_pause(health)
+                            except Exception:
+                                logger.exception("on_pause callback raised; continuing")
+                    else:
+                        self._pending.append(ref)
```
khazic, line 329: Unlike `put`, this path never checks whether the `sample_id` is pending or outstanding. Enqueuing the same ref twice lets the second `acquire` overwrite `_outstanding[sample_id]`, which breaks lease accounting. Apply the duplicate check atomically before both append paths.

nemo_automodel/components/speculative/streaming/stores/local.py
```diff
@@ -0,0 +1,390 @@
...
+            )
+            return out, handle
+
+    def release(self, handle: StoreHandle) -> None:
+        if handle.store is not self:
+            raise ValueError(
+                f"StoreHandle was issued by a different store (id {id(handle.store):x}), cannot release it here"
+            )
+        with self._lock:
+            count = self._handle_refs.get(handle.sample_id, 0)
```
khazic: Each `get` returns an indistinguishable handle for the same sample, so releasing one handle twice decrements the shared count twice and can delete storage while another handle is still live. Give handles unique identities and record which ones have already been released to make `release` genuinely idempotent.

nemo_automodel/components/speculative/streaming/refs.py
```diff
@@ -0,0 +1,290 @@
...
+            weight resync lands in PR 4).
+        draft_weight_version: Same idea for the draft model's weights, so a
+            consumer can refuse to train against a ref produced before its
+            own weight snapshot. Starts at ``"0"``; PR 4 wires the resync.
+    """
+
+    sample_id: str
+    run_id: str
+    store_uri: str
+    feature_keys: dict[str, str]
```
khazic, line 169: `frozen=True` does not freeze these dictionaries. Callers can mutate `feature_keys` or `feature_specs` after validation, including inserting a tensor, so the advertised immutable tensor-free control-plane contract does not hold. Copy the mappings into an immutable representation during `__post_init__`.

kashif:
Four round-2 review comments addressed in commit `af9c3cea8` on this branch (rebased forward on PR 2 / PR 3 fixes):

1. **Late ACK for expired lease** (queue.ack / fail): `Lease` now carries a per-acquire unique `lease_id`; `_outstanding` is keyed by lease_id and `ack`/`fail` verify the identity before mutating. A late ACK for a reclaimed-then-re-leased ref is a no-op and leaves the new active lease intact.
2. **`put_blocks_until_below` duplicate check**: the blocking put now performs the same atomic duplicate-id check as `put()`, against `_pending_seen` and `_active_by_sample`.
3. **Indistinguishable handles**: `StoreHandle` now carries a per-get unique `handle_id`; `LocalFeatureStore.release` keys by handle_id and is a true per-handle no-op. Sibling handles of the same sample no longer over-decrement.
4. **`SampleRef` dict mutation**: `feature_keys` and `feature_specs` are wrapped in `MappingProxyType(dict(...))` during `__post_init__` so item assignment / deletion raises `TypeError` -- the tensor-free control-plane contract is enforced.

Tests added for each (12 new tests, all passing). PR 2 and PR 3 rebase onto this commit to pick the lease/handle identity work up automatically.

`git log` on this branch:
- `98fdd6059` Merge branch 'main' into kashif/feat/streaming-data-plane-pr1
- older: original PR 1 commits

Wider sweep: 955 passed / 12 skipped.

nemo_automodel/components/speculative/streaming/stores/local.py
```diff
@@ -0,0 +1,408 @@
...
+            draft_weight_version=draft_weight_version,
+        )
+
+    @property
+    def store_uri(self) -> str:
+        # Stable URI the queue's lease / ack protocols can match on. PR 3's
+        # SharedDirFeatureStore will return a different scheme ("file://"),
+        # so the queue can refuse to lease a ref whose URI does not match
+        # its bound store.
+        return f"mem://local-{id(self):x}"
```
khazic: `id(self)` is not a stable identity. CPython reuses addresses, so a store that is garbage collected and a new one allocated at the same address produce the same `store_uri`, and `get()`'s URI check would then accept a ref belonging to the dead store. The module docstring says refs travel over queues, HTTP, and checkpoint metadata, where an address-derived URI is meaningless. A per-instance uuid4 (or an explicit caller-supplied name) would give the URI the stability the contract claims.

nemo_automodel/components/speculative/streaming/queue.py
```diff
@@ -0,0 +1,531 @@
...
+    def pending_count(self) -> int:
+        with self._lock:
+            return len(self._pending)
+
+    def outstanding_count(self) -> int:
+        with self._lock:
+            return len(self._outstanding)
+
+    def close(self) -> None:
+        """Drain the queue; subsequent :meth:`acquire` calls return ``None``.
```
khazic: The docstring says subsequent `acquire` calls return `None`, but `acquire` only returns `None` when `_closed and not _pending and not _outstanding`, so after `close()` it keeps handing out the remaining pending refs. The drain behaviour looks intentional; the docstring should say so, since `FeatureDataLoader` reads `is_closed` to decide when to raise `StopIteration` and the two readings disagree about what happens to the leftover refs.

kashif:
Superseded by #3085: this was the base of the stack, and its contents landed in main as part of the #3085 squash-merge (which was stacked on top of it). Closing as redundant; #3094 carries the remaining async producer + SharedDirFeatureStore work.

PR #3085, opened 2026-07-15
feat(streaming): add EAGLE-3 producer and loader over the local store

kashif:
PR 2 of NVIDIA-NeMo/Automodel#3062. Closes the minimal end-to-end loop on top of NVIDIA-NeMo/Automodel#3084: a colocated target forward through a producer + queue + LocalFeatureStore, materialized by a loader that yields `Eagle3TargetBatch` to the trainer. No recipe or dispatch changes (per the RFC's "library layer first" answer to Q1).

## What does this PR do ?

PR 2 of NVIDIA-NeMo/Automodel#3062. Adds the EAGLE-3 producer/loader pair on top of PR 1's data-plane contracts and verifies numerical parity against the existing colocated `HFEagle3TargetModel.generate_batch` path.

## Changelog

* New `nemo_automodel/components/speculative/streaming/` modules:
  * `eagle3.py`: the EAGLE-3 schema (core features + exactly-one supervision encoding) plus the helpers that map an `Eagle3TargetBatch` into the producer's tensors + per-feature `FeatureSpec` dict. `validate_eagle3_ref` is the gate the loader runs.
  * `producer.py`: `FeatureProducer` wraps the existing colocated `HFEagle3TargetModel`, runs its `generate_batch`, packs the tensors under the schema, and returns a tensor-free `SampleRef`. Algorithm is inferred from the backend's runtime type (`*Eagle3TargetModel`) with an explicit override. Rejects a backend that emits the precomputed draft-vocab encoding (the colocated producer only carries full logits).
  * `loader.py`: `FeatureDataLoader` is a Python iterator over `Eagle3TargetBatch`. Releases the previous batch's lease + store handle on the NEXT pull (so the trainer can hold a batch across one forward without it being freed mid-forward). `consume_now()` releases eagerly after `backward()`. `close()` / context manager releases on iteration end. Refs whose algorithm does not match the loader's are failed back to the queue rather than silently dropped.
* 34 new tests under `tests/unit_tests/speculative/streaming/`:
  * `test_eagle3.py`: schema validator + tensor / feature-spec packing.
  * `test_producer.py`: round-trip, algorithm inference, draft-vocab encoding rejection, pass-throughs.
  * `test_loader.py`: release-previous-on-next, consume_now, close, algorithm-mismatch redelivery.
  * `test_eagle3_parity.py`: bit-identical `Eagle3TargetBatch` across `HFEagle3TargetModel.generate_batch` (colocated) and `FeatureProducer + FeatureDataLoader` (streaming) over the same input. The numerical-equivalence guarantee PR 2 lands on.
* `ruff format` clean, `ruff check` clean.
* Wider sweep: `859 passed, 12 skipped` across `tests/unit_tests/speculative/`. Zero regressions.

## Additional Information

* Builds on NVIDIA-NeMo/Automodel#3084 (PR 1)
* Out of scope (deferred): recipe wiring for `target_model_backend: streaming`, packed-sequence producer metadata threading, async prefetch, draft-vocab encoding producer.


nemo_automodel/components/speculative/streaming/loader.py
```diff
@@ -0,0 +1,201 @@
...
+    def __iter__(self):
+        return self
+
+    def __next__(self) -> Eagle3TargetBatch:
+        self._release_pending()
+        if self._closed:
+            raise StopIteration
+        lease = self._queue.acquire()
+        if lease is None:
+            raise StopIteration
```
khazic: `acquire()` also returns `None` for a transient empty poll while the queue is still open. Raising `StopIteration` here makes a normal producer delay permanently terminate the training loop. Please retry until `queue.is_closed` indicates the stream is drained.

nemo_automodel/components/speculative/streaming/eagle3.py
```diff
@@ -0,0 +1,161 @@
...
+        raise ValueError(
+            f"SampleRef.algorithm must be {FeatureAlgorithm.EAGLE3!r} for an EAGLE-3 consumer, got {ref.algorithm!r}"
+        )
+    keys = set(ref.feature_specs)
+    missing = set(EAGLE3_CORE_FEATURES) - keys
+    if missing:
+        raise ValueError(f"EAGLE-3 ref missing required core features {sorted(missing)}; present={sorted(keys)}")
+    has_logits = "logits" in keys
+    has_draft = "target_probs" in keys and "position_mask" in keys
+    if has_logits == has_draft:
```
khazic: This accepts `logits` together with only one draft-vocab field because `has_draft` is false. That violates the documented `logits` alone contract and later silently selects the logits path. Please reject either draft-vocab key whenever `logits` is present.

nemo_automodel/components/speculative/streaming/queue.py
```diff
@@ -0,0 +1,437 @@
...
+    def ack(self, lease: Lease) -> None:
+        """Mark a leased ref as successfully consumed and free its queue slot.
+
+        Does NOT touch the store -- the consumer's :meth:`FeatureStore.get`
+        return value carries a :class:`~nemo_automodel.components.speculative.streaming.store.StoreHandle`
+        that the consumer must hand to :meth:`FeatureStore.release` to drop
+        the tensors. The queue's responsibility ends at "lease no longer held".
+        """
+        with self._lock:
+            active = self._outstanding.pop(lease.ref.sample_id, None)
```
khazic: A stale lease can pop a newer active lease for the same sample. After visibility reclaim and redelivery, a late ACK from the first consumer removes the second consumer lease. Please verify the stored lease identity or generation before removing it.

kashif:
Three round-2 review comments addressed:

1. **acquire() raises StopIteration on empty poll** (loader.py: `__next__`): the loader now loops with a small `acquire_poll_interval` sleep and exits only on `queue.is_closed` -- the canonical close-signal from PR 1's fix. A producer that briefly stalls no longer terminates the trainer mid-iteration. New ctor kwarg `acquire_poll_interval` (default 50ms) controls the sleep.
2. **validate_eagle3_ref accepts partial draft-vocab** (eagle3.py): the validator now rejects `target_probs` without `position_mask` (and vice versa) via XOR, and rejects any mix of full + draft supervision keys. A producer that emits only one draft key no longer silently takes the wrong encoding path.
3. **Stale lease pop** (queue.ack / fail): same fix as PR 1 #1 -- `Lease.lease_id` identity check. A late ACK from the first consumer no longer pops the second consumer's active lease.

This branch is now `31b8069c3 fix(streaming): loader polls on empty queue, validate partial draft-vocab` on top of PR 1's `af9c3cea8 fix(streaming): lease + handle identity, ref deep-freeze, put duplicate check`. 155 streaming tests pass; 955 wider sweep.

`git log`:
- `31b8069c3` fix(streaming): loader polls on empty queue, validate partial draft-vocab

nemo_automodel/components/speculative/streaming/loader.py
```diff
@@ -0,0 +1,228 @@
...
+            self._queue.fail(lease)
+            raise ValueError(
+                f"FeatureDataLoader is bound to algorithm={self._algorithm!r} but "
+                f"received ref with algorithm={ref.algorithm!r} for sample_id={ref.sample_id!r}"
+            )
+        if self._algorithm.value == "eagle3":
+            try:
+                validate_eagle3_ref(ref)
+            except ValueError:
+                self._queue.fail(lease)
```
khazic: A permanently malformed ref poisons the stream. `fail()` re-enqueues the ref at the tail, and the loader has no retry budget, so the next `__next__` leases the same ref, fails the same validation, and re-enqueues it again. `SampleRefQueue.fail`'s own docstring says "a permanently bad ref is the consumer's problem (drop it after a bounded retry budget)", but that budget is not implemented anywhere. A schema error is deterministic, so this should ack-and-drop (or drop once `lease.redelivery_count` exceeds a threshold) rather than `fail()`.

nemo_automodel/components/speculative/streaming/producer.py
```diff
@@ -0,0 +1,248 @@
...
+            sample_id,
+            tensors,
+            run_id=self._run_id,
+            algorithm=self._algorithm,
+            schema_version=EAGLE3_SCHEMA_VERSION,
+            target_model_version=self._target_model_version,
+            draft_weight_version=self._draft_weight_version,
+            num_tokens=num_tokens,
+        )
+        if ref.feature_specs != feature_specs:
```
khazic, line 237: This self-check runs after `store.put` has already committed the sample. When it trips, the exception propagates and the `SampleRef` is discarded, so nothing can ever `get`/`release` that sample and its bytes stay resident in the store for the life of the process. Either compare the specs before the put, or release the sample in the failure path.

HuiyingLi (approved):

PR #3094, opened 2026-07-15
feat(streaming): async producer + SharedDirFeatureStore

kashif:
PR 3 of NVIDIA-NeMo/Automodel#3062. Producer runs ahead of the trainer in a background thread, gating its pacing through the queue's HWM/LWM hysteresis; a new SharedDirFeatureStore (safetensors-backed) lets producer and trainer live in separate processes on a shared mount.

## What does this PR do ?

PR 3 of NVIDIA-NeMo/Automodel#3062. Adds the asynchronous producer + the file-backed feature store.

## Changelog

* New `nemo_automodel/components/speculative/streaming/` modules:
  * `stores/shared_dir.py`: `SharedDirFeatureStore` writes one `<sample_id>.safetensors` per produced sample into a shared directory. Atomic writes (tmp + os.replace). Per-instance file ownership so release only unlinks files this store put. Same residency caps + `health()` contract as `LocalFeatureStore`. `close()` unlinks every still-owned file.
  * `async_pipeline.py`: `AsyncFeaturePipeline` wraps a `FeatureProducer` (PR 2) in a daemon background thread, draining a `PromptSource` (callable or iterator). Backpressure honors `queue.put_blocks_until_below`. `PromptSource` / producer errors are captured and re-raised from `stop()`. Context-manager entry / exit for clean lifecycle.
* 27 new tests under `tests/unit_tests/speculative/streaming/`:
  * `test_shared_dir_store.py`: round-trip via tmp_path, detached copies, URI mismatch rejection, path-escape refusal, sample + byte caps, ctor validation, refcount + idempotent release, health watermark snapshot, close unlinks owned files, atomic write leaves no leftover tmp files.
  * `test_async_pipeline.py`: prompt drain, producer-runs-ahead-of-loader overlap, idempotent start/stop, thread join, context-manager drain, callable + iterator prompt sources, streaming polling mode, prompt-source / producer error propagation, backpressure interaction with a draining consumer.
* `ruff format` clean, `ruff check` clean.
* Wider sweep: `892 passed, 12 skipped` across `tests/unit_tests/speculative/`. Zero regressions.

## Additional Information

* Builds on NVIDIA-NeMo/Automodel#3084 (PR 1) and NVIDIA-NeMo/Automodel#3085 (PR 2)
* Out of scope: DFlash / DSpark producer, async loader (prefetched batches), `NcclFeatureStore`, multi-replica controller, DP resharding (all PR 4 territory)


nemo_automodel/components/speculative/streaming/loader.py
```diff
@@ -0,0 +1,214 @@
...
+    def __iter__(self):
+        return self
+
+    def __next__(self) -> Eagle3TargetBatch:
+        self._release_pending()
+        if self._closed:
+            raise StopIteration
+        lease = self._queue.acquire()
+        if lease is None:
+            raise StopIteration
```
khazic: `acquire()` also returns `None` for a transient empty poll while the async producer is still running. Raising `StopIteration` here terminates the trainer before later refs arrive. Please retry until the queue is closed and drained.

nemo_automodel/components/speculative/streaming/async_pipeline.py
```diff
@@ -0,0 +1,244 @@
...
+                    input_ids=input_ids,
+                    attention_mask=attention_mask,
+                    loss_mask=loss_mask,
+                    **packing_kwargs,
+                )
+                # put_blocks_until_below honors the queue's HWM/LWM
+                # hysteresis; a fast producer here naturally blocks
+                # when the store is at high watermark.
+                try:
+                    self._queue.put_blocks_until_below(
```
khazic, line 247: The ref has already been written to the store before this watermark check. If that write reaches the HWM, this call waits for the store to drain, but the ref is not queued so no consumer can drain it. Please gate before producing or enqueue this ref before pausing subsequent production.

nemo_automodel/components/speculative/streaming/stores/shared_dir.py
```diff
@@ -0,0 +1,388 @@
...
+            )
+        with self._lock:
+            count = self._handle_refs.get(handle.sample_id, 0)
+            if count <= 0:
+                logger.debug(
+                    "SharedDirFeatureStore release sample_id=%s is a no-op (already released)",
+                    handle.sample_id,
+                )
+                return
+            count -= 1
```
khazic: Release state is tracked only by sample count, not by handle. With two handles, releasing the first handle twice decrements both references and unlinks the file while the second handle is still outstanding. Please make idempotency specific to each handle.

kashif:
Three round-2 review comments addressed:

1. **acquire() raises StopIteration on empty poll** (carried over from PR 2 fix; this branch includes the PR 1 + PR 2 fixes via rebase): same fix as PR 2 #1 -- the loader polls on `queue.is_closed` instead of breaking on transient empty polls.
2. **Ref written to store before watermark check** (async_pipeline.py: `_run`): the producer thread now gates BEFORE calling `produce()`, polling `store.health().low_watermark_hit` until the store reports headroom. The wrapped producer's `store.put` runs inside its forward pass, so producing while the store is at HWM would leave the ref in the store with no consumer awareness. New `_wait_for_capacity` is `stop_event`-aware so shutdown stays responsive. The same fix carries the lease / handle identity from PR 1 via rebase.
3. **Release state by sample count, not by handle** (carried over from PR 1 fix via rebase): `StoreHandle` carries `handle_id`, `release` is per-handle, sibling handles of the same sample no longer over-decrement.

This branch is `674cfad85 fix(streaming): gate async producer before forward, drop PR refs` on top of PR 1 + PR 2 fixes. 155 streaming tests pass with both the project venv (Python 3.12) and the user's venv (Python 3.13); 955 wider sweep pass.

`git log`:
- `674cfad85` fix(streaming): gate async producer before forward, drop PR refs
- `31b8069c3` fix(streaming): loader polls on empty queue, validate partial draft-vocab

nemo_automodel/components/speculative/streaming/stores/shared_dir.py
```diff
@@ -0,0 +1,407 @@
...
+                resident_bytes,
+            )
+
+    def gc(self) -> int:
+        # Release already unlinks files when the last handle drops; a
+        # controller could override this to reclaim files from crashed producers.
+        return 0
+
+    def health(self) -> StoreHealth:
+        sample_count, resident_bytes = self._directory_residency()
```
khazic: This violates the `health()` contract the ABC declares: "Must not block on tensor I/O ... in-memory counters are sufficient. Backends that do background I/O MUST serve health() from cached counters." Every call does an `os.listdir` plus a `getsize` per file, and `SampleRefQueue.put_blocks_until_below` polls `health()` every `poll_interval` (50ms by default) while the producer is paused. On the shared POSIX mount this store exists for, that is a sustained metadata storm against the filesystem. Maintain resident bytes / sample count as in-memory counters updated in `put`/`release` and serve `health()` from those.

nemo_automodel/components/speculative/streaming/stores/shared_dir.py
```diff
@@ -0,0 +1,407 @@
...
+            raise ValueError(f"sample_id must be str, got {type(sample_id).__name__}")
+        if not tensors:
+            raise ValueError("tensors must be non-empty so a SampleRef has at least one feature")
+        path = self._path_for(sample_id)
+        with self._lock:
+            if self._closed:
+                raise RuntimeError("SharedDirFeatureStore is closed; no further puts accepted")
+            if sample_id in self._owned_files or os.path.isfile(path):
+                raise ValueError(f"sample_id already present in store: {sample_id}")
+            sample_count, resident_bytes = self._directory_residency()
```
khazic: The residency caps are per-instance config but measured over the whole shared directory. The class docstring says each process owns its own instance pointing at the same directory, so with N ranks producing concurrently every rank sees the aggregate footprint and hits `max_bytes` / `max_samples` at 1/N of its configured budget. Either scope the count to files this instance owns (`_owned_files`) or document that the caps are directory-global and must be divided across producers.

nemo_automodel/components/speculative/streaming/stores/shared_dir.py
```diff
@@ -0,0 +1,407 @@
...
+                sample_count=sample_count,
+                high_watermark_hit=(self._high_watermark is not None and resident_bytes >= self._high_watermark),
+                low_watermark_hit=(self._low_watermark is not None and resident_bytes <= self._low_watermark),
+            )
+
+    def close(self) -> None:
+        with self._lock:
+            if self._closed:
+                return
+            outstanding = len(self._handle_refs)
```
khazic: `LocalFeatureStore.close()` clears its state unconditionally while this one raises when handles are outstanding. Two implementations of the same ABC with different shutdown semantics will bite whichever caller closes a store generically; the ABC docstring only promises idempotency and says nothing about rejecting the call.

nemo_automodel/components/speculative/streaming/async_pipeline.py
```diff
@@ -0,0 +1,244 @@
...
+                    if self._stop_event.is_set() and (
+                        "aborted during shutdown" in str(e) or "closed while put was waiting" in str(e)
+                    ):
+                        break
+                    raise
+        except BaseException as e:  # noqa: BLE001 -- capture then re-raise from stop()
+            logger.exception("AsyncFeaturePipeline background thread failed")
+            self._error = e
+        finally:
+            self._queue.close()
```
khazic, line 267: Closing the queue in `finally` makes `stop()` terminal, but `start()` is documented as idempotent and clears `_stop_event`/`_joined_event` as if a restart were supported. After a `stop()`/`start()` cycle the first `put_blocks_until_below` raises `RuntimeError: SampleRefQueue is closed`. Either close the queue only when the prompt source is genuinely exhausted, or document that the pipeline is single-use.

nemo_automodel/components/speculative/streaming/async_pipeline.py
```diff
@@ -0,0 +1,244 @@
...
+                if isinstance(prompt, tuple) and len(prompt) == 6:
+                    input_ids, attention_mask, loss_mask, position_ids, seq_lens, doc_remaining = prompt
+                    packing_kwargs = {
+                        "position_ids": position_ids,
+                        "seq_lens": seq_lens,
+                        "doc_remaining": doc_remaining,
+                    }
+                else:
+                    input_ids, attention_mask, loss_mask = prompt
+                ref = self._producer.produce(
```
khazic, line 237: If the `put_blocks_until_below` below aborts (shutdown) or raises, the sample `produce()` just wrote is already resident in the store but its ref is dropped, so nothing can release it. On `SharedDirFeatureStore` that leaves an orphaned `.safetensors` file in the shared directory that only a later `close()` on that same instance would clean up. Worth releasing the sample when the put does not land.

HuiyingLi (dismissed):

HuiyingLi (approved):

PR #2930, opened 2026-07-05
feat(moe): DeepEP v2 dispatcher, Torch 2.13 PP, and selectable EP variants

hemildesai:
# What does this PR do?

This PR upgrades the MoE stack to DeepEP v2 while retaining selectable DeepEP v1
and HybridEP builds. It adds the DeepEP v2 ElasticBuffer dispatcher, Torch 2.13
pipeline/FSDP synchronization fixes, tuned communication defaults, benchmark and
finetuning recipes, HybridEP JIT-cache reuse, and a selectable CUDA development
image.

The branch includes `main` through `a664ea4a242883a394c812232ed46b4e426c2472`.
All 26 commits are signed off.

This replaces the earlier stacked PRs #2850, #2782, #2851, #2853, #2854, and
#2855. Those PRs are closed; this is a standalone PR against `main`.

# Changelog

- **build(moe):** add conflicting `uv` extras for DeepEP v1, DeepEP v2, and
  HybridEP (`deepep_v1`, `deepep_v2`, `hybridep`; `moe` selects v2 by default),
  with updated dependency pins and lockfiles.
- **feat(moe):** add a DeepEP v2 ElasticBuffer dispatcher with non-reentrant
  activation-checkpointing support and SM/QP configuration propagated through
  dispatch and backward combine.
- **fix(moe):** initialize the ElasticBuffer during MoE parallelization with the
  actual EP process group. Non-PP fixed and packed recipes infer capacity from
  local batch size and sequence length during distributed setup; PP uses
  microbatch size and sequence length. Unknown dynamic shapes retain a one-time
  first-dispatch fallback, and the training loop has no dispatcher-specific
  buffer lifecycle.
- **perf(moe):** share the dispatch-owned ElasticBuffer handle checkpoint state
  with the paired combine operation. This prevents non-reentrant checkpoint
  unpack from replaying forward combine communication.
- **fix(pp):** adapt pipeline execution and FSDP MoE gradient synchronization to
  Torch 2.13, including final-microbatch reduction without duplicate reduction.
- **perf(moe):** preserve Qwen throughput after the rebase by disabling the outer
  FSDP wrapper for PP1 Qwen3 30B.
- **test/perf(moe):** add DeepEP v2 dispatcher tests and Qwen3 MoE benchmark and
  finetuning recipes.
- **perf(moe):** reuse HybridEP JIT kernels through a stable cache bridge.
- **build(docker):** add a selectable EP CUDA development image with
[...]

PR #3095, opened 2026-07-16
feat(models): add Inkling VLM MoE support

hemildesai:
## Summary

- add a native Inkling VLM MoE implementation, processor setup, and exact raw-checkpoint state-dict conversion
- register Inkling and add a MedPix fine-tuning recipe using PP8, EP32, FSDP2, non-reentrant activation checkpointing, and `torch_mm` experts
- keep Inkling's fp32 short-convolution parameters in model-local `_fp32_params` holders that reuse the existing FSDP holder path; `components/moe/parallelizer.py` is unchanged from `main`
- support flat multimodal patch batches through the shared VLM collation and pipeline-media paths
- support PyTorch 2.12 pipeline output metadata and reuse PR #2930's centralized PP neighbor-communicator warmup and static-metadata selection; Inkling has no model- or recipe-specific warmup hook
- pin the Transformers commit containing upstream Inkling support and add model coverage documentation

## Validation

- [100-step MedPix fine-tuning after final simplification](https://wandb.ai/nvidia/automodel-inkling/runs/eijh41si)
  - Slurm job `14060480`: 256 H100 GPUs, PP8/EP32, global batch 256, sequence length 2048
  - completed steps 0-99 in 34:23 with exit `0:0` using NCCL's default communicator blocking behavior
  - loss decreased from `2.3000` to `1.7906`; final grad norm `0.7824`, memory `58.53 GiB`, and throughput `4,415.29` tokens/s
  - matched the [previous 100-step baseline](https://wandb.ai/nvidia/automodel-inkling/runs/5kowsm5h), which ended at loss `1.7898`, grad norm `0.8266`, memory `58.54 GiB`, and throughput `4,316.08` tokens/s
- focused unit, parity, PP warmup, mesh, VLM batching, and MoE sharding suite: 122 passed in 8.48s (Slurm job `14060405`, exit `0:0`)
- real two-GPU NCCL PP warmup, forward/backward, static metadata, and sequence-length reset smoke with default communicator blocking (Slurm job `14060410`, exit `0:0`)
- EP2 + FSDP2 + non-reentrant activation-checkpointing optimizer smoke using the unchanged `main` MoE parallelizer (Slurm job `14057832`, exit `0:0`)
- four-layer forward/backward parity and full checkpoint validation
- all 1,392 non-MTP checkpoint keys and converted tensor shapes validated in both directions
- pre-commit hooks, Ruff, `uv lock --check`, `compileall`, and `git diff --check`

## Caveat

- MTP tensors are intentionally ignored because the standard Hugging Face conditional-generation model does not expose the MTP training path.


yuhezhang-ai:
Hi @hemildesai  Thanks for the great work. Since #2998, my earlier TieSupport refactor, has merged, every model with an LM head now needs to declare its tying policy. I added a small follow-up marking Inkling as UNTIED_ONLY with focused tests. This matches its existing separate embed/unembed weights and should not change any of your implementation’s behavior.

docker/common/uv-pytorch.lock
```diff
@@ -4178,7 +4178,7 @@ requires-dist = [
     { name = "torchvision", marker = "sys_platform != 'darwin' and sys_platform != 'linux' and extra == 'diffusion'", index = "https://download.pytorch.org/whl/cpu" },
     { name = "torchvision", marker = "sys_platform == 'linux' and extra == 'diffusion'", index = "https://download.pytorch.org/whl/cu130" },
     { name = "transformer-engine", extras = ["pytorch"], marker = "extra == 'cuda'", specifier = ">=2.14.1" },
-    { name = "transformers", specifier = "==5.12.1" },
+    { name = "transformers", git = "https://github.com/huggingface/transformers.git?rev=c375c5e8666aee864528aa52e8287b3ba82232bb" },
```
HuiyingLi: Are we ready to change to this version?

nemo_automodel/components/datasets/vlm/collate_fns.py
```diff
@@ -424,6 +424,18 @@ def build_labels_from_template(
     processor_type = type(processor).__name__
     tokenizer = getattr(processor, "tokenizer", processor)
 
+    # Inkling closes an assistant message before emitting a separate sampling
+    # terminator. Label through the first sampling terminator, which also keeps
+    # subsequent padding (using the same token id) masked.
+    if processor_type == "InklingProcessor":
```
HuiyingLi, line 430: A separate collate_fn for inkling, wdyt?

docker/common/uv-pytorch.lock
hemildesai: The bump is required: Inkling isn't in released transformers ≤ 5.12.1 — the model code imports `transformers.models.inkling.*` (configuration/modeling/processing), which only exist in newer transformers.

Good news since this pin was added: transformers **5.14.1 is now released on PyPI and ships the Inkling model** (`src/transformers/models/inkling/`). So we no longer need the unreleased git commit — we can point `[project.dependencies]` at `transformers==5.14.1` and drop the `[tool.uv.sources]` git override, then re-lock.

That also clears the current CI failure: the Pip install jobs fail because `pip` ignores `[tool.uv.sources]` (uv-only) and can't resolve `transformers==5.14.0.dev0` from PyPI, so every `Pip - …` job errors with `No matching distribution found for transformers==5.14.0.dev0`. Pinning to the released `5.14.1` makes both pip and uv resolve it.

One heads-up before flipping the repo to 5.14: it surfaces `generate()` API changes unrelated to Inkling — the `nemotron_v3` and `qwen3_5` L0 tests fail with `get_experts_implementation` / `layer_type` AttributeErrors. Those need a compat fix (or test adjustment) as part of the upgrade, so it's worth scoping that alongside the version switch.

nemo_automodel/components/datasets/vlm/collate_fns.py
hemildesai: Inkling actually reuses the shared `default_collate_fn` — the only Inkling-specific touchpoints are the `qwen-vl-utils` import guard (the `!= "InklingProcessor"` check) and the label-marker branch in `build_labels_from_template`, which mirrors the existing per-processor branches there (the Qwen `<|im_start|>` fast path and the Gemma template-derived path).

A dedicated Inkling collate_fn would have to duplicate the ~80 lines of shared batching / label-building / PP-media scaffolding just for those two small hooks, so I kept it on the shared path to stay consistent with how Qwen/Gemma are handled.

If the concern is model-specific logic living in the shared file, I'm happy to instead pull the Inkling marker resolution into a small helper (or into the inkling model package) rather than a full separate collate_fn — let me know which you'd prefer.

nemo_automodel/components/models/inkling/layers.py
```diff
@@ -0,0 +1,294 @@
...
+        shared_gammas = topk_weights[..., -self.n_shared_experts :].contiguous()
+        topk_weights = topk_weights[..., : self.top_k].contiguous()
+        return topk_weights, topk_indices, shared_gammas
+
+    def update_bias(self) -> None:
+        """No-op: Inkling's correction bias is a trained parameter, not an aux-loss update."""
+        return
+
+
+def inkling_swiglu(hidden_states: torch.Tensor, routing_weights: torch.Tensor) -> torch.Tensor:
```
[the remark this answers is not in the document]
hemildesai: Added a Google-style Args/Returns documenting the interleaved `[g0, u0, g1, u1, ...]` gate/up layout (`::2` gate, `1::2` up) and the broadcastable `routing_weights`. Fixed in 2e909a2.

nemo_automodel/components/datasets/vlm/pp_media.py
```diff
@@ -180,6 +180,39 @@ def chunk_step3_media(
     return result
 
 
+def chunk_inkling_media(
```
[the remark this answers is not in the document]
hemildesai: `chunk_inkling_media` was folded into `chunk_step3_media` when `main` was merged in (it now handles the flat-patch layout). Added a Google-style Args/Returns there documenting the flat/ragged `pixel_values` and `num_patches` layout. Fixed in 2e909a2.

nemo_automodel/components/models/inkling/layers.py
```diff
@@ -0,0 +1,294 @@
...
+    def __init__(self, config) -> None:
+        super().__init__()
+        model_dtype = get_dtype(getattr(config, "torch_dtype", None), torch.bfloat16)
+        self.gate_up_proj = nn.Parameter(
+            torch.empty(config.hidden_size, 2 * config.intermediate_size, dtype=model_dtype)
+        )
+        self.down_proj = nn.Parameter(torch.empty(config.intermediate_size, config.hidden_size, dtype=model_dtype))
+        self.global_scale = nn.Parameter(torch.ones(1, dtype=model_dtype))
+
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
```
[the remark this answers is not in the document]
hemildesai: Added a Google-style Args/Returns documenting the `[..., hidden]` in/out contract and the interleaved gate/up projection the `::2`/`1::2` slicing relies on. Fixed in 2e909a2.

docker/common/uv-pytorch.lock
hemildesai: Update: reverted — this PR now keeps `transformers` at main's `5.12.1` and no longer touches the version (`pyproject.toml` and both lockfiles are back to main). Pushed in `1de4e266`.

The Inkling model is registered lazily in `MODEL_ARCH_MAPPING` and only imported on first access, so it stays usable whenever a transformers build that ships Inkling is installed, without forcing the upgrade on every other model. The Inkling unit tests are now `importorskip`-guarded, so they skip cleanly on 5.12.1 and auto-run once a transformers release containing Inkling is pinned. The repo-wide 5.14 upgrade (and its `nemotron_v3` / `qwen3_5` `generate()` fallout) can be handled in a separate PR.

nemo_automodel/components/models/inkling/layers.py
```diff
@@ -0,0 +1,395 @@
...
+        **kwargs: Any,
+    ) -> torch.Tensor:
+        input_dtype = hidden_states.dtype
+        output = self._fp32_params(
+            hidden_states.float(),
+            past_key_values=past_key_values,
+            conv_mask=conv_mask,
+            **kwargs,
+        )
+        return output.to(dtype=input_dtype)
```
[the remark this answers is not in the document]
hemildesai: Added Google-style Args/Returns to both `InklingShortConvolution.forward` and `_InklingShortConvolutionFP32.forward` documenting the `[batch, sequence, hidden]` layout and the `[batch, sequence]` boolean `conv_mask`. Fixed in 6944208.

HuiyingLi (approved):

PR #3096, opened 2026-07-16
fix(checkpoint): gather PEFT adapter across PP stages

hyfine:
## Problem

When training a PEFT (LoRA) model with `pp_size > 1`, the saved adapter only
contains the layers of a single pipeline stage (e.g. a fraction of the model's
layers). Merging such an adapter silently degrades the model.

Root cause: for a plain (non-EP, non-quantized) PEFT model,
`ModelState.state_dict()` uses DCP `get_model_state_dict(full_state_dict=True)`.
Per PyTorch's contract that returns the full dict **only on global rank 0** and
an **empty dict on every other rank**, so PP ranks `1..N-1` contribute nothing
and the on-disk adapter is truncated to rank 0's stage.

The existing local-collection bypass only triggers for expert-parallel or
quantized models (`_has_expert_parallelism` / `_has_quantized_params`).
Grouped-experts MoE that store no `ep_size` therefore fall through to the DCP
path even under pipeline parallelism.

`adapter_config.json` has the mirror bug: `_get_hf_peft_config` reads only
`model_state.model[0]` (the first virtual stage), so `target_modules`
under-lists the trained layers.

## Fix

- `stateful_wrappers.py`: use the local per-rank PEFT collection whenever
  `pp_group is not None` (in addition to EP/quant), then all-gather the adapter
  tensors across the PP group and union by FQN. Layer indices never overlap
  across stages, so the union is exact. A guard warns if the merge did not grow
  beyond a single rank (mis-passed group). No-op when `pp_size == 1`.
- `addons.py`: extract `target_modules` / `target_parameters` across **all**
  local virtual stages (not just `model[0]`) and, when a PP group is supplied,
  union the names across the PP group so the config matches the gathered weights.
- Thread the PP group through `Checkpointer` -> `ModelState` via
  `config.build(pp_group=...)`; recipes obtain it from a new
  `BaseRecipe._get_pp_group()` helper (mirrors the existing `_get_pp_rank()`).

## Impact / compatibility

- `pp_size == 1`: `pp_group is None`, both paths short-circuit — behavior
  unchanged.
- EP / quantized PEFT: unchanged (still take the local bypass).
- Extra communication is one `all_gather_object` of LoRA tensors per checkpoint.

## Tests

- New `tests/unit_tests/checkpoint/test_peft_pp_gather.py` (CPU, fake
  `torch.distributed`): merge/union, dedup to lowest rank, single-rank
  short-circuit, degenerate-gather warning, `target_modules` PP union, and the
  no-PP-group backward-compatible path. Regression: `test_addons.py` passes.
- End-to-end (pp=4): per-rank tensor counts go from `[224, 0, 0, 0]` to
  `[224, 210, 210, 210]`; the saved adapter and `adapter_config.json` both cover
  every layer.


hyfine:
Thanks for kicking off the CI, @akoumpa  - all checks are green. 

Quick context on the fix: when training LoRA with pp_size > 1, the adapter
was saved with only one pipeline stage's layers, because the plain-PEFT path
goes through DCP full_state_dict, which returns data on rank 0 only and an
empty dict on the other PP ranks. Grouped-experts MoE that don't set ep_size
never hit the existing EP/quant bypass, so they fell into this path. This PR
routes PEFT+PP through the local per-rank collection and all-gathers the
adapter across the PP group (with a matching union for adapter_config.json's
target_modules). No-op when pp_size == 1.

Happy to adjust anything if you'd like a different approach. Thanks for the review!

akoumpa:
Thanks a lot @hyfine , I think what you have right now looks to me, no adjustment necessary, when you get the chance, please take a look at the conflicts, and then i can help retrigger the CI. Thanks again for the PR.

hyfine:
Thanks for the review! Addressed finding 1 in 339f2097 - the degenerate-gather
guard now only warns when more than one rank contributes adapters, so a valid
PP layout that places all adapters on a single stage (per_rank=[N, 0, ...]) no
longer false-positives. Added a unit test for that case.

On finding 2 (multi-process coverage): agreed a 2-rank torchrun functional test
would be ideal. There's no existing PP+PEFT functional harness in the repo, so
I'd prefer to add one in a follow-up rather than grow this PR - happy to do that
if you'd like it bundled here instead.

Note the L0_Unit_Tests_GPU failures are unrelated to this change: they're all in
test_lora_kernel.py failing at torch.cuda init with "NVIDIA driver too old
(version 12080)" - an infra/runner issue. This PR only touches the save path,
and the non-kernel checkpoint/PEFT tests pass in the same run.

akoumpa (approved):
Thanks a lot @hyfine

PR #3103, opened 2026-07-16
fix(retrieval): use stock Ministral embedding backbone

oliverholworthy:
# What does this PR do ?

Configure Hugging Face retrieval embedding fallback backbones with bidirectional attention by default while retaining registered custom backbones and existing reranking behavior.

# Changelog

- Configure direct and extracted Hugging Face embedding fallbacks with `is_causal=false`, including the stock `Ministral3Model`.
- Preserve the stock Hugging Face sequence-classification path for standard Ministral score models.
- Keep the custom `ministral3_bidirec` class available only for explicit and legacy checkpoints.
- Preserve native Hugging Face loader options across config discovery and model loading.
- Test generic direct and extracted embedding fallbacks, bidirectional attention, plain Hugging Face save/reload, legacy dispatch, standard score dispatch, and unchanged Llama embedding and cross-encoder dispatch.

**Pre checks**:

# Additional Information

- This is the architecture-loading prerequisite for the Sentence Transformers export PR.
- The generic extracted-export sharding fix discovered during this work has been split into #3423.
- Focused retrieval, registry, and legacy-model verification: 74 passed, 1 skipped.
- A live regression check with `nvidia/llama-nemotron-rerank-1b-v2` confirmed that the existing Llama cross-encoder path still trains and reloads correctly.

### DIFF
--- a/nemo_automodel/_transformers/retrieval.py
+++ b/nemo_automodel/_transformers/retrieval.py
@@ -22,7 +22,7 @@ from typing import Optional
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-from transformers import AutoConfig, AutoModel, AutoModelForSequenceClassification, PretrainedConfig, PreTrainedModel
+from transformers import AutoConfig, AutoModel, AutoModelForSequenceClassification, PreTrainedModel
 from transformers.models.auto.modeling_auto import MODEL_FOR_SEQUENCE_CLASSIFICATION_MAPPING, MODEL_MAPPING
 from transformers.utils import logging
 
@@ -31,15 +31,6 @@ from nemo_automodel.components.loss.intermediate_distill import LayerCapture
 from nemo_automodel.components.models.common.bidirectional import EncoderStateDictAdapter
 
 logger = logging.get_logger(__name__)
-
-
-def _clone_pretrained_config(config: PretrainedConfig) -> PretrainedConfig:
-    """Clone a config while retaining private source-revision metadata."""
-    cloned_config = config.__class__.from_dict(config.to_dict())
-    commit_hash = getattr(config, "_commit_hash", None)
-    if commit_hash is not None:
-        cloned_config._commit_hash = commit_hash
-    return cloned_config
 
 
 def _extract_submodel(model: nn.Module, extract_submodel: str) -> PreTrainedModel:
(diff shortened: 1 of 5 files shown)


nemo_automodel/_transformers/retrieval.py
```diff
@@ -268,6 +280,19 @@ def build_encoder_backbone(
...
         )
 
+    if model_type.lower() == "ministral3" and task == "embedding":
+        backbone = AutoModel.from_pretrained(
+            model_name_or_path,
+            trust_remote_code=trust_remote_code,
+            **hf_kwargs,
+        )
+        backbone.config.is_causal = False
+        if pooling is not None:
```
gabrielspmoreira: This logic is similar to this [one](https://github.com/NVIDIA-NeMo/Automodel/pull/3103/changes#diff-c84a00ac55fe85b398fd7fc8ecc3dd70a5b80f8e66e049959ea84cfd44103d9eR115-R119). Could we maybe extract a common method for consistency?

gabrielspmoreira (commented):
Overall this MR looks good to me, as our intent is avoiding the need to distribute custom code with our models.
If the results from training and evaluating Ministral 3 with is_causal=False match the results using Ministral3BidirectionalModel, I am good merging it

ybabakhin (dismissed):
LGTM

nemo_automodel/_transformers/retrieval.py
oliverholworthy: Thanks for the review @gabrielspmoreira I've removed this part of the change relating to the pooling/temerature since it's unrelated to the core PR purpose

ybabakhin (approved):
Thanks for cleaning up those unused attributes

yuhezhang-ai (approved):
Approved — the implementation looks sound, the reviewed head is unchanged, and CI is green.

P2 (non-blocking): the public retrieval documentation still describes the previous routing behavior. In particular, docs/guides/llm/retrieval-finetune.mdx and docs/model-coverage/embedding/** still describe standard ministral3 as using Ministral3BidirectionalModel, direct scoring as unsupported, and generic AutoModel embedding fallbacks as retaining native causal attention. With this change, stock ministral3 uses Ministral3Model with is_causal=False, stock sequence classification is available for scoring, and ministral3_bidirec is the legacy custom path.

This documentation cleanup does not need to block this PR and can be addressed in the next follow-up docs PR.

PR #3113, opened 2026-07-17
fix(vlm): resolve get_rope_index from the base model for packed mRoPE

khazic:
## What

`FinetuneRecipeForVLM` resolved the mRoPE position-id builder with a plain attribute read:

```python
get_rope_index = getattr(self.model_parts[0], "get_rope_index", None)
```

That resolves to `None` for **every** mRoPE VLM, so packed multimodal training silently trained on 1D positions instead of 3D mRoPE ones. This fixes the lookup. Reported by @HuiyingLi in #3052.

## Why it is always None

Transformers defines `get_rope_index` on the **base model**, not on the `*ForConditionalGeneration` that `model_parts` holds. Checked against transformers `5.13.0.dev0`:

| class | has `get_rope_index` |
|---|---|
| `Qwen2VLModel` / `Qwen2_5_VLModel` / `Qwen3VLModel` / `Qwen3VLMoeModel` | yes |
| `Qwen2VLForConditionalGeneration` | no |
| `Qwen2_5_VLForConditionalGeneration` | no |
| `Qwen3VLForConditionalGeneration` | no |
| `Qwen3VLMoeForConditionalGeneration` | no |

`model_parts[0]` is the `*ForConditionalGeneration`, so the read misses. DDP and MegatronFSDP add a second wrapper (`.module`) that does not proxy attribute reads, which misses it again.

Confirmed on a real `Qwen3-VL-8B-Instruct` config:

```
real model class: Qwen3VLForConditionalGeneration
OLD  getattr(model, "get_rope_index"): None
NEW  resolved via base_model         : Qwen3VLModel.get_rope_index
```

## Impact

`neat_packing_vlm` derives `has_mrope` from that callable:

```python
self.has_mrope = get_rope_index is not None
...
if self.has_mrope and self.get_rope_index is not None:
    mrope_pos = _compute_mrope_position_ids(sample, self.get_rope_index)
```

With `get_rope_index=None`, `has_mrope` is always `False`, so `_compute_mrope_position_ids` never runs and the 3D position IDs are never built. `has_mrope` also gates `_shift_sample` and `_build_packed_vlm_sample`, so the whole mRoPE-preserving path was dead. On `main` this affects the `neat` packing format; the `thd` format proposed in #3052 takes the callable from this same lookup and inherits the fix.

The existing tests always pass `get_rope_index` in explicitly, so they cover `_compute_mrope_position_ids` but never the resolution from a model, which is why this survived.

## Fix

Two commits:

[...]

PR #3114, opened 2026-07-17
chore: CI proxy for #3113

HuiyingLi:
CI-only proxy PR. Do not merge / do not review. Points at the exact head commit of #3113 (28fae72fdb815724ccf4c9d98b8b6f0190252697) so internal CI runs under the internal-contributor queue; results post back to #3113 via the shared SHA. Source of truth: #3113. Close once CI completes.


HuiyingLi:
Closing CI-only proxy because source PR #3113 is merged.

PR #3115, opened 2026-07-17
feat(kernels): add QuACK backend

akoumpa:
## Summary

- add QuACK as a selectable backend for linear, RMSNorm, and Llama-family rotary embedding modules
- extend QuACK linear/RMSNorm selection across registered custom models by replacing residual exact PyTorch modules after construction
- preserve parameter objects, tied aliases, checkpoint keys, dtype/device, and train/eval state during generic backend replacement
- add the QuACK/CuTe dependencies to the main dependencies and synchronize both lockfiles
- preserve batch-specific position semantics while using QuACK's fused in-place RoPE path for the common local-batch-one training case
- add unit coverage for backend routing, real Llama/Qwen2/Baichuan forwards, RoPE layout conversion, and missing optional dependencies
- add matched Qwen3 MoE, Llama 3.1 8B, and Qwen2.5-7B benchmark recipes

This implements [AM-413](https://linear.app/nvidia/issue/AM-413/add-quack-as-a-backend). It provides a CuTe-based alternative to Transformer Engine/PyTorch while preserving the existing backend configuration interface.

## Registry-wide coverage

The model registry currently exposes 49 architecture keys across roughly 41 model packages. Models that already construct linear/RMSNorm layers through the shared backend factories continue to route directly. For registered custom models that still construct exact `torch.nn.Linear` or one-dimensional `torch.nn.RMSNorm` modules, model initialization now applies the requested QuACK backend after construction.

The generic pass intentionally does not rewrite subclasses or specialized model-owned normalization layers (for example zero-centered or gated norms). This keeps model-specific semantics inside each model package while allowing legacy and modern registered models to opt into QuACK for their standard modules.

### Fused Llama RoPE

Measured on 1 node with 8×H100 GPUs using Llama 3.1 8B, sequence length 4096, global batch size 128, local batch size 1, FSDP2, SDPA, and torch linear/RMSNorm backends. Results cover 20 iterations after 10 warmup iterations.

| RoPE backend | Average iteration time | MFU |
| --- | ---: | ---: |
| torch | 10.127s | 31.571% |
| QuACK | 9.690s | 32.996% |

QuACK fused RoPE reduces steady-state iteration latency by **4.32%**, corresponding to a **4.51% throughput increase**, and improves MFU by **1.425 percentage points**.

- torch baseline: [job 366318566](https://nv/nemoci/-/jobs/366318566)
- QuACK result: [job 366318565](https://nv/nemoci/-/jobs/366318565)
- successful matched leaf pipeline: [58460630](https://nv/nemoci/-/pipelines/58460630)

Both artifacts report `exit_code_training: 0`; the QuACK lane completed with finite loss through all 30 iterations.

### Linear and RMSNorm

Measured on 1 node with 8×H100 GPUs using Qwen3-30B-A3B, sequence length 4096, global batch size 512, EP8, grouped GEMM, and DeepEP. Results cover 20 iterations after 10 warmup iterations.

| Backend | Average iteration time | MFU |
| --- | ---: | ---: |
| Transformer Engine | 21.062s | 28.953% |
| QuACK | 20.824s | 29.283% |

QuACK reduces steady-state iteration latency by **1.13%**, corresponding to a **1.14% throughput increase**, and improves MFU by **0.330 percentage points**.

- TE baseline: [job 365369401](https://nv/nemoci/-/jobs/365369401)
- QuACK result: [job 365388362](https://nv/nemoci/-/jobs/365388362)
- successful QuACK leaf pipeline: [58323370](https://nv/nemoci/-/pipelines/58323370)

### Registry-coverage benchmark

The new generic routing path was also exercised end to end on Qwen2.5-7B with 8×H100, sequence length 4096, global batch size 128, local batch size 1, FSDP2, and SDPA. Both lanes trained successfully with finite loss for 30 iterations, using 10 warmup and 20 measured iterations.

| Backend | Average iteration time | MFU |
| --- | ---: | ---: |
| torch | 9.302s | 31.975% |
| QuACK | 9.790s | 30.383% |

This shape is a negative performance result: QuACK is **5.25% slower** and reduces MFU by **1.592 percentage points**. QuACK remains an explicit opt-in; this result is included to distinguish expanded functional coverage from the Qwen3-MoE and Llama RoPE performance wins above.

- torch baseline: [job 366985891](https://nv/nemoci/-/jobs/366985891)
- QuACK result: [job 366985890](https://nv/nemoci/-/jobs/366985890)
- matched leaf pipeline: [58558632](https://nv/nemoci/-/pipelines/58558632)
- exact-commit build pipeline: [58557565](https://nv/nemoci/-/pipelines/58557565)

### Evaluated expansion

Additional QuACK kernels were evaluated but are intentionally excluded from this PR:

- fused Llama SwiGLU MLP completed stably but averaged 10.179s versus 10.116s for torch, a 0.62% regression ([torch](https://nv/nemoci/-/jobs/366265425), [QuACK](https://nv/nemoci/-/jobs/366265424)); QuACK fused gradient accumulation regressed further to 10.210s ([job 366291491](https://nv/nemoci/-/jobs/366291491))
- router top-k became numerically unstable in the matched 30-step Qwen run
- router softmax completed stably but averaged 33.484s versus 33.290s for native softmax, a 0.58% regression ([baseline](https://nv/nemoci/-/jobs/366104373), [QuACK softmax](https://nv/nemoci/-/jobs/366231753))
- QuACK cross entropy produced NaNs on the Qwen vocabulary shape and was 0.77% slower in the matched Llama check

The PR retains only kernel integrations with demonstrated correctness. Backend selection remains explicit so callers can choose QuACK on shapes where it demonstrates a gain.

### Compilation caveat

QuACK kernels compile on first use. The original linear/RMSNorm run took 707.5 seconds to compile; fused RoPE's first iteration took 16.94 seconds versus its 9.69-second steady state. In the Qwen2.5 coverage run, the QuACK warmup took 280.75 seconds versus 134.29 seconds for torch. Benchmarks exclude compilation through their 10 warmup iterations, so reported comparisons are steady-state performance. QuACK recipes request a one-hour CI allocation to leave enough room for compilation and measured iterations.

## Validation

- generic backend override and real-model CPU tests: 13 passed
- focused QuACK RoPE and backend regression tests passed
- real tiny Llama, Qwen2, and legacy Baichuan models completed finite CPU forwards through the generic routing path
- changed Python files: `ruff format --check` and `ruff check` passed
- all 454 example YAMLs parsed successfully
- both new Qwen2.5 recipes passed CI metadata validation
- all matched performance jobs completed successfully on 8×H100 with finite training loss
- `uv lock --check` passed for the standard and PyTorch-image lock paths


akoumpa:
Re-benchmarked PR head `af1dc87129455094fc388dd29e60313a5d6711be` with `akoumpa/quack@aa9b6256f615aabe369ff2c9d445b9be7ca0c233` on 8x H100 (cw-dfw), using 30 iterations with the first 10 excluded as warmup.

| Model / backend pair | Baseline | QuACK fork | Iteration-time delta | Result |
|---|---:|---:|---:|---|
| Llama 3.1 8B, torch RoPE vs QuACK RoPE | 9.202 s, 34.746% MFU | 8.650 s, 36.961% MFU | **-6.00%** | PASS |
| Qwen3 MoE 30B-A3B, TE vs QuACK | 21.642 s, 28.176% MFU | 21.475 s, 28.396% MFU | **-0.77%** | PASS |
| Qwen2.5 7B, torch Linear/RMSNorm vs QuACK | 8.578 s, 34.676% MFU | 9.062 s, 32.823% MFU | **+5.64%** | FAIL |

The complete Llama/Qwen3 run is cw-dfw Slurm job `14171070`; Qwen2 is from job `14170867`. Losses remained finite throughout. A duplicate Llama run on a second node reproduced both averages within 0.002 s.

I profiled the Qwen2 regression on EOS with Nsight Compute (job `5670042`). Aggregate forward+backward device time per isolated train call:

| Shape | torch | QuACK fork | Delta |
|---|---:|---:|---:|
| RMSNorm | 89.632 us | 74.016 us | -17.42% |
| Q/O Linear | 508.576 us | 491.040 us | -3.45% |
| K/V Linear | 89.984 us | 114.368 us | +27.10% |
| MLP up | 2376.416 us | 2351.360 us | -1.05% |
| MLP down | 2398.048 us | 2386.816 us | -0.47% |

K/V is the only isolated GPU-kernel regression, but it is too small to explain the +0.484 s full-iteration gap. EOS cProfile (job `5670147`) identifies the dominant cost as eager Python/custom-autograd dispatch: per train call, QuACK adds about 120 us for RMSNorm, 160 us for K/V, 260 us for Q/O, and 150-160 us for each MLP projection. Across 28 layers and 16 accumulated microbatches, that host overhead is consistent with the end-to-end gap.

I also tested matched whole-model compile on cw-dfw (job `14171624`): torch 8.582 s vs QuACK 9.045 s (**+5.40%**), so it does not remove the regression. Explicit per-module compile was worse for every tested Qwen2 shape.

Because Qwen2 still regresses, I have not changed `pyproject.toml` or the lockfiles to the fork revision; that preserves the requested no-regression gate. The branch remains unchanged at `af1dc871`.

akoumpa:
Re-benchmark complete with `akoumpa/quack@aa9b6256f615aabe369ff2c9d445b9be7ca0c233` on 8×H100 (`cw-dfw`, 10 warmup + 20 measured steps). All final model cases pass the no-regression gate with finite loss:

| Model / QuACK coverage | Baseline iter | QuACK iter | Delta | Baseline MFU | QuACK MFU |
| Llama 3.1 8B / RoPE | 9.2039 s | 8.6490 s | **-6.03%** | 34.7379% | 36.9666% |
| Qwen2.5 7B / RMSNorm + RoPE | 8.5781 s | 8.1536 s | **-4.95%** | 34.6755% | 36.4809% |
| Qwen3 30B-A3B / Linear + RMSNorm | 21.6424 s | 21.4750 s | **-0.77%** | 28.1763% | 28.3960% |

Qwen2 diagnosis and resolution:

- The original QuACK Linear + RMSNorm case regressed by 5.64%.
- EOS NCU showed the aggregate QuACK device kernels were faster except K/V GEMM, while cProfile isolated eager Python/custom-autograd/GEMM dispatch overhead as the dominant whole-model regression.
- Ablation confirmed QuACK Linear + torch RMSNorm still regressed (9.0356 s on EOS; 9.0699 s on cw-dfw), while torch Linear + QuACK RMSNorm passed (8.5451 s EOS; 8.5718 s cw-dfw).
- Qwen2 also ignored `backend.rope=quack`; commit `8139ce48` wires it to the same fused QuACK RoPE path as Llama. The final Qwen2 recipe uses the two passing QuACK paths, RMSNorm + RoPE. Qwen3 retains end-to-end QuACK Linear coverage.
- EOS independently reproduced the final Qwen2 result: 8.1480 s vs 8.5693 s, **-4.92%**.

`pyproject.toml` now pins the fork revision through `[tool.uv.sources]`; CUTLASS DSL is aligned to 4.6.1 for QuACK/FFPA compatibility, and both `uv.lock` files were regenerated and pass `uv lock --check`.

Benchmark jobs: cw-dfw `14171070`, `14172708`; EOS `5670042`, `5670147`, `5670515`. CI context: https://nv/nemoci

akoumpa:
Added broader model-shape and numerical coverage in 107bfb02.

Numerical validation on one H100 (Slurm 14175017): 10/10 Python pytest cases passed in 128.96s. The matrix covers Quack linear forward, input gradients, weight gradients, and bias gradients plus RMSNorm forward, input gradients, and weight gradients against PyTorch BF16 references for Qwen3-Next, Nemotron-H, GPT-OSS, Qwen2, Qwen3-MoE, and Llama dimensions.

Paired steady-state training-kernel timings versus Transformer Engine on H100 (alternating launch order, CUDA events):

| Model-derived shape | Tokens | Quack ms | TE ms | Quack vs TE |
|---|---:|---:|---:|---:|
| Qwen3-Next Q 2048x8192 | 4096 | 0.5970 | 0.6051 | +1.35% |
| Qwen3-Next KV 2048x512 | 4096 | 0.1943 | 0.2066 | +6.34% |
| Nemotron-H MLP 2688x1856 | 4096 | 0.2110 | 0.2234 | +5.85% |
| GPT-OSS Q 2880x4096 | 4096 | 0.4132 | 0.4244 | +2.72% |
| GPT-OSS MLP 2880x2880 | 4096 | 0.3167 | 0.3359 | +6.07% |
| Nemotron-H RMSNorm 2688 | 4096 | 0.1676 | 0.1978 | +18.02% |
| GPT-OSS RMSNorm 2880 | 4096 | 0.2158 | 0.2513 | +16.42% |

I also probed the lower-occupancy regime where TE might lead (Slurm 14175244). Quack remained ahead: Qwen3-Next Q m=512 +4.20%; KV m=128 +6.89%, m=512 +11.05%; Nemotron-H MLP m=512 +3.34%; GPT-OSS Q m=128 +2.42%, m=512 +10.01%; GPT-OSS MLP m=512 +6.76%.

### Qwen2 launch-overhead follow-up

The earlier full-model comparison was Quack versus torch, not Quack versus TE. A controlled H100 rerun with only the linear provider changed confirms that Quack does **not** have more host launch overhead than TE.

Weighted steady-state CPU enqueue time across the actual Qwen2 linear shapes (100 forward+backward calls per shape, three GEMMs per call):

| Linear provider | Host time per train call |
|---|---:|
| Quack | 269.8 us |
| TE | 272.6 us |
| torch | 112.6 us |

Quack is 2.8 us/call lower than TE (effectively equal). Within the Quack path, cProfile attributes 107-112 us/call to the three `gemm()` wrapper calls (~36 us/GEMM), while the actual TVM-FFI `launch_gemm` portion is ~6.7 us/GEMM; the remainder is plan-key handling, output allocation, and autograd wrapper work. TE reaches essentially the same total host time.

Full Qwen2.5-7B training results on 8xH100, sequence length 4096, global batch 128, 20 measured steps after 10 warmup steps, with Quack RMSNorm and RoPE held constant:

| Linear provider | Step time | Delta vs torch |
|---|---:|---:|
| torch | 8.1536 s | - |
| TE | 8.1653 s | +11.7 ms (+0.14%) |
| Quack | 8.8243 s | +670.7 ms (+8.23%) |

The real Q/K/V bias path does not reveal extra host launch overhead either. CUDA-event train timings give Quack-vs-TE deltas of +44.2 us for Q and -9.9 us for each K/V projection. Combining those with O and MLP timings gives +131 us/layer for the seven train linear sequences, or ~58.8 ms/optimizer step over 28 layers and 16 microbatches. Activation-checkpoint recomputation and CPU/GPU overlap make that extrapolation non-additive, but it is still far short of the measured 659.0 ms Quack-vs-TE full-step gap.

So the previous statement that wrapper/host overhead dominated was too strong. Quack and TE have the same measured host overhead; the remaining full-model gap is a model-context critical-path/cache/scheduling effect that these isolated launch and hot-kernel measurements do not yet attribute. The Qwen2 recipe should still keep torch linear on the measured result, but a full-step systems trace is needed before assigning the gap to a specific layer of the stack. Follow-up jobs: host/launch profile 14176942, kernel timing 14176958, TE full-model control 14176943, Quack full-model control 14177182, biased Q/K/V timing 14177213.

Also moved the fork-pinned quack-kernels dependency from the CUDA extra into the default install set and regenerated both lockfiles. `uv export --frozen --no-default-groups` now includes the exact aa9b6256f615aabe369ff2c9d445b9be7ca0c233 source.

akoumpa:
Moved the production-shape QuACK numerical parity tests to akoumpa/quack#3 so kernel compilation no longer consumes AutoModel L0 GPU time: https://github.com/akoumpa/quack/pull/3

nemo_automodel/components/models/common/utils.py
```diff
@@ -194,8 +198,9 @@ class BackendConfig:
     """
 
     attn: Literal["te", "sdpa", "flex", "eager", "tilelang"] = "te" if HAVE_TE and torch.cuda.is_available() else "sdpa"
-    linear: Literal["torch", "te"] = "te" if HAVE_TE and torch.cuda.is_available() else "torch"
-    rms_norm: Literal["torch", "torch_fp32", "te"] = "torch_fp32"
+    linear: Literal["torch", "te", "quack"] = "te" if HAVE_TE and torch.cuda.is_available() else "torch"
+    rms_norm: Literal["torch", "torch_fp32", "te", "quack"] = "torch_fp32"
+    rope: Literal["torch", "quack"] = "torch"
```
yuhezhang-ai, line 203: [P1] Prevent QuACK RoPE from inheriting the fused-cache path

On CUDA builds with Transformer Engine installed, `BackendConfig(rope="quack")` still leaves `rope_fusion=True`. Both Llama and Qwen2 pass that value into `LlamaRotaryEmbedding`, whose fused branch deliberately builds `cos`/`sin` from contiguous `[0, seq_len)` positions and ignores the values in `position_ids`; attention then selects the QuACK branch and consumes those tables. Offset, packed, or per-example positions are therefore silently rotated with the wrong phases unless every caller also remembers to set `rope_fusion=False` (the example YAMLs do, but the config/API does not enforce it). Please derive or validate this combination so QuACK always receives position-gathered tables, and add a regression test using non-contiguous `position_ids` with `BackendConfig(rope="quack")` and its defaults.

pyproject.toml
```diff
@@ -278,6 +279,7 @@ prerelease = "allow"
 
 [tool.uv.sources]
 deep_ep = { git = "https://github.com/deepseek-ai/DeepEP.git", rev = "7febc6e25660af0f54d95dd781ecdcd62265ecca" }
+quack-kernels = { git = "https://github.com/akoumpa/quack.git", rev = "add2869739a745fd004da20fb3985e0753d03d35" }
```
yuhezhang-ai: [P2] Preserve the tested QuACK source in published dependency metadata

This source override is only honored by uv while resolving the checkout; it is not part of the package's published dependency metadata. A wheel/sdist consumer therefore sees only `quack-kernels==0.6.1` from `project.dependencies` and resolves PyPI's upstream 0.6.1, not `akoumpa/quack@add286...` used by this PR's lockfile and benchmarks. `uv export --frozen` preserves the fork for repo deployments, but `pip install nemo-automodel` (and other standards-based installs) does not. Please publish the required fork changes under a distinct version and pin that release, or otherwise make the publishable dependency identify the tested artifact; `uv build/lock --no-sources` is the useful validation mode here. See https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-sources.

yuhezhang-ai (commented):
I re-checked the current head and found two concrete issues worth addressing before merge: one silent RoPE correctness problem and one published-dependency mismatch. Inline details below.

nemo_automodel/components/models/common/utils.py
akoumpa: Fixed in `b18d1223c`. `BackendConfig` now explicitly disables `rope_fusion` whenever `rope="quack"`, before the temporary global fusion override, so the invariant remains enforced after that temporary override is removed. I also added a CPU regression that reproduces the CUDA/TE default with `rope_fusion=True`, uses non-contiguous positions `[0, 5, 2, 9]`, and verifies QuACK receives position-gathered tables rather than the contiguous slice. Focused validation: 63 passed, 2 skipped; Ruff format/lint passed.

pyproject.toml
akoumpa: Fixed in `b18d1223c`. I removed the fork-only `tool.uv.sources` entry and restored CUTLASS DSL 4.6.0, returning the dependency to the original published/tested artifact: PyPI `quack-kernels==0.6.1`. Both lockfiles were regenerated. `uv lock --no-sources-package quack-kernels --check` passes for the standard and PyTorch-image locks, and `uv build --wheel --no-sources` succeeds; the wheel metadata contains `Requires-Dist: quack-kernels==0.6.1; sys_platform == "linux"` with no direct URL. A global source-disabled lock still encounters the unrelated `deep-ep` static-metadata limitation, so the package-scoped lock check isolates the QuACK contract.

yuhezhang-ai (approved):

yuhezhang-ai:
Only non-blocking cleanup: the PR description says QuACK is in the CUDA extra, while pyproject.toml currently makes it a base Linux dependency. The description should be updated if that placement is intentional.

PR #3119, opened 2026-07-17
ci: add HF hub cache preflight check

thomasdhc:
# What does this PR do ?

Adds tests/ci_tests/utils/hf_cache_check.py — walks a recipe YAML for HF repo IDs (pretrained_model_name_or_path, tokenizer_name) and live-inspects \$HF_HOME/hub/ so the nemo-ci launcher wrappers can decide whether to flip HF_HUB_OFFLINE=0 for a run. Replaces the static offline assumption that would fail any recipe referencing a not-yet-cached model.

# Changelog

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/tests/ci_tests/utils/hf_cache_check.py
+++ b/tests/ci_tests/utils/hf_cache_check.py
@@ -16,25 +16,18 @@
 """Check whether every HF repo referenced by a recipe YAML is already present
 in the local Hugging Face Hub cache (``$HF_HOME/hub/``).
 
-This is a *pure inspection* — it never downloads and never modifies the cache.
-The caller (typically a nemo-ci launcher wrapper) reads the JSON output and
-decides whether to flip ``HF_HUB_OFFLINE`` for the actual test run.
+Pure inspection — never downloads, never modifies the cache. The exit code
+IS the decision so the caller shell reduces to::
 
-CLI:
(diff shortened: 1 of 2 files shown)


tests/ci_tests/utils/hf_cache_check.py
```diff
@@ -0,0 +1,153 @@
...
+    cached: list[str] = []
+    missing: list[str] = []
+    for repo_id in _extract_repo_ids(recipe_path):
+        repo_dir = hub_dir / _repo_folder(repo_id)
+        (cached if _has_snapshot(repo_dir) else missing).append(repo_id)
+    return cached, missing
+
+
+EXIT_MISS_DETECTED = 0
+EXIT_DO_NOTHING = 1
```
chtruong814, line 122: Why is EXIT_DO_NOTHING `1`?  That seems to imply it should be `0` instead?

thomasdhc: Yeah I can see how it could be confusing but it's a catch-all. Any failures will result in default behavior and success means we've successfully verified that missing cache was found.

thomasdhc: A script failure should result in the default behavior

chtruong814 (approved):

PR #3130, opened 2026-07-18
feat(checkpoint): add DCP CPU offload option

khazic:
## Summary

- Add the opt-in `checkpoint.cpu_offload` setting, disabled by default.
- Pass the setting to PyTorch DCP model and optimizer state dict options for normal FSDP checkpoints.
- Preserve existing PEFT behavior, which already uses CPU offload.

## Why

DCP can materialize distributed state dict shards on GPU during checkpointing. The new setting provides a supported PyTorch path to move the produced model and optimizer tensors to CPU before they are written.

## Validation

- `uv run pytest tests/unit_tests/checkpoint/test_checkpointing.py -q`: 173 passed, 2 skipped.
- `uv run ruff format --check .` and `uv run ruff check .` passed.
- Single-GPU Qwen3-1.7B FSDP smoke with `save_consolidated=false`: baseline and `checkpoint.cpu_offload=true` both completed a model and optimizer checkpoint with identical artifact sizes.

## Measurement boundary

In the single-GPU smoke, the sampled process peak was about 17.3 GiB for both paths. CPU offload increased the observed save interval from about 4 seconds to about 6 seconds. This validates the option and its checkpoint output, but does not demonstrate a peak-memory reduction in that small configuration. A larger multi-GPU run on an idle node is required to quantify the intended benefit.


khazic:
## Single-GPU FSDP smoke

Both runs used Qwen3-1.7B, one GPU, one training step, synchronous DCP saving, and `save_consolidated=false`. Both produced model and optimizer checkpoint artifacts with matching sizes and exited successfully.

```text
baseline
21:07:59 training memory: 16.05 GiB
21:08:05 Saving checkpoint: epoch 0, step 0
21:08:09 Model checkpoint write completed
21:08:25 Training completed
EXIT_CODE=0

cpu_offload=true
21:09:38 training memory: 16.05 GiB
21:09:43 Saving checkpoint: epoch 0, step 0
21:09:49 Model checkpoint write completed
21:10:05 Training completed
EXIT_CODE=0
```

The sampled process peak was about 17.3 GiB for both runs. This smoke validates correctness and the configuration path. It does not show a peak-memory improvement at this scale.

HuiyingLi:
I've validated the ckpt save correctness by save+resume.

HuiyingLi (approved):

HuiyingLi (approved):

Issue #2267, opened 2026-05-18
Support configurable HSDP in both expert and non-expert groups for MoEs


hemildesai:
From [Cursor 2.5 blog](https://cursor.com/blog/composer-2-5):

> HSDP forms multiple FSDP replicas and all-reduces gradients across corresponding shards. We use separate HSDP layouts for non-expert and expert weights: non-expert weights are comparatively small, so their FSDP groups can stay narrow, often within a node or rack, while expert weights hold most of the parameters and most of the Muon compute, so they use a wider expert sharding mesh.

PR #3135, opened 2026-07-20
fix(moe): preserve HSDP replica gradient synchronization

wangzhxg:
# What does this PR do ?

Preserve the HSDP replica process group when the custom MoE parallelizer constructs its FSDP2 mesh, restoring cross-replica gradient synchronization.

# Changelog

* Detect `dp_shard_cp` when it is registered as a flattened root-mesh dimension, so `MeshContext` propagates both `dp_replicate` and `dp_shard_cp` for HSDP.
* Route the canonical MoE HSDP axes through the existing root-preserving `get_fsdp_dp_mesh()` helper.
* Add focused unit coverage for flattened-axis discovery and MoE FSDP mesh selection.
* Add a 4-rank CPU/Gloo numerical regression that runs real FSDP2 + EP forward, backward, and an optimizer step, then verifies corresponding replica shards remain identical.

**Pre checks**:

- [X] Did you add or update any necessary documentation? No user-facing documentation is required because this does not change configuration or public APIs; the regression rationale is documented in the test.

# Additional Information

* Related to NVIDIA-NeMo/Automodel#2267. This PR fixes correctness of the existing HSDP path; it does not add separately configurable HSDP layouts for expert and non-expert parameters.
* Root cause: `dp_shard_cp` is created with `DeviceMesh._flatten()` and stored in the root mesh mapping rather than in the physical `mesh_dim_names`. The previous MoE path therefore omitted `dp_replicate` and applied FSDP only to each shard group, allowing replicas to update independently.
* Regression sensitivity: with the fix reverted, the numerical test reports all 32 elements of a corresponding parameter shard as different after one optimizer step (maximum absolute difference approximately `8.87e-3`). With the fix applied, both corresponding shard pairs are exactly equal.
* Validation:
  * `87 passed` across the changed mesh and MoE unit-test modules, including the 4-rank numerical regression.
  * `ruff format --check` passes for all five changed files.
  * `ruff check` passes for all five changed files.
  * `git diff --check origin/main...HEAD` passes.


wangzhxg:
The four failed L2 jobs all failed before model execution with CUDA Error 802 / no GPUs found on separate ephemeral GPU runners. This appears to be a CI infrastructure issue rather than a code regression. Could you please rerun the failed jobs?

HuiyingLi (approved):

PR #3144, opened 2026-07-21
fix(datasets): keep system turns in the sharegpt conversation converter

### LINKED ISSUE #3143: System turns are silently dropped by the VLM sharegpt/meta dataset path
## Summary

`_convert_sharegpt_to_conversation` maps only `user` and `assistant` roles. Every other role hits a bare `continue`, so a `system` turn present in the source data is dropped before the conversation ever reaches the chat template. Nothing is logged, the sample count is unchanged, and the loss curve looks normal, so a dataset whose system prompts are silently ignored is indistinguishable from one that trains on them.

This affects every dataset loaded through `make_meta_dataset` (the meta-JSON path used by VLM SFT recipes) and `make_tulu3_dataset`, both of which delegate row conversion to this helper.

https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/components/datasets/vlm/datasets.py#L566-L571

```python
        if role_value == user_tag:
            role = "user"
        elif role_value == assistant_tag:
            role = "assistant"
        else:
            continue
```

## Why this looks like a defect rather than an intended restriction

1. **The target format declares `system` valid.** The conversation format this helper produces is documented in the same module as `{"role": "user"|"assistant"|"system", "content": [...]}`.

2. **A sibling builder in the same module keeps `system` turns.** `make_tulu_magicoder_mix_dataset` filters with `valid_roles = {"system", "user", "assistant"}` and appends system turns to the conversation. Two builders that emit the same format disagree about whether `system` survives.

3. **The whole downstream path already handles `system` correctly.** The collate functions pass `message["role"]` straight into `apply_chat_template`, and `build_labels` supervises assistant spans only, so a system turn is rendered by the template and masked to `-100` automatically. The only thing standing between user data and correct behavior is this one `continue`.

4. **The text-only path supports it.** `nemo_automodel/components/datasets/llm/chat_dataset.py` accepts `system` (and `tool`) roles and carries a role map that includes `"system": "system"`.

5. **There is no supported alternative.** The meta config exposes `columns` (`messages`, `images`, `videos`) and `tags` (`role_tag`, `content_tag`, `user_tag`, `assistant_tag`). There is no field through which a system prompt can be supplied, so on this path system-prompted data is simply unusable.

## Impact

Any ShareGPT-style corpus that carries per-sample system prompts trains as if those prompts were never written. This matters most for data whose system turn carries actual constraints (terminology tables, output-format rules, domain personas): the model is trained on the answers those constraints produced, without ever seeing the constraints.

I hit this on a translation SFT set where roughly 20% of rows carry a system turn holding a required-terminology table. Those rows had been training with the constraint stripped since the dataset was built, and there was no signal anywhere that it was happening.

A second, quieter consequence: `scripts/precompute_tokens.py` applies the same role filter when it writes `_text_tokens`. Once system turns are trained on, that precomputed count under-reports sequence length, which feeds `LengthGroupedSampler` and the packing estimate.

https://github.com/NVIDIA-NeMo/Automodel/blob/main/scripts/precompute_tokens.py#L83-L86

## Reproduction

```python
from nemo_automodel.components.datasets.vlm.datasets import _convert_sharegpt_to_conversation

example = {
    "messages": [
        {"role": "system", "content": "Answer in French."},
        {"role": "user", "content": "Hello"},
        {"role": "assistant", "content": "Bonjour"},
    ]
}
conv = _convert_sharegpt_to_conversation(example)["conversation"]
print([turn["role"] for turn in conv])
```

Actual: `['user', 'assistant']`
Expected: `['system', 'user', 'assistant']`

## Expected behavior

`system` turns are preserved and passed to the chat template, with the role remappable through a `system_tag` entry in `tags`, consistent with how `user_tag` and `assistant_tag` already work. Roles outside the mapped set keep being skipped.

## Note on existing tests

Two current tests encode the drop rather than catch it:

- `TestMakeTulu3Dataset::test_system_turns_dropped_like_meta_path` asserts the roles come back as `["user", "assistant"]`.
- `TestConvertSharegptToConversation::test_unknown_role_skipped` uses `system` as its example of an *unrecognized* role.

Fixing the behavior means updating both. The second one keeps its value if the example role is changed to one that is genuinely unmapped.

## Behavior change to be aware of

With the fix, a conversation containing a system turn is handed to `apply_chat_template` as-is. For a processor whose chat template does not accept a system role, that surfaces as an error instead of being silently discarded. That is the better failure mode (loud rejection beats invisible data loss), but it is a visible change for anyone currently feeding system-prompted data to such a template.

I have a patch with tests ready and will open a PR shortly.

khazic:
Fixes #3143

## What

`_convert_sharegpt_to_conversation` mapped only the `user` and `assistant` roles; every other role hit a bare `continue`. A `system` turn present in the source data was therefore dropped before the conversation ever reached the chat template, with nothing logged and no change in sample count, so a dataset whose system prompts are ignored looked identical to one that trains on them.

This affects every dataset loaded through `make_meta_dataset` (the meta-JSON path used by VLM SFT recipes) and `make_tulu3_dataset`, since both delegate row conversion to this helper.

## Why this is a defect and not an intended restriction

- The conversation format this helper produces is documented in the same module as `{"role": "user"|"assistant"|"system", ...}`.
- The sibling builder `make_tulu_magicoder_mix_dataset` in the same module filters on `valid_roles = {"system", "user", "assistant"}` and keeps system turns, so two builders emitting the same format disagreed.
- The downstream path already handles system turns correctly: the collate functions pass `message["role"]` straight into `apply_chat_template`, and `build_labels` supervises assistant spans only, so a system turn is rendered by the template and masked to `-100` automatically.
- The text-only path (`datasets/llm/chat_dataset.py`) accepts `system` already.
- The meta config exposed no other field through which a system prompt could be supplied, so system-prompted data was simply unusable on this path.

## Changes

**`nemo_automodel/components/datasets/vlm/datasets.py`**

Map the `system` role, remappable through a new `system_tag` entry in `tags` so it stays consistent with the existing `user_tag` and `assistant_tag`. Media placeholders keep being parsed in user turns only, so an `<image>` inside a system turn stays literal text and consumes no image, leaving image-to-turn alignment untouched. Roles outside the mapped set are still skipped.

**`scripts/precompute_tokens.py`**

The worker that writes `_text_tokens` applies the same role filter, so the role set is mirrored there. Without this, the precomputed count under-reports sequence length for rows carrying a system turn, which then feeds `LengthGroupedSampler` and the packing estimate.

## Behavior change to be aware of

A conversation containing a system turn is now handed to `apply_chat_template` as-is. For a processor whose chat template does not accept a system role, that surfaces as an error instead of being silently discarded. That is the better failure mode, but it is a visible change for anyone currently feeding system-prompted data to such a template.

Two existing tests encoded the old behavior and are updated:

- `TestMakeTulu3Dataset::test_system_turns_dropped_like_meta_path` asserted the roles came back as `["user", "assistant"]`. It now asserts the system turn survives, and the class docstring is corrected.
- `TestConvertSharegptToConversation::test_unknown_role_skipped` used `system` as its example of an unrecognized role. The example role is changed to one that is genuinely unmapped, so the test keeps its original purpose.

## Tests

New coverage in `tests/unit_tests/datasets/vlm/test_datasets.py`:

- `test_system_turn_preserved`: a system turn maps to `role="system"` with its text intact.
- `test_system_turn_media_placeholder_not_expanded`: `<image>` inside a system turn stays literal and the image is still consumed by the user turn.
- `test_custom_system_tag`: the system role is remappable like the other role tags.
- `test_custom_system_tag_leaves_default_system_unmapped`: with a remapped `system_tag`, a literal `system` role is unknown and skipped.
- `TestMakeMetaDataset::test_system_turns_survive_the_meta_path`: end to end through a JSONL file, mixing rows with and without a system turn.

New file `tests/unit_tests/scripts/test_precompute_tokens.py` covers `_process_chunk`: system turns counted, unmapped roles excluded, `system_tag` remapping, media placeholders stripped, and the empty-sample case.

```
$ pytest tests/unit_tests/datasets/vlm/ tests/unit_tests/scripts/ -q
358 passed
```

The wider `tests/unit_tests/datasets/` run has 21 pre-existing failures on this machine (object-storage, parquet, and column-mapped tests that need optional dependencies). The failure set is byte-identical before and after this change, verified by running the same selection on a clean `main` checkout.

Lint clean: `ruff format --check` and `ruff check` pass on all four changed files.


khazic:
Thanks, fixed in 2aaf85d2e. The docstring now correctly states that system turns are preserved.

scripts/precompute_tokens.py
```diff
@@ -82,7 +86,7 @@ def _process_chunk(args):
         texts = []
         for msg in sample.get(messages_col, []):
             role = msg.get(role_tag, "")
-            if role not in (user_tag, assistant_tag):
+            if role not in counted_roles:
```
HuiyingLi, line 89: counted_roles include all roles? should the img tag sub only happen to user turn? and not other turn?

HuiyingLi (approved):

PR #3165, opened 2026-07-22
feat(diffusion): add LTX-2.3 video+audio finetuning (training + prepr…

linnanwang:
…ocessing)

Add support for finetuning the LTX-2.3 dual-stream (video+audio) diffusion transformer: preprocessing, a joint flow-matching loss, and distributed training.

- flow_matching: add LTX2Adapter that noises the audio latents with the same per-sample sigma as video and adds the audio flow-matching MSE via a new FlowMatchingAdapter.auxiliary_losses() hook; wire the hook into FlowMatchingPipeline.step() (summed into the total loss and logged per name) and register the "ltx2" adapter type.
- data: carry the LTX-2 dual-stream cache keys (audio_latents, audio_text_embeddings) through the video dataset collation.
- preprocessing: add LTX2Processor (video+audio -> latents) and an audio encoding helper; call processor.encode_audio() from the multiprocess preprocessor and expose the "ltx2" processor choice.
- distributed: register LTX2ParallelizationStrategy (whole-block activation checkpointing) so attention/FFN activations on the long combined video+audio sequence are recomputed instead of OOM-ing.
- recipes/peft: accept model_type "ltx2" for LoRA runs.
- examples: add ltx2_3_t2v_flow.yaml (full FT) and ltx2_3_t2v_flow_lora.yaml (LoRA) recipes.
- tests: CPU unit tests for the adapter (pack/unpack, shared-sigma, aux-loss integration), processor (mel/audio encode, cache contract), dual-stream collation, and LoRA merge; register both recipes in the diffusion nightly CI config.
- tools: add validate_ltx2_roundtrip.py (encode/decode round-trip check with PSNR/audio-level gates) and generate_ltx2_samples.py (memorization eval: regenerate training clips from training captions for base/full-FT/LoRA comparison).
- deps: add PyAV ("av") for audio decoding.

# What does this PR do ?

Add a one line overview of what this PR aims to accomplish.

# Changelog

**Pre checks**:

# Additional Information

- Related to # (issue)


tools/diffusion/generate_ltx2_samples.py
pthombre, line 1: Why do we need this script?

tools/diffusion/validate_ltx2_roundtrip.py
pthombre, line 1: Lets try to keep the tools directory generic. Is there a need for a specific ltx validation script?

tools/diffusion/generate_ltx2_samples.py
linnanwang: Good catch — we don't. Removed in 0a0c3a1.

It duplicated `examples/diffusion/generate/generate.py`, which already covers pipeline loading, full-FT and LoRA checkpoint loading, and distributed inference. The one thing it did that the shared script couldn't was write the generated **audio** track: `run_inference` called `export_to_video(output.frames[0])` and dropped `output.audio`, so LTX-2 came out as silent mp4s.

That capability is now in the shared script, and generically — the branch keys on the output carrying an `audio` field rather than on a pipeline name, so `WanPipelineOutput` / `HunyuanVideoPipelineOutput` / `LTXPipelineOutput` (which declare only `frames`) take the unchanged `export_to_video` path. Sample rate resolves from `output.audio_sample_rate`, falling back to the pipeline vocoder. Added `configs/generate_ltx2.yaml` alongside the other model configs.

Verified end-to-end on a full-FT and a LoRA checkpoint: 768x384, 89 frames, aac 48 kHz stereo muxed in.

tools/diffusion/validate_ltx2_roundtrip.py
linnanwang: Agreed — removed in 0a0c3a1.

This was a dev-time encode/decode harness used while bringing the preprocessing path up; it didn't need to ship. `tools/diffusion/` is back to generic entry points plus the per-model processors.

Its one piece of genuinely reusable logic — the PyAV video+audio muxer — moved into `generate.py`, which is where the audio output path needs it. Encode-side latent normalization stays covered by the processor unit tests (`TestEncodeAudio::test_normalization_applied`), so nothing that ships lost coverage.

linnanwang:
@pthombre thanks for the review — both comments are addressed in 0a0c3a1, replies inline on each thread.

Short version: you were right that neither script needed to exist. Both are deleted, and `tools/diffusion/` is back to generic entry points plus the per-model processors. The only thing the LTX-2 sampler did that `examples/diffusion/generate/generate.py` couldn't was write the generated audio track, so that moved into the shared script — implemented generically (keyed on the pipeline output carrying an `audio` field, not on a model name), which leaves Wan / HunyuanVideo / FLUX on exactly the path they're on today.

Three related changes rode along in the same commit:

- **Double-allocation OOM in `_load_checkpoint_into_attr`.** The replacement transformer was moved to CUDA while the pipeline's existing one was still resident — twice the weights in VRAM. That OOM'd an 80GB device on LTX-2's ~19B transformer (~38GB in bf16). The old module is now released before the swap. This affects any large model loading a consolidated or sharded checkpoint, not just LTX-2.
- **AAC layout.** `add_stream("aac", ...)` defaults to stereo, which silently upmixed a mono waveform. The layout is now declared on the stream.
- **Default checkpoint id** moved from `dg845/LTX-2.3-Diffusers` to `diffusers/LTX-2.3-Diffusers`.

Net **+441 / −735**. New coverage in `test_generate_audio.py` (19 tests): frame/waveform conversion, sample-rate resolution, save routing (asserts video-only pipelines still hit `export_to_video` and are never muxed), and a real PyAV mux round trip gated on the optional `av` extra. Local run of the affected suites is green at 539 passed / 6 skipped, ruff clean.

Generation verified end-to-end on GPU against both a full-FT and a LoRA consolidated checkpoint — 768x384, 89 frames, aac 48 kHz stereo present in every clip.

pthombre (approved):

thomasdhc (approved):

PR #3166, opened 2026-07-22
feat(speculative): add the multimodal speculative decoding (MSD) core

khazic:
## What

Add the multimodal speculative decoding (MSD) core for VLM targets.

MSD extends EAGLE-1/2 feature drafting to vision-language targets. Text positions keep the usual concatenation of target features and next-token embeddings, while image positions feed the target VLM's already-projected image embeddings straight into the draft transformer, preserving the non-causal relationship between visual patches.

## Changelog

- Add a modality-aware EAGLE draft that routes image positions around the text feature projection and consumes projected image embeddings directly.
- Add a frozen Hugging Face VLM target wrapper that captures the embeddings fused before the language backbone, including the nested `model.language_model` layout used by Qwen3-VL.
- Add hidden-state and soft-token supervision, with the frozen target head detached from the draft loss.
- Add the two-stage MSD data curriculum, with deterministic per-step source selection that is identical across ranks and across a resume.
- Add the reference decoding primitives: recursive feature-level tree drafting, engine-ready tree attention and retrieval metadata, greedy posterior verification, and lossless stochastic chain verification.

## Pre-checks

- [x] Read and followed contributor guidelines.
- [x] Added unit tests covering the draft, the target wrapper, the curriculum, and the decoding primitives.
- [x] Ran ruff format and ruff check on changed files.
- [x] Ran the full `tests/unit_tests/speculative` suite on CPU and on GPU.
- [x] Validated on real Qwen3-VL weights: feature capture, loss and backward, a training smoke, and tree decoding.
- [x] Commit includes DCO sign-off.

## Additional information

This PR contains the MSD training core, the target feature interface, and the reference decoding path. A following change will connect it to the VLM recipe and data pipeline.

Because there is no recipe yet, the numbers in this PR establish that the core is numerically sound (feature capture, gradient flow, single-sample convergence, lossless verification) rather than the paper's end-to-end acceptance and speedup results. Those need the recipe, checkpoint integration, iterative generation, and KV cache, and belong to the follow-up change.

One known limitation to record for that follow-up: the target wrapper has no document-boundary awareness, so it must not be paired with the packed VLM collators. Those emit an indexed `attention_mask` (1, 2, ... per document) and apply their autoregressive shift during packing, which the wrapper's single-tail handling does not model. The LLM path solves this by threading `seq_lens` and per-document `position_ids` through `HFEagleTargetModel`; MSD will need the same.


khazic:
## Server validation

Validated commit `c181b9147` on a single node with 8x NVIDIA A100-SXM4-80GB GPUs. The run used a local `Qwen3-VL-30B-A3B-Instruct` checkpoint at `/data/models/Qwen3-VL-30B-A3B-Instruct`, with no model download.

Focused unit tests:

```text
5 passed, 15 warnings in 7.89s
MSD_SERVER_PYTEST_EXIT=0
```

Real CUDA multimodal target forward, with an in-memory 32x32 RGB image:

```text
MSD_MODEL_READY cuda:0
MSD_INPUT_READY {'input_ids': (1, 81), 'attention_mask': (1, 81), 'mm_token_type_ids': (1, 81), 'pixel_values': (256, 1536), 'image_grid_thw': (1, 3)}
MSD_REAL_VLM_FORWARD_OK {'inputs_embeds': (1, 81, 2048), 'input_hidden_states': (1, 81, 2048), 'target_hidden_states': (1, 81, 2048), 'target_logits': (1, 81, 151936), 'image_mask': (1, 81)} image_tokens= 64
MSD_REAL_VLM_EXIT=0
```

This run exposed the nested `model.language_model` layout used by Qwen3-VL. The PR now captures fused image embeddings from that language backbone, with a dedicated unit test.

khazic:
## Real CUDA loss and backward validation

Validated commit `ffbbdae51` against the local `Qwen3-VL-30B-A3B-Instruct` checkpoint on an A100 80GB GPU. The smoke uses a one-layer BF16 MSD draft, an in-memory 32x32 RGB image, and runs the actual MSD objective through `loss.backward()`.

```text
MSD_REAL_LOSS_BACKWARD_OK {'loss': 2.7320215702056885, 'hidden_loss': 1.484375, 'token_loss': 12.476465225219727, 'accuracy': 0.0, 'valid_tokens': 80, 'grad_norm': 3.066777467727661, 'target_lm_head_grad': False}
MSD_REAL_LOSS_EXIT=0
```

The first real loss run surfaced a BF16 attention-probability mismatch. This PR now aligns probabilities with the value dtype and detaches the frozen target LM head. The result above is from the corrected implementation.

khazic:
## 100-step CUDA training smoke

Validated commit `ffbbdae51` on an A100 80GB GPU with local `Qwen3-VL-30B-A3B-Instruct` weights. AdamW updated only a one-layer BF16 MSD draft for 100 steps. This is a fixed single-sample convergence smoke using an in-memory image, not a full dataset run.

```text
step   loss       hidden_loss  token_loss  accuracy
1      2.718215   1.476562     12.416525   0.0000
10     1.876499   1.078125      7.983738   0.1500
20     1.686837   0.945312      7.415250   0.2000
30     1.630313   0.894531      7.357816   0.4500
40     1.593260   0.863281      7.299788   0.4625
50     1.572596   0.843750      7.288455   0.4750
60     1.558358   0.832031      7.263268   0.4625
70     1.546510   0.820312      7.261970   0.4625
80     1.541877   0.816406      7.254702   0.4625
90     1.537416   0.812500      7.249158   0.4750
100    1.537372   0.812500      7.248724   0.4750

MSD_TRAIN_100_DONE {'first_loss': 2.718214988708496, 'last_loss': 1.5373724699020386, 'target_lm_head_grad': False}
MSD_TRAIN_100_EXIT=0
```

The total loss fell by 43.4 percent. Both component losses decreased, backward completed each step, and the frozen target head received no gradients.

khazic:
## MSD two-stage curriculum

Added the two-stage data curriculum required by MSD in commit `762c39b7f`.

`MSDCurriculumConfig` exposes `text_only_steps`, `multimodal_ramp_steps`, `final_multimodal_ratio`, and `seed`. `MSDCurriculum` provides a deterministic data-source selection for each global optimizer step plus checkpointable progress for `next_source()`.

```text
stage 1: steps [0, text_only_steps) select text-only batches
stage 2: multimodal probability rises linearly to final_multimodal_ratio
post-ramp: the configured final multimodal mixture is used
```

The selection is seeded by global step, so every distributed rank and a resumed run choose the same text or multimodal source at a given step. Server validation:

```text
14 passed, 15 warnings in 6.97s
MSD_CURRICULUM_SERVER_PYTEST_EXIT=0
```

khazic:
## MSD tree decoding and verification

Added commit `825d108ae` with the remaining MSD decoding primitives:

- Recursive feature-level drafting after a target-sampled root token.
- Top-k, beam-pruned candidate-tree expansion with engine-ready tree attention and retrieval metadata.
- Greedy posterior verification for tree leaves.
- Lossless stochastic chain verification using `min(1, p/q)` acceptance and normalized positive-residual resampling on rejection.

The Hugging Face reference verifier replays leaf paths for model-agnostic correctness. Serving backends can consume the emitted tree layout for a single tree-attention verification forward.

Real Qwen3-VL-30B-A3B CUDA smoke on an in-memory image:

```text
MSD_TREE_DECODE_OK {'nodes': 4, 'leaf_paths': ((8697, 102282, 56267), (8697, 111720, 56267)), 'tree_mask_shape': (5, 5), 'accepted_draft_tokens': 0, 'emitted_token_ids': (8697, 13)}
MSD_TREE_DECODE_EXIT=0
```

The random draft has no accepted draft tokens, as expected before training. Server unit validation:

```text
23 passed, 15 warnings in 7.24s
MSD_DECODE_SERVER_PYTEST_EXIT=0
```

khazic:
## Review pass: six correctness fixes on the MSD core

A correctness review of this PR surfaced six defects. All are fixed in `556f7c51f`, each with a regression test. Four came from an external review of the earlier commits, two were found while verifying those.

### Fixed

**1. The zero-filled tail was supervised.** Every MSD supervision tensor is shifted left with a zero tail, but the collator's already-aligned `loss_mask` could still mark that final position as supervised, so the draft was trained towards a zero hidden state and a uniform soft-token target. `generate_batch` now drops that column, on a clone so the caller's mask is untouched.

**2. Seeded sampling was impossible on GPU.** `torch.rand` defaulted to CPU while the residual resample followed the logits, so no single generator could serve both calls once verification ran on GPU: a CUDA generator failed the first call, a CPU generator failed the second. The acceptance draw now samples on the logits device, and a generator whose device type does not match is rejected up front instead of failing opaquely part-way through a round.

**3. Acceptance used `draw <= acceptance`.** The lossless rule is a strict `draw < min(1, p/q)` over `torch.rand`'s half-open `[0, 1)`. `random_value` is validated against that same interval now.

**4. A candidate both models rule out was always accepted.** When `target_probs[c]` and `draft_probs[c]` both underflow to exactly zero, the ratio is `NaN`, and `min(1.0, NaN)` returns `1.0` rather than `NaN`, because any comparison against `NaN` is false. With `acceptance == 1.0` every draw is below it, so a token carrying no target mass was emitted unconditionally. `NaN` now maps to zero acceptance. A draft-only zero still divides to `+inf` and correctly accepts.

**5. The rejection branch crashed when the positive residual had no mass.** `residual / residual.sum().clamp_min(eps)` only guards the division: zeros divided by `eps` stay zero, and `torch.multinomial` rejects an all-zero row outright. This does not require the two distributions to be equal. Rounding of the two softmax normalizers is enough to leave `target_probs <= draft_probs` at every index, for example:

```text
target_logits = [41.3499, -19.4971, -34.1979,  2.7798]
draft_logits  = [44.5302,  24.0608, -11.3887, 18.6314]
-> residual.sum() == 0.0 while the distributions are not equal
-> RuntimeError: invalid multinomial distribution (sum of probabilities <= 0)
```

The correction now falls back to a draw from the target, which is the correct correction distribution in that case.

**6. The draft carried a vocabulary-sized embedding it never read.** MSD feeds the target's fused embeddings straight into the draft, so the inherited `embed_tokens` never received a gradient and aborted the default DDP reduction. The base class gains a `builds_token_embeddings` opt-out, so EAGLE-1/2 keeps its table unchanged and MSD never allocates or initializes one. On this config that is 848.8M parameters down to 226.5M.

### Investigated, not defects

**Supervision alignment.** The wrapper does not shift `loss_mask`, unlike the EAGLE-1/2 path. That is correct here and not an off-by-one: every VLM collator in `collate_fns.py` already emits `labels[:, 1:]` against `input_ids[:, :-1]`, so the incoming mask is next-token aligned before `generate_batch` sees it, whereas the LLM datasets pass an unshifted mask that `HFEagleTargetModel` must shift itself. Both paths land supervision on the same underlying token. Only the tail needed fixing, item 1 above.

Padding needs no separate handling either. A label is only non-`-100` inside the real sequence, so `loss_mask[k]` true forces `k+1` to be a real position, which is exactly when the shifted target is real. The one position this cannot cover is the batch's full-width column, which item 1 now drops.

**Constructing the draft from a target config with `attn_implementation="sdpa"`.** This does raise, but not on any real path: `train_eagle1.py` builds the draft config through `target_config.to_dict()` and `from_dict()`, and `to_dict()` drops `_attn_implementation`, so the round-trip resets it to eager. Coercing it inside the draft was also rejected deliberately: `test_eagle_draft_attn_support.py` encodes the convention that a draft must not claim a backend its attention cannot run, and silently ignoring an explicit request is worse feedback than the error.

### Validation

Single node, 8x NVIDIA A100-SXM4-80GB, commit `556f7c51f`, `Qwen3-VL-8B-Instruct` for the real-target checks.

Full `tests/unit_tests/speculative` suite:

```text
CPU:  795 passed, 12 skipped
GPU:  806 passed, 1 skipped
```

EAGLE-1/2 regression, confirming the `builds_token_embeddings` opt-out does not leak into the existing path:

```text
tests/unit_tests/speculative (eagle1/2 files):  66 passed, 1 skipped
tests/unit_tests/recipes/llm -k eagle:         176 passed
```

<details>
<summary>Two-rank DDP check, with a control that re-adds the unused parameter</summary>

```text
step 0: loss 1.57192
  step 1: loss 1.48646
  step 2: loss 1.40526
[PASS] 3 DDP steps completed and fc grads are synchronized
[PASS] control fails as expected :: Expected to have finished reduction in the prior iteration before starting a new one. This error indicates that your module has parameters that were not used in producing loss. You can enable unused p
RESULT: DDP CHECK PASSED
```

</details>

<details>
<summary>CUDA sampling, real Qwen3-VL feature capture, and a 20-step overfit</summary>

```text
==============================================================================
A. pre-fix mechanism: a CUDA generator cannot drive the CPU-default rand
==============================================================================
[PASS] pre-fix rand((), generator=cuda) raises :: Expected a 'cpu' device type for generator but found 'cuda'
[PASS] pre-fix multinomial(cuda, generator=cpu) raises :: Expected a 'cuda' device type for generator but found 'cpu'

==============================================================================
B. post-fix: seeded CUDA sampling works and is reproducible
==============================================================================
[PASS] CUDA generator accepted by accept_or_resample :: 200 draws
[PASS] same seed reproduces the same decisions
[PASS] a different seed diverges
[PASS] verify_stochastic_chain runs on CUDA :: (0,)

==============================================================================
C. losslessness on CUDA: emitted distribution matches the target
==============================================================================
[PASS] speculative sampling stays lossless on CUDA :: total variation = 0.00161 over 200000 draws

==============================================================================
D. draft model: no vocabulary-sized embedding, every parameter gets a grad
==============================================================================
[transformers] `torch_dtype` is deprecated! Use `dtype` instead!
[PASS] MSD draft has no embed_tokens
[PASS] base EAGLE-1/2 draft keeps embed_tokens
[PASS] MSD draft drops the vocabulary table :: 848.8M -> 226.5M, saved 622.3M (1.24 GiB bf16)
[PASS] no draft parameter is left without a gradient :: missing=[]

==============================================================================
E. real Qwen3-VL target: feature capture and loss_mask tail
==============================================================================

Loading weights:   0%|          | 0/750 [00:00<?, ?it/s]
Loading weights:  50%|█████     | 377/750 [00:00<00:00, 3766.71it/s]
Loading weights: 100%|██████████| 750/750 [00:00<00:00, 3747.31it/s]
image_token_id = 151655
sequence length = 85, image tokens = 64
[PASS] fused embeddings captured :: (1, 85, 4096)
[PASS] image mask non-empty :: 64 shifted image positions
[PASS] shifted tail is zero-filled
[PASS] loss_mask never supervises the zero tail :: loss_mask[-1]=False, supervised=84/85
[PASS] caller's loss_mask is not mutated
[PASS] interior mask entries are preserved

==============================================================================
F. 20-step overfit on the real VLM features
==============================================================================
/data/runs/msd_fix_validation/msd_cuda_checks.py:193: UserWarning: Converting a tensor with requires_grad=True to a scalar may lead to unexpected behavior.
Consider using tensor.detach() first. (Triggered internally at /pytorch/torch/csrc/autograd/generated/python_variable_methods.cpp:838.)
  losses.append(float(metrics.loss))
  step  0  loss 2.6289  hidden 1.2969  token 13.3204  acc 0.0000  valid 84
  step  5  loss 1.9503  hidden 1.2188  token 7.3154  acc 0.3095  valid 84
  step 10  loss 1.6948  hidden 1.1719  token 5.2296  acc 0.4167  valid 84
  step 15  loss 1.4811  hidden 1.1250  token 3.5610  acc 0.5119  valid 84
  step 19  loss 1.4029  hidden 1.0938  token 3.0917  acc 0.5476  valid 84
[PASS] draft overfits a single multimodal sample :: 2.6289 -> 1.4029
[PASS] target stays frozen

==============================================================================
RESULT: ALL CHECKS PASSED
==============================================================================
```

</details>

Section A above records the pre-fix failure mode directly, so the CUDA generator defect and its fix are both demonstrated in the same run. Section C is the losslessness check: with a strict `<` and the `NaN` guard in place, the emitted distribution still matches the target to a total variation of 0.00161 over 200000 draws.

This supersedes the description of stochastic verification in my earlier comment on `825d108ae`, which described the acceptance rule before items 3, 4 and 5.

nemo_automodel/components/speculative/eagle/msd_decode.py
```diff
@@ -0,0 +1,563 @@
...
+    if active_length < 1:
+        raise ValueError("MSD verification requires a non-empty prompt.")
+
+    logits_by_leaf = []
+    for path in proposal.candidate_paths():
+        candidate_ids = torch.tensor([path], dtype=input_ids.dtype, device=input_ids.device)
+        extended = dict(model_inputs)
+        extended["input_ids"] = torch.cat((input_ids[:, :active_length], candidate_ids), dim=1)
+        extended["attention_mask"] = torch.ones_like(extended["input_ids"], dtype=attention_mask.dtype)
+        if "mm_token_type_ids" in extended:
```
HuiyingLi, line 489: Preferably model specific knowledge doesnt leak into recipe

khazic: `mm_token_type_ids` is not knowledge specific to one model here, it is this repo's own VLM batch key: it is produced by `collate_fns.py` and `neat_packing_vlm.py` and consumed in `distributed/pipelining/hf_utils.py`. The handling is a presence guard, so a target that never emits it is untouched.

It is also confined to `verify_hf_greedy_tree`, which is the reference correctness path. That function deliberately replays each leaf path through a plain HF forward, so it has to extend every sequence-aligned input alongside `input_ids`, not just `input_ids` itself. The engine-facing output (`proposal.layout`) carries no key names at all, and no recipe imports this module today.

If you would still rather see zero key names in this file, I can generalize it to "extend every batch entry whose last dimension matches `input_ids`", which drops the literal without losing the behavior. Happy to do that in this PR or the follow-up recipe change, whichever you prefer.

HuiyingLi (approved):
Thank you so much!

Issue #2861, opened 2026-06-30
Feature Request: Generic Automodel CP interface for downstream libraries e.g. NeMo RL


sharonyu-115:
## Is Your Feature Request Related To A Problem? Please Describe.

Downstream training libraries such as NeMo RL need to use Automodel context parallelism (CP), but  <br>today they must understand and partially reimplement Automodel's CP internals. The current APIs expose several related but different paths:

* Generic SDPA CP via `create_context_parallel_ctx`.
* Transformer Engine CP, including THD or packed batch layouts.
* Model-owned CP via `prepare_model_inputs_for_cp` and `_cp_make_batch_fn` such as for gemma4.

For example, NeMo-RL can use the generic `create_context_parallel_ctx` path for standard SDPA-style CP. For models such as Gemma4, however, NeMo-RL needs a separate model-owned CP path that calls `prepare_model_inputs_for_cp`, invokes `make_cp_batch_and_ctx`, infers the sequence shard layout, computes local next-token targets, and implements its own token-level gather.

This is brittle for downstream libraries because `prepare_model_inputs_for_cp` is not a reliable capability signal by itself. Some models expose that method only to precompute full-sequence multimodal embeddings before generic CP sharding. Other models use it to attach a model-owned CP sharder and attention implementation. Downstream libraries should not need to infer those semantics from method names or private batch keys.

It would be good to have a public Automodel-owned CP contract that tells downstream libraries:

* Which CP backend was selected.
* Which sequence layout is being used.
* Which local sequence range each rank owns.
* How to gather token-level or sequence-level outputs correctly.
* Which context manager and model inputs should be used for the forward.

## Describe The Solution You'd Like

Add a public Automodel CP preparation interface for downstream training libraries. The interface should accept a model, device mesh, batch, etc., then return a stable CP plan object. So the general idea is that Automodel exposes the CP layout and helper operations needed for NeMo RL to compute losses and logprobs without branching on private CP implementation details.

For example:

```python
cp_plan = prepare_cp_forward(
    model=model,
    device_mesh=device_mesh,
    batch=batch,
    purpose="causal_lm_forward",
)

with cp_plan.forward_context():
    outputs = model(**cp_plan.model_inputs)

local_logits = extract_logits(outputs)
local_targets = cp_plan.local_next_token_targets(
    input_ids=batch["input_ids"],
    seq_dim=1,
)

local_token_logprobs = compute_local_token_logprobs(local_logits, local_targets)
[...]

PR #3187, opened 2026-07-22
refactor(cp): sharder ownership

akoumpa:
# What does this PR do?

Simplifies the context-parallel forward API introduced by NVIDIA-NeMo/Automodel#2937 and makes `ContextParallelSharder` the single recipe-facing owner of CP batch preparation and token layout.

The sharder now:

* derives TE and Magi behavior from the live model instead of accepting `use_te`, `magi`, or backend arguments;
* derives packed THD input from `batch["qkv_format"]`;
* reuses a model-owned `prepare_model_inputs_for_cp` sharder when one is available;
* binds the CP/TP meshes once, so token shard/gather calls do not accept a mesh argument;
* shards the batch through `shard(batch)` and returns `(train_ctx, sharded_batch)`;
* lives under `nemo_automodel.components.distributed.context_parallel`, with only `ContextParallelSharder` exported at the package level.

# API

```python
from nemo_automodel.components.distributed.context_parallel import ContextParallelSharder

cp_sharder = ContextParallelSharder(
    model,
    device_mesh,
    batch,
    padding_token_id=tokenizer.pad_token_id,
)
train_ctx, batch = cp_sharder.shard(batch)

# Optional token-aligned tensors use the same captured CP layout.
local_advantages = cp_sharder.shard_token_tensor(
    advantages,
    fill=0.0,
)

with train_ctx():
    outputs = model(**batch)
    local_logprobs = compute_token_logprobs(outputs, batch["labels"])

full_logprobs = cp_sharder.gather_token_tensor(
    local_logprobs,
    trim=True,
)
```

There is intentionally no separate `.context()` call. `shard(batch)` performs the batch transformation, captures the resulting token layout, and returns the context factory required by the model forward.

`shard_token_tensor(...)` and `gather_token_tensor(...)` use the mesh and layout already bound to the sharder. This keeps auxiliary token tensors aligned with round-robin SDPA, contiguous model-owned layouts, packed THD partitions, and Magi dispatch permutations without adding them to the model batch.

A `DTensor` is not returned because standard `Shard(dim)` placements cannot represent all supported CP layouts, including head/tail ordering, packed THD partitions, Magi permutations, and model-owned re-padding.

# Changelog

* Remove the recipe-facing `prepare_cp_forward` wrapper.
* Rename `ContextParallelismSharder` to `ContextParallelSharder`.
* Move CP implementation modules under `components/distributed/context_parallel/`.
* Update LLM, VLM, dLLM, KD, model-owned CP hooks, tests, and tools to use the new package and API.
* Keep backend-specific and model-specific preparation behind the sharder boundary.

# Validation

* `138 passed` across the focused CP sharder, CP utility, differential coverage, input-embedding, and Mamba CP unit tests.
* `142 passed, 1 skipped` across Magi and model-owned CP tests.
* Ruff checks and syntax parsing passed.
* Verified that the previous root-level CP module import paths are removed.

Local recipe collection is currently blocked before reaching this code by the environment's incompatible `mlflow` and system `cachetools` installations.

# Additional Information

Related to NVIDIA-NeMo/Automodel#2861 and NVIDIA-NeMo/Automodel#2937.


nemo_automodel/components/distributed/context_parallel/sharder.py
```diff
@@ -236,38 +237,146 @@ class ShardLayout:
...
+        self.shard_batch = resolved.shard_batch
+        self.local_token_global_indices = resolved.local_token_global_indices
+        self.shard_layout = resolved.shard_layout
+        self._cp_mesh = resolved._cp_mesh
+        self._tp_mesh = resolved._tp_mesh
+        self._loss_mask = resolved._loss_mask
+        self._padding_token_id = resolved._padding_token_id
+
+    @classmethod
+    def _from_strategy(
```
HuiyingLi: Thanks @akoumpa . We have a init and also a _from_strategy and then bind. I'm wondering whats the use case/if necessary?

nemo_automodel/components/distributed/context_parallel/sharder.py
```diff
@@ -236,38 +238,191 @@ class ShardLayout:
...
```
HuiyingLi: Thanks for you prev commit. Are we still keeping this?

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -972,24 +949,14 @@ def _forward_backward_step(
...
             self.model_parts[0] if hasattr(self, "model_parts") else None,
             self.device_mesh,
             batch,
-            magi=self.magi,
-            use_te=_use_te_value,
             padding_token_id=self.tokenizer.pad_token_id if self.tokenizer else 0,
-            num_chunks=_num_chunks_value,
+            num_chunks=self.pp.pp_batch_size // self.pp.pp_microbatch_size if self.pp_enabled else 1,
         )
+        train_ctx, batch = cp_sharder.shard(batch)
```
HuiyingLi, line 959: Not super sure about using init sharder+sharder.shard().  The sharder init becomes a bit confusing. Could you please explain your consideration for replacing the function call? many thanks

akoumpa: Thanks @HuiyingLi, the goal was to make `ContextParallelSharder` the single owner of CP strategy and mesh state, rather than keeping a procedural wrapper around it.

The object-based API has two advantages:
1. The sharder can be cached on the trainer and reused, avoiding repeated strategy and mesh resolution. Only the batch-dependent `ShardLayout` is refreshed by each `shard()` call.
2. The sharder owns the device mesh and related options, so `shard()`, `shard_token_tensor()`, and `gather_token_tensor()` do not need the mesh and other runtime arguments passed on every call.

The separate `.shard(batch)` operation is intentional because that is where the batch is mutated and any data-dependent layout is computed and retained for subsequent token-aligned sharding or gathering.

PR #3192, opened 2026-07-22
fix(distributed): activation-checkpoint Qwen3-Next linear_attn layers

amolkhanna:
# What does this PR do ?

Adds `"linear_attn"` to the attention attribute-name tuple in `apply_submodule_checkpointing`, so hybrid linear-attention decoder layers (Qwen3-Next / Qwen3.5 / Qwen3.5-MoE Gated DeltaNet) are activation-checkpointed instead of silently skipped.

# Changelog

- `nemo_automodel/components/distributed/activation_checkpointing.py`: append `"linear_attn"` to the attention attr tuple `("self_attn", "attention", "attn")` used by `_wrap_first_existing_attr`. Gated DeltaNet layers name their mixer `linear_attn`, so they were never wrapped — their backward-saved tensors stayed resident and long-context training OOMed (~176 GiB peak at 262k on Qwen3.6-27B, where 48/64 layers are GDN).
- `tests/unit_tests/distributed/test_activation_checkpointing.py`: add a CPU unit test asserting a `linear_attn` mixer is checkpoint-wrapped (and the MLP still wraps).

## Why this is safe

- A hybrid decoder block has **either** `self_attn` (full-attention layer) **or** `linear_attn` (linear-attention layer), never both, and `_wrap_first_existing_attr` wraps only the first match — so no double-wrapping. `"linear_attn"` is placed last, preserving existing precedence.
- `linear_attn` is the canonical mixer name across `qwen3_next`, `qwen3_5`, and `qwen3_5_moe` (their `state_dict_adapter.py` files key on `.linear_attn.`), so this is a generic contract fix, not model-specific branching.
- GDN layers don't use the KV `DynamicCache`, so sharing the `skip=has_kv_sharing` guard (which exists to prevent `self_attn` double-writes to the cache) is harmless for them.

**Pre checks**:

- [x] Did you write any new necessary tests? (CPU unit test in `tests/unit_tests/distributed/test_activation_checkpointing.py`)
- [ ] Did you add or update any necessary documentation? (n/a — internal AC behavior fix)

# Additional Information

- The fix generalizes activation checkpointing to any linear-attention/GDN model already supported in the repo; all callers of `apply_submodule_checkpointing` (FSDP2, DDP, MoE parallelizer) inherit it automatically.
- Verified locally: `ruff format --check` / `ruff check` clean on the changed source file; the new unit test passes and a `self_attn`-only block still wraps (no regression).


yuhezhang-ai:
Thanks for finding and fixing this, it’s a good catch, especially for GDN-heavy models at long context.
One additional path seems to need the same update: when FSDP2 runs with enable_compile=True, components/distributed/parallelizer.py bypasses apply_submodule_checkpointing and uses its own attribute list:
`("self_attn", "attention", "attn", "mlp", "feed_forward", "ffn")`
Could you add linear_attn there as well, ideally with a focused test for the compiled path? Otherwise compiled Qwen3-Next/Qwen3.5 training would still silently leave GDN uncheckpointed.

Minor PR-body clarification: Qwen3.5-MoE with EP already uses whole-block checkpointing through the MoE parallelizer, so this change mainly benefits dense Qwen3-Next/Qwen3.5 and MoE runs using the generic FSDP2 path. It would be helpful to narrow the applicability claim accordingly.

The core change looks good, thanks again for contributing this.

akoumpa (approved):
Thanks @amolkhanna

PR #3193, opened 2026-07-22
feat(checkpoint): preserve intrinsic fp32 during offline consolidation

yuhezhang-ai:
## Summary

This is a minor checkpoint-export feature aimed primarily at config-only pretraining runs. The offline Hugging Face consolidation tool can now produce a mostly-BF16 checkpoint with `--cast-dtype bf16` while preserving tensors that AutoModel explicitly requires to remain FP32.

When a run starts from an HF checkpoint, the original per-tensor dtype metadata remains the baseline. For pretraining from config, model-owned dtype overrides are now recorded even when no original HF safetensors are available. During offline consolidation, those intrinsic FP32 entries take precedence over the broad export cast; other ordinary floating-point tensors still use the requested dtype.

## Why

Pretraining users often save distributed checkpoints first and create a consolidated HF checkpoint offline. Requesting a BF16 export should reduce checkpoint size without violating the model's stricter precision contract for protected tensors.

## Validation

- 201 checkpoint and offline-consolidation unit tests passed
- Ruff formatting and lint checks passed
- Real Qwen3.5 0.8B two-shard DCP checkpoint consolidated with `--cast-dtype bf16`:
  - 452 BF16 tensors
  - 36 checkpoint-mapped FP32 tensors preserved on disk
  - 102 uint8 tensors
  - exported config reloads locally through `AutoConfig`
  - full Hugging Face and AutoModel weight reloads completed on H100
  - both loaders completed the same deterministic text forward pass
  - AutoModel restored its stricter 452-BF16/36-FP32 runtime contract, which intentionally differs from the older checkpoint's HF-oriented dtype mapping
  - Hugging Face and AutoModel logits had 0.999812 cosine similarity

## Tracking

[AM-398](https://linear.app/nvidia/issue/AM-398/minor-feature-make-offline-hf-ckpt-consolidate-tool-preserve-intrinsic)

### DIFF
--- a/nemo_automodel/components/checkpoint/_backports/consolidate_hf_safetensors.py
+++ b/nemo_automodel/components/checkpoint/_backports/consolidate_hf_safetensors.py
@@ -145,11 +145,24 @@ def _resolve_output_dtype(
 ) -> tuple[torch.dtype, str]:
     """Resolve the output dtype for a tensor.
 
-    Explicit cast_dtype wins for ordinary floating-point tensors. Otherwise,
-    original HF dtype metadata restores ordinary floating-point tensors per FQN
-    when available; checkpoints without that metadata keep the saved dtype.
+    Per-FQN fp32 metadata is an intrinsic model contract and remains authoritative
+    when ``cast_dtype`` is set. The explicit cast wins for other ordinary
+    floating-point tensors. Without a cast, per-FQN metadata restores ordinary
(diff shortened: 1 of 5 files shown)


yuhezhang-ai:
/ok to test [608531c](https://github.com/NVIDIA-NeMo/Automodel/pull/3193/commits/608531c4de4b1cc54ff13aeb65742df14a668b24)

akoumpa (approved):

PR #3214, opened 2026-07-24
perf(bagel): grouped MoT routing + fused SwiGLU/RoPE

zyzhou5:
# What does this PR do ?

 Adds three opt-in throughput optimizations for BAGEL packed Mixture-of-Transformers training, selected via new BagelBackendConfig fields under model.backend (mot_grouped, fused_swiglu, fused_rope): 
- mot_grouped reorders the packed sequence once so und/gen tokens form contiguous blocks, letting every MoT layer route the pointwise path (LayerNorm/MLP/QKV/QK-norm/O-proj) by a scalar slice boundary instead of per-layer gather/scatter, while attention still runs in original token order (restored around the kernel) so its block-diagonal FlexAttention mask stays block-sparse — a pure permutation in which each token is still processed by its correct expert
- fused_swiglu and fused_rope compile silu(gate)*up and the rotary apply into single pointwise kernels over activation tensors only (no FSDP2/DTensor params, so no torch.compile recompile thrash). 

All three are threaded through construction like the existing linear/rms_norm backends and the shipped BAGEL pretrain and SFT example configs enable them; the change also fixes an UnboundLocalError in PackedAttentionMoT.forward_train's FlexAttention path.

### DIFF
--- a/examples/multimodal_finetune/bagel/bagel_sft.yaml
+++ b/examples/multimodal_finetune/bagel/bagel_sft.yaml
@@ -25,9 +25,13 @@ dist_env:
 
 model:
   # Transformer Engine is enabled for BAGEL's language-model Linear and RMSNorm layers.
+  # mot_grouped / fused_swiglu / fused_rope are throughput optimizations (see BagelBackendConfig).
   backend:
     linear: te
     rms_norm: te
+    mot_grouped: true
+    fused_swiglu: true
+    fused_rope: true
   _target_: nemo_automodel.NeMoAutoModelForMultimodalLM.from_pretrained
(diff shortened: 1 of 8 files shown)


zyzhou5:
/ok to test [23f1da9](https://github.com/NVIDIA-NeMo/Automodel/commit/23f1da924483e4642e8bee2aa9a0a8de5b4ba19d)

zyzhou5:
/ok to test [87471cd](https://github.com/NVIDIA-NeMo/Automodel/pull/3214/commits/87471cdfe4c4a78ff307ea12da7786319ccfae0c)

akoumpa (approved):

PR #3223, opened 2026-07-25
fix(thd): derive packed padding mask from the pack layout, not token ids

HuiyingLi:
# What does this PR do ?

Makes the packed-THD `padding_mask` come from the pack layout instead of comparing token ids against a `padding_token_id`.

## The bug

Three places built the mask the same way:

```python
padding_mask = input_ids == padding_token_id
```

That is only correct when the pad id never appears as content — something the caller cannot guarantee, and a downstream one already does not: molt leaves the sharder default of `0`, and **token id 0 is `!` in the GLM tokenizer**, so every exclamation mark is masked out of the MoE experts.

Passing GLM's *real* `pad_token_id` is no better. GLM-5.2 sets:

```
pad_token_id = 154820  ->  <|endoftext|>
eos_token_id = [154820, 154827, 154829]  ->  <|endoftext|>, <|user|>, <|observation|>
```

The pad id **is** the first eos. A value comparison would drop every document-final eos from MoE routing — the failure mode this repo already documented in `datasets/utils.py`:

> matching the pad token *value* misclassifies real tokens as padding whenever `pad_token_id` collides with a content token (e.g. `pad_token_id == eos_token_id`), which masks real eos/separator tokens out of the MoE experts -> near-random outputs on chat data

So for GLM there is no value of `padding_token_id` that yields a correct mask.

## The change

On a packed stream the pack metadata is the only reliable source: sequence `i` occupies `[start_i, start_i + padded_i)` and only its first `real_i` slots hold tokens.

- **`thd_utils.py`** — `process_input_for_thd` derives the mask from `seq_lens` / `seq_lens_padded`, which it already computes for `cu_seqlens`.
- **`glm_moe_dsa/cp.py`** — `_slice_thd_chunk_for_cp` index_selects the incoming mask with the same query indices it already applies to `input_ids` / `labels` / `position_ids`, instead of re-deriving it.
- **`context_parallel/utils.py`** — `_shard_thd_chunk_for_te` already partitioned `padding_mask` with `local_indices` and then **overwrote it** with a value comparison one line later. Drop the overwrite.

`padding_token_id` **stays in every signature**. A caller with no pack metadata still needs it, and it remains the tokenizer-provided filler — it is demoted, not deleted. The new `thd_padding_mask_from_token_ids` owns that fallback and validates that the comparison yields a right-padded run, raising when the id also occurs as content, so a colliding pad id fails loudly instead of silently masking content.

## Prior art

This repo has already reached the right answer twice: the shared collater switched to `~attention_mask` in #2633 (with the comment quoted above), and DeepSeek-V4's CP path builds its mask from its own repad bookkeeping rather than token values. Only the THD paths were left behind.

Megatron-Bridge draws the same line in `data/packing/in_batch.py`: it keeps a `pad_token_id` parameter, prefers the mask, and where it must fall back to a value comparison it validates that the result is a right-padded prefix and raises `"Direct sequence packing requires right-padded input rows."` otherwise — never silently mis-masking. This PR follows that shape exactly.

## Validation

New `tests/unit_tests/distributed/test_thd_padding_mask.py` (5 passed) pins the cases that motivated this:

- a pad id that is also `eos_token_id` — the document-final eos stays visible
- a content token equal to the default pad id (`0`) — not masked
- trailing pack pad — masked
- the fallback rejects a pad id that also occurs as content, and accepts a genuine right-padded run

Two existing tests encoded the old semantics and were updated: a GLM CP fixture now carries a pack-derived mask, and `test_padding_mask_correctness` asserts that a pack whose `seq_lens == seq_lens_padded` has no padding at all (previously it asserted the `0`s in `input_ids` were padding — the bug in test form).

`ruff format --check` / `ruff check` clean on all touched files.

Paired before/after run of `tests/unit_tests/distributed`, `models/glm_moe_dsa` and `models/deepseek_v4` on one node/image:

| | failed | passed |
|---|---|---|
| `main` | 9 | 1319 |
| this PR | 9 | 1326 |

`make_cp_batch_for_te`'s `num_chunks > 1` path had no coverage, so this PR adds `test_make_cp_batch_for_te_multi_chunk` for the per-chunk shard call.

The failure **sets are identical** — all nine are tile-kernel tests (`test_glm_moe_dsa_tilelang.py`, `test_dsv4_layers.py::TestDeepseekV4OptimizedKernels`) that fail on unpatched `main` because this image ships no tile kernels. No new failures.

## NeMo-CI validation

- Passed full release GLM-5.2 4k validation: parent pipeline https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/59837375; leaf job https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/375599505 (`50/50`, `SLURM_STATE=COMPLETED`, `SLURM_EXITCODE=0`).


athitten (approved):
LGTM, thanks @HuiyingLi !

PR #3227, opened 2026-07-25
fix(deepseek-v4): preserve fp32 model dtype

HuiyingLi:
## Summary

- preserve an explicit DeepSeek V4 `torch_dtype` after `PretrainedConfig` initialization
- initialize torch-fp32 RMSNorm modules in the requested model dtype so fp32 construction is uniform
- distinguish fp32 parameter/master-weight storage from the FSDP compute policy, while retaining fp32 compute for reference-sensitive islands
- add regression coverage for fp32 construction, bf16 compute with fp32 master weights, and the explicit all-fp32 fast path

## Root cause

`PretrainedConfig.__init__` overwrote the dtype assigned by `DeepseekV4Config`, torch-fp32 RMSNorm construction fell back to bf16, and the DSV4 all-fp32 FSDP fast path conflated uniform fp32 storage/master weights with fp32 compute. Together these prevented a uniform fp32 instantiation and ignored an explicit bf16 FSDP compute policy.

## Validation

- Scoped Ruff format/check: passed
- DSV4 unit suite in `auto2606rc9.sqsh` with TileLang 0.1.9 overlay: `205 passed, 2 skipped`
- Full 43-layer model: 284,332,230,231 total parameters / 284,106,500,055 trainable parameters
- 100-step distributed run: Slurm `14370695`, `COMPLETED`, exit `0:0`, wall time `00:10:19`
  - 16 nodes / 128 H100, EP128, PP1, CP1, HybridEP + TileLang
  - `torch.optim.AdamW`, fp32 parameter/master-weight storage, bf16 FSDP compute
  - fixed-length dataset padding to keep HybridEP all-gather shapes symmetric
  - all 100 logged losses and grad norms finite; no NaN/Inf, OOM, traceback, or NCCL timeout
  - step 0: loss 2.3385, grad norm 7.0413
  - step 99: loss 1.7022, grad norm 1.6035
- W&B: https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/prbffcy5

### DIFF
--- a/nemo_automodel/components/models/deepseek_v4/config.py
+++ b/nemo_automodel/components/models/deepseek_v4/config.py
@@ -133,7 +133,6 @@ class DeepseekV4Config(PretrainedConfig):
         self.attention_dropout = attention_dropout
         self.pretraining_tp = pretraining_tp
         self.initializer_range = initializer_range
-        self.torch_dtype = torch_dtype
 
         super().__init__(
             pad_token_id=pad_token_id,
@@ -143,3 +142,4 @@ class DeepseekV4Config(PretrainedConfig):
             use_cache=use_cache,
             **kwargs,
         )
+        self.torch_dtype = torch_dtype
--- a/nemo_automodel/components/models/deepseek_v4/config.py
+++ b/nemo_automodel/components/models/deepseek_v4/config.py
@@ -91,9 +91,14 @@ class DeepseekV4Config(PretrainedConfig):
         pretraining_tp: int = 1,
         tie_word_embeddings: bool = False,
         initializer_range: float = 0.02,
-        torch_dtype: str = "bfloat16",
+        torch_dtype: str | None = None,
         **kwargs,
     ):
+        dtype = kwargs.pop("dtype", None)
+        resolved_dtype = dtype if dtype is not None else torch_dtype
+        if resolved_dtype is None:
+            resolved_dtype = "bfloat16"
+
         self.vocab_size = vocab_size
         self.hidden_size = hidden_size
         self.moe_intermediate_size = moe_intermediate_size
@@ -140,6 +145,6 @@ class DeepseekV4Config(PretrainedConfig):
             eos_token_id=eos_token_id,
             tie_word_embeddings=tie_word_embeddings,
             use_cache=use_cache,
+            dtype=resolved_dtype,
             **kwargs,
         )
-        self.torch_dtype = torch_dtype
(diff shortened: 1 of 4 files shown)


akoumpa:
Hi @yuhezhang-ai could you review? thanks

yuhezhang-ai (commented):
I found one dtype regression to address.

nemo_automodel/components/models/deepseek_v4/config.py
```diff
@@ -143,3 +142,4 @@ def __init__(
             use_cache=use_cache,
             **kwargs,
         )
+        self.torch_dtype = torch_dtype
```
yuhezhang-ai: This change introduces a regression for the canonical Transformers 5 `dtype` argument. Specifically:

```python
DeepseekV4Config(dtype="float32").dtype
```

returns fp32 on `main`, but bf16 on this PR. `dtype="float32"` is first processed correctly by `super().__init__`; then, because `torch_dtype` was not supplied, its default `"bfloat16"` is assigned afterward. In Transformers 5, `torch_dtype` is an alias for `dtype`, so that assignment overwrites the caller’s fp32 value. The same problem causes an fp32 `to_dict()`/`from_dict()` round trip to return bf16.

Could we address this regression in this PR by resolving `dtype` and `torch_dtype` before calling `super()`—ideally using `torch_dtype=None` to distinguish “not supplied”—and add canonical `dtype` and serialization round-trip tests?

There is also a broader cleanup opportunity because several custom configs still expose the legacy `torch_dtype` API. I’m happy to handle that cross-model migration separately; this PR only needs to preserve both input paths correctly for DeepSeek V4.

HuiyingLi: Thanks, addressed in b95db108. `torch_dtype` now defaults to `None`; the constructor resolves canonical `dtype` first, then an explicit legacy `torch_dtype`, and finally the unchanged default `bfloat16`, before passing one canonical dtype to `PretrainedConfig`.

I added coverage for `DeepseekV4Config(dtype="float32")` and the fp32 `to_dict()`/`from_dict()` round trip. Targeted config/fp32 tests pass (`9 passed`), and a real 4-layer EP8 checkpoint run with fp32 parameter/master storage plus bf16 FSDP compute also completed cleanly (Slurm 14428384, exit 0; finite train and validation losses).

yuhezhang-ai:
/ok to test [b95db10](https://github.com/NVIDIA-NeMo/Automodel/pull/3227/commits/b95db10865598f2c72d56fa668213904955da539)

yuhezhang-ai (approved):

PR #3233, opened 2026-07-26
test(vlm): stabilize Qwen3.5 checkpoint robustness

HuiyingLi:
## Summary

- run Qwen3.5 checkpoint parity with matching torch grouped-MoE backends and preserve `A_log`/`dt_bias` in FP32 for Hugging Face loads
- release the Hugging Face reload model state before resume training and enable expandable CUDA allocator segments for the sequential robustness phases
- give the second `torchrun` a distinct rendezvous ID
- calibrate Qwen3.5 parity thresholds and robustness batch settings against existing Qwen3 MoE recipes and observed old-image results

## Root cause

The checkpoint robustness test loads and trains several 35B model instances sequentially in one worker. The Hugging Face reload phase retained allocator pressure before Phase 6, which could fragment memory and OOM the second training setup. The launcher also reused the completed finetune rendezvous for the robustness `torchrun`.

Qwen3.5 additionally compared different MoE execution implementations and downcast Gated DeltaNet state that Hugging Face keeps in FP32. That introduced expected backend/dtype drift, while the generic parity thresholds were stricter than existing Qwen3 MoE coverage.

These are test-infrastructure and recipe changes only; model common utilities are unchanged.

## Validation

- Exact prior CI image: `gitlab-master.nvidia.com/dl/joc/nemo-ci/main/automodel:pipe.59210842`
- AutoModel proof overlay: `5a8959b5b36f0e1a4a2fcfccff61e5d50c71397c`
- Scoped NeMo-CI parent: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/59644765 (`success`)
- Leaf pipeline: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/59644777 (`success`)
- Exact job: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/374245337 (`success`, Slurm `COMPLETED`)
- Finetune: 342 seconds; robustness: 856 seconds; total: 1201 seconds; no OOM
- `62 passed` in the focused checkpoint-robustness/config-resolver unit tests
- Ruff, `bash -n`, and `git diff --check` pass

### DIFF
--- a/examples/vlm_finetune/qwen3_5_moe/qwen3_5_35b.yaml
+++ b/examples/vlm_finetune/qwen3_5_moe/qwen3_5_35b.yaml
@@ -121,11 +121,10 @@ ci:
   checkpoint_robustness:
     check_source_load_parity: true
     experts_implementation: grouped_mm
-    hf_kl_threshold: 2e-2
-    hf_keep_in_fp32_modules: A_log,dt_bias
     hf_device_map_auto: true
+    hf_kl_threshold: 1e-1
+    loss_fn._target_: nemo_automodel.components.loss.chunked_ce.ChunkedCrossEntropy
     model.backend.experts: torch_mm
-    resume_loss_threshold: 1e-2
     source_load_cosine_threshold: 0.9985
     source_load_kl_threshold: 1e-1
     source_load_mean_kl_threshold: 1e-2
--- a/examples/vlm_finetune/qwen3_5_moe/qwen3_5_35b.yaml
+++ b/examples/vlm_finetune/qwen3_5_moe/qwen3_5_35b.yaml
@@ -122,6 +122,10 @@ ci:
     check_source_load_parity: true
     experts_implementation: grouped_mm
     hf_device_map_auto: true
+    # Qwen3.5 BF16 parity has known model/backend drift: AutoModel applies routing
+    # weights before down_proj while HF applies them after, and DeepEP combine uses
+    # different reduction numerics from HF grouped_mm's FP32 reshape-and-sum.
+    # Focused parity runs measured max/mean source KL of ~0.068/~0.0086 (AM-711).
     hf_kl_threshold: 1e-1
     loss_fn._target_: nemo_automodel.components.loss.chunked_ce.ChunkedCrossEntropy
     model.backend.experts: torch_mm
--- a/examples/vlm_finetune/qwen3_5_moe/qwen3_5_35b.yaml
+++ b/examples/vlm_finetune/qwen3_5_moe/qwen3_5_35b.yaml
@@ -120,7 +120,17 @@ ci:
   # Keep failures blocking; source/reload defects are tracked by AM-711/AM-717.
   checkpoint_robustness:
     check_source_load_parity: true
+    experts_implementation: grouped_mm
+    hf_kl_threshold: 2e-2
+    hf_keep_in_fp32_modules: A_log,dt_bias
     hf_device_map_auto: true
+    model.backend.experts: torch_mm
+    resume_loss_threshold: 1e-2
+    source_load_cosine_threshold: 0.9985
+    source_load_kl_threshold: 1e-1
+    source_load_mean_kl_threshold: 1e-2
+    step_scheduler.global_batch_size: 16
+    step_scheduler.local_batch_size: 1
     tokenizer_name: Qwen/Qwen3.5-35B-A3B
 
 # Uncomment and configure for W&B logging
--- a/tests/ci_tests/scripts/finetune_launcher.sh
+++ b/tests/ci_tests/scripts/finetune_launcher.sh
@@ -100,7 +100,13 @@ if [[ "$HAS_ROBUSTNESS" == "true" ]]; then
       ;;
   esac
 
-  ROBUSTNESS_CMD="${CMD} --tee 3 --log-dir $TEST_DIR/robustness_logs \
+  ROBUSTNESS_LAUNCH_CMD="$CMD"
+  if [[ "$CMD" == torchrun* ]]; then
+    # A completed rendezvous cannot be reused reliably by the second torchrun.
+    ROBUSTNESS_LAUNCH_CMD="${CMD/--rdzv_id=${SLURM_JOB_ID}/--rdzv_id=${SLURM_JOB_ID}-robustness}"
+  fi
+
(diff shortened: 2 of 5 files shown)


yuhezhang-ai:
Thanks for fixing this, and sorry, this exposed a gap in the checkpoint-robustness harness I added. AutoModel intentionally keeps the `GDN A_log/dt_bias` parameters in FP32 under a bulk BF16 load, while the HF reference does not by default. Applying the same strict FP32 contract to HF is therefore the correct parity test.

Could the harness infer these names from AutoModel’s model-owned FP32 contract instead of requiring `hf_keep_in_fp32_modules` in each CI config? That would avoid duplicating model behavior in recipes and the current missing fixture-key registration.

Also, after aligning the FP32 contract and expert backend, are the relaxed KL/cosine and resume-loss thresholds still necessary? It would be preferable to rerun with the existing thresholds and retain only relaxations supported by the resulting metrics.

I’m happy to help with or take this harness change if you prefer.

HuiyingLi:
@yuhezhang-ai Addressed in `7c965bce8`:

- Removed the per-recipe `hf_keep_in_fp32_modules` knob. The HF reference now infers the FP32 names from AutoModel’s model-owned gated-delta-net checkpoint contract.
- Restored the default resume-loss threshold (`5e-3`). In the original phase order, the three resume diffs were `4.78e-4`, `2.29e-3`, and `1.49e-3`.
- Retained only the source/HF parity relaxations supported by reruns: source max/mean KL reached about `0.068/0.0089` with cosine down to `0.99884`, and HF max KL reached `0.065`. Phase 3 remained exact (`KL=0`).

The original-order 8×H100 robustness test passed, along with 140 targeted unit tests.

yuhezhang-ai:
@HuiyingLi Thanks, the harness changes look good, especially the memory cleanup and deriving the HF FP32 behavior from AutoModel’s model-owned contract. One minor follow-up would be to generalize this beyond GDN so other models’ strict-FP32 contracts are handled automatically.

For Qwen3.5 specifically, however, the new source-parity metrics appear effectively identical to the original [AM-711 ](https://linear.app/nvidia/issue/AM-711/qwen35-moe-vlm-source-load-parity-mismatch-in-checkpoint-robustness)result: max/mean KL around 0.0683/0.0089 and cosine around 0.998844. This suggests the FP32 and grouped-MM alignment did not materially change the existing mismatch. Could we investigate that residual further, before treating the 0.1 threshold as expected?

yuhezhang-ai:
@HuiyingLi Correction after reviewing the ERNIE follow-up more closely: the first tiny-model isolation had a precision flaw. It blanket-cast the HF model with `model.to(bfloat16)`, which downcast HF's non-persistent RoPE `inv_freq`; the real HF BF16 checkpoint loader keeps that buffer in FP32. AutoModel's new model-wide cast had independently downcast its `inv_freq`, so the probe made both sides share the same RoPE bug. That is why changing only the MoE routing-weight order could make the two tiny models bit-identical despite the production precision mismatch.

The routing-weight placement difference is still real:

- HF: `down_proj(swiglu_output) * routing_weight`
- AutoModel grouped-expert/DeepEP path: `down_proj(swiglu_output * routing_weight)`

It therefore remains a plausible contributor to Qwen3.5's BF16 residual, but the earlier ERNIE experiment does **not** isolate how much it contributed to ERNIE's production KL. We are preserving ERNIE's RoPE buffer in FP32, restoring the original ERNIE thresholds, and rerunning before attributing any remaining tolerance to MoE arithmetic order.

For Qwen3.5, the proposed focused experiment is still the right discriminator: preserve both models' complete FP32 contracts, move only the routing multiplication, and measure the first differing MoE layer plus final max/mean KL. Until then, I would not use the ERNIE numbers as quantitative support for Qwen3.5's `0.1` max-KL tolerance.

HuiyingLi:
@yuhezhang-ai I ran the focused 8xH100 experiment in the same CI image and configuration, preserving the corrected FP32 contracts on both sides and changing only the routing-weight placement.

The first difference is the first MoE block:

- layer 0 input: elementwise equal (`max_abs=0`, `mean_abs=0`)
- current AutoModel order (`down_proj(swiglu * weight)`), layer 0 output: `max_abs=4.882812e-4`, `mean_abs=1.481347e-5`, cosine `0.99999499`
- HF order (`down_proj(swiglu) * weight`), layer 0 output: `max_abs=4.882812e-4`, `mean_abs=1.251739e-5`, cosine `0.99999654`

Final source-parity results from the matched runs:

| routing placement | max KL | mean KL | cosine |
| --- | ---: | ---: | ---: |
| current AutoModel order | `6.830092e-2` | `8.562144e-3` | `0.99885988` |
| after down projection | `4.022782e-2` | `5.424098e-3` | `0.99882811` |

So the routing placement is a real contributor and materially reduces max/mean KL, but it does not eliminate the residual. One implementation constraint surfaced as well: DeepEP rejects FP32 expert payloads at combine (`Unsupported type`), so the after-down result performs the multiply in FP32 and casts the payload back to BF16 before DeepEP combine. The remaining difference therefore still includes DeepEP/BF16 communication and accumulation effects.

I also pushed `3ed3227b7` for the minor harness follow-up: `_hf_fp32_module_names` now combines the model class's generic `_keep_in_fp32_modules_strict` contract with the GDN bare-name checkpoint mapping, with unit coverage for generic, combined, duplicate, and empty contracts. The targeted harness suite passes (`30 passed`) and ruff is clean.

yuhezhang-ai (approved):
Thanks, the harness changes look good, especially the memory cleanup and deriving the HF FP32 behavior from the AutoModel model-owned strict-FP32 contract. The focused Qwen3.5 experiment establishes routing-weight placement as a material BF16 contributor, and an additional EP=8 control confirms a small DeepEP-versus-HF grouped-reduction difference. We have not exhaustively attributed the remaining KL, but it is stable and reasonably small, so I do not think further isolation should block this harness/OOM fix. I added commit 49720410b to document the known reasons next to the relaxed thresholds.

yuhezhang-ai (approved):
Re-approving signed-off replacement commit 15e60345f. Its tree is identical to the previously reviewed 49720410b, so the review reasoning above is unchanged.

yuhezhang-ai:
/ok to test [15e6034](https://github.com/NVIDIA-NeMo/Automodel/pull/3233/commits/15e60345f943272f666cb338d49c2ad210799776)

thomasdhc (approved):

PR #3239, opened 2026-07-26
docs(speculative): document Gemma4, MSD, CP, and during-training tools

khazic:
## What

`examples/speculative/README.md` has drifted from what `main` ships. Documentation only; no code changes.

The gaps were found by diffing the README against the registries and the component tree, so each item below is a concrete mismatch rather than a stylistic preference.

### Gemma4 is missing from the EAGLE-3 target list

This is the one that actively misleads. `EAGLE3_DRAFT_REGISTRY` maps `Gemma4ForConditionalGeneration`, and there are four shipped configs (`gemma4_e2b`, `e4b`, `31b`, `26b_a4b`), but the supported-target list stops at Llama, Phi-3, Qwen3, and Qwen3-MoE. A reader concludes EAGLE cannot target Gemma4.

Added, with the limitation stated rather than left to be discovered: the target is multimodal, but the draft is built from `config.get_text_config()` and consumes post-block hidden states, so it trains on the text backbone only and images do not participate in drafting.

### MSD is undocumented

`eagle/msd.py`, `msd_target.py`, `msd_curriculum.py`, and `msd_decode.py` are in `main` and the README does not mention MSD at all.

Putting it in the methods table would imply it is launchable, which it is not: nothing wires it to a recipe or config. Leaving it out entirely leaves four modules unexplained. It is described below the table with an explicit "not yet runnable" note and a pointer to what is missing.

### `decode_eval` and `regen` were invisible

Both are user-facing config blocks with real capability behind them, and neither appeared anywhere in the README, so the features could only be found by reading source.

Documented with their required keys, including why `cuda_visible_devices` has no default: both runners launch a detached engine and must be given GPUs the training run is not using.

### Context parallelism had one table cell

`cp_size` appeared once, in the `distributed` row of the config reference. The draft-side ring attention, its zig-zag layout, and `target_cp.py` were undocumented.

Added a section with the constraint matrix. This is worth stating because every combination listed **raises at setup**, and a reader who hits one otherwise has no way to know the failure was expected:

| Constraint | Applies to |
|---|---|
| No sequence packing | EAGLE-1/2, EAGLE-3, DFlash, DSpark |
| Colocated target backend only | EAGLE-3 |
| No tensor parallelism | DFlash |
| Dense Qwen3-style draft only | DSpark |
| Draft attention must be the ring path (the DeepSeek MLA draft raises) | EAGLE-3 |

### Implementation tree gaps

Added `draft_gemma`, `msd*`, `ring_attention` / `zigzag_ring_attention`, the engine-backed target modules, `regen_loop.py`, `decode_eval.py`, and `target_cp.py`. The `core(.py/_v12)`-style glob entries already cover the `_v12` and `peagle_*` files and were left alone.

## Verification

Every factual claim added here was checked against the source rather than written from memory:

- target lists diffed against `eagle/registry.py`
- config keys (`every_steps`, `cuda_visible_devices`, `target_model`, `input_data`, `num_speculative_tokens`) confirmed present in the resolvers, and `num_speculative_tokens` confirmed to belong to `decode_eval` only
- each CP constraint read off the gate that enforces it in the corresponding recipe
- the epoch-boundary swap and the skip-when-still-running behavior read off `RegenConfig` and `maybe_launch`
- the four Gemma4 config filenames confirmed to exist

One claim was written and then removed: an earlier draft said both cadence runners align to the restored `global_step` on resume. That is true of `RegenRunner` and not of `DecodeEvalRunner` on this base, so documenting it would have introduced exactly the kind of inaccuracy this PR removes.

## Scope

ViSpec is deliberately not documented here. It is still an open PR, and describing an unmerged feature in `main`'s README would leave readers looking for something that is not there. It can be added with its own change once merged.

### DIFF
--- a/examples/speculative/README.md
+++ b/examples/speculative/README.md
@@ -32,6 +32,16 @@ to P-EAGLE, all from the same draft class. Domino and JetSpec reuse the DFlash
 draft class and checkpoint format; only their training wrappers (and, for
 Domino, the extra head weights) differ.
 
+**Not yet runnable: MSD (multimodal speculative decoding).** `eagle/msd.py`,
+`msd_target.py`, `msd_curriculum.py`, and `msd_decode.py` implement the MSD
+training core, the frozen VLM target wrapper, the two-stage data curriculum, and
+the reference tree-decoding path. MSD extends EAGLE-1/2 feature drafting to
+vision-language targets: text positions keep the usual concatenation of target
+features and next-token embeddings, while image positions feed the target's
+already-projected image embeddings straight into the draft. It is deliberately
+absent from the table above because no recipe or config wires it up yet, so there
+is nothing to launch. The recipe and data pipeline are a follow-up change.
+
 ## Supported target models
 
 A target's `config.architectures` string selects the draft architecture through a
@@ -46,6 +56,13 @@ The shipped example configs only cover a subset.
 - **gpt-oss** (`GptOssForCausalLM`): EAGLE-3 only, via a dedicated draft class.
 - **DeepSeek-V3** (`DeepseekV3ForCausalLM`): EAGLE-3 only, via a dedicated MLA
   draft class (eager attention; sequence packing not supported yet).
+- **Gemma4** (`Gemma4ForConditionalGeneration`): EAGLE-3 only, via a dedicated
+  draft class that reconciles the Gemma4 text config (`hidden_activation`, nested
+  per-attention-type `rope_parameters`). The target is multimodal but the draft is
+  trained on its **text backbone only**: the decoder config is read through
+  `config.get_text_config()` and the draft consumes post-block hidden states, so
+  images do not participate in drafting. Configs: `eagle3/gemma4_e2b_eagle3.yaml`,
+  `gemma4_e4b_eagle3.yaml`, `gemma4_31b_eagle3.yaml`, `gemma4_26b_a4b_eagle3.yaml`.
 - **DFlash / Domino / JetSpec**: `Qwen3ForCausalLM`, `Qwen3MoeForCausalLM`.
 - **DSpark**: `Qwen3ForCausalLM`, `Qwen3MoeForCausalLM`,
   `DeepseekV4ForCausalLM`, GLM-5.2 (`GlmMoeDsaForCausalLM`), Gemma4
@@ -314,6 +331,66 @@ recipe_args:
 Datasets are consumed by `ChatDataset`: a `messages` list of `{role, content}`,
 or a `conversations` column (ShareGPT or OpenAI style) that is auto-converted.
 
+## Context parallelism for draft training
+
+`distributed.cp_size > 1` shards the sequence across ranks for long-context draft
+training. The draft's attention runs as a differentiable ring
+(`eagle/ring_attention.py`, with a load-balanced zig-zag layout in
+`zigzag_ring_attention.py`); `target_cp.py` prepares the target-side inputs.
+
+The constraints are enforced, not advisory: every combination below raises at
+setup rather than training something subtly wrong.
+
+| Constraint | Applies to |
+|---|---|
+| Cannot combine with sequence packing (`packed_sequence_size > 0`); CP shards the sequence and strips the block-causal mask packing relies on | EAGLE-1/2, EAGLE-3, DFlash, DSpark |
+| Colocated target backend only; the remote backend runs the target out-of-process | EAGLE-3 |
+| Cannot combine with tensor parallelism (`tp_size > 1`) | DFlash |
+| Dense Qwen3-style draft only | DSpark |
+| EAGLE-3 draft attention must be the `Eagle3LlamaAttention` ring path; architectures without it (the DeepSeek MLA draft) raise `NotImplementedError` | EAGLE-3 |
+
+## During-training tools (EAGLE-3)
+
+Both are opt-in, off unless their block sets `every_steps`, and both run their
+work in a **detached subprocess on GPUs the training run does not use**, so
+`cuda_visible_devices` is required rather than defaulted.
+
+### Periodic real acceptance length (`decode_eval`)
+
+Training loss and simulated accept length are proxies. The acceptance length that
+matters is the one an inference engine produces from the draft plus the target.
+`decode_eval` snapshots the draft every `every_steps` optimizer steps, serves the
+snapshot, benchmarks it, and reports the engine's real `accept_length` back into
+the training log and W&B.
+
+```yaml
+decode_eval:
+  every_steps: 500
+  cuda_visible_devices: "7"     # required: a GPU the training run is not using
+  target_model: meta-llama/Llama-3.1-8B-Instruct
+  input_data: /path/to/eval.jsonl
+  num_speculative_tokens: 4
+```
+
+A cycle whose predecessor is still running is skipped rather than queued, so a
+cadence faster than one eval takes will simply measure less often.
+
+### On-policy regeneration (`regen`)
+
+The draft is trained against target answers, and those answers age as the draft
+changes what it is asked to draft. `regen` periodically regenerates a slice of the
+corpus with the target and swaps the fresh shards into the dataloader. The swap
+happens at an epoch boundary (a mid-epoch dataloader rebuild would disturb the
+sampler), so a completed cycle waits for the next boundary.
+
+```yaml
+regen:
+  every_steps: 1000
+  cuda_visible_devices: "6"
+  target_model: meta-llama/Llama-3.1-8B-Instruct
+  input_data: /path/to/prompts.jsonl
+```
+
 ## Serve and benchmark a trained draft
 
 ### SGLang
@@ -505,12 +582,18 @@ Implementation:
 ```
 nemo_automodel/components/speculative/
   eagle/        core(.py/_v12), draft_llama(.py/_v12), draft_gpt_oss,
-                draft_deepseek, backend, registry, target(.py/_v12),
-                peagle_*, remote/
+                draft_deepseek, draft_gemma, backend, registry,
+                target(.py/_v12), peagle_*, remote/,
+                msd*                        # MSD core (no recipe yet)
+                ring_attention, zigzag_ring_attention   # draft-side CP
+                {sglang,vllm}_target, *_runner          # engine-backed targets
   dflash/       core, domino_core, jetspec_core, draft_qwen3, registry, target
   dspark/       core, draft_qwen3, draft_deepseek_v4, draft_glm_5_2,
                 draft_gemma4, draft_minimax_m3, markov_head, registry, target
   regenerate.py            # dataset regeneration with the target model
+  regen_loop.py            # on-policy regeneration during training (regen block)
+  decode_eval.py           # periodic real accept-length eval (decode_eval block)
+  target_cp.py             # target-side context-parallel input prep
   precompute_eagle3.py     # offline target-output cache
   serve_target.py          # remote target server (HTTP + NCCL)
   serve_sglang.py          # serve a trained draft via SGLang
--- a/examples/speculative/README.md
+++ b/examples/speculative/README.md
@@ -1,4 +1,4 @@
-# Speculative decoding: draft-model training in NeMo AutoModel
+# Train Draft Models for Speculative Decoding in NeMo AutoModel
 
 This directory trains the **draft models** used for speculative decoding. A draft
 proposes several future tokens cheaply, the frozen **target** model verifies them
@@ -6,19 +6,19 @@ in one forward pass, and every accepted token is a step the target never had to
 run autoregressively. The faster and more accurate the draft, the higher the
 acceptance length and the larger the inference speedup.
 
-AutoModel trains the draft; you then serve `draft + target` together in an
+AutoModel trains the draft. You then serve the draft and target together in an
 inference engine (SGLang or vLLM). The training code lives in
 `nemo_automodel/components/speculative/`, the recipes in
 `nemo_automodel/recipes/llm/`, and ready-to-run configs in this folder.
 
-## Supported methods
+## Supported Methods
 
-| Method | What the draft does | Recipe (`recipe:` key) | Configs |
+| Method | What the Draft Does | Recipe (`recipe:` Key) | Configs |
 |---|---|---|---|
-| **EAGLE-1** | Single decoder block that predicts the next target hidden state; supervised by SmoothL1 hidden-state loss plus a token loss through the frozen target `lm_head`. | `TrainEagle1Recipe` | `eagle1/` |
+| **EAGLE-1** | Single decoder block that predicts the next target hidden state. Supervised by SmoothL1 hidden-state loss and a token loss through the frozen target `lm_head`. | `TrainEagle1Recipe` | `eagle1/` |
 | **EAGLE-2** | Same training objective as EAGLE-1 (it differs only in the inference-time tree policy), so the recipe is a thin subclass. | `TrainEagle2Recipe` | `eagle2/` |
 | **EAGLE-3** | Draft with its own `lm_head` over a (optionally compressed) draft vocab, fed three auxiliary target hidden states, trained with a **test-time-training (TTT)** unroll. | `TrainEagle3Recipe` | `eagle3/` |
-| **EAGLE-3.1** | EAGLE-3 plus two drafter toggles (`fc_norm`, `norm_output`), matching the vLLM EAGLE-3.1 architecture. | `TrainEagle3Recipe` | `eagle3_1/` |
+| **EAGLE-3.1** | EAGLE-3 with two drafter toggles (`fc_norm`, `norm_output`), matching the vLLM EAGLE-3.1 architecture. | `TrainEagle3Recipe` | `eagle3_1/` |
 | **P-EAGLE** | Parallel-drafting EAGLE-3: predicts all `num_depths` tokens in one forward over a COD-subsampled sequence instead of the TTT unroll. Serves on vLLM only. | `TrainEagle3Recipe` (`parallel_drafting: true`) | `p-eagle/` |
 | **DFlash** | Block-parallel drafting: drafts a whole block of `block_size` tokens in one non-causal "denoising" forward over `[anchor, MASK, MASK, ...]`. | `TrainDFlashRecipe` (LLM) / `DiffusionLMSFTRecipe` (DLLM SFT) | `dflash/`, `../dllm_sft/*dflash*` |
 | **Domino** | DFlash backbone plus a serial GRU correction head (`prefix_gru` / `embed_proj`) that refines each block position on the previous ones. | `TrainDominoRecipe` | `dflash/qwen3_domino.yaml` |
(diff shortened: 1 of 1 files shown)


examples/speculative/README.md
jgerh:
```suggestion
# Train Draft Models for Speculative Decoding in NeMo AutoModel
```

jgerh:
```suggestion
AutoModel trains the draft. You then serve the draft and target together in an
```

jgerh:
```suggestion
## Supported Methods
```

jgerh:
```suggestion
| Method | What the Draft Does | Recipe (`recipe:` Key) | Configs |
```

jgerh:
```suggestion
| **EAGLE-1** | Single decoder block that predicts the next target hidden state. Supervised by SmoothL1 hidden-state loss and a token loss through the frozen target `lm_head`. | `TrainEagle1Recipe` | `eagle1/` |
```

jgerh:
```suggestion
("v1/v2") implementation. The unsuffixed files are the EAGLE-3/3.1 path. Inside
```

examples/speculative/README.md
```diff
@@ -32,6 +32,16 @@ to P-EAGLE, all from the same draft class. Domino and JetSpec reuse the DFlash
 draft class and checkpoint format; only their training wrappers (and, for
```
jgerh:
```suggestion
draft class and checkpoint format. Only their training wrappers (and, for
```

examples/speculative/README.md
jgerh:
```suggestion
| **EAGLE-3.1** | EAGLE-3 with two drafter toggles (`fc_norm`, `norm_output`), matching the vLLM EAGLE-3.1 architecture. | `TrainEagle3Recipe` | `eagle3_1/` |
```

examples/speculative/README.md
```diff
@@ -32,6 +32,16 @@ to P-EAGLE, all from the same draft class. Domino and JetSpec reuse the DFlash
...
+`msd_target.py`, `msd_curriculum.py`, and `msd_decode.py` implement the MSD
+training core, the frozen VLM target wrapper, the two-stage data curriculum, and
+the reference tree-decoding path. MSD extends EAGLE-1/2 feature drafting to
+vision-language targets: text positions keep the usual concatenation of target
+features and next-token embeddings, while image positions feed the target's
+already-projected image embeddings straight into the draft. It is deliberately
+absent from the table above because no recipe or config wires it up yet, so there
+is nothing to launch. The recipe and data pipeline are a follow-up change.
+
 ## Supported target models
```
jgerh:
```suggestion
## Supported Target Models
```

examples/speculative/README.md
```diff
@@ -46,6 +56,13 @@ The shipped example configs only cover a subset.
 - **gpt-oss** (`GptOssForCausalLM`): EAGLE-3 only, via a dedicated draft class.
 - **DeepSeek-V3** (`DeepseekV3ForCausalLM`): EAGLE-3 only, via a dedicated MLA
   draft class (eager attention; sequence packing not supported yet).
+- **Gemma4** (`Gemma4ForConditionalGeneration`): EAGLE-3 only, via a dedicated
```
jgerh:
```suggestion
- **gpt-oss** (`GptOssForCausalLM`): EAGLE-3 only, using a dedicated draft class.
- **DeepSeek-V3** (`DeepseekV3ForCausalLM`): EAGLE-3 only, using a dedicated MLA
  draft class (eager attention, sequence packing not supported yet).
- **Gemma4** (`Gemma4ForConditionalGeneration`): EAGLE-3 only, using a dedicated
```

examples/speculative/README.md
jgerh:
```suggestion
and MiniMax M3 VL (`MiniMaxM3SparseForConditionalGeneration`, including a
multimodal data path through `recipe_args.multimodal: true`).
```

jgerh:
```suggestion
frozen through the same expert-parallel / FSDP paths their fine-tuning recipes use.
Per-target notes are in `dspark/README_*.md`.
```

jgerh:
```suggestion
## Quickstart
```

jgerh:
```suggestion
The CLI is `automodel` (alias `am`). It drives `torchrun` internally.

Train an EAGLE-3 draft model:
```

jgerh:
```suggestion
Run DFlash **DLLM SFT** configs under `../dllm_sft/` with the standard AutoModel
SFT entry script:
```

jgerh:
```suggestion
## Operators, Kernels, and Attention Backends
```

jgerh:
```suggestion
Select them in the config. All degrade gracefully when a dependency is missing.
```

jgerh:
```suggestion
### Draft Attention Backend
```

jgerh:
```suggestion
| Method | Backends | How to Select |
```

jgerh:
```suggestion
extension columns for cached TTT steps are merged in log space through `logaddexp`.
```

jgerh:
```suggestion
### Fused Triton Soft Cross-Entropy
```

jgerh:
```suggestion
**### Draft-Vocab Compression (d2t / t2d)**
```

jgerh:
```suggestion
`components/datasets/llm/eagle3.py`. Set `recipe_args.draft_vocab_size` to enable,
```

jgerh:
```suggestion
### FP8 Draft Training
```

jgerh:
```suggestion
top-level `compile:` block (`CompileConfig`). The draft is compiled in place
(`nn.Module.compile()`, so checkpoint keys are unchanged) after the FP8 swap.
This matters for FP8 throughput: Float8Linear's per-GEMM cast/scale ops are
memory-bound and only pay off when inductor fuses them into the GEMM
prologue. In eager mode FP8 draft training is typically slower than BF16
(measured 0.76x on an H100 EAGLE-3 run). Compiled, the same A/B measured
FP8 at 1.03x over BF16 with byte-equivalent convergence. Expect the FP8
```

jgerh:
```suggestion
as a plain draft speedup (measured ~1.34x over the eager BF16 baseline in
```

jgerh:
```suggestion
### LoRA Draft Adaptation (EAGLE-3 Only)
```

jgerh:
```suggestion
base draft is frozen and only `lora_A`/`lora_B` adapters train. Checkpoints are
adapter-only (`adapter_model.safetensors` through the standard PEFT checkpoint
```

jgerh:
```suggestion
initialized draft are pointless. `draft_weights_path` also works without
```

jgerh:
```suggestion
base run's token mapping must be reused through `selected_token_ids_path` (the
frozen `lm_head` rows are tied to it). A differing mapping fails fast at
load. The final checkpoint of a LoRA run also exports the merged
```

jgerh:
```suggestion
draft's forward and backward GEMMs run in FP8. Requires SM89+ (H100 or newer);
`emulate: true` runs the FP8 numerics on older GPUs for testing. The frozen
target model is never converted (it already supports FP8-quantized checkpoints
through dequant-on-load in the DSpark V4 / GLM path). Linears with a weight dim not
divisible by 16 are skipped automatically. Use `filter_fqns` to exclude more
(for example, `["lm_head"]`). On DSpark's FSDP2 path, `enable_fsdp_float8_all_gather`
and `precompute_float8_dynamic_scale_for_fsdp` reduce the cost of computing
scales at each step. The SFT recipes use the same optimization. See
```

jgerh:
```suggestion
or in the DFlash-family / DSpark / EAGLE-1/2 recipes (rejected explicitly.
Their drafts carry trainable non-LoRA heads that the freeze would silently
```

jgerh:
```suggestion
## Target Backends
```

jgerh:
```suggestion
The frozen target produces the supervision signal (aux hidden states and the
```

jgerh:
```suggestion
| Backend | `recipe_args.target_model_backend` | When to Use |
```

jgerh:
```suggestion
| **Co-located (default)** | `colocated` | Target and draft share the same GPUs. Simplest, default for every config. |
| **Remote** | `remote` | Target served on separate GPUs/host. Training streams supervision over HTTP (control) and NCCL (data, with a binary wire fallback). Numerically identical to co-located. |
```

jgerh:
```suggestion
then consumed through `cached_target_path`. DSpark also supports a text-only offline cache through `precompute_dspark.py`
```

jgerh:
```suggestion
through the same expert-parallel and FSDP2 path as online training and writes the
```

jgerh:
```suggestion
training. Each rank forwards its own contiguous slice of the dataset and writes its
```

jgerh:
```suggestion
### Two-Step Flow
```

jgerh:
```suggestion
must already be served. The examples use SGLang. vLLM or any other
```

jgerh:
```suggestion
1. Start the target server (use `--tp 2` or higher to shard a multi-GPU
   target):
```

jgerh:
```suggestion
   ```bash
   python -m sglang.launch_server --model-path meta-llama/Llama-3.1-8B-Instruct --port 30000
   ```
```

jgerh:
```suggestion
2. Regenerate against the running server:
```

jgerh:
```suggestion
### Regeneration Tuning Knobs
```

jgerh:
```suggestion
| `--split` | `train` | Supports HF slice syntax, e.g., `train[:10000]`. |
```

jgerh:
```suggestion
### Regeneration Pitfalls
```

jgerh:
```suggestion
- **Wrong model name.** `--model` is the name sent in the OpenAI payload. It must
```

jgerh:
```suggestion
- **Server not warm.** Send one curl request to the server first. Otherwise the
```

examples/speculative/README.md
```diff
@@ -314,6 +331,66 @@ recipe_args:
 Datasets are consumed by `ChatDataset`: a `messages` list of `{role, content}`,
 or a `conversations` column (ShareGPT or OpenAI style) that is auto-converted.
 
+## Context parallelism for draft training
```
jgerh:
```suggestion
## Context Parallelism for Draft Training
```

examples/speculative/README.md
```diff
@@ -314,6 +331,66 @@ recipe_args:
...
+
+`distributed.cp_size > 1` shards the sequence across ranks for long-context draft
+training. The draft's attention runs as a differentiable ring
+(`eagle/ring_attention.py`, with a load-balanced zig-zag layout in
+`zigzag_ring_attention.py`); `target_cp.py` prepares the target-side inputs.
```
jgerh:
```suggestion
`zigzag_ring_attention.py`). `target_cp.py` prepares the target-side inputs.
```

examples/speculative/README.md
```diff
@@ -314,6 +331,66 @@ recipe_args:
...
+
+The constraints are enforced, not advisory: every combination below raises at
+setup rather than training something subtly wrong.
+
+| Constraint | Applies to |
```
jgerh:
```suggestion
| Constraint | Applies To |
```

examples/speculative/README.md
```diff
@@ -314,6 +331,66 @@ recipe_args:
...
+|---|---|
+| Cannot combine with sequence packing (`packed_sequence_size > 0`); CP shards the sequence and strips the block-causal mask packing relies on | EAGLE-1/2, EAGLE-3, DFlash, DSpark |
+| Colocated target backend only; the remote backend runs the target out-of-process | EAGLE-3 |
```
jgerh:
```suggestion
| Cannot combine with sequence packing (`packed_sequence_size > 0`). CP shards the sequence and strips the block-causal mask packing relies on | EAGLE-1/2, EAGLE-3, DFlash, DSpark |
| Co-located target backend only; the remote backend runs the target out-of-process | EAGLE-3 ||
```

examples/speculative/README.md
```diff
@@ -314,6 +331,66 @@ recipe_args:
...
+| Cannot combine with tensor parallelism (`tp_size > 1`) | DFlash |
+| Dense Qwen3-style draft only | DSpark |
+| EAGLE-3 draft attention must be the `Eagle3LlamaAttention` ring path; architectures without it (the DeepSeek MLA draft) raise `NotImplementedError` | EAGLE-3 |
+
+## During-training tools (EAGLE-3)
```
jgerh:
```suggestion
## During-Training Tools (EAGLE-3)
```

examples/speculative/README.md
```diff
@@ -314,6 +331,66 @@ recipe_args:
...
+
+Both are opt-in, off unless their block sets `every_steps`, and both run their
+work in a **detached subprocess on GPUs the training run does not use**, so
+`cuda_visible_devices` is required rather than defaulted.
+
+### Periodic real acceptance length (`decode_eval`)
```
jgerh:
```suggestion
### Periodic Real Acceptance Length (`decode_eval`)
```

examples/speculative/README.md
```diff
@@ -314,6 +331,66 @@ recipe_args:
...
+
+Training loss and simulated accept length are proxies. The acceptance length that
+matters is the one an inference engine produces from the draft plus the target.
```
jgerh:
```suggestion
matters is the one an inference engine produces from the draft and the target.
```

examples/speculative/README.md
```diff
@@ -314,6 +331,66 @@ recipe_args:
...
+  cuda_visible_devices: "7"     # required: a GPU the training run is not using
+  target_model: meta-llama/Llama-3.1-8B-Instruct
+  input_data: /path/to/eval.jsonl
+  num_speculative_tokens: 4
+```
+
+A cycle whose predecessor is still running is skipped rather than queued, so a
+cadence faster than one eval takes will simply measure less often.
+
+### On-policy regeneration (`regen`)
```
jgerh:
```suggestion
### On-Policy Regeneration (`regen`)
```

examples/speculative/README.md
```diff
@@ -314,6 +331,66 @@ recipe_args:
...
+
+```yaml
+regen:
+  every_steps: 1000
+  cuda_visible_devices: "6"
+  target_model: meta-llama/Llama-3.1-8B-Instruct
+  input_data: /path/to/prompts.jsonl
+```
+
 ## Serve and benchmark a trained draft
```
jgerh:
```suggestion
## Serve and Benchmark a Trained Draft
```

examples/speculative/README.md
jgerh:
```suggestion
After training, serve the target and draft through SGLang:
```

jgerh:
```suggestion
speculative token map from `eagle_meta.pt` when needed. SGLang is not bundled. The
```

jgerh:
```suggestion
output throughput, and a `speedup` ratio compared to an optional non-speculative
```

jgerh:
```suggestion
`serve_vllm.py` is the vLLM companion. It serves EAGLE-3, P-EAGLE (vLLM's
```

jgerh:
```suggestion
### Multi-Dataset Sweep
```

jgerh:
```suggestion
`bench_sglang.py` and `bench_vllm.py` measure one dataset per invocation, but the same draft's acceptance rate can vary sharply by task. A draft tends to accept more tokens on conversational data than on math or code data because their token distributions differ from the training mix.

`bench_sweep.py` drives either engine through several datasets in one pass. It reports a per-dataset table and a completed-weighted aggregate, eliminating the need to run each single-dataset script and collate the results manually:
```

jgerh:
```suggestion
acceptance/speedup on (MT-Bench (first turn), HumanEval (code), GSM8K
(math), and Alpaca (single-turn instruction-following)). None of these ship a
chat-messages column, so each is read through `--prompt-column` (a raw text field
```

jgerh:
```suggestion
## Inference-Engine Compatibility
```

jgerh:
```suggestion
`serve_sglang.py` rejects P-EAGLE drafts with an actionable error. Serve those on
```

jgerh:
```suggestion
## Config Reference (EAGLE-Style Schema)
```

jgerh:
```suggestion
### Top-Level Sections
```

jgerh:
```suggestion
### DFlash-Family Validation Metrics
```

jgerh: suggest a cross link, verify the change

```suggestion
| `fp8` | Optional; torchao FP8 draft training, same surface as the SFT recipes. See [FP8 Draft Training](#fp8-draft-training). |
```

jgerh: suggest a cross link, verify the change

```suggestion
| `peft` | Optional, EAGLE-3 only; LoRA draft adaptation (`PeftConfig`). See [LoRA Draft Adaptation (EAGLE-3 Only)](#lora-draft-adaptation-eagle-3-only). |
```

jgerh:
```suggestion
### `recipe_args` Common to All Methods
```

jgerh:
```suggestion
### Method-Specific `recipe_args`
```

jgerh:
```suggestion
| `markov_rank`, `markov_head_type`, `confidence_head_alpha`, `confidence_head_with_markov`, `ce_loss_alpha` | DSpark | Markov / confidence-head knobs on top of the block-drafting set (`block_size`, `num_anchors`, `mask_token_id`, `target_layer_ids`, and so on). |
```

jgerh:
```suggestion
## Directory Layout
```

jgerh:
```suggestion
The following tree shows the configs in this folder:
```

jgerh:
```suggestion
The following tree shows the implementation layout:
```

examples/speculative/README.md
```diff
@@ -505,12 +582,18 @@ Implementation:
...
   dflash/       core, domino_core, jetspec_core, draft_qwen3, registry, target
   dspark/       core, draft_qwen3, draft_deepseek_v4, draft_glm_5_2,
                 draft_gemma4, draft_minimax_m3, markov_head, registry, target
   regenerate.py            # dataset regeneration with the target model
+  regen_loop.py            # on-policy regeneration during training (regen block)
+  decode_eval.py           # periodic real accept-length eval (decode_eval block)
+  target_cp.py             # target-side context-parallel input prep
   precompute_eagle3.py     # offline target-output cache
   serve_target.py          # remote target server (HTTP + NCCL)
   serve_sglang.py          # serve a trained draft via SGLang
```
jgerh:
```suggestion
  serve_sglang.py          # serve a trained draft through SGLang
```

examples/speculative/README.md
jgerh:
```suggestion
  serve_vllm.py            # serve a trained draft through vLLM (EAGLE-3 / P-EAGLE / DFlash)
```

jgerh: suggest a cross link, verify the change

```suggestion
| `target_model_backend`, `remote_urls`, `target_prefetch_depth`, `remote_timeout`, `remote_max_retries` | EAGLE-3 remote | See [Target Backends](#target-backends). |
```

jgerh:
```suggestion
   ```bash
   python -m nemo_automodel.components.speculative.regenerate --input-data Aeala/ShareGPT_Vicuna_unfiltered --output-dir ./regenerated/sharegpt_llama31_8b --target-server http://localhost:30000/v1 --model meta-llama/Llama-3.1-8B-Instruct --concurrency 64 --shard-size 1000
   ```
```

jgerh:
```suggestion
`1 + K` query tokens per sequence, so a large block size might need a smaller
`--max-num-seqs` (forward it through the trailing extra args, for example,
```

jgerh (commented):
Completed tech pubs review of examples/speculative/README.md and provided copyedits, style updates, and suggested text revisions.

jgerh (commented):
Completed tech pubs review of examples/speculative/README.md and provided copyedits, style updates, and suggested text revisions.

HuiyingLi (approved):
Thank you!

PR #3240, opened 2026-07-26
fix: harden checkpoint torch loads

HuiyingLi:
## Summary

- Add `safe_torch_load`, which defaults checkpoint sidecar and `.bin` restores to `torch.load(weights_only=True)`.
- Gate legacy pickle fallback behind `checkpoint.allow_legacy_pickle_restore`, defaulting to disabled.
- Migrate `StatefulRNG` checkpoints to a weights-only-loadable plain dictionary while retaining legacy `RNGState` restore support after explicit unsafe loading.
- Harden AutoModel sidecar, metadata, and Hugging Face `.bin` loads in BaseRecipe, Checkpointer, EAGLE, DSpark, and DFlash recipes.

## Root cause

AMINT-195 / ZDI-CAN-31759 is the PyTorch pickle RCE class: restoring AutoModel `.pt` sidecars with `torch.load(..., weights_only=False)` can execute code if the checkpoint is untrusted. The RNG sidecar was one reason AutoModel had to use unsafe pickle because it saved a custom `RNGState` dataclass containing Python and NumPy RNG state.

## Security notes

- Default restore now refuses legacy pickle-only artifacts instead of silently loading them unsafely.
- Trusted legacy checkpoints can still be restored by explicitly setting `checkpoint.allow_legacy_pickle_restore=True`.
- PyTorch DCP metadata and `BYTE_IO` pickle handling remain a separate trusted-checkpoint surface and are not changed in this scoped PR.

## Validation

Ran on existing interactive H100 allocation `14397630` (`pool0-01978`) using container `/lustre/fsw/portfolios/coreai/users/huiyingl/container_images/auto2606rc9.sqsh`:

- `pytest -q tests/unit_tests/training/test_rng.py tests/unit_tests/checkpoint/test_safe_torch_load.py` -> 11 passed
- Focused EAGLE, DFlash, and DSpark recipe suites -> 154 passed
- `ruff check` on touched files -> passed
- `git diff --check` -> passed


nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -90,6 +91,52 @@
...
+        if not allow_legacy_pickle_restore:
+            raise RuntimeError(_format_legacy_pickle_restore_error(path, description)) from err
+
+    logger.warning(
+        "Loading %s from %r with torch.load(weights_only=False). This can execute code; "
+        "only proceed for trusted legacy checkpoints.",
+        description,
+        path,
+    )
+    unsafe_kwargs = {"weights_only": False}
```
akoumpa: @HuiyingLi does this maintain the issue? 🤔

akoumpa (dismissed):

akoumpa (approved):

PR #3140, opened 2026-07-20
fix(moe): preserve top-k routing under full AC

yuhezhang-ai:
# What does this PR do ?

Preserve MoE top-k routing decisions under full activation checkpointing.

The existing `ignore_router_for_ac=True` policy saved the router projection but recomputed `aten.topk`. Exact tied scores could therefore select different experts during backward recomputation, changing DeepEP routed-token buffer shapes and raising `CheckpointError`.

# Changelog

- Save `aten.topk` outputs alongside the router projection in the full-AC MoE policy.
- Clarify the `ignore_router_for_ac` routing-stability contract.
- Extend focused unit coverage for router projection and top-k policy decisions.

# Before your PR is "Ready for review"

**Pre checks**:

- [x] Read and followed the contributor guidelines.
- [x] Added focused unit coverage.
- [x] Updated the affected configuration and API documentation.

# Validation

- `pytest tests/unit_tests/moe/test_parallelizer.py -q` — 74 passed
- `ruff format --check .`
- `ruff check .`

# Additional Information

The motivating failure was observed with tie-heavy `GroupedExpertsDeepEP` routing under EP: recomputed top-k assignments changed the routed-token count by one.

PR #3247, opened 2026-07-27
fix(moe): preserve gate load across activation recompute

akoumpa:
## Summary

- preserve each MoE gate's cumulative expert-load state around activation-checkpoint recomputation
- compose the state guard with full, router-selective, and operator-selective checkpoint contexts
- add a 1-GPU functional regression test covering activation checkpointing off/on output, input-gradient, parameter-gradient, and load-accounting parity

## Root cause

Activation checkpointing replays the MoE block during backward. The original forward changes `_cumulative_expert_load` from `None` to a tensor, so the replay enters the in-place `+=` path that was absent from the original forward. Selective activation checkpointing then fails with `aten.add_.Tensor encountered during backward but not found in storage`; without that failure, replay also double-counts expert load.

The checkpoint context now snapshots the pre-forward load, restores it for recomputation so the replay follows the original operation sequence, and restores the committed post-forward load afterward.

Observed in [nemo-ci job 374246076](https://nv/nemoci/dl/JoC/nemo-ci/-/jobs/374246076).

## Validation

- `CUDA_VISIBLE_DEVICES=0 pytest tests/functional_tests/moe/test_gate_activation_checkpointing.py -vv` — 1 passed
- `pytest tests/unit_tests/moe/test_layers.py -q` — 89 passed, 1 skipped
- `uv run --no-sync ruff check --fix nemo_automodel/components/moe/parallelizer.py tests/functional_tests/moe/test_gate_activation_checkpointing.py`
- `uv run --no-sync ruff format nemo_automodel/components/moe/parallelizer.py tests/functional_tests/moe/test_gate_activation_checkpointing.py`
- `git diff --check`

The standalone `test_parallelizer.py` stub suite is not independently runnable from a clean checkout because its synthetic `torch.distributed.fsdp` module omits symbols imported by the current distributed config; this PR does not modify that harness.


yuhezhang-ai:
Minor / non-blocking (P2): the regression currently calls apply_ac(...) with its defaults, so it covers only the ignore_router=True router-selective path. Since this change also wires the state guard into selective=True and ignore_router=False, could we parameterize the test across all three modes? The core load-accounting mismatch is observable on CPU, so a small CPU unit test could cover the matrix while retaining the GPU case for the CUDA-specific aten.add_ failure. This might also address the codecov.

Also, @akoumpa, could you help review [#3140](https://github.com/NVIDIA-NeMo/Automodel/pull/3140) when you have a chance? It is complementary to this fix: #3140 preserves tied top-k routing decisions, while this PR preserves cumulative gate-load state during recomputation.

akoumpa:
CI: https://nv/nemoci/-/pipelines/59941205

PR #3250, opened 2026-07-27
fix(checkpoint): save all PEFT EP optimizer parts

yuhezhang-ai:
# What does this PR do ?

Fix PEFT+expert-parallel optimizer checkpointing when an interleaved pipeline
rank owns multiple local model and optimizer parts.

The failure in AMINT-202 happened because the native fallback inspected only
the first local model part and then saved only the first optimizer. When that
first part was dense, the code incorrectly entered PyTorch's model-aware DCP
optimizer path, which could not construct a consistent parameter-ID-to-FQN
mapping for LoRA parameters over EP-sharded experts and raised `KeyError: 0`.

# Changelog

- Select the native PEFT optimizer fallback from the actual EP topology and all
  local model parts.
- Save and restore every local optimizer part while preserving the existing
  single-optimizer checkpoint schema.
- Namespace multi-part optimizer state by global pipeline stage so different PP
  ranks cannot publish unrelated Adam tensors under the same DCP key.
- Materialize lazy Adam/AdamW state placeholders so DCP can restore native
  optimizer state into a fresh optimizer.
- Reject multi-part checkpoint/layout count mismatches with a clear error.
- Add CPU regression coverage for topology selection, compatibility, DCP
  round-trip restoration, and mismatch handling.

**Pre checks**:

- [x] Added focused unit tests.
- [x] No documentation update is required for this internal checkpoint fix.
- [x] `ruff format` and `ruff check` pass for the changed files.
- [x] `185 passed, 2 skipped` in `test_optimizer_state.py` and
  `test_checkpointing.py`.

# Additional Information

- Related Linear issue:
  https://linear.app/nvidia/issue/AMINT-202/optimizer-checkpoint-save-fails-with-keyerror-0-when-accessing-fqn-pid
- Scoped 64-GPU Ling NeMo CI passed:
  [ling_1t_lora_pp job 375536631](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/375536631).
  It completed all 50 steps and the final optimizer checkpoint save at step 49.
- CI tested commit `f20a46b` before the DCO-only amendment. Current PR head
  `a9178c1` has the identical source tree (`48fb0ccb`).

### DIFF
--- a/nemo_automodel/components/checkpoint/checkpointing.py
+++ b/nemo_automodel/components/checkpoint/checkpointing.py
@@ -612,6 +612,8 @@ class Checkpointer:
         model: nn.Module | list[nn.Module],
         weights_path: str,
         scheduler: Optional[Any] = None,
+        *,
+        optimizer_part_ids: list[int] | None = None,
     ) -> None:
         """
         Save optimizer (and optional scheduler) state to `weights_path/optim` using DCP.
@@ -621,6 +623,8 @@ class Checkpointer:
             model: Model or pipeline model parts providing partitioning context.
             weights_path: Base directory for checkpoints.
             scheduler: Optional LR scheduler to include.
+            optimizer_part_ids: Global pipeline-stage indices corresponding to
+                per-model-part optimizers.
         """
         optimizer_path = os.path.join(weights_path, "optim")
         _ensure_dirs(optimizer_path, process_group=self.process_group)
@@ -631,6 +635,7 @@ class Checkpointer:
             is_peft=self.config.is_peft,
             cpu_offload=self.config.cpu_offload,
             has_expert_parallelism=self.moe_mesh is not None,
+            optimizer_part_ids=optimizer_part_ids,
         )
         state_dict = optimizer_state.state_dict()
         self._optim_ctx.future = self._do_save(state_dict, optimizer_path)
@@ -641,6 +646,8 @@ class Checkpointer:
         model: nn.Module | list[nn.Module],
         weights_path: str,
         scheduler: Optional[Any] = None,
+        *,
+        optimizer_part_ids: list[int] | None = None,
     ) -> None:
         """
         Load optimizer (and optional scheduler) state from `weights_path/optim` using DCP.
@@ -650,6 +657,8 @@ class Checkpointer:
             model: Model or pipeline model parts providing partitioning context.
             weights_path: Base directory for checkpoints.
             scheduler: Optional LR scheduler to populate.
+            optimizer_part_ids: Global pipeline-stage indices corresponding to
+                per-model-part optimizers.
         """
         optimizer_state = OptimizerState(
             model,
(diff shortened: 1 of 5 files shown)


yuhezhang-ai:
/ok to test [87289b2](https://github.com/NVIDIA-NeMo/Automodel/pull/3250/commits/87289b2cffa2d8abfbde48ed0c291dcc75fb321b)

yuhezhang-ai:
/ok to test [a9178c1](https://github.com/NVIDIA-NeMo/Automodel/pull/3250/commits/a9178c15181ac50055eab51e453d0c1257578e17)

athitten:
@yuhezhang-ai the code LGTM, thanks! qq: do we know why this doesn't show up in other models ?

yuhezhang-ai:
> @yuhezhang-ai the code LGTM, thanks! qq: do we know why this doesn't show up in other models ?

@athitten This is a relatively narrow checkpointing issue rather than something specific to Ling. It requires PEFT + EP + PP, particularly interleaved PP where each rank owns multiple model and optimizer parts, with optimizer checkpointing enabled.
Ling has exactly this setup: LoRA, PP=8, EP=8, five pipeline parts per rank, and checkpointing enabled. Its first local part can be dense, so the old code failed to detect the global EP topology and selected the wrong optimizer-state path. After fixing that, we also found that local optimizer-part indices collided across PP ranks, which is why the final fix uses global pipeline-stage IDs.
I checked the other PP+EP+PEFT recipes. They have similar potentially affected topologies, but all set checkpoint.enabled: false, so their CI runs never exercise optimizer saving. Ling is the only confirmed reproducer; the others are at risk but unverified. A targeted PP+EP+PEFT save/load/resume test might be worth adding to close this coverage gap.

athitten:
> > @yuhezhang-ai the code LGTM, thanks! qq: do we know why this doesn't show up in other models ?
> 
> @athitten This is a relatively narrow checkpointing issue rather than something specific to Ling. It requires PEFT + EP + PP, particularly interleaved PP where each rank owns multiple model and optimizer parts, with optimizer checkpointing enabled. Ling has exactly this setup: LoRA, PP=8, EP=8, five pipeline parts per rank, and checkpointing enabled. Its first local part can be dense, so the old code failed to detect the global EP topology and selected the wrong optimizer-state path. After fixing that, we also found that local optimizer-part indices collided across PP ranks, which is why the final fix uses global pipeline-stage IDs. I checked the other PP+EP+PEFT recipes. They have similar potentially affected topologies, but all set checkpoint.enabled: false, so their CI runs never exercise optimizer saving. Ling is the only confirmed reproducer; the others are at risk but unverified. A targeted PP+EP+PEFT save/load/resume test might be worth adding to close this coverage gap.

Thanks for the info @yuhezhang-ai ! Yes might be good to create an issue for the others to close the coverage gap.

athitten (approved):
Thank you, LGTM @yuhezhang-ai !

PR #2598, opened 2026-06-16
fix(models): audit fp32 protected tensors

yuhezhang-ai:
## Summary

- Audit model-specific fp32-protected tensors from #2570 and keep the remaining model-specific buffers/params protected through dtype casts.
- Preserve MoE gate/rotary fp32 buffers for MiniMax, HY, Qwen3-Omni, and Qwen3-VL paths.
- Move Nemotron V3 Mamba `A_log`/`dt_bias`/`D` into fp32 holder modules, with HF-compatible state-dict adapter routing for the public checkpoint keys.
- Keep callable fp32 holders usable under real FSDP2 by materializing full tensors from FSDP DTensor params and keeping strict holder subtrees unresharded during the parent forward.

## Why

`model.torch_dtype=bf16` and broad dtype casts can round small tensors that are part of a model's numerical/checkpoint contract. Trainable fp32 params that live directly on mixed modules also cannot be isolated cleanly by FSDP, so this follows the `_fp32_params` holder pattern used by the stacked Qwen GDN branch.

The real 2-GPU FSDP smoke exposed one more holder-specific issue: a callable holder that returns its parameter can hand the parent module a sharded DTensor. The fix keeps storage sharded/fp32, but returns the full fp32 tensor value to the caller.

## Notes

This PR is stacked on `yuhez/fix/qwen-gdn-fp32-precision`. After that branch lands, this should be rebased onto `main`.

## Validation

- `uv run --no-sync pytest tests/unit_tests/distributed/test_parallelizer_utils.py tests/unit_tests/distributed/test_fp32_compute_contract.py -q` (`25 passed`)
- focused GPT-OSS/Nemotron/DeepSeek holder tests (`30 passed`)
- `uv run --no-sync pytest tests/unit_tests/models/nemotron_v3/test_nemotron_v3_state_dict_adapter.py tests/unit_tests/distributed/test_fp32_compute_contract.py -q` (`30 passed`)
- focused Nemotron V3 Mamba/CP dtype tests (`9 passed`)
- focused GPT-OSS holder/state-dict tests (`5 passed, 1 skipped`)
- focused fp32 protected dtype regression suite (`11 passed`)
- 2-GPU H100 Slurm smoke: `fp32-holder-fsdp-smoke-c938fb8-r3`, job `12863187`, exit `0:0`
- `ruff check` on touched files
- `git diff --check`

PR #3255, opened 2026-07-27
fix(distributed): preserve ERNIE router and dense Qwen3.5 SSM precision

yuhezhang-ai:
# What does this PR do?

Preserve model-owned FP32 storage and compute contracts across BF16 construction, checkpoint loading, and FSDP2, including ERNIE 4.5's MoE router and dense Qwen3.5's SSM gate.

HF keeps ERNIE's router `gate.weight`, correction bias, and non-persistent RoPE `inv_freq` in FP32 during a BF16 checkpoint load. AutoModel previously stored `gate.weight` in BF16. Casting that value back to FP32 during routing cannot recover precision lost during loading.

The MoE sharder previously handled FP32 parameters only when models placed them in dedicated `_fp32_params` holder modules. That covers contracts such as Qwen3.5's GDN parameters, but not ordinary strict-FP32 parameters such as ERNIE's router weight. It now reuses the shared dtype-aware FSDP grouping, which groups model-owned parameters by storage and compute dtype, isolates compatible FP32 units, and preserves FP32 parameters, reductions, and outputs.

# Changelog

- Declare `mlp.gate.weight` and `mlp.gate.e_score_correction_bias` as ERNIE's strict FP32 parameter contract.
- Preserve ERNIE's `rotary_emb` buffers in FP32 during the model-wide BF16 cast.
- Default ERNIE router computation to FP32 without mutating the caller-owned `BackendConfig`.
- Reuse the shared `fully_shard_by_dtype()` path for MoE transformer blocks while excluding EP-owned experts.
- Apply FP32 parameter, reduction, and output policies to model-owned FP32 compute units, with module-owned input casting so activation-checkpoint forward and recompute observe the same parent activation dtype.
- Retain model-specific sharders such as DeepSeek V4's and fail early when one parameter owner requests compute dtypes that cannot be represented by a single FSDP policy.
- Restore ERNIE's HF-reload KL threshold from the relaxed `1.1e-2` to the harness default `5e-3` after validating the complete precision contract.
- Add model, policy, and two-rank EP/FSDP2 regressions, including explicit FP32 router/top-k numerical parity before and after full-model sharding and FP32 storage/gradient checks.

**Pre checks**:

- [x] Added focused unit and distributed regression coverage.
- [x] No user-facing documentation changes are required.

# Additional Information

- Linear: [AMINT-207](https://linear.app/nvidia/issue/AMINT-207/kl-divergence-exceeds-threshold-when-reloading-ernie4-model-from-hf)

## Validation

- Rebased with `main` in `039385b06`, including DeepSeek V4's model-owned FSDP handling from #3227.
- `ruff check` and `git diff --check` pass.
- Post-rebase focused suite: 56 passed, 3 skipped.
- Two-rank H100 EP/FSDP2/activation-checkpointing regression passes in Slurm job `14486166`.
- The exact production `ernie4_5_21b_a3b_hellaswag` checkpoint-robustness job passed: [leaf job 376922646](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/376922646), [pipeline 60000628](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/60000628). The test reported `1 passed` in 460.99 seconds.
- The scoped Qwen3.5 release validation passed on PR head `372099ad3`: [dense `qwen3_5_4b` leaf job 378345604](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/378345604), [MoE `qwen3_5_35b` leaf job 378345605](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/378345605), [VLM leaf pipeline 60187305](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/60187305), and [parent pipeline 60187081](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/60187081).

## Checkpoint-robustness metrics

The successful production job used the harness-default HF KL threshold of `5e-3` and reported:

- source-load maximum per-token KL: `1.187141e-3` (threshold `5e-3`)
- source-load mean per-token KL: `2.602303e-4` (threshold `1e-3`)
- source-load p95 KL: `8.557549e-4`
- source-load cosine similarity: `0.99977142` (threshold `0.9995`)
- AutoModel consolidated-reload maximum KL: `0.0` (threshold `0.0`)
- HF-reload maximum per-token KL: `1.280195e-3` (threshold `5e-3`)
- resume-loss differences: `0.0` for all three compared steps

The CI ran on validation SHA `4b21fcdfb8e8305a5a85a8751b73c177e4f45b27`, based on PR head `712a937dd`, with only two validation changes: restoring `hf_kl_threshold: 5e-3` and adding pytest `-rP` so passing-test metrics were visible. The validated threshold change is included in this PR as `497d864ef`.

These corrected metrics show that the earlier ERNIE-specific `1.1e-2` relaxation is no longer justified.

## Why the earlier tiny probe matched exactly

The diagnostic probe called `HF_model.to(bfloat16)` and restored only HF's persistent strict-FP32 router tensors. That blanket cast also rounded HF's non-persistent `inv_freq`, while AutoModel's model-wide cast rounded its copy. The state-dict comparison could not see `inv_freq` because it is non-persistent. After switching the MoE multiplication order, the probe compared two models that shared the same RoPE bug, so exact equality was conditional and did not prove the production residual came only from MoE ordering.

The probe now preserves and asserts the HF-style FP32 RoPE contract. The checkpoint-robustness harness already constructs HF with its dtype-aware `from_pretrained(..., torch_dtype=...)` path rather than blanket-casting an instantiated model.

## Why FP32 holders may reshard normally

FP32 holders were initially forced to `reshard_after_forward=False` because returning raw parameter storage could let FSDP reshard that storage before the parent finished consuming it. PR #2598 subsequently changed Nemotron and shared GDN holder outputs to cloned tensors and intentionally restored normal caller-controlled resharding. `NemotronV3MambaFP32Params.forward()` calls `_full_tensor_if_dtensor()` for every returned parameter, and that helper always clones the value while preserving the gradient path. The returned tensors therefore do not alias storage that FSDP may reshard.

### DIFF
--- a/examples/llm_finetune/ernie4_5/ernie4_5_21b_a3b_hellaswag.yaml
+++ b/examples/llm_finetune/ernie4_5/ernie4_5_21b_a3b_hellaswag.yaml
@@ -104,7 +104,7 @@ ci:
   time: "00:20:00"
   checkpoint_robustness:
     check_source_load_parity: true
-    hf_kl_threshold: 1.1e-2
+    hf_kl_threshold: 5e-3
     source_load_kl_threshold: 5e-3
     source_load_cosine_threshold: 0.9995
     tokenizer_name: baidu/ERNIE-4.5-21B-A3B-PT
--- a/nemo_automodel/components/distributed/parallelizer_utils.py
+++ b/nemo_automodel/components/distributed/parallelizer_utils.py
@@ -99,11 +99,15 @@ def iter_maximal_uniform_dtype_subtrees(
 def _group_params_by_dtype(
     layer: nn.Module,
     dtype_of: Optional[Callable[[torch.Tensor], torch.dtype]] = None,
+    ignored_params: Optional[Set[nn.Parameter]] = None,
 ) -> Dict[torch.dtype, List[nn.Parameter]]:
     if dtype_of is None:
         dtype_of = lambda t: t.dtype
+    ignored_param_ids = {id(param) for param in ignored_params or ()}
     ans: Dict[torch.dtype, List[nn.Parameter]] = {}
     for name, param in layer.named_parameters():
+        if id(param) in ignored_param_ids:
+            continue
         dtype = dtype_of(param)
         if dtype not in ans:
             ans[dtype] = []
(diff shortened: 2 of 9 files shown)


yuhezhang-ai:
/ok to test [039385b](https://github.com/NVIDIA-NeMo/Automodel/pull/3255/commits/039385b065d6076f92d7944c9ebc1f02269c57f3)

yuhezhang-ai:
/ok to test [bce9f1e](https://github.com/NVIDIA-NeMo/Automodel/pull/3255/commits/bce9f1eb436f6639a14d01a5b3926ff5ab5ca0b9)

akoumpa (approved):

PR #3267, opened 2026-07-28
fix(speculative): align the GLM-5.2 DSpark draft RoPE with its target

khazic:
## What

Make the GLM-5.2 DSpark draft derive its RoPE and attention scale from the same
config fields `GlmMoeDsaModel` / `GlmMoeDsaAttention` use.

## Why

The draft is trained on the target's hidden states, so a mismatch in positional
encoding or attention temperature never raises and never shows up in the loss
curve: the draft converges against a slightly different model than the one it
will speculate for, and only acceptance suffers. Three such mismatches existed:

| | target (`glm_moe_dsa`) | draft (before) |
|---|---|---|
| `rope_theta` | `rope_parameters["rope_theta"]`, else the top-level field (`model.py:184`) | same, but **falls back to 10000.0** when absent |
| rope table | `precompute_freqs_cis(..., rope_scaling=getattr(config, "rope_scaling", None))` (`model.py:189`) | `rope_scaling=None` hardcoded, so a YaRN target gets an unscaled draft |
| softmax scale | `qk_head_dim ** -0.5`, times `mscale ** 2` under YaRN (`layers.py:457`) | `qk_head_dim ** -0.5` only |

A draft silently trained at theta=10000 against a theta=1e6 target is the worst
of these, so the fallback is now a hard error instead of a default.

`qk_head_dim` also follows the target's precedence (an explicit config field
wins over the nope+rope sum), so both sides size `q_b_proj` and the scale the
same way.

## Notes

The mscale derivation moves into `mla_softmax_scale` in a new
`glm_moe_dsa/rope_utils.py`, shared by the GLM target attention and the draft:
the pair whose drift this PR is about cannot separate again. The byte-identical
copies in the DeepSeek V3 and V3.2 attention layers are deliberately left alone
(their unit tests patch the imported `yarn_get_mscale` symbol); hoisting the
helper for them is a mechanical follow-up.

The remaining asymmetry is intentional and mirrors the target: the scale reads
`rope_parameters` (falling back to `rope_scaling`), while the frequency table
reads `rope_scaling`.

## Testing

CPU unit tests only.

- new tests for `mla_softmax_scale` (plain, YaRN, inside the original window,
  older `rope_scaling`-only configs) and for the draft's scale, frequency table,
  `rope_theta` source, missing-`rope_theta` failure, and `qk_head_dim`
- `pytest tests/unit_tests/speculative tests/unit_tests/models/glm_moe_dsa
  tests/unit_tests/models/deepseek_v3 tests/unit_tests/models/deepseek_v32` passes

No GPU run is needed: the released GLM-5.2 config ships `rope_type: default`, so
this is a no-op for that checkpoint and only changes behavior on a YaRN-scaled
or theta-relocated config, where the previous behavior was wrong.


nemo_automodel/components/models/deepseek_v3/rope_utils.py
```diff
@@ -24,6 +24,35 @@ def yarn_get_mscale(scale=1, mscale=1):
     return 0.1 * mscale * math.log(scale) + 1.0
 
 
+def mla_softmax_scale(config, qk_head_dim: int) -> float:
```
HuiyingLi: This can be moved to glm_moe_dsa instead

khazic: Moved: `mla_softmax_scale` now lives in `nemo_automodel/components/models/glm_moe_dsa/rope_utils.py` (its only two callers are the GLM attention and the GLM draft), and `deepseek_v3/rope_utils.py` plus its tests are back to untouched. Tests moved with it to `tests/unit_tests/models/glm_moe_dsa/test_glm_moe_dsa_rope_utils.py`.

HuiyingLi (approved):

khazic:
The `L0_Unit_Tests_GPU_1_of_2` failure on this PR is unrelated to the change.

The only failing test is `tests/unit_tests/speculative/test_eagle3_cp_shift.py::test_cp_shift_left_zigzag_single_rank_matches_local_shift` (1 failed, 5908 passed), and it fails at process-group setup, before any assertion:

```
dist.init_process_group("gloo", rank=rank, world_size=world_size)
torch.distributed.DistNetworkError: The server socket has failed to listen on any local network address.
port: 39275, useIpv6: false, code: -98, name: EADDRINUSE, message: address already in use
```

That test's `_free_port()` helper binds port 0, reads the assigned port, closes the socket, and only then passes the port to the spawned worker, which re-binds it via `TCPStore`. Between the close and the re-bind the port can be taken by anything else on the runner, which is what happened here. The test came in with the draft-model context-parallelism work and lives on `main`; this PR touches only the GLM-5.2 DSpark draft RoPE and does not go near it.

Could you re-run CI on this commit? Happy to send a separate PR making `_free_port()` hold the socket open until the child binds, so the race stops recurring.

HuiyingLi (approved):

khazic:
The red CI here is infrastructure, not this change.

Both failing shards (`L0_Unit_Tests_CPU` and `L0_Unit_Tests_GPU_2_of_2`) fail on the same two tests, neither of which this PR touches:

```
tests/unit_tests/tools/test_sync_model_coverage_tables.py::test_model_release_docs_pages_exist_in_nightly_navigation
tests/unit_tests/tools/test_sync_model_coverage_tables.py::test_embedding_and_reranking_releases_are_discovered_from_recipes

ValueError: Git history is incomplete for model support generation; use fetch-depth: 0.
Recipe additions fall on a shallow boundary: examples/audio_finetune/qwen2_5_omni_asr/ami_sft_3b.yaml,
examples/audio_finetune/qwen2_5_omni_asr/ami_sft_7b.yaml, examples/audio_finetune/qwen3_omni_asr/ami_sft.yaml,
examples/convergence/tulu3/models/gemma4-31b/gemma4_31b_tulu3_packed2k_cp1_gbs32_4node.yaml, ...
```

The rest of the suite is green: `2 failed, 11228 passed, 1146 skipped`. Every test covering the RoPE and softmax-scale changes in this PR passes.

Root cause: `tests/ci_tests/utils/sync_model_coverage_tables.py` walks `git log` to find the commit that introduced each `examples/**/*.yaml`. The CI checkout is shallow, and a batch of recipe-adding commits sits exactly on the shallow boundary, so the helper raises. The error message names the fix itself: `fetch-depth: 0` on the checkout step.

This is not specific to this branch. #3263 hits the identical two failures in the same two shards on its latest run, with `L0_Unit_Tests_GPU_1_of_2` green in both cases.

Happy to open a separate one-line CI fix for the checkout depth if that is the preferred route; I would rather not fold a workflow change into this PR.

HuiyingLi:
@khazic thanks, there is https://github.com/NVIDIA-NeMo/Automodel/pull/3472

PR #3284, opened 2026-07-29
fix(#1755): Fix PEFT adapter outputs for v5 support

### LINKED ISSUE #1755: Fix PEFT adapter outputs for v5 support
https://github.com/NVIDIA-NeMo/Automodel/pull/1727 CI caught an issue where Qwen3 state dict adapter outputs unfused lora adapters (as was the behavior in transformers v4). In v5, the expectation for Qwen3MoE is to have fused adapters saved for the expert layers. Since it is currently close to release, I limited blast radius to Qwen3, but I have a strong suspicion it affects other MoE custom models as well. It remains to be investigated + fixed accordingly.

Mr-Neutr0n:
Fixes NVIDIA-NeMo/Automodel#1755

Moved ParamWrapper v5 PEFT adapter logic (fused LoRA format) from the Qwen3-specific adapter into the shared `MoESplitExpertsStateDictMixin` so all MoE models (DeepSeek V3/V3.2/V4, GLM4 MoE, Qwen3 Omni/Next, etc.) benefit from v5 fused adapter output. Also generalized `_extract_target_parameters` and `_extract_target_modules` in `addons.py` from Qwen3-only to all mixin-based MoE models.

Local test infra unavailable in CI sandbox.

---

This change was prepared with AI assistance under human direction and review.


Mr-Neutr0n:
Thanks for the review — all three findings addressed.

**1. Dead `logger`.** Removed, along with the now-unused `import logging`. No diagnostic was intended; it was left over from an earlier draft.

**2. Duplicated conversion methods.** Deleted the Qwen3 overrides so the mixin is the single source of truth. One correction worth recording, though: they were *not* byte-for-byte identical. The Qwen3 copies carried explanatory comments the mixin versions lacked — the ParamWrapper nesting rule (`"down_proj" < "gate_up_proj"`, so down_proj is outer) and the per-branch shape transforms:

```
+        # PEFT ParamWrapper nesting: target_parameters are sorted alphabetically
+        # and wrapped in order. The FIRST wrapped becomes the OUTER ParamWrapper.
         if suffix == "lora_gate_and_up_B":
+            # (E, r, 2*I) -> (r*E, 2*I)
```

I checked equivalence at AST level rather than by eye before touching anything:

```
_convert_lora_to_paramwrapper:  AST-equivalent (comments ignored) = True
_convert_paramwrapper_to_native: AST-equivalent (comments ignored) = True
```

So the executable code was genuinely identical and safe to collapse — but a straight deletion would have silently lost the better documentation. I moved those comments into the mixin first, then removed the copies. Net −128 lines from the Qwen3 adapter.

Doing it here rather than as a follow-up since this PR introduced the duplication; leaving it to drift for a release seemed worse than the slightly wider diff. Happy to split it out if you'd rather keep this PR to the v5 change.

**3. Test gap on the v5 defaults.** Added `TestExtractTargetsV5Defaults` — CPU-only, no CUDA gate, since none of it touches a device. Your read was right that this was the real hole: every existing MoE assertion passes `v4_compatible=True`, so the shipped defaults had no coverage at all. Measured on the tiny 2-layer fixture:

```
target_parameters (v5 default): ['mlp.experts.gate_up_proj', 'mlp.experts.down_proj']
target_parameters (v4):         []
target_modules   (v5 default):  n=0     <- per-expert expansion correctly skipped
target_modules   (v4):          n=24
```

Six cases: the fused paths, the v4 empty case, no `experts.<N>.` names in v5, a mutual-exclusivity check (exactly one mechanism describes the experts in each mode), the non-MoE model case, and the PP list-of-parts form.

## Test state

`tests/unit_tests/moe` + `tests/unit_tests/checkpoint`: **819 passed, 74 skipped**. `ruff check` and `ruff format --check` clean on all changed files.

One pre-existing failure to flag: `test_optimizer_state.py::test_native_pipeline_optimizer_state_dcp_round_trip_across_ranks` fails on my machine with a `ProcessRaisedException`. It fails identically on unmodified `main`, so it's my macOS box and multiprocess distributed, not this change.

akoumpa:
Hi @Mr-Neutr0n , thank you for the PR!

I see there's a test failing in https://github.com/NVIDIA-NeMo/Automodel/actions/runs/30533391202/job/91391697153?pr=3284 could you take a look? I can then help rerun the CI. thank you

Mr-Neutr0n:
Addressed the review: the failing merge helper was implicitly relying on the legacy per-expert format, so it now passes `v4_compatible=True` explicitly while production keeps the fused v5 default. The failing functional file and MoE unit tests now pass locally: 36 passed, 2 skipped. https://github.com/Mr-Neutr0n/Automodel/commit/fff1427b

nemo_automodel/components/checkpoint/addons.py
```diff
@@ -303,30 +303,23 @@ def _get_automodel_peft_metadata(peft_config: "PeftConfig") -> dict:
...
-    model = model[0] if isinstance(model, (list, tuple)) else model
+    model_part = model[0] if isinstance(model, (list, tuple)) else model
     if v4_compatible:
         return []
-    if _is_qwen3_moe(model):
-        return ["mlp.experts.gate_up_proj", "mlp.experts.down_proj"]
+    adapter = getattr(model_part, "state_dict_adapter", None)
+    if adapter is not None and isinstance(adapter, MoESplitExpertsStateDictMixin):
+        expert_segment = adapter._expert_path_segment
+        return [f"{expert_segment}.gate_up_proj", f"{expert_segment}.down_proj"]
```
yuhezhang-ai: [P1] Handle non-gated MoE target parameters

This always advertises `gate_up_proj`, but Nemotron V3 is a non-gated ReLU² MoE whose Transformers model exposes `mixer.experts.up_proj` and `down_proj`. With the current list, PEFT wraps `down_proj` but silently omits the expert up projection, so the saved expert adapter is incomplete. Please derive the projection names from `adapter._is_gated_moe` (or an adapter-owned target-parameter hook) and add a non-gated test that expects `up_proj`.

nemo_automodel/components/moe/state_dict_mixin.py
```diff
@@ -742,11 +876,15 @@ def _convert_single_merged_expert_to_hf_split_experts(
...
+        # When v4_compatible=True: per-expert split keys (v4 format).
+        # When v4_compatible=False (default): fused ParamWrapper format (v5).
+        v4_compatible = kwargs.get("v4_compatible", False)
         for suffix in _LORA_EXPERT_SUFFIXES:
             if f".{expert_segment}.{suffix}" in fqn and fqn.endswith(f".{suffix}"):
-                return self._convert_lora_expert_to_hf(fqn, tensor, n_experts, inter_dim, expert_segment)
+                if v4_compatible:
+                    return self._convert_lora_expert_to_hf(fqn, tensor, n_experts, inter_dim, expert_segment)
+                else:
+                    return self._convert_lora_to_paramwrapper(fqn, tensor)
```
yuhezhang-ai, line 916: [P1] Do not quantize Kimi ParamWrapper LoRA tensors

For Kimi K2.5, this new default produces expert keys ending in `lora_A.weight` / `lora_B.weight`. `KimiK25VLStateDictAdapter.convert_single_tensor_to_hf` subsequently forces quantization on, and `_is_quantized_expert_key` matches any `mlp.experts.*` key containing `.weight`. These LoRA tensors are therefore rewritten as `weight_packed`, `weight_scale`, and `weight_shape`, which PEFT cannot load. Please exclude LoRA/ParamWrapper keys from Kimi expert-weight quantization and add a v5 export test asserting that the LoRA keys remain ordinary `.weight` tensors.

nemo_automodel/components/moe/state_dict_mixin.py
```diff
@@ -229,6 +232,134 @@ def _concatenate_expert_weights(
...
+            out = tensor.reshape(-1, tensor.shape[2]).contiguous()
+            pw_suffix = "lora_A.weight"
+        elif suffix == "lora_down_A":
+            # (E, I, r) -> permute(1,2,0) -> (I, r, E) -> (I, r*E)
+            out = tensor.permute(1, 2, 0).contiguous().reshape(tensor.shape[1], -1)
+            pw_suffix = "lora_B.weight"
+        else:
+            return [(fqn, tensor)]
+
+        out_fqn = f"{prefix}layers.{layer_num}.{expert_segment}.{pw_suffix}"
```
yuhezhang-ai, line 314: [P1] Apply adapter-specific HF prefix normalization

This copies the prefix directly from the native FQN. Nemotron V3 native keys use `model.layers...`, while its HF format requires `backbone.layers...`; the `model` → `backbone` rename currently occurs only in the non-expert branch after this helper has returned. The new expert ParamWrapper keys can consequently use a different prefix from the rest of the adapter and fail to attach during PEFT loading. Please route the generated FQN through adapter-specific HF prefix normalization (or provide a hook) and cover the Nemotron prefix in a round-trip test.

yuhezhang-ai (commented):
I found three correctness issues in the generalized v5 ParamWrapper path. Leaving the release rollout/scope decision for a separate discussion.

yuhezhang-ai:
@Mr-Neutr0n Thanks very much for the PR!

@akoumpa I’d like to discuss the scope because this change affects a broad set of MoE adapters.

The PR makes the Transformers v5-compatible fused adapter format the default for all shared MoE adapters, while our checkpoint-robustness coverage validates only a subset of the existing PEFT examples. Expanding and stabilizing that coverage before the release would be substantial work.

For this release, I recommend making the new format opt-in per adapter:

- Retain it for Qwen3-MoE, which already has the fused path and checkpoint coverage.
- Enable it for Nemotron V3, MiniMax M2, and Kimi K2.5 after the identified fixes and targeted PEFT save/reload tests pass.
- Keep the legacy format for Qwen3 Next, GLM4 MoE, DeepSeek V3, and other unvalidated adapters whose legacy paths currently work.

Once @Mr-Neutr0n updates the PR, I can help select and run scoped checkpoint CI.

Does this scope sound good to you?

Mr-Neutr0n:
Addressed the review: scoped v5 ParamWrapper export to Qwen3 through an adapter-owned opt-in, so every unvalidated MoE adapter stays on the legacy format. I also merged current main, routed Nemotron expert LoRA keys through its backbone/up_proj normalization, and included the upstream Kimi LoRA quantization guard. Focused result: 128 passed, 15 skipped. https://github.com/NVIDIA-NeMo/Automodel/commit/ecb5d6c08f30e8e75b760985b2522c5a62f10664

yuhezhang-ai:
/ok to test [ecb5d6c](https://github.com/NVIDIA-NeMo/Automodel/pull/3284/commits/ecb5d6c03adf151981fa2041d2673c9c482f76e3)

yuhezhang-ai:
Thanks — the latest changes address my review concern. I ran the scoped `qwen3_moe_30b_lora` checkpoint-robustness CI on the exact current head (`ecb5d6c03adf151981fa2041d2673c9c482f76e3`), and it passed: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/387389341

That job covers source reference/load parity, train/save, AutoModel reload, and Hugging Face reload, so the Qwen3-MoE Transformers v5 PEFT path in this PR is ready from my review perspective.

I will follow up separately for Nemotron V3 and MiniMax M2, since their legacy PEFT paths have independent model-specific issues that should be fixed and validated outside this PR. Kimi K2.5 is being covered through its repaired legacy path in #3295 and #3431.

stanley1208:
hey! been following this from #1755,  the per-adapter scoping is a nice touch.

I think I found one gotcha though, and it's a sneaky one. I ran the fused export shapes against the real Qwen3MoeExperts on transformers 5.12.1, and peft wants them the other way around:

    peft creates:  base_layer.lora_A (r*E, hidden), base_layer.lora_B (2*inter, r*E)
    pr emits:      base_layer.lora_A (r*E, 2*inter), base_layer.lora_B (hidden, r*E)

so loading gives a size mismatch. the tricky part is it lines up perfectly whenever 2*moe_inter equals hidden, so on some configs everything looks fine. qwen3-30b isn't one of them though (1536 vs 2048), which makes me think the validation went through automodel's own save/resume, that path is self consistent so it wouldn't notice. an external PeftModel.from_pretrained should show it right away.

when I was testing for the survey, this orientation matched a manual reference merge exactly:

    lora_A = A.permute(0, 2, 1).reshape(r * E, in_features)
    lora_B = B.permute(2, 1, 0).reshape(out_features, r * E)

same idea for the down pair. the probe script is about 20 lines and runs on cpu, happy to share it or help test.

yuhezhang-ai:
@stanley1208 Thanks for checking this! I reproduced it with the real `Qwen3MoeExperts` using the repository’s pinned Transformers 5.12.1 and PEFT 0.18.1, and I’m seeing the opposite orientation.

For `gate_up_proj` shaped `[E, 2I, H]`, PEFT’s `ParamWrapper` creates:

```text
base_layer.lora_A: [rE, 2I]
base_layer.lora_B: [H, rE]
```

Those shapes exactly match the tensors emitted by this PR. A tiny CPU `PeftModel.from_pretrained` test loaded successfully, and `merge_and_unload()` matched the native AutoModel `A @ B` expert-weight delta within `4.8e-7`.

The scoped checkpoint-robustness job also did run the HF reload phase—not just AutoModel resume. It loaded with `PeftModel.from_pretrained`, matched all 578 saved adapter tensors, and completed a forward pass. 

I suspect the difference may come from applying the normal `nn.Linear` `[out, in]` convention, whereas `ParamWrapper` derives its LoRA dimensions directly from the raw 3-D parameter layout. Could you share your probe and the exact PEFT/Transformers versions and `target_parameters` order? I’d be happy to compare them directly.

yuhezhang-ai (approved):

PR #3287, opened 2026-07-29
feat(speculative): add Kimi K3 DFlash draft training

khazic:
Adds Kimi K3 support to the DFlash speculative-decoding training path.

Re-opens #3265, which GitHub closed automatically when its base branch `huiyingl/feat/kimi-k3` was merged and deleted. Same work, now re-targeted at `main`; the review history lives on #3265.

## The draft

`KimiK3DFlashDraftModel` is a small dense K3 MLA stack trained against the frozen K3 text backbone. It keeps the target's MLA projections (Q-LoRA, compressed KV latent), output gate, fp32-variance RMSNorm, and SiTU MLP. KDA linear attention, routed and shared experts, MTP, and the learned attention-residual mixer stay in the target only, and the draft-config builder switches each of them off explicitly so the serialized draft config describes the draft rather than the target.

Three things are specific to this backbone:

- **The attention subclasses the target's `KimiMLAAttention`** and overrides only `forward`, so the draft's projection layout cannot drift from the layers whose hidden states it consumes.
- **NoPE.** K3 requires `mla_use_nope=True`, so no rotary is applied anywhere in its full-attention layers and the draft ignores `position_ids`.
- **`F.scaled_dot_product_attention` instead of the target's eager matmul.** A DFlash block batch makes both the query and the key axis sequence-scale (`num_anchors * block_size` by `seq_length + num_anchors * block_size`), so materializing the score matrix is not affordable. The dense additive DFlash mask is passed through unchanged, so visibility semantics are identical; verified numerically against the eager reference in fp64 (max abs diff 5.6e-17).

## Recipe changes

Rather than branching on the model, the per-target divergence moves into `DFlashDraftSpec`, which now carries `build_draft_config`, `build_target_kwargs`, `attention_backends`, and `supports_context_parallel`. As a result `train_dflash.py` contains no model names:

- The Qwen3 draft-config surgery moves out of the recipe into `draft_qwen3.py` as `build_qwen3_dflash_draft_config` (a faithful move, no behavior change).
- The text config is resolved generically with `checkpoint_config.get_text_config()`, which unwraps K3's multimodal wrapper and is the identity for Qwen3.
- The sequence-packing gate keys on the target's own `_owns_packed_attention` flag rather than a model name.
- `attention_backend` is validated against the spec: the K3 draft declares `("sdpa",)` because it has no FlexAttention path.

Two gates are new and worth calling out for existing users:

- **`pp_size > 1` is now rejected for every DFlash target.** Online hidden-state capture hooks one complete, non-pipelined target forward, so a pipelined target would fire the hooks on a subset of ranks with partial activations. This previously failed silently or confusingly; no shipped config or test sets it.
- **`cp_size > 1` is rejected for K3**, which declares `_owns_cp_attention` and shards the batch itself, so it never routes through the CP key/value-gather hook the DFlash target wrapper installs. The check runs before the pre-existing `target_force_hf` / `sdpa` CP gates, because those would otherwise report a misleading error first and their suggested fix (`force_hf=true`) routes K3 off its native class entirely. `test_context_parallel_is_rejected_before_the_generic_cp_gates` pins that ordering and asserts no weight is loaded first.

## Example and docs

`examples/speculative/dflash/kimi_k3_dflash.yaml` plus `README_kimi_k3.md`. `num_anchors` is lower than the Qwen3 configs' because the dense mask spans `num_anchors * block_size` queries by `seq_length + num_anchors * block_size` keys, so anchors cost quadratically more than on the sparse FlexAttention path. `examples/speculative/README.md` is updated with the new supported target and the `pp_size` constraint.

## Testing

`tests/unit_tests/speculative/test_dflash_kimi_k3.py` (13 CPU tests): the config builder (including that it does not mutate the target config, and that it rejects the multimodal wrapper), registry resolution for both K3 architectures, the three rejection paths (projector head, grouped K/V heads, RoPE), the CP-gate ordering, `target_layer_ids` defaulting, mask-driven context isolation in the forward, and an end-to-end forward plus backward through `DFlashTrainerModule` on the SDPA mask path.

Full local run over `tests/unit_tests/speculative`, `tests/unit_tests/models/kimi_k3` and `tests/unit_tests/recipes/llm`: the failure list is identical to the one `main` produces on its own (`test_benchmark.py` plus one K3 test needing the `fla` extra, both local environment gaps). `ruff format` and `ruff check` clean on the changed files.

Not yet validated on GPU. The frozen K3 target needs an expert-parallel allocation, so the multi-node run against the real checkpoint is still outstanding; the draft, the mask contract, and every gate are covered on CPU.

## Re-targeting onto main

`main` now carries the merged K3 model, so this branch merges `main` and keeps main's version of every K3 file. One fix was needed on top:

- `KimiMLAAttention.__init__` now takes a `BackendConfig` and dispatches its forward on `backend.attn`. `KimiK3DFlashAttention` subclasses it, so the draft threads a backend through its decoder layers and passes `attn="eager"`: the draft replaces the forward with its own SDPA path over the dense mask, and eager keeps the parent from building attention modules the draft never uses.


HuiyingLi:
GPU validation completed successfully on the full Kimi K3 target.

### Run

- Validated implementation commit: `7a690a573a849bf4d757bd855f410a2166d48fba`
- Recipe memory fix now on the PR branch: `06c48a53d851070bb9dc71f91eb5a578092f5321`
- Slurm job: `6014922` — `COMPLETED (0:0)`, 52/52 training steps
- Topology: 64 GB200 GPUs (16 nodes x 4), FSDP2, EP64 / PP1 / CP1 / TP1
- Runtime: 2K sequence length, micro-batch size 1, 3-layer DFlash draft, 128 anchors x block size 16, HybridEP target dispatcher, FLA KDA, and SDPA draft attention
- Peak sampled GPU memory: **160,875 MiB (157.10 GiB)**
- W&B: https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/1g9nw94a
- Rank-0 run log: `/lustre/fs1/portfolios/coreai/projects/coreai_dlalgo_llm/users/huiyingl/k3/pr3287/logs/full_ep64_s2048_52steps_6014922_nvl72167-T01.out`

### Result

- Train loss: `15.5000` at step 1 -> `6.2500` at step 52
- Validation loss: `36.3594` after epoch 1 -> `6.4370` after epoch 13
- Final validation accuracy: `0.0645`
- Final validation acceptance length: `1.0603`
- The preceding one-node / 4-layer EP4 smoke also completed 8/8 steps in job `6014768` before the full run.

Final log excerpt:

```text
epoch=12 step=52 loss=6.2500 acc=0.0562 lr=1e-05
Finished epoch 13/13 completed_steps=4 skipped_short_micro_batches=0 val_loss=6.4370 val_accuracy=0.0645 val_accept_len=1.0603
```

`reshard_after_forward: true` is required for this frozen target. The identical full Kimi K3 FSDP target path was already measured to OOM near the 184 GiB device limit with `false`: the target runs under `torch.no_grad()`, so no backward pass releases the all-gathered block parameters. This DFlash run reached 157.10 GiB with `true`. I pushed the recipe change in `06c48a53d`.

One data caveat: this run used a local 256-train / 32-validation Tulu3 slice, not responses regenerated from Kimi K3 at temperature 0.7. It validates full-target loading, EP64 execution, dense-mask SDPA draft training, metric aggregation, memory behavior, and loss movement; the acceptance value should not be treated as final PerfectBlend/serving-quality data.

HuiyingLi (approved):

PR #3204, opened 2026-07-23
build: cache CUDA extension source builds

akoumpa:
## What changed

- Mount uv's `/opt/uv_cache` during the main Docker dependency sync and stop disabling uv caching in the container and media install paths.
- Mount pip's wheel cache for the deploy image's source builds of `causal-conv1d` and `mamba-ssm`.
- Persist BuildKit cache mounts through GitHub Actions with `main` as the canonical writer and copied PR branches as restore-only consumers.
- Add `flash-attn` to the persistent CUDA wheelhouse alongside `causal-conv1d`, `mamba-ssm`, and `nv-grouped-gemm`.
- Verify all four heavy wheels exist and re-enable the `fa` installation-test lane.

## Why

The source-only CUDA dependencies are expensive to rebuild. BuildKit cache mounts speed up repeated builds, while the Actions-backed CUDA wheelhouse provides durable reuse across ephemeral CI runners. `flash-attn` was the missing wheel and its install test had been disabled because compiling it on every run took roughly two hours.

GitHub Actions cache storage is shared at repository scope. Saving multi-GB copies under every `pull-request/*` ref caused the repository to exceed its cache limit and evict useful entries. Restricting persistent writes to `main` prevents cache thrashing while still allowing copied PR branches to restore an exact cache from their current or default-branch scope.

## Impact

`main` owns the canonical BuildKit and CUDA-wheelhouse caches. Copied PR branches restore exact matching caches but do not create branch-scoped persistent copies. If a PR changes an ABI-keyed build input and no exact cache exists on `main`, that run rebuilds the missing wheels and passes them to downstream install jobs through the workflow artifact; the first post-merge `main` run then populates the canonical key.

## Validation

- YAML parse of `.github/workflows/cicd-main.yml`, `.github/workflows/install-test.yml`, and `.github/actions/build-container/action.yml`
- Cache lifecycle assertions for main/branch hit and miss paths
- `docker buildx build --check -f docker/Dockerfile .`
- `docker buildx build --check -f docker/Dockerfile.deploy .`
- `bash -n scripts/install_media.sh`
- `git diff --check`
- `actionlint` comparison against the previous head; no new findings

Runtime validation plan:

1. On the copied PR head, confirm the cache steps use `actions/cache/restore@v4` and do not expose post-job cache-save steps.
2. Confirm a wheelhouse cache miss still builds/verifies all required wheels and supplies the `fa` install lane through the workflow artifact.
[...]

PR #3289, opened 2026-07-29
ci: prototype OCI-backed CUDA wheelhouse

akoumpa:
## What changed

- replace the persistent `actions/cache` CUDA wheelhouse with an exact-tagged OCI artifact in `ghcr.io/nvidia-nemo/automodel-cuda-wheelhouse`
- retain the existing lock-derived ABI fingerprint as the OCI tag, so only an exact Torch/CUDA/Python/architecture/package match can restore wheels
- keep PR jobs read-only with `packages: read`
- add a separate `main`-only publisher with `packages: write`
- continue using the one-run GitHub artifact to fan wheels out to the CUDA installation matrix
- add policy tests that prevent PR registry writes and accidental reintroduction of `actions/cache` for the wheelhouse

## Why this is necessary after three cache PRs

This does not replace the correctness work in the previous PRs. It addresses the storage boundary that remains after them:

- [#3204](https://github.com/NVIDIA-NeMo/Automodel/pull/3204) introduced persistent CUDA wheel reuse, added Flash Attention, and made `main` the canonical writer so PR refs stopped creating duplicate multi-GB entries.
- [#3253](https://github.com/NVIDIA-NeMo/Automodel/pull/3253) stopped newer `main` pushes from cancelling the only run capable of finishing and saving the canonical wheelhouse.
- [#3269](https://github.com/NVIDIA-NeMo/Automodel/pull/3269) made the key ABI-correct and exact, avoiding unrelated invalidations and unsafe prefix restores.

Those changes fixed duplication, cancellation, and key correctness, but the canonical wheelhouse still lives in the repository-scoped Actions cache. The repository was observed above the default 10 GB cache limit with several roughly 1.35 GB wheelhouses, so GitHub can evict the expensive canonical entry even when the key and ownership policy are correct. A cold exact-key build still takes about 3.5 hours, dominated by Flash Attention.

GHCR is durable artifact storage rather than an LRU dependency cache. This prototype preserves the manifest and ownership rules from the earlier PRs while moving the multi-GB payload out of the constrained Actions cache.

## Behavior

1. The CUDA job computes the existing lock-derived fingerprint and forms an exact OCI tag.
2. It authenticates to GHCR with read-only package permission and restores the artifact when that exact tag exists.
3. On a miss it builds and verifies the wheelhouse as before, then uploads the one-day workflow artifact for downstream install jobs.
4. Only a successful `main` miss enables the publisher job, which downloads that workflow artifact and publishes the exact OCI tag.
5. The publisher checks for an existing tag first, making concurrent `main` completion idempotent.

PR code cannot publish packages. The only job with `packages: write` is gated to `refs/heads/main`.

## Validation

- `9 passed`: existing lock/fingerprint tests plus the new registry-policy and summary tests
- Ruff check and format check passed
- `ty` passed for the helper and new test
- workflow YAML parsed successfully
- `git diff --check` passed
- `regctl v0.11.5` completed a local OCI artifact put/head/get/extract byte-for-byte round trip using the same artifact media type as the workflow
- `actionlint v1.7.7` reports only the existing custom-runner-label and pre-existing UV matrix diagnostics; no new registry-step diagnostic

## Runtime proof

The GHCR write and final read-only restore paths were proven without merging:

1. [Bootstrap job 90814087358](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/30525100295/job/90814087358) authenticated with `packages: write` and published the verified wheelhouse under its exact ABI tag. The temporary bootstrap job was removed before the final run and is not present in the PR diff.
2. After merging current `main`, [wheelhouse job 90816196544](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/30525722346/job/90816196544) ran with the final `packages: read` PR policy and completed in 3m55s.
3. `Restore CUDA wheelhouse from GHCR` succeeded, `Build cached CUDA dependency wheels` was skipped, and wheel verification passed.
4. The base, `media`, `all`, `vlm`, `fa`, and `cuda` installation lanes all passed using the restored wheels.
5. The `main`-only publisher remained skipped on the copied PR branch.


akoumpa:
### OCI wheelhouse proof complete

This has now been validated without merging to `main`.

#### Write path

[Bootstrap job 90814087358](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/30525100295/job/90814087358) reused the verified wheelhouse artifact from the original cold-build run and successfully:

- authenticated to GHCR with `packages: write`;
- computed the same lock-derived ABI tag;
- published the OCI artifact to `ghcr.io/nvidia-nemo/automodel-cuda-wheelhouse`.

The temporary PR-only bootstrap job was then removed. It is not present in the current diff.

#### Final read-only restore path

After merging current `main` and resolving the action-pinning conflict, head `366360f0d8187fd3f919fa9fc86569951fef09da` ran with the final policy:

- the CUDA job has only `packages: read`;
- the `main`-only publisher is skipped on the PR;
- [wheelhouse job 90816196544](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/30525722346/job/90816196544) completed successfully in **3m55s**;
- `Restore CUDA wheelhouse from GHCR` succeeded;
- `Build cached CUDA dependency wheels` was **skipped**;
- wheel verification and workflow-artifact upload passed.

All six wheelhouse-consuming CUDA pip lanes passed:

- [base](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/30525722346/job/90817134439)
- [`media`](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/30525722346/job/90817134448)
- [`all`](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/30525722346/job/90817134462)
- [`vlm`](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/30525722346/job/90817134464)
- [`fa`](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/30525722346/job/90817134473)
- [`cuda`](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/30525722346/job/90817134498)

This proves the intended steady state: an exact OCI restore replaces the roughly 3.5-hour native-extension rebuild while preserving the existing ABI fingerprint and PR-read-only/main-writer ownership model.

.github/workflows/install-test.yml
```diff
@@ -417,14 +483,17 @@ jobs:
           RUN_ID: ${{ github.run_id }}
           SKIPPING_IS_ALLOWED: ${{ needs.pre-flight.outputs.docs_only == 'true' || needs.pre-flight.outputs.is_deployment_workflow == 'true' }}
         run: |
-          FAILED_JOBS=$(gh run view $GITHUB_RUN_ID --json jobs --jq '[.jobs[] | select(.status == "completed" and .conclusion != "success")] | length') || echo 0
+          FAILURE_CONCLUSIONS='["failure","cancelled","timed_out","action_required","startup_failure"]'
```
chtruong814, line 486: Could we make this fail closed instead of enumerating only known failure conclusions? This currently treats a completed `stale` job as success, even though GitHub uses `stale` when a check took too long. The change only needs to permit the skipped publisher, so please explicitly accept `success`, `skipped`, and (if intended) `neutral`, and count every other completed conclusion as a failure. The focused test should cover `stale` as well.

.github/workflows/install-test.yml
```diff
@@ -200,32 +207,43 @@ jobs:
...
+      # The fingerprint is the complete ABI key, so only an exact OCI tag may
+      # be restored. GitHub Actions cache is intentionally not used here:
+      # wheelhouses are multi-GB and compete for the repository's 10 GB cache.
+      - name: Restore CUDA wheelhouse from GHCR
+        id: restore-cuda-wheelhouse
+        env:
+          WHEELHOUSE_REFERENCE: ${{ steps.cuda-wheelhouse-ref.outputs.reference }}
+        shell: bash
+        run: |
+          if regctl manifest head "${WHEELHOUSE_REFERENCE}" >/dev/null 2>&1; then
```
chtruong814, line 235: This collapses every `manifest head` error into a cache miss. Authentication failures, rate limits, and GHCR outages would therefore silently trigger the roughly 3.5-hour cold build. Could we distinguish a genuine not-found response from operational errors, report the latter clearly, and only set `registry-hit=false` for an actual missing tag?

.github/workflows/install-test.yml
```diff
@@ -200,32 +207,43 @@ jobs:
...
-        if: github.ref != 'refs/heads/main'
-        uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
+      - name: Install regctl
+        uses: regclient/actions/regctl-installer@c70ad64367908075211b10dcd2ab9fad4bfa1816
         with:
-          path: ${{ env.WHEELHOUSE_DIR }}
-          key: ${{ steps.cuda-wheelhouse-key.outputs.key }}
+          release: v0.11.5
+
+      - name: Log in to GHCR
```
chtruong814, line 218: Could we use a full-SHA-pinned `docker/login-action` here and in the publisher instead of handling `GITHUB_TOKEN` through custom shell login code? `regctl` can consume Docker's credential configuration, and the action provides the standard GHCR login/logout lifecycle. The existing job-level `packages: read` / `packages: write` split should remain the authorization boundary.

.github/workflows/install-test.yml
```diff
@@ -248,6 +266,54 @@ jobs:
...
+        shell: bash
+        run: |
+          if regctl manifest head "${WHEELHOUSE_REFERENCE}" >/dev/null 2>&1; then
+            echo "Wheelhouse already published by another main run: ${WHEELHOUSE_REFERENCE}"
+            exit 0
+          fi
+
+          archive="${RUNNER_TEMP}/cuda-wheelhouse.tar"
+          tar -cf "${archive}" -C /tmp/cuda-wheelhouse .
+          regctl artifact put \
```
chtruong814, line 311: Each fingerprint publishes another roughly 1.35 GB package version, and GHCR does not provide an Actions-style `retention-days` setting. Could we add a separate scheduled/manual cleanup workflow using the GitHub Packages REST API? A safe policy would preserve the current-main tag, the newest three versions, and every version younger than 60 days, deleting only versions outside all three protections. Please keep cleanup separate from publication and limit deletion authority to that trusted cleanup job.

chtruong814 (commented):
The OCI wheelhouse direction makes sense: it preserves the exact ABI fingerprint and read-only PR policy while moving the multi-GB payload out of the Actions cache eviction boundary. The final read-only restore path is convincingly demonstrated.

I left focused comments on four remaining areas: fail-closed summary conclusions, distinguishing a registry miss from registry failures, using the standard pinned GHCR login action, and adding bounded scheduled retention cleanup.

One remaining validation caveat: the successful write used the temporary bootstrap job, while the final `main`-gated publisher stayed skipped in the proof run. Please monitor and report the first post-merge `main` miss to confirm the final permissions, workflow-artifact handoff, and publication path operate together as written.

PR #3292, opened 2026-07-29
feat(distributed): migrate Megatron FSDP to MCore v2

wujingyue:
## What changed

- replace the standalone Megatron-FSDP adapter behind `strategy: megatron_fsdp` with Megatron Core MFSDP v2
- shard configured blocks bottom-up and the model root with flat DP placements
- restrict the initial implementation to DP-only ZeRO-3 and reject unsupported legacy options/topologies
- materialize meta models and load large checkpoints on CPU before GPU sharding
- preserve protected FP32 parameter groups, tie/freeze before sharding, and adapt optimizers in place
- remove legacy wrapper back-references and parameter attribute repair
- replace the `megatron-fsdp` dependency with a pinned compatible Megatron Core revision
- update existing unit, configuration, mesh, infrastructure, functional-registry, documentation, and skill coverage

## Why

The existing `megatron_fsdp` strategy used the retired standalone adapter and depended on wrapper-specific parameter metadata. Moving it to the Megatron Core v2 APIs provides the maintained sharding implementation and enables CPU checkpoint staging for models that cannot be materialized fully on GPU.

## User impact

`strategy: megatron_fsdp` remains the configuration entry point, but initially requires:

- `zero_dp_strategy: 3`
- `tp_size`, `cp_size`, `pp_size`, and `ep_size` equal to 1
- default values for legacy options that MFSDP v2 does not support

Large pretrained models are initialized on meta, materialized and loaded on CPU, and then sharded onto the DP mesh.

## Validation

- Ruff format and lint on all changed Python files
- 439 affected AutoModel unit tests passed
- unchanged Megatron Core MFSDP v2 bucket passed on two ranks: 51 passed, 11 topology-dependent skips per rank
- 12 affected functional tests collected successfully
- two-GPU Qwen3.5-35B-A3B LoRA smoke completed two optimizer steps with finite losses and gradients
  - peak memory: 44,638 MiB and 44,470 MiB
  - losses: 2.997448 and 1.7897

## Known validation gap

The smoke test's first loss did not match the older custom-FSDP2 reference of `3.044429` within `atol=0.01`; the observed difference was approximately `0.04698`. The old DP-only FSDP2 reference path also currently fails from clean `main` while converting MTP expert checkpoint tensors because it assumes an EP mesh. This PR is therefore intentionally opened as a draft.


akoumpa:
Thank you, @wujingyue, for putting this PR together!

My preference would be to hold off on integrating it until Megatron Core MFSDP v2 supports CUDA Graphs and symmetric memory. I see those as important capabilities, and their absence is currently a key gap compared with FSDP2.

Would you be comfortable pausing the integration until that support is available? I’d be happy to revisit it then.

Thanks again!

wujingyue:
> their absence is currently a key gap compared with FSDP2.

Hey @akoumpa, not really -- I already fixed https://github.com/NVIDIA/Megatron-LM/issues/5717 and https://github.com/NVIDIA/Megatron-LM/issues/5655. 

Caveat: as per https://github.com/NVIDIA/Megatron-LM/issues/5655#issuecomment-5132671561, we can't yet enable symmetric memory and CUDA graph **simultaneously**. However, FSDP2 has the same problem. 

Also, this PR is currently in draft, helping me finding bugs and working with agents. So would you mind me keeping it?

akoumpa:
Hi @wujingyue yeah no problem at all to keep the current PR. What i meant is to post-pone upgrading until the symmem+cg feature has landed, which is the key feature fsdp2 does not currently have.

wujingyue:
> What i meant is to post-pone upgrading until the symmem+cg feature has landed

Why? Neither MFSDP v2 nor PyTorch FSDP2 support symm_mem + cg. Am I missing something? cc @akoumpa

PR #3293, opened 2026-07-29
fix(pp): route Qwen3.5 MoE pre-embedded inputs across stages

HuiyingLi:
Summary:
- Add Qwen3.5 MoE pipeline-stage routing for pre-embedded inputs.
- Preserve the FSDP root pre-embedding path across PP boundaries.
- Handle split-FP8 expert checkpoint tensors used by this PP path.

Validation:
- qwen3_5_moe unit tests
- local 4-GPU EP+PP smoke tests


HuiyingLi:
Closing in favor of a clean model-only PR.

PR #3303, opened 2026-07-30
feat(training): add weighted domain mixtures

akoumpa:
# What does this PR do ?

Add a typed `domain_mixture` objective for Megatron continual pretraining so sampling frequency and desired domain contribution can be configured independently.

For normalized sampling weight `q_i` and objective weight `w_i`, supervised tokens from domain `i` receive the importance multiplier `w_i / q_i`. This preserves one optimizer and LR scheduler while making the expected mixed-data loss match the configured domain objective.

# Changelog

- Add typed named-domain sampling/objective configuration with validation against explicit Megatron blend weights.
- Thread per-token objective weights through masked, fused-linear, chunked, Transformer Engine parallel, and MTP cross-entropy paths before global token normalization.
- Log achieved supervised-token counts/fractions per domain.
- Combine named per-domain validation losses into a `weighted` metric and use it as the default best-checkpoint metric.
- Reject unvalidated combinations early: sequence packing, pipeline parallelism, context parallelism, JSON blends, implicit unweighted blends, and non-sum losses.
- Document the multi-domain CPT configuration and its current support boundary.

# Validation

- `pytest -q tests/unit_tests/training/test_domain_mixture.py tests/unit_tests/loss/test_masked_ce.py tests/unit_tests/loss/test_chunked_ce.py tests/unit_tests/loss/test_linear_ce.py tests/unit_tests/loss/test_te_parallel_ce.py tests/unit_tests/recipes/test_train_ft.py tests/functional_tests/llm_pretrain_and_kd/test_domain_mixture_objective.py` — 140 passed.
- Real two-rank Gloo DDP gradient and SGD-step parity against a single-process weighted reference.
- Real CUDA fused cross-entropy loss and gradient parity against PyTorch.
- `ruff check .` and `ruff format --check .`.
- `lint-imports --no-cache` — component import contract kept.
- `make -C docs/fern docs-check` — 0 errors.

**Pre checks**:

# Additional Information

- Addresses the use case raised in [discussion #2940](https://github.com/NVIDIA-NeMo/Automodel/discussions/2940).


docs/guides/llm/pretraining.mdx
jgerh:
```suggestion
NeMo AutoModel provides an end-to-end recipe to run LLM pretraining with Hugging Face–native models and Megatron Core-style datasets.
```

jgerh:
```suggestion
This guide explains how to pretrain OpenAI’s `GPT2-124M` model on a FineWeb-Edu subset of 10 billion tokens.
```

jgerh:
```suggestion
[FineWeb-Edu](https://huggingface.co/datasets/HuggingFaceFW/fineweb-edu) is a dataset consisting of 1.3T tokens of educational web pages filtered from the larger [FineWeb](https://huggingface.co/datasets/HuggingFaceFW/fineweb) dataset. The educational web pages were filtered from the main dataset using a fine-tuned [BERT](https://huggingface.co/docs/transformers/en/model_doc/bert)-like classifier. You can find more information about the filtering process in the [FineWeb-Edu blog post](https://huggingface.co/spaces/HuggingFaceFW/blogpost-fineweb-v1).
```

jgerh:
```suggestion
The following example shows the data format:
```

jgerh:
```suggestion
This guide uses the FineWeb-Edu 10BT sample, which is a collection of approximately 10 billion tokens randomly drawn from the full FineWeb-Edu dataset. To prepare the data, run the following commands:
```

jgerh:
```suggestion

```bash
```

jgerh:
```suggestion

NeMo AutoModel provides a tool that tokenizes data and saves it in the Megatron Core dataset format. Use the tool as follows:
```

jgerh:
```suggestion

```bash
```

jgerh:
```suggestion
Set `--workers` to the number of CPU cores to use for parallel tokenization.
```

jgerh:
```suggestion
- `setup()`: Builds all training components from the config.
```

jgerh:
```suggestion
- `run_train_validation_loop()`: Executes training and validation steps.
```

jgerh:
```suggestion
To add weights to the dataset blends, pass a list. For example, `paths: ["30", "fineweb_edu/megatron_gpt2/processed_data_0_text_document", "70", "fineweb_edu/megatron_gpt2/processed_data_1_text_document"]`.
```

docs/guides/llm/pretraining.mdx
```diff
@@ -312,6 +312,68 @@ If you want to add weights to the dataset blends, you can do so by passing in a
...
+
+The `sampling_weight` values must match the explicit weights in
+the `MegatronPretraining` dataset's `dataset.paths`, and domain order must
+match the blend order. JSON blend files and implicit unweighted blends are not
+supported. Sequence packing is not supported with `domain_mixture` because
+packed batches do not currently preserve per-token domain IDs. Pipeline
+parallelism is also not currently supported because its target contract does
+not carry per-token objective weights. Context parallelism is disabled until
+its distributed gradient contract is covered by parity tests. The configured
+loss must use `reduction: sum`; the built-in masked, fused-linear, chunked, and
```
jgerh:
```suggestion
loss must use `reduction: sum`. The built-in masked, fused-linear, chunked, and
```

docs/guides/llm/pretraining.mdx
```diff
@@ -312,6 +312,68 @@ If you want to add weights to the dataset blends, you can do so by passing in a
...
+Transformer Engine parallel cross-entropy losses accept these objective
+weights.
+
 ## Load Large Models
 In distributed training, the typical model-loading pipeline has each GPU load the entire model and then retain only the shard it needs. This approach becomes problematic when the model size exceeds the memory capacity of a single GPU. For instance, a 70B-parameter model requires about 140GB of memory for its parameters when using the BF16 data type (2 bytes per parameter). Since most widely used GPUs are limited to 80GB, the full model cannot be directly loaded onto a single device.
```
jgerh:
```suggestion

In distributed training, the typical model-loading pipeline has each GPU load the entire model and then retain only the shard it needs. This approach becomes problematic when the model size exceeds the memory capacity of a single GPU. For example, a 70B-parameter model requires about 140 GB of memory for its parameters when using the BF16 data type (2 bytes per parameter). Because most widely used GPUs are limited to 80 GB, the full model cannot be loaded directly onto a single device.
```

docs/guides/llm/pretraining.mdx
jgerh:
```suggestion
In these scenarios, you can pass `is_meta_device: true` in the model config. The model is then instantiated using [PyTorch's Meta device](https://docs.pytorch.org/docs/stable/meta.html), which loads no data but stores the parameter metadata necessary for sharding the model. After the model is sharded, only the weights required by each model shard are loaded.
```

jgerh:
```suggestion
If you saved or plan to use the provided configuration at `examples/llm_pretrain/megatron_pretrain_gpt2.yaml`, run the following command:
```

jgerh:
```suggestion
During training, stepwise logs report loss, memory usage, and tokens per second. The recipe saves checkpoints under the configured `checkpoints/` directory.
```

jgerh:
```suggestion
As training progresses, you should observe the model loss beginning to converge. To verify your results, compare your convergence curves with the baseline benchmarks in the [llm.c repository](https://github.com/karpathy/llm.c/discussions/481).
```

jgerh:
```suggestion
This example demonstrates how to pretrain an LLM by using the NeMo AutoModel library. It uses the LLM [training recipe](https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/recipes/llm/train_ft.py), specifically `TrainFinetuneRecipeForNextTokenPrediction`. This recipe orchestrates loading, dataset preparation, optimizer setup, distributed training, checkpointing, and logging.
```

jgerh:
```suggestion
A recipe in NeMo AutoModel is a self-contained orchestration module that wires together all
components needed to perform a specific task, such as pretraining.
It is equivalent to a `Trainer` class but is highly modular, stateful, and reproducible.
```

jgerh, line 312:
```suggestion

<Tip>
```

jgerh:
```suggestion
The following is a complete configuration based on `examples/llm_pretrain/megatron_pretrain_gpt2.yaml`:
```

jgerh (commented):
Completed the tech pubs review of docs/guides/llm/pretraining.mdx and provided a few copyedits.

yuhezhang-ai (commented):
The LLM pretraining/finetuning recipe is a reasonable scope for this feature. I left six inline comments: the ignore-index integration needs a fix before merge; the domain-order guard, reduction contract, test portability, and runnable example/setup coverage are worth addressing; aggregate-name reuse is a small nonblocking cleanup.

Validation: 22 focused CPU tests passed, one failed because the optional fused-loss symbol was absent, and one CUDA test was skipped. Small CPU reproductions also confirmed the ignore-index crash/normalization mismatch, path-substring false positive, and missing-reduction guard issue. GPU/distributed tests were not rerun.

nemo_automodel/components/loss/masked_ce.py
```diff
@@ -80,9 +96,13 @@ def forward(
         if isinstance(labels, DTensor):
             labels = labels.full_tensor()
 
-        loss = F.cross_entropy(logits, labels, reduction=self.reduction)
+        reduction = "none" if loss_weights is not None else self.reduction
+        loss = F.cross_entropy(logits, labels, ignore_index=self.ignore_index, reduction=reduction)
```
yuhezhang-ai: **Blocking: make the ignore-index contract consistent.**

Honoring `self.ignore_index` exposes two integration problems:

- With `ignore_index=0`, existing collator padding and default MTP targets containing `-100` now raise `IndexError: Target -100 is out of bounds`. MTP's ignore index is configurable, but defaults to `-100` independently of the loss.
- Training and validation still normalize using `labels != -100` (`train_ft.py:1408,1603`). With labels `[[0,1],[0,2]]`, the denominator is 4 instead of 2, halving the loss and gradients.

Please align label construction, MTP masking, and normalization with the loss contract, or keep this shared behavior change out of this PR. Add regression coverage for both cases.

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -182,6 +183,85 @@ def _maybe_downgrade_loss_fn(loss_fn: nn.Module, probe_module: nn.Module, pp_ena
...
+    Names need not be derived from paths, so this only fires on evidence of an
+    actual mismatch: a name that matches some *other* position's prefix but not
+    its own. Names that match no prefix at all are unverifiable and pass.
+    """
+    if not isinstance(blend_prefixes, (list, tuple)) or len(blend_prefixes) != len(domain_mixture.names):
+        return
+    lowered = [str(prefix).lower() for prefix in blend_prefixes]
+    for position, name in enumerate(domain_mixture.names):
+        key = name.lower()
+        matches = [index for index, prefix in enumerate(lowered) if key in prefix]
```
yuhezhang-ai: **Highly recommended: remove path-substring inference from domain-order validation.**

This rejects correctly ordered configurations. For domains `[code, web]` and paths `[/data/python/train, /data/web/encoded_text_document]`, `code` matches the substring in `encoded`, causing setup to fail. I reproduced this rejection.

Since names need not derive from paths, substring matches cannot establish the mapping. Please preserve the documented positional contract, or validate an explicit binding, and add this false-positive case.

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -541,6 +622,49 @@ def setup(self):
...
+            if self.cfg.dataloader.emits_thd or self.cfg.dataloader.packing is not None:
+                raise ValueError(
+                    "domain_mixture does not support sequence packing because packed batches do not preserve "
+                    "per-token dataset_id provenance"
+                )
+            if not _supports_loss_weights(self.loss_fn):
+                raise ValueError(
+                    f"domain_mixture requires a loss function that accepts loss_weights, got {type(self.loss_fn).__name__}"
+                )
+            if getattr(self.loss_fn, "reduction", "sum") != "sum":
```
yuhezhang-ai: **Highly recommended: require an explicit sum-reduction contract.**

Defaulting a missing `reduction` attribute to `"sum"` allows a custom mean-reduced loss through both guards if it names a `loss_weights` parameter. A reproduction through `LossFromFactoryConfig` passes validation and produces one-quarter of the expected loss for four tokens when the mean is divided by the supplied global token count.

Please reject missing reduction metadata rather than assuming sum, and test missing, mean, and valid sum cases.

tests/unit_tests/loss/test_linear_ce.py
```diff
@@ -250,3 +250,61 @@ def _fake_linear_ce(hidden, weight, targets=None, **kwargs):  # noqa: D401,E501
...
+    from nemo_automodel.components.loss import linear_ce as linear_ce_mod
+
+    monkeypatch.setattr(linear_ce_mod, "HAVE_CUT_CROSS_ENTROPY", True)
+    captured = {}
+
+    def _fake_linear_ce(hidden, weight, targets=None, **kwargs):
+        captured["reduction"] = kwargs["reduction"]
+        return torch.tensor([[1.0, 2.0], [3.0, 4.0]])
+
+    monkeypatch.setattr(linear_ce_mod, "linear_cross_entropy", _fake_linear_ce)
```
yuhezhang-ai: **Test fix: keep this unit test independent of the optional fused-loss package.**

This test fails with `AttributeError` when `cut_cross_entropy` is unavailable because `linear_cross_entropy` is never bound. Please use `monkeypatch.setattr(..., raising=False)`, matching the neighboring normalization test. This was the failing test in the focused CPU run.

docs/guides/llm/pretraining.mdx
```diff
@@ -307,31 +310,111 @@ lr_scheduler:
...
 </Tip>
 
+### Balance Sampling and Domain Objectives
+
+Sampling weights control how often a domain appears in a batch. They do not
+need to be the same as the domain's desired contribution to the optimization
+objective. Configure both values with `domain_mixture`:
+
+```yaml
+dataset:
```
yuhezhang-ai, line 330: **Highly recommended: add a runnable example and setup smoke test.**

Please add a minimal `domain_mixture` YAML under `examples/` and a focused test that exercises configuration resolution through the recipe's setup validation.

This documentation snippet is partial, and the new tests exercise configuration parsing and individual helpers separately. A setup smoke test should cover a valid mixture and rejection of unsupported loss reductions, packing, and parallelism settings. This would verify that users can reach the supported training path from the documented configuration.

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -755,6 +879,17 @@ def materialize_loader(config):
             )
             for name, dl_config in self.cfg.validation_dataloaders.items()
         }
+        if self.domain_mixture is not None and self.val_dataloaders:
+            if "weighted" in self.val_dataloaders:
```
yuhezhang-ai: **Nonblocking cleanup: reuse `WEIGHTED_AGGREGATE_NAME`.**

`domain_mixture.py` already defines the reserved aggregate name, but this recipe repeats `"weighted"` for validation-name checks, the default checkpoint metric, logger creation, and aggregate logging. Please import and reuse the constant at those sites so the reservation and consumers stay consistent. The current values agree; this is a small maintainability cleanup.

nemo_automodel/recipes/llm/train_ft.py
akoumpa: Setup now requires an explicit `reduction == "sum"`; missing reduction and `mean` are both rejected and covered by the YAML setup smoke tests.

nemo_automodel/components/loss/masked_ce.py
akoumpa: Removed the shared non-default MaskedCrossEntropy change and its tests from #3303, so this PR keeps the existing `-100` contract. The complete Masked CE, MTP, and normalization fix is split into #3897.

tests/unit_tests/loss/test_linear_ce.py
akoumpa: Updated the optional fused-loss monkeypatch to use `raising=False`.

nemo_automodel/recipes/llm/train_ft.py
akoumpa: Reused `WEIGHTED_AGGREGATE_NAME` for the reserved validation key, default checkpoint metric, logger lookup, and aggregate metric logging.

nemo_automodel/recipes/llm/train_ft.py
akoumpa: Removed path-substring inference. Domain names are positional labels, and the docs and regression test now make that contract explicit.

docs/guides/llm/pretraining.mdx
akoumpa: Added a complete runnable GPT-2 domain-mixture YAML and a setup smoke test covering the valid configuration plus unsupported reduction, sequence packing, context parallelism, and pipeline parallelism.

yuhezhang-ai:
@akoumpa Thanks for the updates! Splitting the loss fix into #3897 makes sense, and the other review comments look addressed. My focused checks passed.

I noticed one small issue in the new example: training uses `split: "1,0,0"` and validation uses `"0,1,0"` on the same dataset prefixes, so both select the full corpus.

Could you use the same split ratio, such as `"0.99,0.01,0"`, in all three blocks, keeping their respective `splits_to_build` values? That would give validation a held-out partition.

With that correction and green CI, I’m happy to approve.

akoumpa:
@yuhezhang-ai Updated in d14a51688: the training and both validation blocks now use the same `0.99, 0.01, 0` split, while retaining their respective `train` and `validation` `splits_to_build` values. This keeps validation held out. The focused example setup tests pass.

yuhezhang-ai (approved):

PR #3290, opened 2026-07-29
fix(pp): use static metadata with PyTorch 2.13

yuhezhang-ai:
# What does this PR do?

Restore static pipeline metadata on PyTorch 2.13 when the replacement
`_user_meta` API is available. This prevents PyTorch's dynamic real-forward
metadata probe from executing GLM stages with uninitialized, value-sensitive
top-k carry tensors and failing in `scatter_`.

The original fallback remains in place for PipelineStage APIs that expose
neither supported static metadata interface.

# Changelog

- Recognize both legacy `_configure_outputs_meta` and PyTorch 2.13 `_user_meta`
  stage metadata APIs.
- Preserve dynamic metadata inference as the fallback for unsupported future
  APIs.
- Add regression coverage for the dynamic fallback and GLM-style hidden/top-k
  carry metadata.

# Validation

- Exact failing runtime: PyTorch `2.13.0a0+8145d630e8.nv26.06`, Transformers
  `5.12.1`: 118 passed (Slurm job `14572939`, exit `0:0`).
- Rebased affected suite: 74 passed, 44 skipped.
- Ruff format/check and `git diff --check` pass.
- Original 32-node GLM reproducer:
  [scoped job 378419119](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/378419119)
  ([pipeline 60195855](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/60195855))
  passed all 50 training steps plus validation with static pipeline shapes
  (`SLURM_STATE=COMPLETED`, 961s); no `ScatterGatherKernel`,
  index-out-of-bounds, device-side assert, or traceback occurred.

# Before your PR is "Ready for review"

- [x] Read and followed the contributor guidelines.
- [x] Added the necessary regression tests.
- [x] Documentation is not required for this internal compatibility fix.
- [x] All commits include DCO sign-off.

# Additional Information

- Related to [AMINT-208](https://linear.app/nvidia/issue/AMINT-208/out-of-bounds-index-access-in-scattergatherkernel-causing-device-side).
- The stale guard originated in the PyTorch 2.13 container upgrade after the
  `_user_meta` compatibility path had landed on `main`.

PR #3318, opened 2026-07-30
fix(glm_moe_dsa): keep the PP top-k carry gradient dense for P2P

athitten:
# What does this PR do ?

Fixes `ValueError: Tensors for P2P must be non-overlapping and dense` on every non-last pipeline stage of the GLM-5 PP benchmark recipes (AMINT-210).

# Why / root cause

GLM MoE DSA threads its non-differentiable IndexShare top-k selection across pipeline stages as a float32 carry, keeping it in the autograd graph with a zero-weight link so `torch.distributed.pipelining` gets a defined gradient:

```python
hidden = hidden + (carry_in.float().sum() * 0.0).to(hidden.dtype)
```

`sum()`'s backward returns `grad.expand(carry_in.shape)` — a **stride-0, overlapping** view. `PipelineStage` reports each stage's input-gradient metadata (shape, **stride**, dtype) upstream, and the previous stage allocates its backward recv buffer with `torch.empty_strided(shape, stride)`. A stride-0 gradient makes that buffer overlapping, so NCCL rejects it on the first `RECV_B`.

This was always malformed; it was masked until torch 2.13 by the static-metadata path, which derives grad metas from the stage's (contiguous) *outputs* rather than the downstream stage's real gradients. #2983 disabled that path on torch >= 2.12, exposing the bug. Independent of #3257 restoring static metadata — the gradient is wrong in either mode, so this fix is needed regardless.

# The fix

Scale before reducing, via a documented `_zero_with_dense_grad()` helper used at all three carry sites: `(carry_in * 0.0).sum()`. The multiply's backward allocates a fresh contiguous zero tensor, so the reported stride is dense.

|  | grad stride | P2P buffer |
|---|---|---|
| `x.float().sum() * 0.0` | `(0, 0, 0)` | rejected |
| `(x * 0.0).sum()` | `(12, 4, 1)` | accepted |

Values are unchanged, and the forward is slightly cheaper than the previous fp32 upcast of the full hidden state.

# Validation

- **nemo-ci performance pipeline (16 + 128 nodes, eos): https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/60154318** — `glm5_lora` and `glm_5_te_deepep` both green, full benchmark summaries, zero P2P errors.
- Single-node A/B (8xH100, scaled-down GLM-5, pp4/ep2, interleaved1f1b, same CI container): pristine `main` reproduces the identical traceback; this branch runs 5 clean steps.
- New regression test fails on `main` with `grad(carry_in) stride (0, 0, 0) is not dense`, passes here. Full `glm_moe_dsa` suite: 154 passed.

# Changelog

- `_zero_with_dense_grad()` in `glm_moe_dsa/model.py`; used for the carry-in links on the last and non-last stages and the carry-out link from `hidden`.
- `test_pp_stage_carry_in_gradient_is_dense`, parametrized over last / non-last stage.

# Pre checks

- [x] Added a regression test
- [ ] Documentation — n/a

# Additional Information

- Fixes AMINT-210
- Related: #2983 (exposed it), #3257 (restores the static-metadata path)

### DIFF
--- a/nemo_automodel/components/models/glm_moe_dsa/model.py
+++ b/nemo_automodel/components/models/glm_moe_dsa/model.py
@@ -44,6 +44,25 @@ from nemo_automodel.components.utils.model_utils import squeeze_input_for_thd
 from nemo_automodel.shared.utils import dtype_from_str as get_dtype
 
 
+def _zero_with_dense_grad(tensor: torch.Tensor) -> torch.Tensor:
+    """Return a ``0.0`` scalar that keeps ``tensor`` in autograd with a *dense* gradient.
+
+    Scaling first makes the multiply's backward allocate a fresh contiguous zero tensor.
+    Reducing first (``tensor.sum() * 0.0``) instead hands back a stride-0 expanded gradient,
+    and since the pipeline's backward P2P recv buffer is allocated from that reported stride,
+    NCCL fails with ``ValueError: Tensors for P2P must be non-overlapping and dense``.
+
+    Args:
+        tensor: Float tensor of arbitrary shape; only its autograd edge, shape, and
+            dtype are used.
+
+    Returns:
+        0-dim tensor holding ``0.0`` with ``tensor``'s dtype, connected to ``tensor``
+        so that ``tensor`` receives a contiguous all-zero gradient.
+    """
+    return (tensor * 0.0).sum()
+
+
 class Block(nn.Module):
     def __init__(self, layer_idx: int, config: GlmMoeDsaConfig, moe_config: MoEConfig, backend: BackendConfig):
         super().__init__()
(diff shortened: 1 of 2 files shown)


athitten:
Just noticed PR #3290 which fixed a CI bug for glm 5.2 and it might be fixing the bug mentioned in this PR for glm 5. Checking if this PR is still needed. Hence converted to draft

athitten:
This PR not needed, glm5_lora and glm_5_te_deepep with latest main run fine. Here's the gitlab pipeline: [https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/379819029](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/379819029)

PR #3290 fixes the issue

Issue #3325, opened 2026-07-31
[bug] MoE + expert parallel + `gradient_checkpoint full` fails at step 0: CheckpointError from HybridEP dispatch token-count drift


NancyFyong:
## Problem

Any MoE model trained with `--fsdp.ep_size > 1` and the `hybridep` dispatcher
fails on the **first** backward pass with:

```
torch.utils.checkpoint.CheckpointError: torch.utils.checkpoint: Recomputed values
for the following tensors have different metadata than during the forward pass.
```

The mismatched dimension is the **permuted token count** after expert dispatch —
forward and recompute disagree by a few tokens, so the grouped-GEMM activations
change shape between the two passes.

This is the default configuration of every shipped flagship MoE recipe, so those
recipes cannot run as written:

| recipe | ep_size | AC |
|---|---|---|
| `examples/scripts/slurm/sft_qwen3_6_35b.sh:57,116` | 8 | `full` |
| `examples/scripts/quick_start/sft_qwen3_6_35b.sh:43,78` | 8 | `full` |
| `examples/scripts/slurm/sft_omni3_30b.sh:62,76` | 8 | `full` |
| `examples/scripts/slurm/rl_qwen3_6_35b.sh:42,47` | 8 | `full` |
| `examples/scripts/slurm/rl_glm5_2_753b.sh:42,45` | 256 | `full` |

### The existing mitigation does not prevent it

`molt/trainer/fsdp/strategy.py:242-262` sets `ignore_router_for_ac=True`
specifically to avoid this error, on the theory that recomputing the router
re-routes near-tie tokens and shifts per-expert counts.

Instrumentation shows that mitigation **works as designed and is still not
sufficient**:

1. The router matcher fires. `_is_router_projection` in
   `nemo_automodel/components/moe/parallelizer.py:459-477` does match the router
   projection, which is therefore marked `MUST_SAVE`:
   ```
   [DBG-AC] want hidden=256 experts=8 | aten.mm.default args=((56, 256), (256, 8)) MATCHED=True
   ```
2. Routing is byte-identical across forward and recompute. Probing right after
   `weights, indices, aux_loss = self.gate(...)` in
   `nemo_automodel/components/moe/layers.py`, the same layer object (`id=`) yields
   identical token counts, index sums, masks, and per-expert bincounts:
   ```
   forward   id=...157750 ntok=58 isum=309 mask=57 counts=[0,55,0,1,49,11,0,0]
[...]

PR #3328, opened 2026-07-31
fix(fsdp): resolve fp32 master-weight compute dtype per parameter

### LINKED ISSUE #3327: [bug] Custom MoE + fp32 master weights: bf16/fp32 dtype mismatch crash in forward
## Problem

Custom MoE models trained with **fp32 master weights** (`--fsdp.param_dtype bf16`  <br>with fp32 parameter storage) crash in the **first forward pass** with:

```
RuntimeError: expected mat1 and mat2 to have the same dtype, but got:
c10::BFloat16 != float
```

The failing matmul is the Qwen3.5-MoE linear-attention `in_proj_qkv` (`nn.Linear`,  <br>text hidden 2048 -> 8192): the incoming activation is bf16 but the weight computes  <br>in fp32. Reproduced with LoRA SFT on `Qwen3.6-35B-A3B`  <br>(`Qwen3_5MoeForConditionalGeneration`), `ep_size=4`, and reproduces under every  <br>`gradient_checkpoint` mode (`none`/`selective`/`full`) — it is a plain forward  <br>crash, not an activation-checkpoint issue.

This is **distinct from** NVIDIA-NeMo/Automodel#3325: NVIDIA-NeMo/Automodel#3325 is the post-dispatch `CheckpointError`  <br>that appears only with `gradient_checkpoint full` *after* this crash is resolved.  <br>With `full`, this dtype crash fires first; both must be fixed for the shipped  <br>flagship MoE recipes to run.

## Root cause

`_make_compute_dtype_fn` in  <br>`nemo_automodel/components/distributed/parallelizer_utils.py` resolves each  <br>parameter's *compute* dtype by precedence: (1) pinned fp32  <br>(`_keep_in_fp32_modules_strict`), (2) `tensor._hf_compute_dtype` recorded at  <br>checkpoint load, (3) a fallback.

1. **Custom models never populate precedence** [#2](<https://github.com/NVIDIA-NeMo/Automodel/issues/2>)**.** Registered architectures are  <br>built through the custom-model path in `_transformers/model_init.py`, which  <br>returns the constructed module *before* weight loading and never calls  <br>`_restore_loaded_model_dtype`. So `tensor._hf_compute_dtype` is never set and  <br>**every** parameter of a custom model falls through to the [#3](<https://github.com/NVIDIA-NeMo/Automodel/issues/3>) fallback.
2. **The fallback used a module-global uniformity flag.** It applied the policy  <br>compute dtype (bf16) to unhinted fp32 storage only when *all* floating params  <br>in the module shared one storage dtype (`storage_is_uniform`).
3. **A single bf16 sibling flips that flag.** In a Qwen3.5-MoE decoder layer with  <br>fp32 master weights, every floating param is fp32 storage **except**  <br>`mlp.shared_expert_gate.weight`, which is constructed bf16. So the module's  <br>storage set is `{float32, bfloat16}` and `storage_is_uniform=False`.

   Runtime probe of the layer holding `in_proj_qkv`:

   ```
   storage_is_uniform=False  floating_storage_dtypes={torch.float32, torch.bfloat16}
   linear_attn.in_proj_qkv.weight   storage=torch.float32  pinned=False  hf=None
   mlp.shared_expert_gate.weight    storage=torch.bfloat16  pinned=False  hf=None
   ```
4. With the flag `False`, the fallback returns each param's own storage dtype, so  <br>the fp32 master weights (`in_proj_qkv`, `out_proj`, ...) are classified as  <br>**fp32 compute** instead of bf16.
5. NVIDIA-NeMo/Automodel#3255 (`preserve ERNIE router and dense Qwen3.5 SSM precision`) set  <br>`cast_forward_inputs=False` on fp32 FSDP units. The misclassified-fp32  <br>`in_proj_qkv` unit therefore no longer casts its incoming bf16 activation, and  <br>the `bf16 @ fp32` matmul crashes. Before NVIDIA-NeMo/Automodel#3255 this misclassification only  <br>caused harmless extra fp32 compute; NVIDIA-NeMo/Automodel#3255 turned it into a hard failure.

## Suggested fix

Resolve the [#3](<https://github.com/NVIDIA-NeMo/Automodel/issues/3>) fallback **per parameter** instead of via a module-global flag: an  <br>unhinted fp32 storage under a lower-precision policy is an fp32 master weight and  <br>should compute in the policy dtype; any other storage keeps its own dtype. A  <br>single genuinely-bf16 sibling (`shared_expert_gate`) then no longer forces the  <br>layer's fp32 master weights into fp32 compute. The `(storage, compute)` grouping  <br>already keeps each FSDP unit storage-uniform, so `_init_mp_dtypes` is unaffected.

A PR implementing this follows.

NancyFyong:
## Summary

Fixes NVIDIA-NeMo/Automodel#3327. Custom MoE models with fp32 master weights (`--fsdp.param_dtype bf16`  <br>with fp32 parameter storage) crash in the **first forward pass** with a `bf16 @ fp32`  <br>matmul at the Qwen3.5-MoE linear-attention `in_proj_qkv`.

## Root cause

`_make_compute_dtype_fn` resolves each parameter's *compute* dtype by precedence:  <br>(1) pinned fp32, (2) `tensor._hf_compute_dtype` recorded at load, (3) a fallback.

* Custom architectures build via the custom-model path in `_transformers/model_init.py`,  <br>which returns the module **before** weight loading and never calls  <br>`_restore_loaded_model_dtype`, so `_hf_compute_dtype` is never set — every parameter  <br>hits the [#3](<https://github.com/NVIDIA-NeMo/Automodel/issues/3>) fallback.
* The fallback keyed on a **module-global** `storage_is_uniform` flag. A single  <br>native-bf16 sibling (`mlp.shared_expert_gate.weight`) makes the layer's storage set  <br>`{float32, bfloat16}`, flipping the flag false, so **all** the fp32 master weights  <br>are classified as fp32 compute.
* Since NVIDIA-NeMo/Automodel#3255 set `cast_forward_inputs=False` on fp32 FSDP units, the misclassified  <br>fp32 `in_proj_qkv` no longer casts its incoming bf16 activation, and the matmul  <br>crashes. Before NVIDIA-NeMo/Automodel#3255 this misclassification was only harmless extra fp32 compute.

## Fix

Resolve the [#3](<https://github.com/NVIDIA-NeMo/Automodel/issues/3>) fallback **per parameter**: unhinted fp32 storage under a lower-precision  <br>policy is an fp32 master weight and computes in the policy dtype; any other storage keeps  <br>its own dtype. A single bf16 sibling no longer forces the layer's fp32 master weights into  <br>fp32 compute. The `(storage, compute)` grouping already keeps each FSDP unit  <br>storage-uniform, so `_init_mp_dtypes` is unaffected. Net −9 lines.

## Testing

LoRA SFT on `Qwen3.6-35B-A3B` (`Qwen3_5MoeForConditionalGeneration`), `ep_size=4`,  <br>`param_dtype=bf16`, fp32 master weights:

* **Before:** `RuntimeError: expected mat1 and mat2 to have the same dtype ... BFloat16 != float`  <br>at the first forward, under every `gradient_checkpoint` mode.
* **After:** forward passes. With `gradient_checkpoint=none`, training runs end-to-end  <br>(2 epochs; loss falls 0.28 → < />1e-5, grad-norm and aux-loss healthy).

> Note: with `gradient_checkpoint=full` the forward now passes but the **separate**  <br>`CheckpointError` tracked in NVIDIA-NeMo/Automodel#3325 (post-dispatch token-count drift) is hit next.  <br>That is out of scope for this PR.

---

## Follow-up (per reviewer feedback)

Two Qwen3.5-MoE-specific gaps the per-parameter resolver exposes, plus a regression test:

* **Pin** `_fp32_params` **compute.** The MoE parallelizer keys the fp32 compute pin on  <br>`_keep_in_fp32_modules_strict`, which `Qwen3_5MoeForConditionalGeneration` never set,  <br>so the SSM-gating `A_log`/`dt_bias` fell to the policy dtype (bf16). Added the strict  <br>entry, matching the `inkling`/`kimi_k3` siblings.
* **Build** `shared_expert_gate` **in** `config.dtype`**.** It was constructed at the  <br>`initialize_linear_module` default (bf16) even under fp32 master-weight storage — the  <br>lone bf16 sibling the fallback above has to tolerate. Now built in `config.dtype`.  <br>(Shared `moe/layers.py`, so the fix also covers the other MoE families using it.)
* **Regression test** for the mixed-storage, no-HF-hint layout (fp32 masters + a  <br>genuinely-bf16 `shared_expert_gate` + the fp32 holder): ordinary weights and the gate  <br>compute bf16; only the pinned `A_log`/`dt_bias` stay fp32.


yuhezhang-ai:
@NancyFyong  Thanks for tracking this down and contributing the per-parameter fix! The approach looks good. Qwen3.5-MoE just needs two small contract updates for it to be safe:

- Construct `shared_expert_gate` with `dtype=config.dtype`; it currently defaults to BF16 even with FP32 master-weight storage.
- Add `_fp32_params` to `_keep_in_fp32_modules_strict`, so `A_log`/`dt_bias` continue to compute in FP32.

The resolver already prioritizes the strict contract, so no further resolver changes should be needed.

Could you also add a focused regression test showing that under an FP32-storage/BF16-compute policy:

- ordinary weights and `shared_expert_gate.weight` compute in BF16;
- `_fp32_params.A_log` and `dt_bias` compute in FP32?

NancyFyong:
> @NancyFyong Thanks for tracking this down and contributing the per-parameter fix! The approach looks good. Qwen3.5-MoE just needs two small contract updates for it to be safe:
> 
> * Construct `shared_expert_gate` with `dtype=config.dtype`; it currently defaults to BF16 even with FP32 master-weight storage.
> * Add `_fp32_params` to `_keep_in_fp32_modules_strict`, so `A_log`/`dt_bias` continue to compute in FP32.
> 
> The resolver already prioritizes the strict contract, so no further resolver changes should be needed.
> 
> Could you also add a focused regression test showing that under an FP32-storage/BF16-compute policy:
> 
> * ordinary weights and `shared_expert_gate.weight` compute in BF16;
> * `_fp32_params.A_log` and `dt_bias` compute in FP32?

Thank you for suggestion! I'll add a focused regression test in this pr.

NancyFyong:
> @NancyFyong Thanks for tracking this down and contributing the per-parameter fix! The approach looks good. Qwen3.5-MoE just needs two small contract updates for it to be safe:
> 
> * Construct `shared_expert_gate` with `dtype=config.dtype`; it currently defaults to BF16 even with FP32 master-weight storage.
> * Add `_fp32_params` to `_keep_in_fp32_modules_strict`, so `A_log`/`dt_bias` continue to compute in FP32.
> 
> The resolver already prioritizes the strict contract, so no further resolver changes should be needed.
> 
> Could you also add a focused regression test showing that under an FP32-storage/BF16-compute policy:
> 
> * ordinary weights and `shared_expert_gate.weight` compute in BF16;
> * `_fp32_params.A_log` and `dt_bias` compute in FP32?

Hi @yuhezhang-ai, I update this pr. Please take a look when you have time.

yuhezhang-ai:
@NancyFyong Thanks, the latest update fixes the original Qwen3.5/3.6 issue and LGTM!

I found one additional affected sibling while checking the generalized fallback: `Qwen3NextForCausalLM` uses the same intrinsic-FP32 `_fp32_params` holder, but still declares it only in `_keep_in_fp32_modules`. Without a strict declaration, its `A_log`/`dt_bias` will now compute in BF16.

Could you add `_fp32_params` to Qwen3-Next’s `_keep_in_fp32_modules_strict` as well?

The new resolver test currently passes `("_fp32_params",)` directly, so it cannot catch a model forgetting to declare the contract. Could you also add a model-owned assertion covering the strict declarations for both Qwen3.5-MoE and Qwen3-Next?

I’m also happy to push this small follow-up to your branch if you prefer.

NancyFyong:
> @NancyFyong Thanks, the latest update fixes the original Qwen3.5/3.6 issue and LGTM!
> 
> I found one additional affected sibling while checking the generalized fallback: `Qwen3NextForCausalLM` uses the same intrinsic-FP32 `_fp32_params` holder, but still declares it only in `_keep_in_fp32_modules`. Without a strict declaration, its `A_log`/`dt_bias` will now compute in BF16.
> 
> Could you add `_fp32_params` to Qwen3-Next’s `_keep_in_fp32_modules_strict` as well?
> 
> The new resolver test currently passes `("_fp32_params",)` directly, so it cannot catch a model forgetting to declare the contract. Could you also add a model-owned assertion covering the strict declarations for both Qwen3.5-MoE and Qwen3-Next?
> 
> I’m also happy to push this small follow-up to your branch if you prefer.

Thanks for your review — I'll add it as you requested.

yuhezhang-ai:
/ok to test [87102b8](https://github.com/NVIDIA-NeMo/Automodel/pull/3328/commits/87102b88cf98e65dbe53b06dc43b57f2c6c98bcb)

yuhezhang-ai:
/ok to test [e0c6d9c](https://github.com/NVIDIA-NeMo/Automodel/pull/3328/commits/e0c6d9c48e3b346783b2664b494b5dc274da53c9)

yuhezhang-ai (approved):

PR #3330, opened 2026-07-31
fix(distributed): stabilize SAC replay with TE and FSDP

akoumpa:
## Summary

- restore Transformer Engine's attention-backend cache to its forward-entry state during activation-checkpoint recomputation
- keep FSDP2 parameter-lifecycle communication, allocation, layout, and copy ops outside selective-AC saved-op replay accounting
- centralize SAC ignore-set mutation in `ignore_sac_ops(...)`, with profiler and FSDP callers supplying their operator lists
- add focused regressions for TE cache-dependent op sequences and recompute-only FSDP unshard setup

## Validation

- `python -m pytest -q tests/unit_tests/distributed/test_activation_checkpointing.py tests/unit_tests/moe/test_parallelizer.py` (`105 passed`)
- `ruff format .` and `ruff check --fix .`
- [nemo-ci job 380509607](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/380509607): `nemotron_ultra_v3_squad` completed 17/17 steps on 256 H100 GPUs; Slurm `COMPLETED`, exit code 0

The failure was reproduced on current `main` in [job 380347397](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/380347397) before applying the fix.

## Tracking

- [AMINT-32](https://linear.app/nvidia/issue/AMINT-32/missing-saved-tensor-for-checkpointed-atenne-operation-causes)

### DIFF
--- a/nemo_automodel/components/moe/parallelizer.py
+++ b/nemo_automodel/components/moe/parallelizer.py
@@ -465,14 +465,19 @@ def apply_ac(
             from nemo_automodel.components.distributed.activation_checkpointing import (
                 SELECTIVE_AC_WRAPPER_FLAG,
                 make_selective_checkpoint_context_fn,
+                transformer_engine_attention_backend_snapshot_context_fn,
             )
 
             selective_context_fn = make_selective_checkpoint_context_fn()
+            attention_context_fn = functools.partial(
+                transformer_engine_attention_backend_snapshot_context_fn,
+                selective_context_fn,
+            )
             for parent_layers, layer_id, block in iter_transformer_and_mtp_blocks(model):
                 block = ptd_checkpoint_wrapper(
                     block,
                     preserve_rng_state=True,
-                    context_fn=_preserve_gate_load_during_recompute(block, selective_context_fn),
+                    context_fn=_preserve_gate_load_during_recompute(block, attention_context_fn),
                 )
                 # Tag so _apply_per_layer_compile compiles the wrapper OUTER (keeping the
                 # selective policy visible to the partitioner) instead of unwrapping and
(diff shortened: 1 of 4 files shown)


nemo_automodel/components/moe/parallelizer.py
```diff
@@ -465,14 +465,19 @@ def apply_ac(
             from nemo_automodel.components.distributed.activation_checkpointing import (
                 SELECTIVE_AC_WRAPPER_FLAG,
                 make_selective_checkpoint_context_fn,
+                transformer_engine_attention_backend_snapshot_context_fn,
             )
 
             selective_context_fn = make_selective_checkpoint_context_fn()
+            attention_context_fn = functools.partial(
```
yuhezhang-ai, line 472: **[P1] Apply the TE snapshot to the shared selective-AC path.** This composition is only added in the MoE parallelizer. `apply_selective_checkpointing_to_layers()` still passes `make_selective_checkpoint_context_fn()` directly to `checkpoint_wrapper`, so dense FSDP2/DDP models using TE attention plus selective AC enter recomputation with the post-forward TE cache and can dispatch the same extra comparison op that this PR is fixing. Please compose the snapshot in the generic selective wrapper (without double-wrapping the MoE path) and add a regression through `apply_selective_checkpointing_to_layers()` rather than calling the snapshot helper directly.

tests/unit_tests/distributed/test_activation_checkpointing.py
```diff
@@ -133,6 +133,103 @@ def test_snapshot_context_fn_recompute_ctx_restores_captured_backend_set():
...
+
+    def policy(ctx, func, *args, **kwargs):
+        if func == torch.ops.aten.ne.Tensor:
+            return CheckpointPolicy.MUST_SAVE
+        return CheckpointPolicy.PREFER_RECOMPUTE
+
+    def selective_context_fn():
+        return create_selective_checkpoint_contexts(policy)
+
+    def attention(x: torch.Tensor) -> torch.Tensor:
```
yuhezhang-ai: **[P3] Add tensor-layout contracts to the new test helpers.** The new `attention(x: torch.Tensor)` helper accepts and returns tensors but has no `Args`/`Returns` layout documentation; `checkpointed_module` below has the same issue. Please add the required Google-style contracts, for example: `x: Tensor of shape [features].` and `Returns: Tensor of shape [features].`

yuhezhang-ai (changes requested):
Requesting changes for one functional coverage gap. The TE cache snapshot is only composed in the MoE activation-checkpoint path, so the generic dense FSDP2/DDP selective-AC path can still reproduce the same cache-dependent replay mismatch. I also noted the required tensor-layout documentation on the new test helpers.

yuhezhang-ai (approved):

PR #3337, opened 2026-07-31
fix(tokenizer): make special token insertion opt-in

akoumpa:
## Summary

- stop forcing `add_bos_token=True` and `add_eos_token=True` in the default NeMo tokenizer wrapper
- preserve the underlying HuggingFace tokenizer's native BOS/EOS behavior by default
- retain `add_bos_token=True` and `add_eos_token=True` as explicit opt-ins
- update tokenizer documentation and compatibility tests to reflect the new contract

## Why

`NeMoAutoTokenizerWithBosEosEnforced.from_pretrained()` defaulted both flags to `True`, overriding tokenizers that intentionally disabled BOS or EOS insertion. This changed token sequences even when callers did not request the behavior.

After this change, NeMo only enables the additional insertion when the caller opts in. Tokenizers that natively add special tokens continue to do so.

## Validation

- `pytest -q tests/unit_tests/_transformers/test_auto_tokenizer.py -k 'not test_converts_tiktoken_tokenizer'` — 57 passed
- isolated TikToken conversion regression test — 1 passed
- `pytest -q tests/functional_tests/data/llm/test_save_pretrained_v4_compat.py` — 5 passed
- Ruff checks passed for all changed files
- `git diff --check` passed

### DIFF
--- a/examples/llm_finetune/baichuan/baichuan_2_7b_mock_fp8.yaml
+++ b/examples/llm_finetune/baichuan/baichuan_2_7b_mock_fp8.yaml
@@ -46,6 +46,8 @@ tokenizer:
   _target_: nemo_automodel.NeMoAutoTokenizer.from_pretrained
   pretrained_model_name_or_path: baichuan-inc/Baichuan2-7B-Chat
   trust_remote_code: true
+  add_bos_token: true
+  add_eos_token: true
 
 checkpoint:
   enabled: true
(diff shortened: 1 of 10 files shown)


yuhezhang-ai:
Changing the tokenizer default to preserve Hugging Face behavior makes sense, but the NeMo-owned recipes that relied on the previous implicit insertion need migration.

In particular, the three Baichuan SQuAD recipes do not set `add_bos_token` or `add_eos_token`. Baichuan’s native settings disable both, and `format_prompt_completion()` does not append an EOS separately. After this change, those recipes will stop supervising a terminal EOS token.

Could we add explicit BOS/EOS settings to the affected recipes, or add a focused test confirming that removing EOS is intentional?

yuhezhang-ai (approved):

PR #3354, opened 2026-08-01
feat(dllm): add SCDD training and sampling

kashif:
Adds SCDD (Self-Correcting Discrete Diffusion) as a dLLM mode.
Paper: https://openreview.net/forum?id=zQKlzKB6I9

The forward process mixes absorbing [MASK] noise with uniform token
transitions, so the model trains on contexts holding wrong but plausible
tokens and learns to overwrite them. That self-correction is what sustains
quality when many positions are decoded per step.

What is in here:

* `corrupt_mix` mixed mask + uniform forward kernel
* `SCDDSchedule` and `SCDDLoss` (discrete-time NELBO). The ELBO needs
  p_theta over the full vocab so it cannot use a fused CE, so the vocab
  reduction is chunked and checkpointed to keep the fp32 intermediates off
  the backward tape.
* `SCDDStrategy`, registered as `mode: scdd`
* `SCDDSampler` in `examples/dllm_generate/generate.py`
* `examples/dllm_sft/llada_scdd.yaml` as the reference recipe

Schedule defaults are taken from the authors' released checkpoint config
(T=1000, ratio 0.1, gamma 1, t_peak 0.5). Setting `uniform_ratio: 0` reduces
the objective back to MDLM exactly, which there is a test for.

Tests cover the kernel, schedule, loss (including chunking equivalence and a
DTensor case), strategy, sampler and the shipped recipe, plus a single GPU
functional smoke.

Two things to flag: context parallelism is not supported and
`SCDDStrategy.setup_extra` raises on `cp_size > 1`, and the authors keep an
EMA of the weights that the dLLM recipe does not currently wire, so expect
somewhat noisier samples than their released checkpoints.


kashif:
cc @laaaarrywang would you be able to review as well?

docs/guides/dllm/finetune.mdx
jgerh:
```suggestion
| **1. Install** | [Install NeMo AutoModel](#install-nemo-automodel) | Install the package with `uv` or Docker |
```

docs/guides/dllm/finetune.mdx
```diff
@@ -103,6 +105,55 @@ dataset:
...
+dataset:
+  unshifted: true
+```
+
+The schedule values above are taken from the config shipped with the authors' released
+`scdd_pu_0.1` checkpoint. Their only other released setting differs solely in `uniform_ratio: 0.2`,
+so that is the first knob to sweep.
+
+Decode SCDD checkpoints with `--sampler scdd`, passing the same `uniform_ratio`, `schedule_shape`,
+and `schedule_peak` used during training — the sampler rebuilds the reverse posterior from that
```
jgerh:
```suggestion
and `schedule_peak` used during training. The sampler rebuilds the reverse posterior from that
```

docs/guides/dllm/finetune.mdx
```diff
@@ -141,6 +192,12 @@ The following table describes the key configuration fields for dLLM fine-tuning.
 | `dllm.block_length` | Diffusion block size for the I-DLM block-diffusion mask (paper curriculum 1→2→3). I-DLM mode only |
 | `dllm.clean_loss_weight` | Fixed `α` on the clean-copy verification CE (paper `0.2`). I-DLM mode only |
 | `dllm.auto_balance_clean_loss` | Replaces `α` with `(CE_noisy/CE_clean).detach()` (paper Equation 2, b3 stage). I-DLM mode only |
```
jgerh:
```suggestion
| `dllm.block_length` | Diffusion block size for the I-DLM block-diffusion mask (paper curriculum 1→2→3). I-DLM mode only. |
| `dllm.clean_loss_weight` | Fixed `α` on the clean-copy verification CE (paper `0.2`). I-DLM mode only. |
| `dllm.auto_balance_clean_loss` | Replaces `α` with `(CE_noisy/CE_clean).detach()` (paper Equation 2, b3 stage). I-DLM mode only. |
```

docs/guides/dllm/finetune.mdx
jgerh:
```suggestion
The existing `nemo_automodel.components.speculative.regenerate` script handles this process. Start an SGLang server that hosts the target model, and then regenerate the assistant turns.
```

jgerh:
```suggestion
The recipe uses a **strategy pattern** to handle differences between model families. The `dllm.mode` field in the YAML configuration selects the strategy.
```

jgerh:
```suggestion
Alternatively, use the prebuilt Docker container.
```

jgerh:
```suggestion
See [`qwen3_8b_idlm.yaml`](https://github.com/NVIDIA-NeMo/Automodel/blob/main/examples/dllm_sft/qwen3_8b_idlm.yaml) for the full working configuration. The following example shows the key I-DLM-specific sections.
```

jgerh (commented):
Completed tech pubs review of docs/guides/dllm/finetune.mdx and provided a few copyedits.

nemo_automodel/components/datasets/dllm/corruption.py
```diff
@@ -151,6 +152,105 @@ def corrupt_all_masked(
...
+    transitioned &= input_ids != mask_token_id
+
+    # Uniform over the vocabulary minus {mask_token_id, input_ids}: draw over
+    # vocab_size - 2 slots, then shift past each excluded id in sorted order.
+    draw = torch.randint(0, vocab_size - 2, (B, L), device=device, dtype=input_ids.dtype, generator=generator)
+    mask_id_t = torch.full_like(input_ids, mask_token_id)
+    lo = torch.minimum(input_ids, mask_id_t)
+    hi = torch.maximum(input_ids, mask_id_t)
+    draw = draw + (draw >= lo).to(draw.dtype)
+    draw = draw + (draw >= hi).to(draw.dtype)
```
khazic, line 259: When the clean token already is `mask_token_id`, `lo == hi`, so a draw at or above that id gets shifted twice and the result can reach `vocab_size`, which is out of range. It is currently harmless only because `transitioned &= input_ids != mask_token_id` guarantees those entries are never selected by the `torch.where` below. That is an implicit dependency between two separated statements; a `clamp_max(vocab_size - 1)` or an explicit comment would keep a future edit from turning it into an out-of-bounds token id.

nemo_automodel/components/datasets/dllm/corruption.py
```diff
@@ -151,6 +152,105 @@ def corrupt_all_masked(
...
+        "different non-``[MASK]``" replacement); they may still be absorbed,
+        which leaves them unchanged and hence out of ``noise_mask``.
+    """
+    if vocab_size < 3:
+        raise ValueError(f"corrupt_mix requires vocab_size >= 3 (got {vocab_size})")
+
+    B, L = input_ids.shape
+    device = input_ids.device
+
+    mask_prob = mask_prob.reshape(B, 1).to(torch.float32)
```
khazic, line 233: The docstring says `mask_prob + uniform_prob` must not exceed 1 but nothing checks it. When it does, `transitioned`'s upper bound silently clips against the `u < 1` range and the realised uniform mass is lower than the schedule asked for, which shows up as a quiet mismatch with the ELBO weights rather than an error.

nemo_automodel/recipes/dllm/strategy.py
```diff
@@ -200,6 +203,144 @@ def prepare_batch(self, batch, noisy_input_ids, noise_mask, clean_input_ids):
...
+            raise ValueError("SCDDStrategy.create_loss_fn must run before corruption (it captures dllm.vocab_size).")
+
+        batch = input_ids.shape[0]
+        # t ~ U(eps, 1) snapped onto the discrete grid {1/T, ..., 1}: SCDD is
+        # derived in discrete time, and the ELBO weights compare t against the
+        # previous grid point s = t - 1/T.
+        u = torch.rand((batch,), device=input_ids.device, generator=generator)
+        u = (1.0 - eps) * u + eps
+        t = ((u * self._num_timesteps).to(torch.int64).float() + 1.0) / self._num_timesteps
+        t = t.clamp(0.0, 1.0 - 1e-4)
```
khazic: The clamp takes `t` off the `1/T` grid that the two lines above just snapped it onto: the top bucket becomes 0.9999, and `SCDDLoss` then computes `s = t - 1/T` assuming both land on grid points. Effect is small, but it contradicts the comment about the discrete grid. Clamping to `1.0 - 1.0 / num_timesteps` would stay on the grid.

khazic:
Heads up: the DCO check is failing. CONTRIBUTING requires every commit to carry a `Signed-off-by` line, so this needs a `git rebase --signoff` before it can be merged.

laaaarrywang:
> cc @laaaarrywang would you be able to review as well?

Thanks for adding SCDD to NeMo AutoModel! I went through the SCDD-specific implementation and overall it looks consistent with our released implementation.

One small documentation issue: the `SCDDSampler` docstring says that a visible token can be “sent back to [MASK]”. In SCDD, already-visible tokens are not remasked. The implementation itself already seems correct (`numer[:, self.mask_id] = 0`), so I think this is just a wording issue.

kashif:
thanks both.

@khazic all three fixed in 9b0070e: corrupt_mix now raises when mask_prob + uniform_prob exceeds 1, the uniform draw is clamped so a clean [MASK] cannot produce an out-of-range id, and the top diffusion time clamps to the previous grid point instead of 1 - 1e-4 so t and s both stay on the 1/T grid. That last one also means T=1 is no longer usable, so SCDDLoss requires num_timesteps >= 2 now. Tests added for each. DCO also fixed, the branch is rebased with sign-off.

@laaaarrywang you are right, the docstring was wrong and the code was not. Reworded in b5d3845 to say a visible token is never sent back to [MASK].

HuiyingLi (dismissed):

examples/dllm_generate/generate.py
```diff
@@ -532,6 +555,179 @@ class IDLMSampler(DLLMSampler):
...
+            # A posterior with no reachable state (possible only for a degenerate
+            # schedule, e.g. uniform_ratio=0) keeps the current token.
+            keep = torch.zeros_like(rows).scatter(-1, current[:, None], 1.0)
+            rows = torch.where(rows.sum(-1, keepdim=True) > 0, rows, keep)
+            x = x.clone()
+            x[canvas] = torch.multinomial(rows, 1).squeeze(-1)
+
+        # Residual [MASK] positions cannot survive into the output; the final
+        # denoise reads them off the model's own distribution.
+        leftover = canvas & (x == self.mask_id)
```
zyzhou5: Based on agent's review against the reference: 

Final denoise only repairs [MASK], so injected noise reaches the output. The chain ends at t = cfg.eps (L664), where ~0.70% of canvas positions still hold uniform-channel noise. The cleanup here skips them: leftover = canvas & (x == self.mask_id)   # only the 0.099% masked

The reference's noise-removal pass argmaxes every position and is on by default (noise_removal: True). It generates unconditionally, so its "every position" is this sampler's canvas; the canvas restriction here is right, but & (x == self.mask_id) has no counterpart there and narrows the pass from "remove noise" to "fill blanks."

zyzhou5 (commented):
Overall LGTM

zyzhou5:
/ok to test [a82ca3c](https://github.com/NVIDIA-NeMo/Automodel/pull/3354/commits/a82ca3cd1aa6f50c6af910b6204bb1dbbd13a319)

zyzhou5:
/ok to test [598a16f](https://github.com/NVIDIA-NeMo/Automodel/pull/3354/commits/598a16f6b9057ad6c22ffe42a4bcb70161931335)

HuiyingLi (dismissed):

HuiyingLi:
Thanks @kashif ! Could you please take a look at the ci err? appreciate it!

zyzhou5:
/ok to test [f54b348](https://github.com/NVIDIA-NeMo/Automodel/pull/3354/commits/f54b34877592864010709dc5ea0b959b62c03583)

zyzhou5 (approved):

PR #3369, opened 2026-08-02
perf(checkpoint): reduce distributed save overhead

akoumpa:
# What does this PR do ?

Reduces metadata, planning, and synchronization overhead in distributed exact-resume checkpoint saves without changing checkpoint payloads or restore semantics.

# Changelog

- Create shared DCP checkpoint directories only on process-group rank zero, while preserving per-rank creation for node-local auxiliary state, then synchronize the group once the directories are visible.
- Give model and optimizer DCP save planners separate cache namespaces so alternating saves do not evict each others cached plans.
- Collapse async checkpoint publication from three distributed barriers to one after rank-zero pointer updates and retention pruning.
- Add focused coverage for group-rank directory creation, planner cache separation, and the reduced finalization barrier sequence.

# Performance impact

The shared-filesystem metadata path was measured in a controlled 8-rank A/B on cw-dfw Lustre. Each mode ran eight trials of 25 checkpoint setups with four directory groups per checkpoint.

| Metric | Before | This PR | Improvement |
| --- | ---: | ---: | ---: |
| `mkdir` calls per directory on 8 ranks | 8 | 1 | 87.5% fewer |
| `mkdir` calls for four directory groups | 32 | 4 | 87.5% fewer |
| Median Lustre directory setup per checkpoint | 7.89 ms | 3.06 ms | 61.3% lower; 2.58x faster |
| Async publication barriers | 3 | 1 | 66.7% fewer |
| Alternating model/optimizer plan-cache hit rate after warmup | 0% | 100% | Cross-eviction eliminated |
| Checkpoint tensor payload | Unchanged | Unchanged | No format or exact-resume tradeoff |

The planner-cache result comes from alternating model and optimizer plans through the same sequence used by checkpoint saves. With the shared cache namespace, all four post-warmup requests missed; with separate namespaces, all four hit and sent zero local plan items.

These are checkpoint control-plane improvements. This PR does not reduce the approximately 93 GiB exact-resume payload in the motivating Qwen3-8B configuration, so an end-to-end training TPS improvement during background checkpoint writes is not claimed yet. A full model baseline-vs-PR run is still needed to quantify checkpoint drain time, checkpoint-boundary step latency, and sustained TPS while I/O is active.

# Validation

- `219 passed, 5 skipped`: checkpoint, optimizer-state, and async-finalization tests.
- The suite includes a true two-process Gloo DCP optimizer save/restore round trip with restored-state equality.
- Ruff format and lint checks passed for all changed Python files.
- `git diff --check` passed.

**Pre checks**:

- [x] Make sure you read and followed the Contributor guidelines.

# Additional Information

This intentionally leaves checkpoint tensor payloads unchanged. Distributed optimizer state and exact resume remain intact; the change only removes redundant shared-filesystem metadata operations, cache thrashing, and intermediate barriers.

### DIFF
--- a/nemo_automodel/components/checkpoint/checkpointing.py
+++ b/nemo_automodel/components/checkpoint/checkpointing.py
@@ -591,7 +591,7 @@ class Checkpointer:
         should_write_consolidated = _should_write_consolidated_safetensors(self.config, is_final_checkpoint)
         consolidated_dir = os.path.join(model_dir, "consolidated") if should_write_consolidated else None
         hf_metadata_dir = os.path.join(model_dir, ".hf_metadata") if _should_write_hf_metadata(self.config) else None
-        _ensure_dirs(model_dir, consolidated_dir, hf_metadata_dir, process_group=self.process_group)
+        _ensure_shared_dirs(model_dir, consolidated_dir, hf_metadata_dir, process_group=self.process_group)
 
         # Because this call lies outside of the dcp save call, we need to consolidate on all ranks on the main process
         # of all ranks, which lies on the critical path. Therefore, we can only do this outside of async mode.
@@ -719,7 +719,7 @@ class Checkpointer:
                 per-model-part optimizers.
         """
         optimizer_path = os.path.join(weights_path, "optim")
-        _ensure_dirs(optimizer_path, process_group=self.process_group)
+        _ensure_shared_dirs(optimizer_path, process_group=self.process_group)
         optimizer_state = OptimizerState(
             model,
             optimizer,
@@ -1383,7 +1383,7 @@ class Checkpointer:
         DTensor metadata and writes all shards correctly.
         """
         state_dir = os.path.join(path, state_name)
-        _ensure_dirs(state_dir, process_group=self.process_group)
+        _ensure_shared_dirs(state_dir, process_group=self.process_group)
         state_dict = state.state_dict()
         planner = dcp.DefaultSavePlanner(enable_plan_caching=True)
         process_group = getattr(self, "process_group", None)
@@ -1438,8 +1438,8 @@ class Checkpointer:
         Returns:
             The populated state dictionary (may be replaced for PEFT).
         """
-        # Both model and optimizer saving is done in this function
-        is_model = True if "/model" in path else False
+        # Both model and optimizer loading is done in this function.
+        is_model = _is_model_checkpoint_path(path)
         # PEFT loading is broadcasted from rank0 so it is a special case
         if self.config.is_peft and is_model and (not is_init_step):
             state_dict = _load_safetensors(_adapter_path(path))
@@ -1467,8 +1467,8 @@ class Checkpointer:
         Returns:
             Optional Future object if async mode is enabled.
         """
-        # Both model and optimizer saving is done in this function
-        is_model = True if "/model" in path else False
+        # Both model and optimizer saving is done in this function.
+        is_model = _is_model_checkpoint_path(path)
         # PEFT saving is done on rank0 so it is a special case
         if self.config.is_peft and is_model:
             if not torch.distributed.is_initialized() or torch.distributed.get_rank(group=self.process_group) == 0:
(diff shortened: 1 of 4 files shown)


nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -1918,11 +1927,13 @@ def _ensure_dirs(*dirs: Optional[str], process_group: torch.distributed.ProcessG
         *dirs: One or more directory paths that should exist.
         process_group: Ranks that must observe the directories before continuing.
     """
-    for d in dirs:
-        if d:
-            if not is_cloud_path(d):
+    is_dist_initialized = torch.distributed.is_initialized()
+    is_group_rank_0 = not is_dist_initialized or torch.distributed.get_rank(group=process_group) == 0
+    if is_group_rank_0:
```
HuiyingLi: This breaks multi-node checkpointing when the checkpoint path is node-local (for example, `/tmp`). `_ensure_dirs` now creates the directory only on process-group rank 0, but `save_on_dp_ranks` subsequently calls `torch.save` on every TP/PP rank 0. On other nodes, the barrier does not make rank 0's local directory visible.

I reproduced this in the PR container with a 2-node, 2-rank GPU Slurm job (job `14930820`, one rank per node), saving to `/tmp/.../rng`:

- rank 0 on `pool0-01774`: save succeeded
- rank 1 on `pool0-01898`: `RuntimeError: Parent directory /tmp/am_pr3369_14930820_pr/rng does not exist.`

In an A/B run within the same job, the immediate pre-PR `_ensure_dirs` behavior saved successfully on both nodes.

nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -1470,7 +1478,8 @@ def _do_save(
             return
 
         ret = None
-        planner = dcp.DefaultSavePlanner(enable_plan_caching=True)
+        planner_cls = _ModelSavePlanner if is_model else _OptimizerSavePlanner
```
HuiyingLi, line 1491: Planner selection still relies on the substring-based `is_model = "/model" in path`. An optimizer checkpoint under a parent directory such as `/opt/models/qwen/step_100/optim` is therefore classified as a model save, because `/models` contains `/model`. Both the model and optimizer saves then use `_ModelSavePlanner`, so they share the same plan-cache namespace and the cache separation introduced here does not take effect for these paths.

nemo_automodel/components/checkpoint/checkpointing.py
akoumpa: i'll fix, but I'm wondering if this a realistic scenario, for example, i'd expect if different nodes where writing on different dirs (ie., non-shared storage) to hang.

HuiyingLi:
I found a reproducible cross-`Checkpointer` cache correctness regression at the current head (`131db9cb8`).

A concrete failing sequence in one Python process is:

```python
# First run/checkpointer: torch_save
torch_ckpt._do_save(model_state, "run_a/step_100/model")
torch_ckpt._do_save(optimizer_state, "run_a/step_100/optim")

# Second run/checkpointer: same model tensor layout, but safetensors
safe_ckpt._do_save(
    model_state,
    "run_b/step_100/model",
    safe_ckpt._get_storage_writer(None, None, None, "run_b/step_100/model"),
)
```

This is a realistic pattern for notebooks, in-process recipe runners/services, or a pytest worker that constructs multiple checkpointers sequentially.

On the merge base (`e7c73a734`) this sequence succeeds. On this PR it fails during the second model save with:

```text
TypeError: argument of type '_StoragePrefix' is not iterable
```

The root cause is that DCP's planner caches are process-global and `DefaultSavePlanner._cached_plans_key` is derived from the planner class name. The new fixed `_ModelSavePlanner` key survives the intervening optimizer save and is also reused by the second `Checkpointer`. Since the tensor layout is unchanged, it reports a cache hit and `finish_plan()` returns the stale final plan created by the torch-save writer. That plan contains a filesystem `_StoragePrefix`, while `_HuggingFaceStorageWriter.write_data()` expects dictionary storage data.

The new unit test mocks `dcp.save`, so it only proves that the model and optimizer class keys differ; it cannot exercise writer plan preparation/reuse and misses this failure.

Could we scope the cache namespace per `Checkpointer` (with separate model/optimizer suffixes), or otherwise include the complete writer/topology context? The namespace needs to remain stable across saves by one checkpointer and serialize with the planner into the async subprocess. An unmocked regression test covering the sequence above would catch this.

HuiyingLi (approved):

PR #3371, opened 2026-08-03
feat(datasets): support SynthTraces agent SFT

### LINKED ISSUE #2422: Support Agent SFT with agent trace dataset
Extend current tc training to popular agent trace dataset such as https://huggingface.co/datasets/julien-c/synthtraces
First generate traces on nemo repos, then perform agent sft on the generated traces.

khazic:
## Summary

- preserve native OpenAI `tool_calls`, `tool_call_id`, tool names, and reasoning content in agent chat datasets
- add an optional Hugging Face dataset revision so raw agent trace repositories can use generated Parquet rows
- optionally filter incomplete traces that contain no assistant supervision
- trim unfinished trailing user turns without dropping completed assistant targets
- filter traces that remain overlong after whole-exchange history truncation
- add a generation-marked Qwen2.5 template that supervises reasoning and tool calls directly
- keep the training and validation dataset slices mutually exclusive
- add a Qwen2.5-3B SFT recipe for `julien-c/synthtraces`
- document how to generate Pi traces on NeMo repositories before SFT
- cover valid, missing-ID, empty, malformed, incomplete, overlong, and real-shaped OpenAI records with unit tests

## Root cause

`AgentChatDataset` handled Swift and ShareGPT tool-call roles, but treated OpenAI assistant messages as plain text. It dropped embedded `tool_calls` and the matching IDs on tool responses, so SynthTraces did not supervise tool use. The dataset also stores raw Pi event streams on `main`; its parsed session rows are published under `refs/convert/parquet`. Some parsed traces exceed the sequence budget even after exchange-level history truncation. Token-level left truncation changes rendered chat prefixes and makes answer-only masks unsafe, so the recipe filters those traces before lazy formatting. The stock Qwen2.5 template does not render `reasoning_content` or expose assistant masks, so this recipe supplies a generation-marked template for reasoning, assistant text, and tool calls.

## Validation

- `70 passed` in the focused agent chat unit-test suites on the Linux training server
- `ruff check` passed for the changed Python files
- `ruff format --check` passed for the changed Python files
- a real SynthTraces row produced `713` supervised tokens and included both `<think>` and `<tool_call>` spans
- the full dataset filter processed `2479` assistant-bearing traces in `3m57s` and retained `2270` trainable traces
- an 8-GPU Qwen2.5-3B smoke run completed one training step on SynthTraces with loss `1.0589` and `11720` supervised tokens

<details>
<summary>8-GPU SynthTraces smoke log</summary>

Command overrides:

```text
--nproc-per-node 8
--step_scheduler.max_steps 1
--step_scheduler.global_batch_size 8
--dataset.split train[:16]
--validation_dataset.split train[16:24]
```

Relevant output:

```text
World size: 8
Backend: nccl
Recipe: TrainFinetuneRecipeForNextTokenPrediction
Model name: Qwen/Qwen2.5-3B
Filter: 100%|██████████| 11/11 [00:00<00:00, 15.57 examples/s]
Max train steps: 1
step 0 | epoch 0 | loss 1.0589 | grad_norm 3.8713 | lr 1.00e-06 | mem 5.76 GiB | tps 1886.51(235.81/gpu) | num_label_tokens 11720
Training: 100%|██████████| 1/1 [00:17<00:00, 17.03s/step, loss=1.0589, lr=1.00e-06, tps=1887]
Successfully exported consolidated HF safetensors to checkpoints/epoch_0_step_0/model/consolidated.
Process exit code: 0
```

</details>

<details>
<summary>Full-dataset preprocessing log</summary>

```text
Filter: 100%|██████████| 2479/2479 [03:57<00:00, 10.43 examples/s]
FILTERED_ROWS 2270

real    4m7.091s
user    4m18.341s
sys     0m3.350s
Process exit code: 0
```

</details>

Closes NVIDIA-NeMo/Automodel#2422


docs/guides/llm/agent-sft.mdx
jgerh:
```suggestion
This guide explains how to fine-tune **Qwen2.5-3B** for **multi-turn agentic tool use** with [NeMo AutoModel](https://github.com/NVIDIA-NeMo/Automodel). The model receives tool definitions and a conversation that interleaves tool calls and responses. It learns to emit correct `tool_calls` (tool name and arguments) and use the returned results across several turns.
```

jgerh:
```suggestion
This is the **multi-turn** counterpart to the single-turn [Function Calling with FunctionGemma](/recipes-e2e-examples/function-calling) guide. That guide maps one user query to one set of tool calls using `make_xlam_dataset`. This guide handles full multi-turn agent traces (`user → tool_call → tool response → assistant → ...`) using `make_agent_chat_dataset`.

```

jgerh:
```suggestion
## Multi-Turn Agent SFT
```

jgerh:
```suggestion
A multi-turn agent trace contains the following elements in order:
```

jgerh:
```suggestion
- A **user** request
- One or more **assistant** turns that might emit `tool_calls` (with parallel calls in a single turn)
- **Tool** responses paired back to those calls
- A final **assistant** answer that uses the tool results
```

jgerh:
```suggestion
The training data is a set of such traces plus the **tool schema** available to the model. The dataset adapter renders each trace through the tokenizer's chat template with `answer_only_loss_mask=True`. Only assistant and tool-call tokens contribute to the loss. The adapter masks `user` and `tool` tokens.
```

jgerh:
```suggestion
## Define the Fine-Tuning Task
```

jgerh:
```suggestion
You fine-tune on the [llamafactory/glaive_toolcall_en](https://huggingface.co/datasets/llamafactory/glaive_toolcall_en) dataset, which contains ShareGPT function-calling traces.
```

jgerh:
```suggestion
The following table compares the target behavior with the base model's behavior:

| Model | Behavior |
```

docs/guides/llm/agent-sft.mdx
```diff
@@ -42,6 +42,47 @@ We fine-tune on the [llamafactory/glaive_toolcall_en](https://huggingface.co/dat
...
+
+```bash
+hf upload <owner>/<dataset> dataset . --repo-type dataset --exclude '.git/*'
+```
+
+Point the SynthTraces recipe at that dataset by overriding `dataset.dataset_name` and `validation_dataset.dataset_name`. The dataset must expose parsed OpenAI-format rows through its normal dataset revision or through `refs/convert/parquet`.
+
+<Note>
+Trace generation is intentionally maintained by the SynthTraces project. NeMo AutoModel consumes the generated traces and does not vendor the Pi runtime, llama.cpp server, or remote inference provider.
+</Note>
```
jgerh, line 160:
```suggestion

</Note>
```

docs/guides/llm/agent-sft.mdx
jgerh:
```suggestion
Review the following hardware requirements:

- **Full-parameter SFT**: The example below shards across 8 GPUs with FSDP2. Qwen2.5-3B can also train unsharded on a single 80 GB GPU because FSDP2 applies only across multiple GPUs.
- **PEFT with LoRA**: You can train on a single GPU. See [Use PEFT with LoRA](#use-peft-with-lora).
- **Dataset size**: The `glaive_toolcall_en` dataset contains a few thousand traces, so a full pass is fast.
```

jgerh:
```suggestion
## Set Up Your Environment
```

jgerh, line 181:
```suggestion
After the container starts, log in to Hugging Face and change to the repository directory:

```

jgerh, line 192:
```suggestion

</Tip>
```

jgerh:
```suggestion
## Explore the `glaive_toolcall` Dataset
```

jgerh:
```suggestion
The following output is illustrative:
```

jgerh:
```suggestion
```text
```

jgerh:
```suggestion
### Review How the Adapter Renders a Trace
```

jgerh:
```suggestion
`make_agent_chat_dataset` converts each trace to the OpenAI chat-completions format. It merges consecutive `function_call` turns into one assistant message with parallel `tool_calls` and pairs `observation` turns with those calls. The following example demonstrates this process and tokenizes the trace with `answer_only_loss_mask=True`:
```

jgerh:
```suggestion
    truncation=True,   # See the seq_length note in Configure Your Training.
```

jgerh:
```suggestion
The code prints the following output, including the fourth `___PAD_TOKEN_IDS___` key that the dataset injects:
```

jgerh:
```suggestion
Only the assistant and tool-call spans are supervised. The user prompt and tool responses are `-100` in `labels`.
```

jgerh:
```suggestion
## Evaluate the Base Model Before Fine-Tuning
```

jgerh:
```suggestion
Use the following code to render a held-out prompt **up to the point the model should call a tool**, pass the `tools` schema through the chat template, and generate the response:
```

jgerh:
```suggestion
Instead of a clean `get_stock_price(symbol="AAPL")` tool call, the base model answers in prose and fabricates content. Fine-tuning corrects this failure.
```

jgerh:
```suggestion
Greedy output is deterministic for a given model snapshot and Transformers version, so you might see slight variation from the trace above.

```

jgerh:
```suggestion
For a rigorous, repeatable score, NeMo AutoModel ships a generation-based **tool-call accuracy** evaluator (`nemo_automodel.components.eval.tool_call_evaluator.ToolCallAccuracyEvaluator`). It builds held-out prompts with `make_agent_chat_eval_samples`, parses generated tool calls with `nemo_automodel.components.eval.tool_call_parser`, and compares them with the ground-truth calls. The [Launch Fine-Tuning](#launch-fine-tuning) workflow runs it automatically during training.

```

jgerh:
```suggestion
## Configure Your Training
```

jgerh:
```suggestion
  output_hidden_states: true   # Required by FusedLinearCrossEntropy below.
```

jgerh:
```suggestion
# The generation-based tool-call accuracy evaluation runs at every validation step alongside val_loss.
```

jgerh:
```suggestion
The agent dataset uses `answer_only_loss_mask` to supervise only the tokens that the model must *produce* (assistant text and `tool_calls`). User turns and tool responses are masked to `-100`, so the model is never trained to "predict the environment." This behavior is enabled by default in `make_agent_chat_dataset`.

```

jgerh:
```suggestion
- `train_on_last_turn_only`: Supervises only the final assistant turn (`mask_history`).
- `mask_reasoning_content`: Renders assistant `reasoning_content` (thinking) into the prompt but excludes it from the loss.
- `drop_history_reasoning_content`: Strips prior-turn thinking from the prompt entirely to match inference. This option works best with `train_on_last_turn_only=true`.

```

jgerh:
```suggestion
`seq_length` only takes effect when **`truncation: true`** (to cap length) or `padding: max_length` (to pad). With the defaults (`truncation: false`, `padding: false`), `seq_length` is ignored, and long traces pass through uncapped. This behavior risks an out-of-memory error. Set `truncation: true` if you want a hard cap.

```

jgerh:
```suggestion
## Launch Fine-Tuning
```

jgerh:
```suggestion
### Monitor Training

Monitor the following metrics:
```

jgerh:
```suggestion
- **`tool_call/accuracy`** (and related `tool_call/*` keys) report generation-based tool-call correctness at each validation step. This metric distinguishes whether the model learned the format or selected the correct tools.
```

jgerh:
```suggestion
Under **FSDP2** (the default strategy here), the in-loop generation evaluation is skipped because generation with sharded weights is expensive. Set `tool_call_eval.run_on_fsdp2: true` to enable it. If you keep it disabled, evaluate tool-call accuracy from a saved checkpoint as described in [Evaluate the Fine-Tuned Model](#evaluate-the-fine-tuned-model).

```

jgerh:
```suggestion
Replace the placeholders in the following illustrative training log with values from your run:
```

jgerh:
```suggestion
```text
```

jgerh:
```suggestion
The example does not override checkpoint settings, so checkpoints are written under the default `checkpoints/` root with `save_consolidated: final`. Every Safetensors checkpoint includes `model/consolidate.sh`. Only the final checkpoint is exported inline to `model/consolidated/`. The root also contains `LATEST` and `LOWEST_VAL` symlinks.
```

jgerh:
```suggestion
## Evaluate the Fine-Tuned Model
```

jgerh:
```suggestion
Use the following code to load the generated Hugging Face-compatible directory and rerun the base-model evaluation prompt:
```

jgerh:
```suggestion
# Use the same messages and tools from the base-model evaluation.
```

jgerh:
```suggestion
The fine-tuned model should emit a structured call such as `get_stock_price` with `{"symbol": "AAPL"}`.
```

jgerh:
```suggestion
## Compare Results
```

jgerh:
```suggestion
Replace the placeholders in the following table with tool-call accuracy values from your run:
```

jgerh:
```suggestion
The numbers above are placeholders that show the expected *shape* of the result (low for the base model and high for the fine-tuned model). Replace them with values from the `tool_call/*` metrics in the training log or a checkpoint evaluation.

```

jgerh:
```suggestion
## Use PEFT with LoRA
```

docs/guides/llm/agent-sft.mdx
```diff
@@ -42,6 +42,47 @@ We fine-tune on the [llamafactory/glaive_toolcall_en](https://huggingface.co/dat
 | **Step 5** | Evaluate the fine-tuned model |
 | **Step 6** | Compare results |
 
+## Generate coding-agent traces on NeMo repositories
```
jgerh: Delete R33 to R44. Replace R45 with new heading.
```suggestion
## Generate Coding-Agent Traces on NeMo Repositories
```

docs/guides/llm/agent-sft.mdx
jgerh:
```suggestion
Use the ready-made config at [`examples/llm_finetune/agent/qwen2_5_3b_function_calling.yaml`](https://github.com/NVIDIA-NeMo/Automodel/blob/main/examples/llm_finetune/agent/qwen2_5_3b_function_calling.yaml). The configuration file contains the following key blocks:
```

jgerh (commented):
Completed tech pubs review of docs/guides/llm/agent-sft.mdx. Copyedited the guide and streamlined its structure by removing the redundant overview section and replacing numbered headings with descriptive, action-oriented headings.

jgerh (commented):
Completed tech pubs review of docs/guides/llm/agent-sft.mdx. Copyedited the guide and streamlined its structure by removing the redundant overview section and replacing numbered headings with descriptive, action-oriented headings.

nemo_automodel/components/datasets/llm/agent_chat.py
```diff
@@ -640,17 +764,41 @@ def make_agent_chat_dataset(
...
+                seq_length=seq_length + 1,
+                padding=False,
+                truncation=True,
+                mask_reasoning_content=mask_reasoning_content,
+                train_on_last_turn_only=train_on_last_turn_only,
+                drop_history_reasoning_content=drop_history_reasoning_content,
+                truncate_history=truncate_history,
+                trim_incomplete_last_turn=trim_incomplete_last_turn,
+            )
+            return len(tokenized["input_ids"]) < seq_length
```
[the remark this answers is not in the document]
khazic: Fixed: the probe renders with a seq_length + 1 budget, so an over-long trace comes back at seq_length + 1. Changed to <= and added a boundary case to the filter test.

docs/guides/llm/agent-sft.mdx
khazic: Applied. Note the other Note blocks in this file use the compact form without the blank line; happy to normalize them all if that is the house style.

khazic:
@jgerh applied all of your suggestions in 760a26834: task-style headings without the `Step N` prefixes, the guide-overview table dropped (it duplicated the headings once they were renamed), `Step 2/3/5` cross-references replaced with named anchors, `text` fences on the illustrative output blocks, and the prose rewrites as written. The `#peft-lora-variant` anchor in Hardware Requirements was updated to `#use-peft-with-lora` to follow the renamed heading.

@claude the off-by-one on the seq_length filter was fixed earlier in 7ad351f71: the predicate is now `<= seq_length`, so a trace that renders to exactly the budget is kept.

khazic:
Thanks for the thorough editorial pass. All 50 documentation suggestions are now reflected in `docs/guides/llm/agent-sft.mdx` as of 26080d22c.

Most of them had already been folded in when the guide was restructured after your review, which is why the threads show as outdated. The ones that were still missing are in this commit: active voice in the intro paragraph and the Note, the section rename to `Multi-Turn Agent SFT` and `Define the Fine-Tuning Task`, the lead-in sentences for the trace-element and hardware lists, labelled hardware list items, and the `Model` header on the behavior comparison table.

I verified each suggestion body appears verbatim in the current file rather than relying on the stale line anchors, and checked the page still has balanced code fences, balanced MDX components, and no dead in-page anchors.

Separately, the off-by-one note on `_chat_fits_sequence_length` was addressed earlier: both call sites render with a `seq_length + 1` budget and keep traces with `<= seq_length` tokens, with the boundary reasoning written into the comment at `agent_chat.py:798`.

Resolving these threads now.

docs/guides/llm/agent-sft.mdx
```diff
@@ -4,72 +4,110 @@ description: "End-to-end multi-turn, function-calling SFT on Qwen2.5-3B with the
...
+BATCH_SIZE=8 CONCURRENCY=2 node src/generate.ts
+```
+
+Upload the generated dataset with the current Hugging Face CLI:
+
+```bash
+hf upload <owner>/<dataset> dataset . --repo-type dataset --exclude '.git/*'
+```
+
+Point the SynthTraces recipe at that dataset by overriding `dataset.dataset_name` and `validation_dataset.dataset_name`. The dataset must expose parsed OpenAI-format rows through its normal dataset revision or through `refs/convert/parquet`.
```
HuiyingLi: Can we remove the Glaive path from this guide and make SynthTraces the single end-to-end story? This section tells readers to point the SynthTraces recipe at their dataset, but the rest of the page switches back to `llamafactory/glaive_toolcall_en` and launches `qwen2_5_3b_function_calling.yaml`; the new `qwen2_5_3b_synthtraces.yaml` is never linked or run. A reader who follows the guide will generate and upload traces, then train on a different dataset. Since this PR specifically adds SynthTraces support, I suggest updating dataset exploration, configuration, launch, and evaluation to use SynthTraces throughout and removing the Glaive/stock-price path.

docs/guides/llm/agent-sft.mdx
```diff
@@ -4,72 +4,110 @@ description: "End-to-end multi-turn, function-calling SFT on Qwen2.5-3B with the
...
+
+```bash
+npm install
+```
+
+```bash
+npm run pull
+```
+
+Start the OpenAI-compatible llama.cpp server used by the user model, then generate a small validation batch before scaling up:
```
HuiyingLi: This trace-generation workflow is not runnable end to end as written. The generator explicitly requires `HF_TOKEN`, relies on a llama.cpp server configured for `LOCAL_MODELS[0]`, and needs concrete `constants.ts`/`pull.ts` changes to support owner/repository pairs. Please include the required `HF_TOKEN` setup, the exact llama-server command and model, a copy-pastable repository-matrix change, and a quick check that the uploaded dataset exposes parsed `messages` rows before training. The upstream README currently does not fill in these gaps.

docs/guides/llm/agent-sft.mdx
```diff
@@ -230,41 +270,47 @@ tool_call_eval:
...
 ```
 
-### What to watch
+### Monitor Training
+
+Monitor the following metrics:
 
 - **`loss`** should fall steadily as the model learns the tool-call format.
-- **`tool_call/accuracy`** (and related `tool_call/*` keys) report generation-based tool-call correctness at each validation step — this is the metric that distinguishes "format learned" from "tools actually correct".
+- **`tool_call/accuracy`** (and related `tool_call/*` keys) report generation-based tool-call correctness at each validation step. This metric distinguishes whether the model learned the format or selected the correct tools.
```
HuiyingLi: The default 8-GPU FSDP2 run below will not emit this metric because `tool_call_eval.run_on_fsdp2` defaults to false, yet the tutorial asks readers to monitor it and the illustrative log includes it. Please make the evaluation path executable: either enable `run_on_fsdp2` in the documented configuration/command or provide the actual offline evaluator command against a consolidated checkpoint. Manually checking one Apple prompt later in the guide is not an accuracy evaluation.

docs/guides/llm/agent-sft.mdx
```diff
@@ -230,41 +270,47 @@ tool_call_eval:
...
 </Note>
 
 <Tip>
 `make_agent_chat_dataset` exposes three flags worth knowing for agent traces:
 
-- `train_on_last_turn_only` — supervise only the final assistant turn (`mask_history`).
-- `mask_reasoning_content` — render assistant `reasoning_content` (thinking) into the prompt but exclude it from the loss.
-- `drop_history_reasoning_content` — strip prior-turn thinking from the prompt entirely (matches inference). Most coherent together with `train_on_last_turn_only=true`.
+- `train_on_last_turn_only`: Supervises only the final assistant turn (`mask_history`).
+- `mask_reasoning_content`: Renders assistant `reasoning_content` (thinking) into the prompt but excludes it from the loss.
```
HuiyingLi: The SynthTraces recipe does more than supervise tool calls: its custom template places `reasoning_content` inside the `{% generation %}` block, so the default `mask_reasoning_content=false` trains the model to emit `<think>` traces. Please call this out in the SynthTraces section and document the choice to supervise or mask reasoning, including the expected inference behavior. Otherwise readers may reasonably assume this is ordinary answer-only tool-call SFT.

HuiyingLi (commented):
Four documentation issues found while validating the SynthTraces tutorial end to end. The implementation and dataset path look sound, but the guide should use one coherent SynthTraces workflow.

khazic:
Thanks for validating the tutorial end to end. All four points are addressed in 5f6700f1c and adb461b63. I checked each claim against the source before rewriting the text.

**Trace generation is now runnable (`agent-sft.mdx:59`).** I read the generator rather than paraphrasing it:

- `src/single-session.ts` raises `HF_TOKEN is not set` because the coding agent goes through the Hugging Face router, so the guide now exports the token.
- `src/constants.ts` reads `LLAMA_BASE_URL` and falls back to `http://localhost:8080/v1`, so the guide gives the exact `llama-server` command for that port and mentions the variable.
- `src/generate.ts` always picks `LOCAL_MODELS[0]`, so the guide names that entry instead of leaving the choice open.
- `src/pull.ts` builds every clone URL as `https://github.com/huggingface/<name>.git`. The guide now explains that this hard-coded owner is the reason the fork needs owner and repository pairs.

**`tool_call` metrics (`agent-sft.mdx:304`).** You are right, and the problem is worse than the default flag. The SynthTraces recipe declares no `tool_call_eval` block at all, so `tool_call/*` cannot appear for it under any setting. A 30-step run confirms it: the log carries `loss` and `val_loss` only. I removed the invented `tool_call/accuracy` values from the sample log, stated plainly that the documented FSDP2 command does not emit those keys, and added a warning in the SynthTraces section.

**Reasoning supervision (`agent-sft.mdx:281`).** Confirmed and quantified. On the published `julien-c/synthtraces` training split (2428 rows after this recipe's filters), 1537 rows (63.3%) carry reasoning content, and in 1424 of them the `<think>` span holds a real label rather than `-100`. The recipe therefore trains the model to emit `<think>` on most examples. The SynthTraces section now warns about this and documents `mask_reasoning_content` and `drop_history_reasoning_content`.

**One coherent workflow (`agent-sft.mdx:71`).** I fixed the incoherence but stopped short of deleting the Glaive path, and I would like your call on that. SynthTraces now has its own subsection with the recipe linked, its own launch command, an observed log, and its caveats, and the hardware section no longer describes Glaive while introducing SynthTraces. I kept Glaive because it is the only path in the guide that carries a tool schema and the generation-based accuracy evaluator, so it supplies the base-model comparison and the evidence that fine-tuning actually worked. Removing it would leave the guide with no quantitative outcome at all. If you would still prefer a SynthTraces-only page, I will cut Glaive and relocate the evaluator walkthrough.

**CI.** `validate-nightly-recipes` failed because the new recipe had no top-level `ci:` block. Added `recipe_owner`, `nodes`, and `time`. Both `validate_new_recipe_ci.py` and `validate_generate_ci.py` pass locally, and a rerun confirms the config loader ignores the new key: step-0 loss and final `val_loss` match the pre-change run to four significant digits.

**One finding outside your review, for your judgment.** 201 of those 2428 rows (8.3%) end up with every label masked, so they contribute nothing to the loss and each one logs a warning. Truncation is not the cause: none of them reach the 8192 cap. `filter_no_assistant` and the span the chat template marks with `{% generation %}` disagree about what counts as supervised. I left it out of this PR to keep the diff to docs and CI. Happy to open a separate issue, or to fold a filter fix in here if you would rather have it now.

Could you run `/ok to test adb461b63` when you have a moment?

HuiyingLi:
@khazic Thanks for the update. Is it possible to use vllm/sglang for inference instead of llama.cpp?

khazic:
Yes. Nothing in the generator is tied to llama.cpp. `src/single-session.ts` registers the user model as a generic OpenAI provider:

```ts
modelRegistry.registerProvider(USER_PROVIDER, {
	baseUrl: LLAMA_BASE_URL,
	apiKey: "no-key",
	api: "openai-completions",
	models: [{ id: modelId, name: modelId, ... }],
});
```

`api` is `openai-completions` and `baseUrl` comes from the `LLAMA_BASE_URL` environment variable, so any OpenAI-compatible server works. Pointing it at vLLM or SGLang needs no code change:

```bash
vllm serve Qwen/Qwen3.6-27B --port 8080
```

```bash
export LLAMA_BASE_URL=http://localhost:8080/v1
```

One catch worth flagging: the three entries in `LOCAL_MODELS` are GGUF repositories (`ggml-org/Qwen3.6-27B-GGUF:Q8_0` and the other two). vLLM's GGUF support is limited and SGLang does not serve GGUF, so switching engines also means replacing `LOCAL_MODELS[0]` with the unquantized repository and making sure the id the generator sends matches what the server advertises, for example via `--served-model-name`. The placeholder `apiKey: "no-key"` is fine for both, since neither requires a key by default.

I am happy to document the vLLM path next to llama.cpp in the guide so readers are not tied to one runtime. Say the word and I will add it in this PR.

Separately, one CI note: the `validate-recipe-configs` workflow run for adb461b63 is sitting at `action_required` rather than having executed, so the `ci:` block I added has only been checked by running `validate_new_recipe_ci.py` and `validate_generate_ci.py` locally. Could you approve that workflow run as well, so it gets exercised for real?

HuiyingLi:
If you could add the vllm path that'd be nice. Thanks!

khazic:
Went ahead and added it in c314f5a28, since it belongs to the same section this PR already rewrote and a separate PR would just conflict on the same lines.

The guide now keeps `llama-server` as the default and adds a tip covering the alternative: `src/single-session.ts` registers the user model as a generic OpenAI provider whose base URL is `LLAMA_BASE_URL`, so any OpenAI-compatible server drops in with no generator change. It shows the `vllm serve` equivalent plus the `LLAMA_BASE_URL` export, and calls out the two things that have to stay aligned, namely replacing the GGUF id in `LOCAL_MODELS[0]` with the unquantized repository and matching the served model name to that entry.

Could you run `/ok to test c314f5a28` when you get a chance, and approve the `validate-recipe-configs` workflow run so the `ci:` block gets exercised?

HuiyingLi:
Thanks for adding the vLLM/SGLang alternative. I tested the latest head (`c314f5a28`) with `Qwen/Qwen3.6-27B` on an H100 using vLLM 0.20.2.dev0. The model loaded successfully, `/v1/models` and a real completion request returned 200, and `single-session.ts` produced a JSONL file. However, I do not think the vLLM path is E2E-ready yet for two reasons:

1. The local user-model output is polluted by reasoning. All four generated follow-up user messages in the persisted trace contained `</think>` and were 3,627–7,125 characters long instead of being a short question. This is present in the JSONL, not only in `DEBUG` output. With the current registration, the model is declared as `reasoning: false`, so Pi does not send the Qwen chat-template control that disables thinking, while the documented vLLM command leaves Qwen thinking enabled by default. A likely fix to validate is either:
   - add `--default-chat-template-kwargs '{"enable_thinking": false}'` to the vLLM command, or
   - register the model with Qwen chat-template compatibility (`reasoning: true`, `compat.thinkingFormat: "qwen-chat-template"`) while keeping the session thinking level off.
   Stripping everything through `</think>` in `lastAssistantText()` could be a fallback, but preventing it at inference time seems cleaner.

2. Provider failures are recorded but do not fail the generation task. My HF Router credits were exhausted after the first turn, so turns 2–5 produced assistant messages with `stopReason: "error"` and HTTP 402 messages. Nevertheless, the process exited 0 and printed `Trace saved to ...`. The batch path appears likely to report the same session as completed. It would be safer for `runSession` to detect an error/aborted assistant message after each `prompt()` and throw, so invalid traces are not presented as successful output.

Could you update the tutorial/generator guidance and validate a persisted multi-turn trace, checking that each generated user message contains only the short question and that provider errors fail the task?

khazic:
Thanks for actually running it, that caught both of these. Fixed in 825480122, and I validated a persisted multi-turn trace as you asked.

## 1. Thinking leaking into the generated user turns

Your diagnosis holds, and the reason it never showed up on llama.cpp is that the packaged GGUF build answers without thinking anyway, so nothing had to send the control. A stock checkpoint served directly has it on by default, and the generator registers the user model with `reasoning: false` and opens the session with `thinkingLevel: "off"`, so it never sends the switch either way.

I took the server-side option and verified it rather than assuming. `--default-chat-template-kwargs` exists in vLLM 0.26.0, and it does what we need. Same server, same request, replaying the user-agent turn with its real system prompt:

```
A. no per-request kwargs (server default, thinking disabled by the flag)
   content chars       : 60
   contains <think> tag: False
   content             : 'What is the purpose of the nemo_automodel/recipes directory?'

B. per-request enable_thinking=true
   content chars       : 965
   contains <think> tag: True
   content             : '<think>\nOkay, let me try to understand this. The user is explaining a codebase called...'
```

One detail worth adding to your write-up: in B the `reasoning_content` field is empty and the whole block sits in `content`. That is the actual mechanism, since the provider registration declares no reasoning support and the server parses none, so nothing separates it out and `lastAssistantText()` takes it along with the question. It is also why stripping in `lastAssistantText()` is the weaker fallback, as you said.

## 2. Provider failures not failing the run

Confirmed in the source, both call sites await `prompt()` and discard the result, so nothing ever looks at `stopReason`.

I hit your exact failure while validating, my included credits ran out mid-session with the same 402 text. Unpatched behaviour matched your report: the run exited 0, printed `Trace saved to ...`, and left a trace whose last three assistant messages were errors. With the check added, the same situation exits non-zero on the first failed turn with `agent turn ended with stopReason=error` and never prints `Trace saved`.

One correction to make in passing: the check does not prevent a partial file. Pi streams the session file as the conversation goes, so a partial trace is already on disk by the time the error surfaces. What the check buys is that the run fails, so the batch path cannot count the session as complete, and the leftover file should be discarded rather than uploaded. The guide now says that.

## The validated trace

Ran one full session against a vLLM-served user model (Qwen3-8B, thinking disabled by the flag) with the default agent model on the router. Five generated user turns, all short questions, none carrying a thinking block:

```
user turns: 5
  [1] chars=  51  think_tag=False  'What are the main characteristics of this codebase?'
  [2] chars=  41  think_tag=False  'What is the purpose of the `hf` CLI tool?'
  [3] chars=  45  think_tag=False  'What is the purpose of the `hf auth` command?'
  [4] chars= 224  think_tag=False  'The `hf auth` command is used to manage authentication with the Hugging Face Hub...'
  [5] chars=  56  think_tag=False  'How does the `hf auth` command handle OAuth device flow?'

assistant stopReasons: {'toolUse': 7, 'stop': 2, 'error': 3}
```

The agent side really exercised its tools before the credits ran out, and the follow-ups track the answers rather than drifting, which is what the pollution was destroying before.

## What the commit changes

The vLLM command now carries `--default-chat-template-kwargs '{"enable_thinking": false}'`, with the reason spelled out and the chat-template-compatibility route (`reasoning: true` plus `compat.thinkingFormat: "qwen-chat-template"`) documented as the alternative. Separately there is a warning covering the provider-error check, with the snippet to add in the fork, listed alongside the other fork changes the guide already asks for.

I kept both fixes in the guide rather than in the dataset code. Filtering thinking residue out of user turns at load time is a real gap, but it needs to distinguish assistant reasoning, which this PR deliberately preserves, from thinking that leaked into a user turn, and it deserves its own tests. Happy to do that as a follow-up if you would rather have the safety net in the library.

HuiyingLi (approved):

PR #3374, opened 2026-08-03
perf: remove Python overhead from model hot paths

akoumpa:
## Summary

- reduce Python and host-synchronization overhead in Transformer Engine mask detection and grouped MoE/LoRA dispatch
- cache repeated `forward` signature inspection for stable class methods in the generic and LTX-2 kwarg filters
- preserve dynamic/instance-patched forward semantics by leaving callable instances, partials, and instance-bound methods uncached
- add semantic, cache-lifetime, and live-patching regression coverage

The grouped-expert bias forward/backward optimization was split into #3387 so its numerical-gradient behavior can be reviewed independently.

LLaVA, retrieval, and hard-negative mining paths are intentionally excluded from this PR.

## Bug found during validation

The first cache implementation used an unbounded cache for arbitrary hashable forward callables. That could retain instance-specific callables and their owning module/GPU state. It also treated the implicit receiver correctly only when it was literally named `self`.

The final implementation:

- bounds both caches to 256 stable class-level forward functions
- strips the implicit bound-method receiver regardless of its name
- does not cache callable instances, partials, or instance-assigned `MethodType` overrides
- continues to invalidate naturally when a class's `forward` is replaced

## Microbenchmarks

Same host, RTX 5880 Ada, PyTorch 2.10 + CUDA 12.8; best of 7 for signature filters and blocked autorange for CUDA kernels:

| Path | main | PR | Change |
|---|---:|---:|---:|
| `filter_forward_kwargs` | 14.802 us | 5.593 us | 2.65x faster |
| LTX-2 kwarg filter | 16.747 us | 5.609 us | 2.99x faster |
| TE causal-mask detection (S=1024, fp32 CUDA) | 167.259 us | 155.741 us | 1.07x faster |

These are isolated microbenchmarks, not end-to-end model throughput measurements.

Bytecode inspection also reduced the instruction count in `_detect_causal_mask` from 276 to 239 on Python 3.10.

## Validation

Targeted suites cover every modified production module:

- `tests/unit_tests/_transformers/test_te_attention.py`
- `tests/unit_tests/_peft/test_lora_experts.py`
- `tests/unit_tests/flow_matching/adapters/test_ltx2_adapter.py`
- `tests/unit_tests/moe/test_experts.py`
- `tests/unit_tests/moe/test_layers.py`
- `tests/unit_tests/utils/test_model_utils.py`

Results after the split:

- local CUDA: **298 passed, 35 skipped**
- Ruff lint, format, and `git diff --check`: passed


akoumpa:
| Revision | AutoModel SHA | NeMoCI scope | Pipeline |
|---|---|---|---|
| Before (main baseline) | 70e738bbe99e3aa42e5ebc006834a37ab2d49890 | nightly: TE+DeepEP MoE, DeepEP LoRA, LTX-2 flow | [60857172](https://nv/nemoci/-/pipelines/60857172) |
| After (PR) | f8095c304e5ab99ce2ba09cafdb15602bf0756ee | nightly: TE+DeepEP MoE, DeepEP LoRA, LTX-2 flow | [60857183](https://nv/nemoci/-/pipelines/60857183) |

Both runs use the same NeMoCI main-mirror definition, fresh AutoModel images, and exclude LLaVA and retrieval cases.

akoumpa:
CI babysitting update for signed head `f8095c304e5ab99ce2ba09cafdb15602bf0756ee`:

- GitHub Actions is fully green: **77 passed, 6 skipped, 0 failed/pending**.
- The originally attached NeMoCI parent pipelines were misleadingly green while downstream jobs failed. They also used `TEST_IMAGE=NONE`, so they were not valid before/after acceptance runs.
- I reran the baseline in isolation. The Qwen3 MoE checks fail on the baseline itself during HF reload parity (TE KL `3.738403e-02 > 1e-2`; LoRA also exceeds its threshold), and the LTX-2 job fails before testing the PR because the launcher does not recognize recipe `ltx2_3_t2v_flow`. These are not regressions introduced by this PR.
- I rebuilt the PR image and ran a fresh, isolated PR-head pipeline using a verified local image because EOS could not pull the valid GitLab image (`502 Bad Gateway`). Final pipeline: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/60880601
  - TE leaf: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/383214348
  - LoRA leaf: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/383214355

Both final Slurm workers reported `TIMEOUT`; afterward the EOS runner wrappers remained stuck and `login-eos` was unresponsive, so I canceled only those orphaned wrappers after preserving the terminal worker evidence. Retrying again during the same EOS/Lustre outage would duplicate the infrastructure failure.

Bottom line: the PR's GitHub checks pass on the exact signed head. The attached NeMoCI before/after comparison cannot currently produce a green acceptance signal: baseline tests/harness are already failing, while the isolated PR rerun is blocked by EOS registry/login/Lustre infrastructure rather than a code assertion.

akoumpa:
NeMoCI recheck after GitLab recovery:

- The exact PR head is still `f8095c304e5ab99ce2ba09cafdb15602bf0756ee`.
- The previously built PR image manifest is healthy again (`200`, digest `sha256:90f767c8dcab375568d1f3a7fc36c8ffb7496a9f23d6619bd1f912f3bbda5087`).
- I launched a fresh retry with the prior proven selectors and a new pipeline/checkpoint namespace:
  - parent: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/60904980
  - generator child: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/60904993
  - Qwen leaf: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/60905028
  - TE job: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/383390732
  - LoRA job: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/383390734

Both leaves passed EOS account/Kerberos preflight and were accepted by `runner-eos02`, but they have remained stuck at the Lustre-backed `git fetch` step with no trace growth for 15+ minutes. Direct `login-eos` is also still unresponsive, and numerous unrelated EOS jobs show the same long-running/stuck pattern.

I am leaving this valid retry running so it can resume if the EOS runner/storage path recovers. GitLab API/registry availability is restored; the remaining blocker is the EOS runner/Lustre side.

HuiyingLi (dismissed):
[P1] Preserve deterministic expert-bias gradient reduction

The new `torch.repeat_interleave` path in `_apply_bias` preserves the forward result, but on CUDA its backward accumulates gradients into repeated bias rows nondeterministically. This affects trainable expert biases; GPT-OSS enables `expert_bias=True` and defaults the model dtype to BF16.

I reproduced this with fixed BF16 inputs, probabilities, and upstream gradients:

- Balanced routing (`64` experts x `64` tokens x `512` features): 10 identical backward runs produced 10 distinct `bias.grad` hashes. All 32,768 gradient elements varied, with a maximum run-to-run span of `1.0`.
- Imbalanced routing (4,096 tokens assigned to one expert): the previous split/cat implementation produced one stable hash across 20 runs, while this implementation produced 20 distinct hashes with a maximum run-to-run span of `43.0`.

The added gradient test does not expose this because `result.sum().backward()` gives exactly representable integer-valued bias gradients, so accumulation order is invisible. Please preserve a deterministic segmented bias-gradient reduction (for example, via a custom autograd implementation), or retain the previous backward path for trainable biases, and add a BF16 CUDA regression test with random upstream gradients and imbalanced token counts.

yuhezhang-ai (approved):

PR #3383, opened 2026-08-03
fix(gemma4_moe): reduce-scatter CP KV grads instead of all-pairs p2p

athitten:
# What does this PR do ?

Fixes the CUDA OOM inside NCCL `batch_isend_irecv` in the Gemma4 CP ring-attention backward (AMINT-242), which fails `gemma4_31b_tulu3_text_cp8_16k`.

# Issue

```
ncclUnhandledCudaError: Call to CUDA function failed.
Last error: Cuda failure 2 'out of memory'
  cp_attention.py:229 in _route_kv_grads_to_owners
  cp_attention.py:202 in _direct_exchange -> torch.distributed.batch_isend_irecv(ops)
```

`_route_kv_grads_to_owners` returned each owner's dK/dV by looping `distance = 1..cp_size-1`, sending point-to-point to `cp_rank - distance`. NCCL connections are **directional**, and the forward ring only ever sends to `cp_rank + 1` / receives from `cp_rank - 1` — so every peer/direction the routing uses is one the forward never opened, **distance 1 included**, since it reverses the ring. NCCL allocates connection buffers lazily on first use, outside the caching allocator, at the backward memory peak.

Measured on 8xH100. `allocated` and `reserved` stay flat while free CUDA memory falls ~0.38 GiB per exchange, one exchange per distance:

| point | alloc | reserved | cuda_free |
|---|---|---|---|
| before routing | 19.28 | 72.73 | 1.36 |
| after distance 1 | 19.29 | 72.73 | 0.98 |
| after distance 2 | 19.31 | 72.73 | 0.61 |
| after distance 3 | 19.31 | 72.73 | 0.23 |
| during distance 4 | — | — | 0.01 → OOM |

Flat `alloc`/`reserved` with falling `cuda_free` is the signature of memory going to something other than the caching allocator. The payload is not the problem: all 8 owners' dK+dV total **0.12 GiB**.

# Fix

Summing per owner and handing owner `i` its slice **is** a reduce-scatter, so issue one per tensor. It runs over the group's existing collective channels, opens no new peer connections, and replaces `2*(cp_size-1)` p2p round-trips plus their per-iteration buffers with two collectives.

Owner-major `cat` on dim 0, not `stack`: `reduce_scatter_tensor` splits dim 0 and requires `output.shape[0] * cp_size == input.shape[0]`, which a `[cp_size, batch, ...]` stack satisfies only when `batch == 1` (this recipe's value, so a stacked version passes the GPU run and hides the bug).

# Validation

**nemo-ci, `gemma4_31b_tulu3_text_cp8_16k` on eos (8xH100)** — [pipeline 60942996](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/60942996) / [job 383727225](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/383727225):

| | |
|---|---|
| Result | **success**, 47.8 min |
| Steps | 50/50 (`Max train steps: 50`, steps 0 → 49) |
| NCCL / torch OOM | **0** / **0** |
| Loss | 0.6311 → 0.3336 |
| grad_norm | 44.42 → 1.62 |

Local A/B, same recipe / node / container (`pipe.60708098`), 1 node / 8 GPUs:

| | Result |
|---|---|
| BEFORE (`518ea2804`, unmodified) | 2x `Cuda failure 2 'out of memory'`, 0 steps |
| AFTER (this branch) | exit 0, **0 OOM**, 3 steps, loss 2.1737, grad_norm 32.13 |
| `tests/unit_tests/models/gemma4_moe/` | 8/8 pass |
| `ruff format` + `ruff check` | clean |

New tests: a 4-rank gloo run asserting each rank's result equals the analytic cross-rank sum (at `batch=2`, which is what caught the `stack` bug), and a test that fails if the routing reaches for `batch_isend_irecv` — the communication pattern is the defect, not the arithmetic.

# Changelog

- `_route_kv_grads_to_owners` uses `reduce_scatter_tensor` instead of all-pairs p2p.
- New `tests/unit_tests/models/gemma4_moe/test_cp_route_kv_grads.py`.
- `gemma4_31b_tulu3_text_cp8_16k.yaml` gains `ci: time: "01:00:00"`; it declared `max_steps: 50` but no `ci:` block, so it inherited `.vlm_finetune_test`'s `00:10:00`, which the measured 47.8 min run cannot fit.

# Pre checks

- [x] Added tests
- [ ] Documentation — n/a

# Additional Information

- Fixes AMINT-242
- Related: AMINT-241 (separate defect: allocator fragmentation, 13.10 GiB FSDP bucket vs 52.48 GiB reserved-but-unallocated)

### DIFF
--- a/examples/vlm_finetune/gemma4/gemma4_31b_tulu3_text_cp8_16k.yaml
+++ b/examples/vlm_finetune/gemma4/gemma4_31b_tulu3_text_cp8_16k.yaml
@@ -108,3 +108,8 @@ wandb:
   entity: <your_wandb_entity>
   name: <your_wandb_name>
   dir: <your_wandb_save_dir>
+
+ci:
+  # max_steps 50 at ~46 s/step measured 47.8 min end-to-end on 8xH100; the
+  # .vlm_finetune_test default of 00:10:00 cannot fit it.
+  time: "01:00:00"
--- a/nemo_automodel/components/models/gemma4_moe/cp_attention.py
+++ b/nemo_automodel/components/models/gemma4_moe/cp_attention.py
@@ -211,34 +211,39 @@ def _route_kv_grads_to_owners(
     cp_rank: int,
     cp_size: int,
 ) -> tuple[torch.Tensor, torch.Tensor]:
-    """Sum each KV owner's dK/dV from every rank whose queries attended it.
+    """Sum each KV owner's dK/dV across every rank that attended it; return the local shard.
 
-    In the ring forward, rank ``r``'s queries attend the chunks owned by ranks
-    ``r, r-1, ..., 0``, so backward produces a dK/dV contribution for each of those
-    owners on rank ``r``. This sends each owner's contribution back to that owner
-    (p2p) and sums them, returning the local rank's accumulated ``(grad_key,
-    grad_value)``. Shared by the Flex and FFPA-varlen ring backward passes.
+    Summing per owner and giving owner ``i`` its slice is a reduce-scatter. It must not be
+    hand-rolled from p2p: that sends to ``cp_rank - distance`` for ``distance 1..cp_size-1``,
+    and NCCL connections are directional, so every peer/direction it uses is one the forward
+    ring (send ``cp_rank + 1``, recv ``cp_rank - 1``) never opened -- including distance 1,
+    which reverses the ring. NCCL allocates those connection buffers lazily on first use,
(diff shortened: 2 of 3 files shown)


athitten:
Closing as not needed. #3408 fix suffices.

PR #3387, opened 2026-08-04
perf(moe): vectorize deterministic expert bias gradients

akoumpa:
## Summary

- replace Python `split` / `zip` / `cat` bias application with a tensorized `repeat_interleave` forward
- preserve deterministic trainable expert-bias gradients with one FP32 `segment_reduce` in the custom backward
- keep the backward compatible with full-graph AOTAutograd and Inductor compilation
- cover list and tensor biases, probability-weighted bias, empty experts, higher-rank outputs, and imbalanced BF16 CUDA gradients

This change was split from #3374 so the generic Python hot-path cleanup can be reviewed independently from the MoE bias-gradient behavior.

## Root cause

The original grouped-expert bias path split tensors per expert, iterated over them in Python, and concatenated the results. Replacing that forward directly with `repeat_interleave` removes the Python and host-synchronization overhead, but its CUDA backward accumulates repeated BF16 bias rows nondeterministically.

The custom autograd path keeps the tensorized forward and reduces bias gradients by their already-contiguous expert segments. FP16/BF16 gradients accumulate in FP32 before casting back; reducing directly in BF16 was deterministic but materially less accurate for heavily imbalanced routing.

## Microbenchmark

Same host, RTX 5880 Ada, PyTorch 2.10 + CUDA 12.8, 64 experts and a `4096 x 512` BF16 tensor:

| Path | main | PR | Change |
|---|---:|---:|---:|
| Forward | 0.553 ms | 0.050 ms | 11.0x faster |
| Forward + backward | 3.321 ms | 0.678 ms | 4.9x faster |

These are isolated microbenchmarks, not end-to-end model throughput measurements.

## Validation

- `CUDA_VISIBLE_DEVICES=0 pytest tests/unit_tests/moe/test_experts.py tests/functional_tests/moe/test_expert_bias_gradient.py -q`
  - **73 passed, 26 skipped**
- ten imbalanced BF16 CUDA backward passes were bitwise identical to the previous split/cat reference
- probability-weighted value, bias, and routing-probability gradients matched exactly
- `torch.compile(..., fullgraph=True)` passed through AOTAutograd; local Inductor validation also passed
- Ruff lint, format, and `git diff --check`: passed


nemo_automodel/components/moe/experts.py
```diff
@@ -163,47 +207,29 @@ def _apply_bias(value, bias, tokens_per_expert, permuted_probs=None):
...
-                    )
-                ]
-            )
-            .view(shape)
-            .to(value.dtype)
-        )
-    return output
+        expanded_bias = torch.repeat_interleave(bias, token_counts, dim=0, output_size=flat_value.shape[0])
+    if permuted_probs is not None:
+        expanded_bias = expanded_bias * permuted_probs
```
HuiyingLi, line 366: The probability weighting needs to stay in FP32 through the segmented reduction. In the real grouped-MM path, `permuted_probs` is FP32 (`flat_weights = weights.float()`), while `expanded_bias` is normally BF16. Because this multiplication is outside the custom `Function`, autograd casts the gradient for the function's BF16 output back to BF16 before `backward` runs; converting `grad_output` to FP32 there cannot recover the lost multiplication precision. With 4096 positive FP32 probabilities and BF16 upstream gradients on PyTorch 2.10, I get 222/512 bias-gradient elements differing from the FP32 reference (max absolute difference 0.5), while the previous split path matches the reference exactly. Please either include the probability weighting in the custom function so its backward reduces `grad_output.float() * permuted_probs.float()`, or promote the function input/output so the weighted gradient reaches the reduction in FP32. A BF16-bias + FP32-probabilities regression test should cover this.

HuiyingLi (changes requested):
One blocking issue: the probability-weighted BF16 bias-gradient path rounds before the intended FP32 segmented reduction.

akoumpa:
Adversarial review and `segment_reduce` findings for the bias-gradient backward:

- PyTorch 2.10 CUDA `segment_reduce` uses one output thread per `[expert, hidden]` element for the 2-D lengths path and accumulates that segment serially. It is deterministic, but its error grows with expert imbalance: https://github.com/pytorch/pytorch/blob/v2.10.0/aten/src/ATen/native/cuda/SegmentReduce.cu
- A realizable 4096-token weighted-gradient adversary (`2048 * +1`, then `2048 * -0.9999`) has FP64 sum `0.204833984375`. The old 2-D CUDA segment reduction produces `0.2373197079` in FP32, which rounds to BF16 `0.2373046875` instead of `0.205078125` (about 16% error).
- In a 32-seed random scan at 4096x512, the old path had 20 BF16 mismatches against the rounded FP64 oracle and max FP32 error `3.14e-4`. At 262144x128 it reached max error `9.43e-3`.
- The old path also serializes the longest expert: one-expert reductions measured about 2.14 ms at 65536x128 and 17.65 ms at 262144x128 on RTX 5880 Ada.

Commit `7cceabe7d77eb902a8db4d535e6490d71cd02cea` replaces that backward with a Triton kernel using `tl.make_tensor_descriptor`. Each program exclusively owns one `[expert, hidden-block]` tile (no atomics), uses a balanced 128-token block reduction, and Kahan-compensates the block totals. Unsupported dtypes, descriptor-misaligned rows, >=2^31 tokens, higher-order gradients, CPU, or unavailable Triton take the differentiable fallback.

Numerical gate:

- All CUDA parity assertions use an FP64 oracle cast to BF16 with `torch.testing.assert_close(..., rtol=0, atol=0)`.
- The 4096-token cancellation adversary is bit-exact, as are empty experts, 127/128/129 block edges, a 4096-token expert, and noncontiguous upstream gradients.
- Ten repeated launches are checked for bitwise determinism.
- CPU FP32 cancellation and double-backward gradcheck/gradgradcheck cover the fallback.

Performance gate on RTX 5880 Ada, reduction-only versus the accurate 1-D fallback:

- 4096x512 balanced: 0.0848 ms vs 0.1144 ms (1.35x).
- 4096x512 one expert: 0.0787 ms vs 0.1116 ms (1.42x).
- End-to-end `_apply_bias` forward+backward versus the pre-PR split/cat reference measured 9.01x balanced and 3.73x one-expert unweighted; weighted measured 11.4x and 12.78x.

The numerical gate intentionally takes precedence where a faster reduction is inaccurate: at 65536-token cancellation, the 1-D fallback rounds to BF16 `3.25`, while the descriptor kernel and FP64 oracle both produce `3.28125`.

nemo_automodel/components/moe/experts.py
akoumpa: Addressed in 7cceabe7d77eb902a8db4d535e6490d71cd02cea. The backward now uses a deterministic compensated FP32 Triton reduction with `tl.make_tensor_descriptor` for supported CUDA inputs, plus a differentiable fallback. The weighted BF16 tests now compare against an FP64 oracle with `torch.testing.assert_close(..., rtol=0, atol=0)`, including the cancellation case that makes the old 2-D `segment_reduce` path wrong by about 16%. I also added empty-expert/block-edge/noncontiguous-gradient coverage and gradcheck/gradgradcheck.

HuiyingLi (approved):

PR #3420, opened 2026-08-05
fix(ci): run Kimi K3 HellaSwag on GB200

akoumpa:
## Summary

- reserve the GB200 cluster for the Kimi K3 HellaSwag recipe
- restore the validated 64-node x 4-GPU topology
- link Huiying's successful 100-step W&B run in the recipe

## Root cause

The recipe was validated on 64 nodes with 4 GB200 GPUs per node, but its CI metadata requested 32 nodes without constraining the accelerator. Release CI therefore scheduled 32 nodes with 8 H100 GPUs per node. Although both layouts have 256 GPUs, the historical run used about 129 GiB per GPU while H100 provides only 79.11 GiB, causing the first backward pass to OOM.

Passing validation: https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/olnfgmh1

Tracks: https://linear.app/nvidia/issue/AMINT-243/cuda-oom-allocating-moe-experts-module-during-backward-pass

## Validation

- `python -m pytest tests/unit_tests/ci_tests/test_generate_ci_tests.py -q` -- 9 passed
- generated release job variables:
  - `TEST_NODE_COUNT=64`
  - `RESERVED_CLUSTER_TAG=/gb200/`
- `git diff --check`

### DIFF
--- a/examples/llm_finetune/kimi/k3_hellaswag.yaml
+++ b/examples/llm_finetune/kimi/k3_hellaswag.yaml
@@ -14,6 +14,7 @@
 
 # Full-parameter, 93-layer Kimi K3 text SFT run on 256 GPUs:
 #   64 nodes x 4 GPUs, EP32 x PP8, DP32, 100 steps, lr=1e-5.
+# Passing GB200 validation: https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/olnfgmh1
 # PP stages own the complete attention-residual blocks:
 #   [0,12), [12,24), [24,36), [36,48), [48,60), [60,72), [72,84), [84,93).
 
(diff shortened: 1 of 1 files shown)


examples/llm_finetune/kimi/k3_hellaswag.yaml
```diff
@@ -14,6 +14,7 @@
 
 # Full-parameter, 93-layer Kimi K3 text SFT run on 256 GPUs:
 #   64 nodes x 4 GPUs, EP32 x PP8, DP32, 100 steps, lr=1e-5.
+# Passing GB200 validation: https://wandb.ai/Nemo-automodel/huiyingl_workspace/runs/olnfgmh1
```
akoumpa:
```suggestion
```

HuiyingLi (approved):
thanks for catching it

PR #3423, opened 2026-08-05
fix(checkpoint): shard extracted exports by size

oliverholworthy:
# What does this PR do ?

This fixes consolidated safetensors export for extracted submodels. The motivating case is the Ministral3 retrieval recipe, which extracts `language_model` from `mistralai/Ministral-3-3B-Instruct-2512-BF16`.

The parent visual language model (VLM) checkpoint indexes weights under names such as `language_model.layers.*`, while the extracted text model exports them as `layers.*`. Pruning the parent index therefore removes every entry. Previously, all exported tensors then defaulted to shard 1, defeating the 5 GiB shard target, concentrating consolidation on one rank, and potentially producing an oversized file or excessive memory pressure.

The fix rebuilds the mapping from logical tensor sizes. Under pipeline parallelism, it gathers size metadata—not tensor data—so every stage derives the same complete shard mapping.

Although this issue was discovered while working on #3103, the fix applies generically to extracted-submodel exports, so it has been split into this focused PR.

# Changelog

- Rebuild a fully pruned source shard mapping using the existing 5 GiB size-based partitioning helper.
- Gather logical tensor-size metadata across pipeline stages so every rank derives the same consolidated shard mapping without moving tensor data.
- Preserve the existing source mapping whenever at least one exported key remains.
- Keep explicitly excluded tied-weight aliases out of the rebuilt mapping.
- Add mocked mapping coverage and a real two-rank CPU/Gloo regression for pipeline-parallel agreement.

**Pre checks**:

# Validation

- `uv run --frozen pytest tests/unit_tests/checkpoint/test_checkpointing.py -q --cpu` — 210 passed
- `uv run --frozen ruff format --check nemo_automodel/components/checkpoint/checkpointing.py tests/unit_tests/checkpoint/test_checkpointing.py`
- `uv run --frozen ruff check nemo_automodel/components/checkpoint/checkpointing.py tests/unit_tests/checkpoint/test_checkpointing.py`
- `git diff --check origin/main...HEAD`


nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -1707,6 +1707,12 @@ def _maybe_build_consolidated_index(
                 num_shards = max(fqn_to_file_index_mapping.values()) if fqn_to_file_index_mapping else 1
                 fqn_to_file_index_mapping = _equally_divide_layers(num_shards, pre_shard_hf_state_dict_keys)
             else:
+                # Decide whether the size metadata collective is needed from inputs that are
+                # identical on every PP rank. Rank-local exclusions below must not control
+                # collective participation, or one stage could wait forever for another.
+                fallback_key_sizes = None
+                if set(fqn_to_file_index_mapping).isdisjoint(pre_shard_hf_state_dict_keys):
```
yuhezhang-ai, line 1714: Could we make this collective-participation decision safe when neither `_pre_shard_hf_state_dict_keys` nor `model_state_dict_keys` is available? In that existing fallback, line 1704 uses each PP stage’s local `state_dict` keys, so a partially matching source index can make one rank skip `_collect_global_tensor_sizes` while another rank enters it and waits indefinitely. Deriving the decision from a global key set (or having all PP ranks gather when the global list is absent) would keep collective order symmetric. A two-rank test with missing global key metadata and a partially overlapping source mapping would cover this case.

nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -2591,6 +2638,52 @@ def _divide_keys_by_size(
...
 
 
+def _collect_global_tensor_sizes(
+    state_dict: dict[str, torch.Tensor],
+    process_group: torch.distributed.ProcessGroup | None,
+) -> dict[str, int]:
+    """Collect logical tensor sizes across pipeline stages without moving tensor data.
+
+    Args:
+        state_dict: Mapping of tensor names to rank-local tensors of arbitrary shape. Tensors remain on their
```
yuhezhang-ai, line 2648: Could we make the distributed size contract explicit here? In particular, `state_dict` contributes the current PP stage’s key subset, while each reported size needs to be the tensor’s global logical size (for example, a DTensor’s global `numel()`) rather than a per-rank TP/FSDP shard size. Documenting that distinction would make it clear why the 5 GiB partitioning remains correct under composed PP + TP/FSDP.

yuhezhang-ai (commented):
Thanks for the focused fix. I left two PP-related comments: one correctness concern around collective participation and one request to clarify the distributed tensor-size contract.

nemo_automodel/components/checkpoint/checkpointing.py
oliverholworthy: I’ve updated the docstring to clarify that state_dict contains the current PP stage’s keys, while the reported sizes are for the full logical tensors rather than the local TP/FSDP shards

nemo_automodel/components/checkpoint/checkpointing.py
oliverholworthy: I've changed this so that when the global key metadata is missing, all PP ranks gather the size metadata first and make the decision from the same key set. I also added the two-rank partial-overlap test you suggested.

yuhezhang-ai (dismissed):

oliverholworthy:
@yuhezhang-ai please take another look, i've updated the tests to improve the test coverage of this change

nemo_automodel/components/checkpoint/checkpointing.py
```diff
@@ -1731,6 +1757,7 @@ def _maybe_build_consolidated_index(
                 fallback_keys,
                 state_dict,
                 _DEFAULT_HF_CONSOLIDATED_SHARD_SIZE_BYTES,
+                key_size_mapping=_collect_global_tensor_sizes(state_dict, self.pp_group),
```
yuhezhang-ai: Thanks for addressing the earlier feedback—the latest update looks good overall, but I just noticed one more edge case to address.

When the HF reference path is missing and neither `_pre_shard_hf_state_dict_keys` nor `model_state_dict_keys` is available, `fallback_keys` still comes from this PP stage’s local `state_dict`, even though the size mapping gathered here is global. With disjoint PP stages, each rank therefore builds a different incomplete FQN-to-shard mapping. I reproduced this by running the existing worker with `source_mapping=None` and `has_global_keys=False`; the global-mapping assertion fails, and consolidation can later fail on remote-stage FQNs missing from rank 0’s mapping.

Could we keep the gathered mapping in a local variable and use `list(global_key_sizes)` for `fallback_keys` when the global key metadata is absent? Covering that fourth combination in the existing two-rank test would close the gap; ideally the cases could reuse a single spawned process group to avoid adding more CPU-CI startup time.

oliverholworthy: Thanks for catching this, addressed in [df0651583](<https://github.com/NVIDIA-NeMo/Automodel/commit/df0651583f7638b5a3b99677f0a5c747906d09dc>). The fallback now uses `list(global_key_sizes)` when global key metadata is absent.

And I've added an additional test case and confirmed it fails before the fix and passes afterward, and updated so that the different cases share one spawned process group.

yuhezhang-ai:
/ok to test [2906168](https://github.com/NVIDIA-NeMo/Automodel/pull/3423/commits/2906168c1eb2a3819fc3ea645892703b9a929cf5)

yuhezhang-ai (approved):

PR #3431, opened 2026-08-05
fix(kimi_k25_vl): keep PEFT expert LoRA keys under the base_model.model. prefix

### LINKED ISSUE #3385: Kimi K2.5 VL peft save writes expert lora keys with mangled prefix, silently dropped on resume
**Describe the bug**

follow up from NVIDIA-NeMo/Automodel#3282. now that the quantization thing is fixed (NVIDIA-NeMo/Automodel#3295) the expert lora keys for kimi k2.5 vl actually survive the save, but I noticed their names come out wrong.

the adapter wraps expert conversion results with a "language_model." prefix (line 215 of state_dict_adapter.py). works fine for full finetune saves since those keys are in native format. but on a peft save the keys already have the "base_model.model." prefix on them, and the lora expert conversion keeps it, so the "language_model." ends up bolted on the outside and you get stuff like

```
language_model.base_model.model.model.language_model.model.layers.5.mlp.experts.0.gate_proj.lora_A.weight
```

language_model shows up twice in there. the attention lora keys from the exact same save look normal, it's only the expert ones. so on resume nothing matches and the expert adapters just quietly disappear, peft can't attach them either.

no gpu needed:

```
import torch
from nemo_automodel.components.models.kimi_k25_vl.model import KimiK25VLConfig
from nemo_automodel.components.models.kimi_k25_vl.state_dict_adapter import KimiK25VLStateDictAdapter
from nemo_automodel.components.moe.config import MoEConfig

moe = MoEConfig(
    n_routed_experts=4, n_shared_experts=1, n_activated_experts=2,
    n_expert_groups=1, n_limited_groups=1, train_gate=False,
    gate_bias_update_factor=0.0, aux_loss_coeff=0.0, score_func="softmax",
    route_scale=1.0, dim=64, inter_dim=128, moe_inter_dim=16, norm_topk_prob=True,
)
adapter = KimiK25VLStateDictAdapter(
    KimiK25VLConfig(), moe, BackendConfig(linear="torch", rms_norm="torch", attn="sdpa")
)

# grouped expert lora params exactly as ModelState.state_dict() emits them on a peft save
base = "base_model.model.model.language_model.model.layers.5.mlp.experts"
sd = {
    f"{base}.lora_gate_and_up_A": torch.randn(4, 8, 64),
    f"{base}.lora_gate_and_up_B": torch.randn(4, 32, 8),
    f"{base}.lora_down_A": torch.randn(4, 8, 16),
    f"{base}.lora_down_B": torch.randn(4, 64, 8),
}

out = adapter.to_hf(dict(sd), exclude_key_regex=r".*_extra_state.*", quantization=False)
print(sorted(out)[0])
print("keys starting with base_model.model.:", sum(k.startswith("base_model.model.") for k in out), "/", len(out))   
```

gives:

```
language_model.base_model.model.model.language_model.model.layers.5.mlp.experts.0.down_proj.lora_A.weight
keys starting with base_model.model.: 0 / 24
```

**Expected behavior**

expert lora keys should look like the attention lora keys from the same save, under base_model.model., so resume and peft can actually find them.

**Additional context**

same family of bug my survey in NVIDIA-NeMo/Automodel#1755 kept finding, a model specific prefix getting applied outside the peft prefix instead of inside. fix should be contained to the kimi adapter's key handling. can put up a PR like last time if that works.

stanley1208:
Fixes NVIDIA-NeMo/Automodel#3385.

two parts. on save, the expert lora conversion keeps the incoming key prefix (which already has base_model.model. on it for peft saves), so the extra "language_model." wrap was landing on the outside and corrupting every expert adapter key. now the wrap is skipped for lora keys, full finetune expert weights still get it since the mixin resets those to the bare model. prefix.

on load, from_hf now routes base_model. keys through the lora recombiner so the per expert keys merge back into the grouped tensors the model holds. before this, even well formed keys wouldn't have resumed.

after the fix a peft save comes out with all keys under base_model.model. (was 0 out of 48), and save then resume round trips exactly, values identical.

3 new cpu tests, 2 of them fail without the change, plus a regression guard for the full finetune path.

### DIFF
--- a/nemo_automodel/components/models/kimi_k25_vl/state_dict_adapter.py
+++ b/nemo_automodel/components/models/kimi_k25_vl/state_dict_adapter.py
@@ -212,7 +212,12 @@ class KimiK25VLStateDictAdapter(MoESplitExpertsStateDictMixin, StateDictAdapter)
 
         expert_result = self._convert_single_merged_expert_to_hf_split_experts(fqn, tensor, **kwargs)
         if expert_result is not None:
-            result = [("language_model." + k, v) for k, v in expert_result]
+            # The mixin's weight branch resets expert keys to the bare "model." prefix,
+            # so full-fine-tune saves need "language_model." re-applied here. Its LoRA
+            # branch instead PRESERVES the incoming fqn (which already contains
+            # "language_model." and, on PEFT saves, the "base_model.model." outer
+            # prefix), so wrapping those again would corrupt the adapter keys.
+            result = [((k, v) if ".lora_" in k else ("language_model." + k, v)) for k, v in expert_result]
         else:
             result = [(fqn, tensor)]
 
@@ -363,9 +368,16 @@ class KimiK25VLStateDictAdapter(MoESplitExpertsStateDictMixin, StateDictAdapter)
         effective_state_dict.update(dequantized)
 
         llm_keys = {}
+        peft_keys = {}
 
         for key, value in effective_state_dict.items():
-            if key.startswith("language_model.model."):
+            if key.startswith("base_model."):
+                # PEFT adapter namespace. These keys are saved with their native-format
+                # path preserved (including the "base_model.model." outer prefix), so no
+                # HF->native renaming applies — only the per-expert expert-LoRA keys need
+                # recombining into the grouped tensors the model actually holds.
+                peft_keys[key] = value
+            elif key.startswith("language_model.model."):
                 llm_key = key.replace("language_model.model.", "model.")
                 llm_keys[llm_key] = value
             elif key.startswith("language_model.lm_head."):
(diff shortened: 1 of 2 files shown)


yuhezhang-ai:
/ok to test [4e731ef](https://github.com/NVIDIA-NeMo/Automodel/pull/3431/commits/4e731eff63ebaca1c434978658694940d14d69a3)

yuhezhang-ai:
Thanks for the fix! I just added a focused CPU PEFT save/load integration test to this branch. It writes the converted Kimi expert adapters as a real PEFT checkpoint, reloads them through PeftModel.from_pretrained on the Transformers v5 stack, checks that no adapter tensors are silently dropped, and verifies that merge_and_unload numerically updates every expert projection. The full Kimi state-dict adapter test file passes locally: 78 passed, 1 skipped.

yuhezhang-ai (approved):

PR #3435, opened 2026-08-05
fix(kimi_k3): apply the HF renames to expert LoRA keys on save

### LINKED ISSUE #3434: Kimi K3 expert lora keys skip the HF renames on save
**Describe the bug**

the k3 adapter renames expert weights to the model's real layout (language_model. prefix, block_sparse_moe.experts.N.w1 naming) but expert lora keys skip all of it. the rename regexes anchor on (w1|w2|w3).weight$ and there's no lora branch, so on a peft save the expert adapter keys come out with automodel's internal naming instead.

that means the saved adapter has keys like mlp.experts.0.gate_proj.lora_A.weight which don't exist anywhere in the actual model, the real modules are language_model.model.layers.N.block_sparse_moe.experts.M.w1. so peft or vllm can't attach those adapters, the keys just don't match anything.

to be fair, automodel's own save then resume round trips fine, I checked. this only bites when the adapter file gets used outside automodel.

no gpu needed:

```
import torch
from types import SimpleNamespace

from nemo_automodel.components.models.kimi_k3.state_dict_adapter import KimiK3StateDictAdapter
from nemo_automodel.components.moe.config import MoEConfig

moe = MoEConfig(
    n_routed_experts=4, n_shared_experts=1, n_activated_experts=2,
    n_expert_groups=1, n_limited_groups=1, train_gate=False,
    gate_bias_update_factor=0.0, aux_loss_coeff=0.0, score_func="softmax",
    route_scale=1.0, dim=64, inter_dim=128, moe_inter_dim=16, norm_topk_prob=True,
)
adapter = KimiK3StateDictAdapter(
    SimpleNamespace(), moe, BackendConfig(linear="torch", rms_norm="torch", attn="sdpa")
)

# a regular expert weight gets the HF renames
w = adapter.to_hf({"model.layers.5.mlp.experts.gate_and_up_projs": torch.randn(4, 64, 32)})
print("weight save:  ", sorted(w)[0])

# grouped expert lora params exactly as ModelState.state_dict() emits them on a peft save
base = "base_model.model.model.layers.5.mlp.experts"
sd = {
    f"{base}.lora_gate_and_up_A": torch.randn(4, 8, 64),
    f"{base}.lora_gate_and_up_B": torch.randn(4, 32, 8),
    f"{base}.lora_down_A": torch.randn(4, 8, 16),
    f"{base}.lora_down_B": torch.randn(4, 64, 8),
}
out = adapter.to_hf(dict(sd))
print("lora save:    ", sorted(out)[0])
print("keys using block_sparse_moe/w1 naming:", sum("block_sparse_moe" in k for k in out), "/", len(out))
```

gives:

```
weight save:   language_model.model.layers.5.block_sparse_moe.experts.0.w1.weight
lora save:     base_model.model.model.layers.5.mlp.experts.0.down_proj.lora_A.weight
keys using block_sparse_moe/w1 naming: 0 / 24
```

**Expected behavior**

expert lora keys should get the same renames the adapter already does for the expert weights, so the saved adapter matches the model it was trained on.

**Additional context**

same family as NVIDIA-NeMo/Automodel#3385 and the stuff my NVIDIA-NeMo/Automodel#1755 survey found, rename regexes anchored on .weight$ that miss the lora suffixes. k3 only landed a few days ago (NVIDIA-NeMo/Automodel#3259) so figured better to flag it early. can put up a PR.

stanley1208:
Fixes #3434.

the rename regexes only matched plain native keys ending in .weight, so lora keys slipped past the entire rename chain. no w1/w2/w3, no block_sparse_moe, no language_model prefix, the adapter file just had names that don't exist in the model.

fix is to let the regexes take any prefix and an optional .lora_A/.lora_B bit, and for peft keys the language_model. part now goes inside base_model.model. since that's how peft names modules on the actual model. while I was in there I noticed attention lora keys were missing the prefix too, same change covers them.

from_hf undoes all of it on load so save then resume still round trips, I checked, values come back identical. normal weight saves and checkpoint loads behave exactly like before.

3 new cpu tests. the naming one fails without the change, the other two guard the round trip and the weight layout.


HuiyingLi:
Hi @stanley1208 , did you get a chance to run lora with K3? If so would you mind also attaching a working peft config? Many thanks!

stanley1208:
Hi @HuiyingLi, I don't have a gpu so I couldn't run a real k3 training, the pr tests only cover the save/resume path on cpu.

for a config, k3_hellaswag.yaml plus this on top should do it:

    peft:
      target_modules: ["*"]
      dim: 32
      alpha: 32
      dropout: 0.0
      moe_rank_scaling: true

    checkpoint:
      enabled: true
      checkpoint_dir: checkpoints/
      model_save_format: safetensors

checkpoint needs to be on since that's where the bug lives. would be great if you could give it a run on your side.

HuiyingLi:
@stanley1208  Sure I can run the peft thank you!

HuiyingLi:
Running the test now

HuiyingLi:
Apologies. Cluster maintenance, wouldn't be able to test it today.

HuiyingLi:
I validated this change with a real 4-layer Kimi K3 expert-LoRA run on a single node.

The core state-dict conversion in this PR works as intended:

- The exported adapter contained 16,128 expert LoRA tensors.
- All expert keys were converted to the official HF K3 module layout under `block_sparse_moe.experts.{expert_id}.w1/w2/w3`.
- No internal `mlp.experts.*.{gate,up,down}_proj` keys remained in the exported safetensors.
- An AutoModel checkpoint resume also succeeded, loading all 128 expected adapter parameters and skipping 0 keys.
- The updated unit tests pass (6/6).

There is one remaining interoperability issue: `adapter_config.json.target_modules` is not converted together with the state-dict keys. It still contains internal module names such as:

```text
model.layers.*.mlp.experts.*.gate_proj
model.layers.*.mlp.experts.*.up_proj
model.layers.*.mlp.experts.*.down_proj
```

Against the official HF K3 model, 8,079 of the 8,122 exported target-module entries could not be resolved, including all 8,064 expert entries. The safetensor weight keys themselves resolve correctly, but a standard HF PEFT/vLLM loading path may still fail to create the adapter modules from this config.

So the expert weight-key fix is correct for AutoModel save/resume, but the export is not yet fully HF PEFT-compatible unless the adapter target-module metadata is converted as well.

stanley1208:
thank you so much for running it! I got a update days ago, could you re run it when you got time? thanks again

HuiyingLi:
I pulled and retested the latest head (`ff4f3b64`). The new target-module hook fixes the main per-expert metadata issue, but 15 target modules are still left in the native layout.

Using the 8,122 `target_modules` from the real 4-layer K3 expert-LoRA export and resolving them against the official Kimi-K3 checkpoint layout:

- Before this commit: 8,079 targets did not resolve.
- After the current mapping: 8,107 resolve and 15 still do not resolve.
- The 8,064 per-expert entries now correctly map to `block_sparse_moe.experts.{E}.w1/w2/w3`.

The remaining entries are five modules in each of layers 1, 2, and 3:

```text
model.layers.N.mlp.routed_expert_down_proj
model.layers.N.mlp.routed_expert_up_proj
model.layers.N.mlp.shared_experts.down_proj
model.layers.N.mlp.shared_experts.gate_proj
model.layers.N.mlp.shared_experts.up_proj
```

The official paths use `model.layers.N.block_sparse_moe...`. These entries are material: the exported adapter contains 30 corresponding LoRA A/B tensors, and their safetensor keys are already converted to `block_sparse_moe`, so the config and weights remain inconsistent.

The cause is that `map_peft_target_module_to_hf()` returns non-per-expert paths unchanged, while state-dict conversion applies `_native_moe_key_to_hf()` to `shared_experts` and `routed_expert_{up,down}_proj` as well. Applying the same MoE child-path conversion after the per-expert `w1/w2/w3` mapping should cover these entries. The test expectation that `model.layers.5.mlp.shared_experts.gate_proj` passes through unchanged should also be updated.

The focused unit suite passes (30/30), but it currently does not check resolution of every exported target against the official layout.

stanley1208:
@HuiyingLi pushed the fix for those  remaining 15 target modules , could you rerun it when you got time? thank you so much

HuiyingLi (approved):
Thank you!

nemo_automodel/components/checkpoint/addons.py
```diff
@@ -530,6 +530,14 @@ def _extract_target_modules(
                 name[len("model.") :] if name.startswith("model.") else name for name in final_target_modules
             }
 
+    # Adapters whose HF checkpoint layout renames modules (e.g. Kimi K3's
+    # mlp.experts.{E}.gate_proj -> block_sparse_moe.experts.{E}.w1) convert the
+    # state-dict keys on save, so adapter_config.json's target_modules must get
+    # the same treatment or PEFT can't resolve them against the real model.
+    map_target_module = getattr(adapter, "map_peft_target_module_to_hf", None)
```
HuiyingLi, line 537: Hi I think this might be more general than just k3. Would you be ok with have another pr making this an interface in StateDictAdapter? So we dont getattr here?

akoumpa:
https://github.com/NVIDIA-NeMo/Automodel/actions/runs/32285980047/job/96196206110?pr=3588

nemo_automodel/components/checkpoint/addons.py
stanley1208: sure absolutely

PR #3427, opened 2026-08-05
test(checkpoint): redesign resume robustness around a shared trajectory

yuhezhang-ai:
# What does this PR do?

Redesign checkpoint-resume robustness so resumed training is compared with the uninterrupted continuation of the same checkpoint-producing trajectory. Independent-run reproducibility remains visible as a separately named, non-blocking, opportunistic diagnostic that reuses existing training phases instead of launching another baseline run.

The previous resume oracle compared independently initialized runs, which conflated checkpoint restoration correctness with broader training reproducibility. This made resume thresholds recipe-specific and could hide missing optimizer, scheduler, RNG, or dataloader restoration behind a loose loss tolerance.

# Changelog

- Continue checkpoint Phase 1 past the checkpoint boundary and compare a fresh restore against the same per-rank post-boundary batches and losses.
- Validate exact optimizer step/group state, LR scheduler state, RNG state, LR, and weight decay before continuing training; verify stateful-dataloader position behaviorally through exact per-rank resumed batch identity.
- Use a strict first-post-resume loss threshold (`1e-6` by default) and a separate later-step BF16 envelope (`5e-3` by default). The distributed retrieval contrastive-loss recipe uses a calibrated `5e-2` later-step envelope while keeping its first resumed step strict.
- Remove the independently trained resume-baseline phase from CI generation and process-isolated execution.
- Record the existing normal finetune and checkpoint Phase 1 for independent-run reproducibility without an additional training run.
- Compare independent runs only when model/init, seed, data ordering, batch/topology, optimizer, scheduler, loss, and backend fingerprints match; phase-specific differences report `not_comparable` rather than pretending to provide coverage.
- Retain the previously calibrated recipe thresholds under `training_reproducibility_loss_threshold`; comparable-run violations emit a prominent non-blocking alert and a machine-readable `report.json`.
- Calibrate Qwen source-load parity and retrieval later-step resume envelopes from scoped CI observations without changing rank-local RNG persistence in this PR.
- Cover the shared trajectory, omitted restored state, strict first-step tolerance, configuration mismatch, batch divergence, loss-envelope alerting, CI config resolution, and phase generation with focused unit tests.
- Update checkpoint-robustness and CI documentation.

# Validation

Local validation:

- `pytest tests/unit_tests/ci_tests/test_resume_trajectory.py tests/unit_tests/ci_tests/test_config_resolver.py tests/unit_tests/ci_tests/test_generate_ci_tests.py -q` — 61 passed, 2 warnings
[...]

PR #3437, opened 2026-08-06
fix(checkpoint): save ranked RNG state per global rank

yuhezhang-ai:
## Summary

- save ranked training RNG state once per global process rank
- restore rank-local RNG state exactly under tensor and pipeline parallelism
- retain legacy per-DP RNG checkpoint loading with an explicit warning
- keep DFlash and DSpark on their intentionally DP-scoped RNG checkpoint path

## Why

`StatefulRNG(ranked=True)` gives each global process rank a distinct RNG stream. The existing auxiliary checkpoint helper only wrote the TP0/PP0 stream for each DP rank, so TP/PP peers restored the wrong RNG state after resume.

## Validation

- four-process CPU/Gloo round trip with TP2 x PP2 and DP1
- checkpointer, BaseRecipe, and EAGLE unit suites: 338 passed, 7 skipped
- DFlash and DSpark guardrail suites: 132 passed
- Ruff format and check on all touched files

This is intentionally separate from the checkpoint-robustness harness changes in #3427.


yuhezhang-ai:
/ok to test [9610417](https://github.com/NVIDIA-NeMo/Automodel/pull/3437/commits/961041742d89095671c0566bb08ce90bd2edaf22)

akoumpa (approved):

PR #3461, opened 2026-08-07
fix(training): all-reduce grad-norm scalars on the mesh device

zyzhou5:
# What does this PR do ?

Under FSDP2 CPUOffloadPolicy the local shard of ``p.grad`` lives in host RAM, so ``_clip_grad_norm_impl`` derives ``target_device = cpu`` and allocates its norm accumulators there. Those scalars were then all-reduced over the mesh's NCCL process group, which has no CPU backend, aborting training on every rank at the first optimizer step with:

    RuntimeError: No backend type associated with device type cpu

Split the compute device from the communication device: the norm math stays on the gradients' own device, and only the 0-dim accumulator hops to ``mesh.device_type`` for the collective before coming straight back. Simply allocating the scalar on CUDA is not sufficient -- the scaled-sum pass divides the CPU gradient by that scalar, and PyTorch's 0-dim promotion is one-way (a CPU scalar mixes with a CUDA tensor, a CUDA scalar does not mix with a CPU tensor).

All five collectives in the clip path are routed through the new helper; fixing only the first one moves the failure to the second.

When the scalar already sits on the mesh device -- every configuration without CPU offload, plus genuinely-CPU gloo meshes -- the helper performs the same in-place all_reduce on the same tensor object, so those paths are unchanged.


zyzhou5:
/ok to test [c5270ff](https://github.com/NVIDIA-NeMo/Automodel/pull/3461/commits/c5270ff4455d54a377bdd7ea36a91b2beeff91cf)

zyzhou5:
/ok to test [ddcb9f7](https://github.com/NVIDIA-NeMo/Automodel/pull/3461/commits/ddcb9f7340b6ce3db148efa6f3f89243d2c41d84)

akoumpa (dismissed):

akoumpa (approved):

PR #3467, opened 2026-08-09
fix(deepseek_v4): avoid TileLang boolx8 backward codegen

HuiyingLi:
# What does this PR do ?

Fix DeepSeek-V4 TileLang sparse-MLA backward compilation on Blackwell GPUs.

TileLang 0.1.11 vectorizes the per-block mask fragment. When the fragment is
stored as `bool`, CUDA code generation attempts to materialize `boolx8`, which
has no supported CUDA storage type. The Python interface already normalizes the
validity mask to 0/1 `int32`, so this change preserves that representation in
the kernel and forms scalar predicates only where they are consumed.

# Changelog

- Store the sparse-MLA backward validity fragment as `int32` instead of `bool`.
- Sum the normalized 0/1 mask directly when detecting all-invalid blocks.
- Preserve the existing masking semantics when computing attention probabilities.

# Validation

- 1x GB200, TileLang 0.1.11, AutoModel 26.06 container:
  `python3 -m pytest -q -s tests/unit_tests/models/deepseek_v4/test_dsv4_layers.py -k "test_sparse_attention_tilelang_backend_matches_torch"`
  - 2 passed, including the causal-padding shape.
  - Both cases compiled `sparse_mqa_bwd_kernel` and compared forward output plus Q, KV, and attention-sink gradients against the torch reference.
  - Slurm allocation: `5987874`.
- CPU DSV4 layer suite: 70 passed, 10 skipped.
- `ruff check`, `ruff format --check`, and `git diff --check`: passed.

**Pre checks**:

- [x] Existing forward/backward parity tests cover the changed kernel contract.
- [x] No documentation update is required for this internal code-generation fix.

# Additional Information

- The change is isolated to the vendored DeepSeek-V4 TileLang sparse-MLA backward kernel.


akoumpa (approved):

nemo_automodel/components/models/deepseek_v4/kernels/tilelang_sparse_mla_bwd.py
```diff
@@ -162,7 +162,13 @@ def sparse_mqa_bwd_kernel(
...
             KV_shared = T.alloc_shared([BS, D], dtype)
             dO_shared = T.alloc_shared([block_H, D], dtype)
-            mask = T.alloc_fragment([BS], "bool")
+            # TileLang 0.1.11 vectorizes this fragment store.  Keeping the
+            # predicate as bool makes CUDA codegen try to materialize boolx8,
+            # which has no CUDA storage type on Blackwell.  ValidMask is
+            # already normalized to 0/1 int32 by the Python interface, so keep
+            # it in that representation and only form scalar predicates at use
+            # sites.
+            mask = T.alloc_fragment([BS], T.int32)
```
akoumpa, line 171: @HuiyingLi any smaller int type?

PR #3472, opened 2026-08-10
fix(docs): generate model coverage tables at build time

akoumpa:
## What changed

- Fetch full Git history when building the CI test image so model support dates are available inside `/opt/Automodel`.
- Remove generated model-coverage tables from tracked MDX files, leaving only generation markers.
- Generate tables at docs validation, preview, and publish time using the existing generator integration.
- Consolidate each category overview into one generated **Supported Models** table.
- Derive recipe-backed support dates from the first commit containing the current checkpoint ID, including across recipe renames.
- Preserve every documented model family and its architecture metadata in the consolidated table; documentation-only rows use the family-page addition date.
- Use `Date | Type | Model | Recipe` in the combined release log and `Date | Model | Architectures | Recipe` on category pages.
- Add `--clean` to strip generated output before committing and a unit-test guard that rejects populated generated blocks.

## Why

Model coverage is derived from recipes, documentation metadata, and Git history. Committing generated tables duplicates those canonical sources and allows generated output to become stale. Separate static model-family and dated support-log tables also presented overlapping information on category overviews.

The support date must follow the current checkpoint, not merely the age of its recipe path. For example, the Ministral BF16 checkpoint entered its existing recipe on 2026-08-07, so it must not inherit that recipe files earlier date. Consolidation must also retain supported families that do not currently have a recipe.

## Impact

Git stores only canonical sources and empty generation markers. Local docs commands and Fern CI populate one dated Supported Models table before validation or rendering. Recipe-backed rows show the checkpoints actual introduction date and link their YAML from a dedicated Recipe column; documentation-only families remain visible with architecture metadata.

## Validation

- `22 passed` for `tests/unit_tests/tools/test_sync_model_coverage_tables.py`
- Regression coverage for replacing a checkpoint in an existing recipe
- Regression coverage ensuring every documented family appears in its generated category table
- Generated Ministral BF16 date verified as `2026-08-07`
- Generate/clean round trip restores empty markers
- `make -C docs/fern docs-check` — 469 MDX files validated, zero Fern errors
- Ruff passes for modified Python files
- `git diff --check`


akoumpa:
passing https://github.com/NVIDIA-NeMo/Automodel/actions/runs/31386325502/job/93454824940?pr=3472

PR #3484, opened 2026-08-10
fix(docker): harmonize non-root runtime contract

ko3n1g:
## Background / motivation

- ARC and other sandboxed runtimes may override the published image to UID/GID 65532.
- The current image exposes `uv` and `uvx` only below `/root`, while the checkout, virtual environment, and uv cache are not writable by that runtime identity.
- The published image must continue to run as root by default.
- Use the same non-root runtime contract as NVIDIA/Megatron-LM#6421.

## What changed

- Install pinned `uv` and `uvx` executables on the system path from architecture-specific release archives with pinned SHA-256 verification.
- Conditionally create the UID/GID 65532 runtime identity when the base image does not already define it.
- Explicitly own the runtime home, checkout, virtual environment, and uv cache paths needed by Automodel's documented `uv sync` workflow.
- Register `/opt/Automodel` as a system-wide Git `safe.directory` so the root default user can still read the runtime-owned checkout.
- Exercise the non-root runtime contract during the image build, then restore root as the final image user.
- Bound the FlashAttention 3 wheel build to `MAX_JOBS=16 NVCC_THREADS=2` so a 48-core builder does not exhaust node memory.

## Details

- Build stages and the final image remain root by default.
- `TARGETARCH=amd64` maps to `uv-x86_64-unknown-linux-gnu`; `TARGETARCH=arm64` maps to `uv-aarch64-unknown-linux-gnu`.
- uv 0.11.24 archive digests are pinned to `5ce1ad074a78f96c5c8122088bb85a12eb282195bc1453151a48762e4fc31fed` (amd64) and `e22c66d36a0098b17cff80a8647e0b8c58202af899d4e9eb820fc7ad126435a1` (arm64).
- `--user 65532:65532` can execute Python, `uv`, and `uvx` and can write the paths needed by the native test launcher.
- The build-time probe sets `HOME` and XDG cache paths explicitly, matching the launcher contract.
- Automodel retains `COPY --chown` and full virtual-environment ownership because its documented mounted-checkout workflow mutates both with `uv sync`.
- The change intentionally supports the sandbox's declared UID/GID; it does not claim arbitrary-UID write access.
- `COPY --chown=${RUNTIME_UID}:${RUNTIME_GID} . /opt/Automodel` also hands `/opt/Automodel/.git` to UID 65532, but CI runs the container with no `--user`, so tests execute as root. Git 2.35.2+ then rejects the checkout for dubious ownership and exits 128; root is not exempt, since the exemption keys off `SUDO_UID`.
- `git config --system` (`/etc/gitconfig`) is used instead of `--global` so the exception applies to root and to UID 65532 alike, `docker/Dockerfile:271-275`.
- This surfaced as three failures in `tests/unit_tests/tools/test_sync_model_coverage_tables.py`, which shell out to `git rev-parse --is-shallow-repository` via `tests/ci_tests/utils/sync_model_coverage_tables.py:174`. That helper re-raises without the captured stderr, so the log shows only `returned non-zero exit status 128`.

## Tested

- `docker buildx build --check --platform linux/amd64 --build-arg BASE_IMAGE=pytorch --target automodel_final -f docker/Dockerfile .` — completed with no warnings against exact head `9cf81d7fcc6a8ec05047b48339cedca7e8d1b5c8`.
- `git diff --check` — clean.
- Signed commit verification and DCO check — passed.
- Cold OCI build `oci-6ffd5837961f4154ddb3fe53` against exact head `9cf81d7fcc6a8ec05047b48339cedca7e8d1b5c8`, base `8b4db0075182b2f9d82e9c8f9342d911aea9ed6c`:
  - source reviewer passed with `{"findings":[],"verdict":"pass"}`;
  - pinned base resolved to `nvcr.io/nvidia/pytorch:26.06-py3@sha256:769ab7dfe6e04010060f2a4c010e98b8025cc07e160f4932e9c8c29a84396826`;
  - uv verification printed `/tmp/uv-x86_64-unknown-linux-gnu.tar.gz: OK`;
  - FA3 used `MAX_JOBS=16 NVCC_THREADS=2`, completed in 2,364 seconds, and produced wheel SHA-256 `a8a80ee9f518b898103e3cd39476c7a5410844010652b01b1858ff7f95f2f8a1`;
  - FA3 stayed near 16 CPU and at or below approximately 44 GiB with a current node lease and zero pod restarts, avoiding the prior unbounded-build node loss;
  - all 16 final-image steps passed, including the UID/GID 65532 write/executable probe; final image user remains root;
  - OCI export, ECR push, build-status AgentRun, and build-receipt AgentRun succeeded;
  - published `766267172432.dkr.ecr.us-east-1.amazonaws.com/nemo-autobot/agent-sandbox@sha256:ba736a999d94cf45d23fc069746b2748ac70d9543f8b060168441a10385107d1` (14,041,580,844 bytes);
  - owner-scoped cleanup AgentRun succeeded with `cleanup_errors: []`; the runner was intentionally interrupted before downstream functional/GPU scenarios.
- Reproduced and fixed the `safe.directory` regression inside `automodel` built from this branch, with a real `.git` directory owned by 65532 and the process running as root:
  ```
  git version 2.43.0
  # before
  $ git -C /tmp/repo rev-parse --is-shallow-repository
  fatal: detected dubious ownership in repository at '/tmp/repo'
  exit=128
  # after: git config --system --add safe.directory /tmp/repo
  $ git -C /tmp/repo rev-parse --is-shallow-repository      # root
  false
  exit=0
  $ setpriv --reuid=65532 --regid=65532 --clear-groups \
      git -C /tmp/repo rev-parse --is-shallow-repository    # runtime uid
  false
  exit=0
  ```
- Run [31880769309](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/31880769309) showed the pre-fix state: `L0_Unit_Tests_CPU` (3 failed, 11436 passed) and `L0_Unit_Tests_GPU_2_of_2` (3 failed, 6210 passed), with the same three `test_sync_model_coverage_tables.py` tests as the only failures.

## Pre-checks

- [x] Kept the default runtime identity unchanged.
- [x] Added a build-time non-root access probe.


thomasdhc (approved):
LGTM.

### Review notes
- Sourcing `/opt/venv/env.sh` now places `/usr/local/bin` ahead of `/opt/venv/bin`, reversing the venv-first precedence the file previously established.
- The system-wide `safe.directory` entry keeps the 65532-owned checkout readable by both root and the runtime identity, and the final image user is restored to root.

PR #3491, opened 2026-08-11
fix(models): preserve lm-head dtype boundaries

jQizhang:
# What does this PR do ?

Prevent mixed-dtype LM-head projections in Gemma4 and expose the nested Nemotron Omni output head through the standard model wrapper contract.

## Problem

With FSDP2 mixed precision, a custom model wrapper can receive FP32 hidden states while its LM-head weight is BF16. These wrapper-owned projections execute outside an autocast region, so `nn.Linear` fails with a `mat1 and mat2 must have the same dtype` error.

Gemma4 has two such custom projection paths. Nemotron Omni also keeps its output head under `language_model.lm_head`, while training consumers expect the conventional top-level `model.lm_head` contract.

# Changelog

- Add an opt-in, keyword-only `match_lm_head_dtype` path to `compute_lm_head_logits` that casts only the sliced projection input to the actual LM-head weight dtype.
- Enable that path for both Gemma4 custom LM-head projections.
- Expose Nemotron Omni's nested LM head through a read-only property without re-registering the module or changing checkpoint keys.
- Add focused CPU regression tests for dtype alignment, Gemma4 logits slicing, and Nemotron Omni module ownership.

# Validation

- `pytest tests/unit_tests/models/common/test_model_common_utils.py::TestComputeLmHeadLogits tests/unit_tests/models/gemma4_moe/test_gemma4_moe_tied_weights.py tests/unit_tests/models/nemotron_omni/test_nemotron_omni_layers.py -q` (`36 passed`)
- `ruff format --check` on all six changed files
- `ruff check` on all three changed production files
- NeMo RL end-to-end regressions: Gemma4 26B (20 steps), Gemma4 31B (20 steps), and Nemotron Omni MMPR (10 steps) completed successfully.

**Pre checks**:

- [x] Did you add or update any necessary documentation? (No documentation changes are required for this internal model-boundary fix.)

# Additional Information

This PR targets `main`. The `r0.6.0` label should be retained so the repository's post-merge cherry-pick workflow can create the release backport.

### DIFF
--- a/nemo_automodel/components/models/common/utils.py
+++ b/nemo_automodel/components/models/common/utils.py
@@ -20,6 +20,7 @@ from typing import Any, Literal
 
 import torch
 from torch import nn
+from torch.distributed.fsdp import FSDPModule
 from transformers.modeling_outputs import CausalLMOutputWithPast
 
 from nemo_automodel.shared.import_utils import safe_import_from
@@ -848,8 +849,6 @@ def compute_lm_head_logits(
     is_thd: bool = False,
     fp32_lm_head: bool = False,
     output_hidden_states: bool = False,
-    *,
-    match_lm_head_dtype: bool = False,
 ) -> CausalLMOutputWithPast:
     """Project hidden states through ``lm_head`` and wrap the result.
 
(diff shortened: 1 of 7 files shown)


nemo_automodel/components/models/common/utils.py
```diff
@@ -914,6 +918,11 @@ def compute_lm_head_logits(
                 sliced = hidden_states[:, slice_indices, :]
         if fp32_lm_head:
             logits = lm_head(sliced.float()).to(hidden_states.dtype)
+        elif match_lm_head_dtype:
+            lm_head_weight = getattr(lm_head, "weight", None)
+            if lm_head_weight is None:
+                raise ValueError("match_lm_head_dtype requires lm_head to expose a weight tensor")
+            logits = lm_head(sliced.to(lm_head_weight.dtype))
```
akoumpa: Hi @jQizhang , I'm wondering if we should just make this the default case ie
```
        if fp32_lm_head:
            logits = lm_head(sliced.float()).to(hidden_states.dtype)
        else:
            logits = lm_head(sliced.to(lm_head_weight.dtype))
```

-- 
if lm_head is wrapped with fsdp, then it's trickier, but my agent pushed a fix to handle this case.

akoumpa: ah ok looks like RL does
```
MixedPrecisionPolicy(
      param_dtype=torch.bfloat16,
      reduce_dtype=torch.float32,
      output_dtype=torch.float32,
  )
  ```
  so it's the hidden state that need to be down-cast to bf16

akoumpa (approved):

nemo_automodel/components/models/common/utils.py
jQizhang: Hi @akoumpa , Thanks for the review and the fix!

Issue #3507, opened 2026-08-12
MiniMax-M3: dense layers 0-2 silently lose context parallelism when ep_size > 1


coderaBruce:
# MiniMax-M3: dense layers 0-2 silently lose context parallelism when `ep_size > 1`

## Summary

When MiniMax-M3 is trained with both expert parallelism and context parallelism
(`ep_size > 1` and `cp_size > 1`), the three dense attention layers (0-2) receive **no CP
setup at all**. They attend only over this rank's shard of the sequence, while the 57 sparse
layers correctly attend the full sequence.

Nothing raises. The run trains and the loss descends, because 57 of 60 layers are still
correct — so the defect is invisible in normal monitoring.

**Both of the shipped M3 CP recipes are affected:**

| recipe | `cp_size` | `ep_size` | `backend.attn` | layers 0-2 see |
|---|--:|--:|---|---|
| `examples/vlm_finetune/minimax_m3/minimax_m3_vl_sft_cp2_medpix_2k.yaml` | 2 | 32 | `sdpa` | 1/2 of the sequence |
| `examples/vlm_finetune/minimax_m3/minimax_m3_vl_sft_tulu3_text_cp8_16k.yaml` | 8 | 32 | `sdpa` | 1/8 of the sequence |

(`minimax_m3_vl_lora_pp4ep8_8node.yaml` is unaffected: it runs `cp_size: 1`.)

## Root cause

Two files make assumptions about each other that do not hold together for a MoE model whose
attention classes are heterogeneous. Line numbers are from `f59b2425`; `main` is unchanged in
the relevant files.

1. **The sequence is sharded before the layer stack.**
   `components/models/minimax_m3_vl/model.py:757-758` round-robin shards `inputs_embeds`, and
   `model.py:195` then iterates all 60 layers. Every layer receives only this rank's shard.

2. **M3 uses two different attention classes.**
   `components/models/minimax_m3_vl/layers.py:563` gives the 57 sparse layers
   `MiniMaxM3CPSparseAttention`, which defines `setup_cp_attention`. `layers.py:567` gives
   dense layers 0-2 plain `MiniMaxM3Attention`, which does not.

3. **`apply_cp` dispatches by capability and has no fallback.**
   `components/moe/parallelizer.py:783-803` routes each block: TE `DotProductAttention` → TE's
   CP group; a module exposing `setup_cp_attention` → call it; anything else → `logger.warning`
[...]

PR #3508, opened 2026-08-12
fix(moe): raise instead of silently skipping CP setup for unsupported attention

coderaBruce:
## What

Turn the `else` branch of `apply_cp`'s attention dispatch from a `logger.warning` into a<br>`ValueError`.

## Why

By the time `apply_cp` runs, the sequence has already been sharded across the CP ranks. A block<br>that reaches this branch gets no CP wiring at all, so its queries attend **only this rank's****<br>****shard** — a fraction of the sequence, with a causal structure that no longer matches global<br>positions. Because this usually affects a minority of blocks, the loss still descends and the<br>run looks healthy, so the model comes out quietly wrong instead of visibly broken.

MiniMax-M3 hits this today, and both shipped M3 CP recipes are affected. Full diagnosis with<br>line references in NVIDIA-NeMo/Automodel#3507; the short version:

* `minimax_m3_vl/layers.py` gives layers 3-59 `MiniMaxM3CPSparseAttention` (has<br>`setup_cp_attention`) and layers 0-2 plain `MiniMaxM3Attention` (does not);
* all three M3 recipes run `backend.attn: sdpa`, so the TE branch does not apply either;
* with `ep_size > 1`, `infrastructure.py` gates off the generic dense-SDPA CP pass, so nothing<br>covers layers 0-2;
* they land here and are skipped with a single warning line.

## Scope of this PR

This makes the failure loud. It does **not** implement CP for M3's dense layers — that is the<br>model-side fix discussed in NVIDIA-NeMo/Automodel#3507, and I am happy to write it if maintainers agree on the<br>shape.

## Compatibility

This is a behaviour change: any configuration that currently reaches this branch will now fail<br>at setup instead of training. That is intentional — such a configuration is computing attention<br>over a fraction of its sequence — but it does mean a run that "worked" before may now stop.

Two things I would like maintainer judgement on:

1. Whether a raise is right here, or whether you would prefer an opt-out (env var / config flag)<br>for anyone who has a legitimate reason to skip a block.
2. Whether any in-tree model currently relies on the skip. The branch is only reachable for<br>`layer_type in ("full_attention", "sliding_attention")` with CP enabled, and such a block<br>always needs CP, so I believe the answer is no — but I have only checked the M3 path directly.

## Testing

The change is a single control-flow edit with no new logic; I verified the file parses and the<br>message renders. I have not run the full CI matrix — I do not have the multi-node hardware the<br>CP recipes need (`ep_size: 32` implies 4+ nodes). I have run MiniMax-M3 on a single 8×B300 node<br>at `cp_size=8, ep_size=8, backend.attn: sdpa` with LoRA, which is the configuration that<br>exposed the underlying problem.

### DIFF
--- a/nemo_automodel/components/moe/parallelizer.py
+++ b/nemo_automodel/components/moe/parallelizer.py
@@ -901,10 +901,19 @@ def apply_cp(model: torch.nn.Module, cp_mesh: DeviceMesh, cp_comm_type: str = "p
                 # installs its own SDPA hook + mask handling.
                 self_attn.setup_cp_attention(cp_mesh)
             else:
-                logger.warning(
-                    "Skipping CP setup for block with unsupported attention module "
-                    "(neither TE DotProductAttention nor model-owned setup_cp_attention): %s",
-                    type(attn_module).__name__ if attn_module is not None else type(self_attn).__name__,
+                # Skipping is not a safe default: the sequence is already sharded across the CP
+                # ranks by the time we get here, so an attention block left without CP silently
+                # attends only this rank's shard. The loss still descends when most blocks are
+                # set up correctly, which makes the resulting model quietly wrong rather than
+                # visibly broken. Fail instead, and say what to do about it.
+                raise ValueError(
+                    f"Context parallelism is enabled (cp_size={cp_mesh.size()}) but block "
+                    f"{_layer_id} has an attention module that provides no CP implementation: "
+                    f"{type(attn_module).__name__ if attn_module is not None else type(self_attn).__name__}. "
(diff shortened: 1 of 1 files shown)


coderaBruce:
Closing: the premise is wrong. @HuiyingLi ran cp1 vs cp8 and got loss parity, and they are right — I missed that torch's `context_parallel` installs the ring by monkeypatching `F.scaled_dot_product_attention` globally (see the `cp_dispatcher_suspended` docstring), so SDPA blocks that reach this `else` are still covered and nothing is silently skipped. Raising here would have broken working M3 CP runs. Full correction in #3507.

PR #3511, opened 2026-08-12
fix(minimax_m3_vl): declare and honour logits_to_keep so fused losses are kept

### LINKED ISSUE #3510: MiniMax-M3: forward does not declare logits_to_keep, so fused losses are silently replaced
### Describe the bug

`MiniMaxM3SparseForConditionalGeneration.forward` does not declare `logits_to_keep`; it absorbs it into `**kwargs`. Because `_supports_logits_to_keep` inspects the forward signature, the check returns `False` and the recipe silently replaces the configured loss:

[https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/recipes/vlm/finetune.py#L554-L556](<https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/recipes/vlm/finetune.py#L554-L556>)

```python
if not _supports_logits_to_keep(model) and not isinstance(self.loss_fn, MaskedCrossEntropy):
    logger.warning("logits_to_keep not found in model.forward. Using MaskedCrossEntropy instead.")
    self.loss_fn = MaskedCrossEntropy()
```

Two consequences:

1. **The full logits tensor is materialised.** M3's vocab is 200,064, so `[tokens, 200064]` is large at long context — at `cp=8` with a 200,704-token sequence that is 25,088 local tokens x 200,064 = ~10 GB in bf16.
2. **The configured** `loss_fn`**'s settings are dropped with it.** The replacement is a *fresh* `MaskedCrossEntropy()`, so a recipe that sets `fp32_upcast: false` loses it and the same tensor is upcast to fp32, roughly doubling the cost.

Callers already pass the kwarg unconditionally ([`finetune.py#L930`](<https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/recipes/vlm/finetune.py#L930>), [`#L1179`](<https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/recipes/vlm/finetune.py#L1179>)), so today it is accepted and then ignored — `**kwargs` swallows it and nothing in the forward reads it.

### Why this looks like an oversight rather than a design choice

The stricter check is correct for this kwarg. `_supports_seq_lens`, immediately below it in the same file, *does* accept `**kwargs` — because merely accepting `seq_lens` is harmless. Accepting `logits_to_keep` is not equivalent to honouring it: the model has to return hidden states instead of logits, so requiring an explicit parameter is a reasonable proxy for "this model implements the contract".

Every other custom model in `components/models` declares it — `minimax_m2`, `glm_moe_dsa`, `deepseek_v4`, `gemma4_moe`, `qwen3_moe`, `gpt_oss`, `llama`, `qwen3_next`, `glm4_moe`, `step3p7`. MiniMax-M3 is the only one that does not.

### Steps/Code to reproduce bug

Configure any M3 recipe with a fused loss, e.g.:

```yaml
loss_fn:
  _target_: nemo_automodel.components.loss.linear_ce.FusedLinearCrossEntropy
```

The run logs `logits_to_keep not found in model.forward. Using MaskedCrossEntropy instead.` and proceeds with full fp32 logits. Alternatively, without launching a job:

```python
import inspect
from nemo_automodel.components.models.minimax_m3_vl.model import (
    MiniMaxM3SparseForConditionalGeneration,
)

sig = inspect.signature(MiniMaxM3SparseForConditionalGeneration.forward)
print("logits_to_keep" in sig.parameters)  # False
```

### Expected behavior

`FusedLinearCrossEntropy` (and `ChunkedCrossEntropy`) should be usable with M3: `forward` declares `logits_to_keep` and, when it is set, returns `{"hidden_states": hidden}` without applying `lm_head`, so the logits tensor is never built and the configured loss survives.

### Environment overview

* Container: `nvcr.io/nvidia/nemo-automodel:26.06`
* Automodel: reproduced against `main` (`6551ee4`) and `f59b2425`
* Hardware: 8xB300, single node
* Model: `MiniMaxAI/MiniMax-M3`, `num_mtp_modules: 0`

### Additional context

Found while fine-tuning M3 with LoRA + `cp=8`. Separately reported: NVIDIA-NeMo/Automodel#3507 (dense layers 0-2 losing context parallelism at `ep_size > 1`).

`MiniMaxM3SparseForCausalLM` has the same gap, but it is not in the architecture registry and its `thd` branch reshapes the return value, so a fix there would need its own validation.

A PR with the minimal fix follows.

coderaBruce:
Fixes NVIDIA-NeMo/Automodel#3510.

### What

`MiniMaxM3SparseForConditionalGeneration.forward` now declares `logits_to_keep` and, when it is set, returns `{"hidden_states": hidden}` without applying `lm_head`.

### Why

The kwarg was absorbed by `**kwargs`, so `_supports_logits_to_keep` — which inspects the signature — returned `False` and `recipes/vlm/finetune.py:554-556` silently swapped any memory-efficient loss for a fresh `MaskedCrossEntropy()`. That both materialises the `[tokens, 200064]` logits tensor (\~10 GB bf16 at 25,088 local tokens) and discards the configured loss's `fp32_upcast: false`, upcasting it to fp32. Callers already pass `logits_to_keep=1` unconditionally, so the parameter was accepted and then ignored.

The signature check is deliberately stricter than `_supports_seq_lens` (which does accept `**kwargs`), and reasonably so: absorbing the kwarg does not mean the model returns hidden states. Every other custom model under `components/models` declares it — `minimax_m2`, `glm_moe_dsa`, `deepseek_v4`, `gemma4_moe`, `qwen3_moe`, `gpt_oss`, `llama`, `qwen3_next`, `glm4_moe`, `step3p7` — so this aligns M3 with the existing convention rather than introducing a new one.

### Notes on the two judgement calls

* **MTP raises rather than silently falling through.** `mtp_logits` needs full logits, so returning a `hidden_states` dict would break it. Raising `NotImplementedError` with the config knob to set matches the MTP-under-PP guard a few lines earlier in the same forward. Falling through to the logits path instead would reintroduce a silent downgrade, which is the bug being fixed.
* **The pipeline-parallel early return is untouched.** Non-final PP stages already forward hidden states as a plain tensor (`lm_head is None`); the new branch is gated on `not is_pp_stage` so that contract is unchanged.

`MiniMaxM3SparseForCausalLM` has the same gap but is not in the architecture registry, and its `thd` branch reshapes the return value, so it is deliberately left out of this change.

### Testing

* `inspect.signature(MiniMaxM3SparseForConditionalGeneration.forward)` now contains `logits_to_keep`, i.e. `_supports_logits_to_keep` returns `True` and the recipe no longer replaces the loss.
* Verified on 8xB300 (`nvcr.io/nvidia/nemo-automodel:26.06`) that an M3 LoRA run at `cp=8`, `ep=8`, 200,704 tokens trains with `FusedLinearCrossEntropy` retained; peak memory drops from 185.76 GiB to 140.12 GiB against the same config without this and two other local fixes, and step-0 loss/grad_norm stay in the same range.

I could not run the repo's full test suite on this hardware; happy to add a signature-level unit test in `tests/` if you would like one alongside.


athitten:
@coderaBruce thank you for the PR. Could you please add tests for this change, the code coverage is low.

coderaBruce:
@athitten added in 4a0ef77 — five tests in `tests/unit_tests/models/minimax_m3_vl/test_minimax_m3_logits_to_keep.py`, all CPU, on the existing tiny-config fixtures (~34 s locally):

- **`test_recipe_detects_logits_to_keep_support`** — asserts `_supports_logits_to_keep(model)`. This is the probe that silently downgraded the loss, so it is the regression guard that matters most: if the parameter ever slips back into `**kwargs`, this fails.
- **`test_logits_to_keep_returns_hidden_states_and_skips_lm_head`** — the return is a mapping containing `hidden_states` shaped `[.., hidden_size]`, not `[.., vocab_size]`, i.e. `lm_head` was not applied and the logits tensor was never built.
- **`test_without_logits_to_keep_still_returns_logits`** — default path unchanged.
- **`test_logits_to_keep_with_mtp_raises`** / **`test_mtp_without_logits_to_keep_is_unaffected`** — the MTP guard fires only on the fused path, and MTP training itself still returns `logits` + `mtp_per_depth_logits`.

One fixture added: `mtp_vlm_model`. The shared `mtp_model` builds `MiniMaxM3SparseForCausalLM`, whereas the branch under test is on the conditional-generation class.

athitten (dismissed):
LGTM, thank you @coderaBruce ! Will merge once the CI passes.

athitten:
@coderaBruce thanks for adding tests. Could you please fix linting issues ? The [CONTRIBUTING.md](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md) has info on how to fix it. Thank you!

coderaBruce:
@athitten done in d25af4b, and thank you for the return-type fix and the main merge.

The linting failure was the closing paren of the `forward` signature sitting at column 0 after the annotation change in 20c491d5 — valid Python, but `ruff format --check` wants it at the method's indentation. One-line fix:

```diff
-) -> torch.Tensor | MiniMaxM3CausalLMOutput | dict[str, torch.Tensor]:
+    ) -> torch.Tensor | MiniMaxM3CausalLMOutput | dict[str, torch.Tensor]:
```

`ruff check` and `ruff format --check` now both pass on the two files this PR touches.

One thing worth flagging rather than quietly fixing: with ruff 0.16.2, `ruff format --check .` reports **12 files** already needing reformatting on `upstream/main`, and 13 on this branch — the 13th being the one above. I left main's 12 alone so this PR stays scoped, but you may want to check whether CI pins an older ruff than `pip install ruff` currently resolves to.

athitten (approved):

PR #3514, opened 2026-08-12
feat(datasets): Datum record and collate-time THD sequence packing

HuiyingLi:
## Summary

Adds `Datum` — the typed single-example input record for training — and
collate-time THD sequence packing, so unpacked datasets get cross-sample
packing without an offline `pack_dataset` preprocessing pass.

Extracted from #2556 (Engine training API) so the data contract can be
reviewed on its own; the Engine is the follow-up consumer.

## What's added

- **`components/datasets/datum.py`**
  - `Datum(input_ids, loss_inputs)` — one sequence plus the float
    side-channel the loss consumes (`weights` / `logprobs` / `advantages`,
    and `target_tokens` → `labels` with the `weights==0 → ignore_index`
    masking convention). Mirrors the equivalent record in the tinker
    training API (and NeMo-RL's `DatumSpec` vocabulary).
  - `collate_datums(datums, packed=...)` — delegates to the **existing
    canonical collaters**: `default_collater` (padded `[B, T]`) or, with
    `packed=True`, one flat `[1, total_tokens]` THD pack. Float side-inputs
    ride the same axis (padded+stacked, or concatenated flat), and
    `seq_lens` lets callers split flat outputs back per example.
- **`components/datasets/utils.py`**
  - `pack_features_for_thd(features)` — the one concatenation helper: flat
    tokens/labels, per-sequence `position_id` resets, `seq_lens`. It emits
    the pre-packed record schema `packed_sequence_thd_collater` already
    accepts, so there is **no second packing implementation** — cu_seqlens,
    CP chunking, and TE kwargs all stay in the existing THD pipeline.
  - `thd_packing_collater(batch)` — YAML-selectable `collate_fn` built on
    the same helper: the sampler decides which examples share the batch;
    this packs them into a single THD row at collate time (e.g. gives
    `ChatDataset` real packing, which today falls into the
    one-sequence-per-pack synthesized path).

## Design notes

- Binning (which sequences share a pack / microbatch) deliberately stays
  with the caller — the sampler for SFT, the microbatcher for RL
  frameworks; offline `pack_dataset` remains the density-optimal option for
  large fixed datasets.
- `Datum` lives in `components/datasets` precisely so collation can reuse
  the canonical collaters instead of forking padding/packing math.

## Validation

- 76 new/updated CPU unit tests (`test_datum.py`, `test_utils.py`): record
  validation and masking convention, padded and packed collation schemas,
  side-input alignment and per-sample split round-trip, position-id resets,
  attention-mask trimming, error cases.
- Full `tests/unit_tests/datasets` suite: **1302 passed, 0 failed** (25.06
  container).

### DIFF
--- a/nemo_automodel/components/datasets/datum.py
+++ b/nemo_automodel/components/datasets/datum.py
@@ -1,0 +1,201 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Typed input contract for training: :class:`Datum` and :func:`collate_datums`.
+
+A ``Datum`` is the single-example input boundary between user/algorithm code
+(SFT, RL post-training) and the training loop. It lives in ``components.datasets``
+because feeding and collating examples is a data concern — and, crucially,
+because that lets :func:`collate_datums` **reuse the canonical collaters**
+(``default_collater`` for padded ``[B, T]`` and ``packed_sequence_thd_collater``
+for THD) instead of forking a second padding/packing implementation that could
+drift from them.
+
+The companion output contract (``ModelOutput`` and the per-token extraction
+helpers) lives in ``components.training`` — that side touches model logits, so
+it is a forward concern, not a dataset one.
+
+Conventions
+-----------
+* A ``Datum`` holds **one** sequence. ``input_ids`` is 1-D, shape ``[T]``.
+* ``loss_inputs`` carries everything the loss needs, aligned to ``input_ids``
+  token positions (length ``T``) for per-token entries:
+
+  ===============  =======================================================
+  key              meaning
+  ===============  =======================================================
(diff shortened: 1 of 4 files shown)


nemo_automodel/components/datasets/utils.py
```diff
@@ -400,6 +400,56 @@ def packed_sequence_thd_collater(batch):
     }
 
 
+def pack_features_for_thd(features: list[dict], *, ignore_index: int = -100) -> dict:
```
akoumpa, line 403: @HuiyingLi do we need this function? I'm not sure if it's correct for all cases (e.g., neat packing, etc).

nemo_automodel/components/datasets/datum.py
```diff
@@ -0,0 +1,201 @@
...
+        """
+        features: dict[str, list[int]] = {
+            "input_ids": self.input_ids.tolist(),
+            "attention_mask": [1] * self.seq_len,
+        }
+        if "target_tokens" in self.loss_fn_inputs:
+            labels = self.loss_fn_inputs["target_tokens"].clone()
+            weights = self.loss_fn_inputs.get("weights")
+            if weights is not None:
+                labels = labels.masked_fill(weights == 0, ignore_index)
```
akoumpa, line 121: does this work with answer-only masking or multi-turn?

nemo_automodel/components/datasets/utils.py
HuiyingLi: this function is used in the colate_dataum. It only put already binned/bundled examples together. We have no control of how to pack (e.g. neat packing) in the rl scenario it's in their hand.

nemo_automodel/components/datasets/datum.py
HuiyingLi: We have no knowledge of answer only or multiturn, we only turn the weights into masks.

akoumpa (approved):

PR #3521, opened 2026-08-12
refactor(models): re-organize models

akoumpa:
## What

Organize model code by the boundary that owns it:

```text
nemo_automodel/
├── retrieval/                  # retrieval and embedding task APIs and model families
├── components/flow_matching/  # reusable flow-matching pipeline and adapter contract
├── _transformers/models/      # transformer-bridge model implementations
└── _diffusers/models/         # diffusion-bridge model implementations and concrete adapters
```

The new `retrieval/` package owns:

- bi-encoder and cross-encoder auto-model entry points
- retrieval modeling and SentenceTransformers export
- in-batch-negative and retrieval state-dict helpers
- `llama_bidirectional`, `llama_nemotron_vl`, and `ministral_bidirectional`

`components/flow_matching/` remains in place. The reusable flow-matching pipeline, context, and base adapter contract stay there; only concrete architecture adapters live with the diffusion bridge under `_diffusers/models/`.

## Motivation

Bridge packages answer how a model is integrated with an upstream runtime. Retrieval is a user-facing task domain that can span backbones and runtimes, so making it a top-level package avoids treating its current Transformers implementation as permanent ownership. Components remain reusable algorithms and contracts rather than a home for model families.

This gives each package one clear reason to change:

- `retrieval/` changes when retrieval behavior or model coverage changes.
- `_transformers/` and `_diffusers/` change when bridge integration changes.
- `components/flow_matching/` changes when the reusable algorithm or adapter contract changes.

## Compatibility

Existing import paths continue to resolve lazily to the same module and class objects, with deprecation warnings where appropriate:

- `nemo_automodel.components.models.*`
- `nemo_automodel.models.*`
- former `_transformers` retrieval modules and model-family paths
- former `components.flow_matching.adapters.<architecture>` paths

Canonical retrieval imports and shipped retrieval recipe targets now point to `nemo_automodel.retrieval.*`. Registry loading is import-order safe, and the compatibility aliases do not eagerly load model packages.

Generic checkpoint code consumes a state-dict-adapter capability instead of importing a concrete retrieval adapter, so `components` does not depend upward on the task package.

## Verification

- Ruff lint and formatting: clean
- Python compilation: clean
- import-linter: 653 files, 1,541 dependencies, 0 broken contracts
- auto-model and package-location suite: 75 passed
- focused retrieval suite: 33 passed, 9 deselected
- broader retrieval/model relocation suite: 188 passed, 9 deselected
- registry suite: 61 passed, 1 skipped
- recipe suite: 64 passed
- checkpoint addon suite: 22 passed
- built-wheel content and isolated-wheel import/identity checks: passed

The 9 deselected SentenceTransformers round-trip cases require a newer SentenceTransformers module layout than the locally installed `sentence-transformers==2.6.1`; the failing import path was already present at the prior PR head and is unrelated to this relocation.


nemo_automodel/_transformers/models/deepseek_v32/__init__.py
HuiyingLi, line 1: Could we unify the notation? e.g. deepseek v3.2 -> deepseek v3_2? or something like that as long as it is unified. I believe we also have "3p5" for 3.5. Would be nice to taken care of altogether in this one. Thanks

nemo_automodel/retrieval/auto_model.py
```diff
@@ -0,0 +1,280 @@
...
+import gc
+import logging
+from typing import TYPE_CHECKING, List, Optional
+
+import torch
+from huggingface_hub import constants as hf_constants
+from torch.nn.attention import SDPBackend
+from transformers import PreTrainedModel
+
+from nemo_automodel._transformers.auto_model import (
```
oliverholworthy, line 26: One thing I’m wondering about is the dependency direction here. `retrieval.auto_model` uses private `_transformers` infrastructure, while `_transformers.registry` imports the concrete models from `retrieval.models`. 

That makes the task and bridge packages aware of each other, even though the split is intended to give each package one reason to change. Would it be cleaner to keep the Transformer-backed implementations under `_transformers/models`, with `retrieval/` owning the task APIs, modeling and export? 

Alternatively, could retrieval depend on a small public bridge contract rather than private helpers?

PR #3522, opened 2026-08-12
fix(moe): stabilize async DeepEP checkpointing

pzelasko:
## Summary

- apply the asynchronous setting to DeepEP combine as well as dispatch
- avoid replaying DeepEP dispatch/combine during activation-checkpoint recomputation
- stabilize zero-element routing metadata across mixed-precision recomputation

## Motivation

Asynchronous DeepEP execution previously covered dispatch but not combine. In addition, activation checkpointing could replay expert-parallel communication and reject harmless dtype differences on empty routed tensors during recomputation.

The changes keep non-empty tensor metadata checks intact while handling the zero-token case narrowly.

## Validation

- `pytest -q tests/unit_tests/moe/test_experts.py tests/unit_tests/moe/test_moe_shard_placement.py tests/unit_tests/moe/test_parallelizer.py tests/unit_tests/moe/test_token_dispatcher.py`
  - 179 passed, 26 skipped
- Ruff checks passed for all changed files
- Python syntax and `git diff --check` passed

### DIFF
--- a/nemo_automodel/components/models/common/utils.py
+++ b/nemo_automodel/components/models/common/utils.py
@@ -212,8 +212,8 @@ class BackendConfig:
             "uccl_ep" uses UCCL-EP for token dispatch across heterogeneous GPUs and NICs.
         dispatcher_share_token_dispatcher: Whether flex token dispatchers share a communication
             manager instance across MoE layers.
-        dispatcher_async_dispatch: Whether DeepEP/UCCL-EP dispatch should return asynchronously
-            and allocate dispatched tensors on the communication stream.
+        dispatcher_async_dispatch: Whether DeepEP/UCCL-EP dispatch and combine should return
+            asynchronously and allocate their outputs on the communication stream.
         enable_deepep: Removed and ignored. Logs a warning if set; configure "dispatcher"
             and "experts" explicitly instead.
         fake_balanced_gate: If True, replace the learned Gate with FakeBalancedGate
--- a/nemo_automodel/components/moe/experts.py
+++ b/nemo_automodel/components/moe/experts.py
@@ -857,6 +857,24 @@ def get_expert_activation_for_deepep(config: MoEConfig):
         raise ValueError(f"Invalid expert activation: {config.expert_activation}")
 
 
+def _stabilize_empty_routing_probs_dtype(
+    permuted_probs: torch.Tensor, compute_dtype: torch.dtype
+) -> torch.Tensor:
+    """Keep empty expert probabilities stable across checkpoint recomputation.
+
+    DeepEP may return an empty probability tensor in either the router dtype or
+    the activation dtype. With non-reentrant activation checkpointing, a later
+    recomputation can choose the other representation, and PyTorch checkpointing rejects
+    the resulting saved-tensor metadata change even though both tensors contain
+    zero elements. Empty probabilities have no numerical contribution, so
+    canonicalize only that case to the dispatched activation dtype. Non-empty
+    probabilities retain their original (normally FP32) router precision.
+    """
+    if permuted_probs.numel() == 0 and permuted_probs.dtype != compute_dtype:
+        return permuted_probs.to(compute_dtype)
+    return permuted_probs
+
+
 class GroupedExpertsDeepEP(nn.Module):
     """
     Sparse MoE implementation using grouped GEMM with DeepEP token dispatch.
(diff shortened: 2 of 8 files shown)


pzelasko:
/ok to test [7c11097](https://github.com/NVIDIA-NeMo/Automodel/pull/3522/commits/7c110978e97ddbd511d80ffdf608909ed7463adc)

pzelasko:
/ok to test [44065bd](https://github.com/NVIDIA-NeMo/Automodel/pull/3522/commits/44065bda03cb662e3bbf03d71c052e304f35c948)

HuiyingLi:
## The selective-AC change is dead code — `moe_comm_ops` is always empty

Running the expression this PR adds, in the training container (torch
`2.12.0a0+...nv26.04`, `deep_ep` installed and importable):

```
torch.ops.deepep.dispatch -> None
torch.ops.deepep.combine  -> None
moe_comm_ops = set()
[n for n in torch._C._dispatch_get_all_op_names() if "ep::" in n] -> []
```

`torch.ops.deepep` is a lazily-created `_OpNamespace`, and
`getattr(..., "dispatch", None)` swallows the `AttributeError` it raises for an
unregistered op — DeepEP registers no torch operators at all. So
`func in moe_comm_ops` is always `False` and the policy is byte-for-byte
equivalent to `main`.

Renaming the namespace won't help: `fused_dispatch`/`fused_combine` are
`torch.autograd.Function`s wrapping pybind calls into `Buffer.dispatch()` /
`Buffer.combine()` (`fused_a2a.py:182`, `:243`), and SAC policies run under
`__torch_dispatch__`, which only sees ATen ops. A C++ extension call never
reaches the dispatcher, so no policy can mark it `MUST_SAVE`.

Note the unit test passes only because the fixture stubs
`torch_stub.ops.deepep` into existence — it asserts a property of the stub, not
of the runtime.

To actually stop the replay you'd have to either register the comm as a real op
(`torch.library.custom_op` + `register_autograd`; DeepEP's `handle` is a plain
tensor tuple so it can cross an op boundary, but `recv_x` has a data-dependent
first dim and `MUST_SAVE` would pin the `topk`-expanded activation per layer), or
move dispatch/combine outside the checkpointed region by changing the AC
granularity. Either is a separate PR, and only worth it if the two extra a2a per
layer per backward actually measure — routing is already saved via the `topk`
`MUST_SAVE`, so this is perf, not correctness.

If that's not something you want to take on here, just dropping the
`moe_comm_ops` hunk and its stubbed test is fine — the other two changes fix the
`CheckpointError` on their own and the PR stands without it.

pzelasko:
/ok to test [6c12db0](https://github.com/NVIDIA-NeMo/Automodel/pull/3522/commits/6c12db070c6221b612ab48f358be93c87760a1df)

pzelasko:
@HuiyingLi great catch, sorry for that! I removed `moe_comm_ops` like you suggested.

HuiyingLi:
## nemo-ci before/after results

Ran the PR head (`44065bda`) against its merge-base (`33eb83d7`) on eos, same
recipes, same test selection — convergence (`llm`/`nightly`) and perf
(`benchmark`/`performance`). All 4 pipelines green, 8/8 jobs passed.

| | before (`33eb83d7`) | after (`44065bda`) |
|---|---|---|
| convergence | [62670234](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62670234) | [62670226](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62670226) |
| perf | [62670253](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62670253) | [62670246](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62670246) |

### Convergence — 50 steps

`moonlight_16b_te` is **bitwise identical before/after** — all 50 steps match
exactly on both loss and grad_norm.

`qwen3_moe_30b_te_deepep` tracks closely (max Δloss 3.1e-02 over 50 steps, both
curves land at 1.8865 vs 1.8865 at step 49) with the usual run-to-run jitter from
non-deterministic grouped-GEMM dgrad on the `experts: torch_mm` path. Step 0 loss
is identical, so the forward is unchanged. No convergence regression.

### Perf — Average MFU, excluding 10 warmup iters

| recipe | before | after | Δ |
|---|---|---|---|
| `qwen3_moe_30b_te_deepep` | 28.815% | 29.090% | +0.95% |
| `moonlight_16b_te_deepep` | 35.741% | 35.465% | −0.77% |

Opposite signs, both under 1%, single run each — noise, not signal. Peak memory
is identical before/after (48.11 GiB qwen, 48.03 GiB moonlight), which
independently corroborates that the `moe_comm_ops` hunk is inert: a working
`MUST_SAVE` on dispatch would pin the `topk`-expanded activation per layer and
show up here.

### Coverage gap

`dispatcher_async_dispatch` defaults to `false` and no CI recipe sets it, so
**none of these runs exercise the async-combine change** — the one change in this
PR that can plausibly move numerics or step time. Worth adding a recipe (or an
override) with the flag on before merging, otherwise that hunk ships untested.

HuiyingLi (approved):

PR #3344, opened 2026-07-31
fix(pp): preserve VLM media cursor with static metadata

yuhezhang-ai:
# What does this PR do ?

Fix the VLM pipeline-parallel media cursor for PyTorch's current static metadata API while preserving the static-metadata behavior introduced by #3290, and fix the training-checkpoint resume problem exposed once that path advanced.

PyTorch now exposes configured stage metadata through populated `_user_meta.inputs` and `_user_meta.outputs`. The VLM staging code mistook that state for dynamic metadata inference and installed a one-shot cursor reset hook. Because no metadata-only forward actually runs with static metadata, the hook fired on the first real microbatch and replayed its media for the second microbatch, eventually triggering the CUDA `masked_scatter_size_check` assertion.

This PR keeps static metadata enabled and narrows the cursor reset to the true dynamic-inference path. It also teaches the checkpoint robustness PP probe to discover the locked sequence length from either the legacy `inputs_meta` API or the current `_user_meta.inputs` API.

The later resume failure had a separate but small checkpoint cause: `dequantize_base_checkpoint` was also being applied to training-checkpoint loads. Training checkpoints are saved from the dequantized native model, so their state does not contain source-only FP8 activation-scale metadata. The load path now requests FP8 scale destinations only during base-checkpoint initialization.

# Changelog

- Recognize populated PyTorch `_user_meta` stage metadata as analytical/static metadata.
- Reset the VLM media cursor only when the schedule will actually run a dynamic metadata forward.
- Read the locked PP sequence length from both legacy and current PyTorch metadata APIs in checkpoint robustness testing.
- Request source FP8 scale metadata only when initializing from a base checkpoint, not when resuming a dequantized training checkpoint.
- Raise the tracked Mistral4 checkpoint-robustness CI allocation from 30 minutes to one hour because the resume phase begins near the old limit.
- Add focused CPU coverage for static and dynamic metadata behavior and for short-prompt padding under both APIs.
- Add a real CPU DCP save/resume round trip that fails if a training resume requests source-only FP8 activation scales.

# Before your PR is "Ready for review"

**Pre checks**:

- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [x] Did you add or update any necessary documentation? (No user-facing documentation change is required.)

Validation:

[...]

PR #3529, opened 2026-08-13
test(pp): add gemma4 parallelism parity tests [1/4]

athitten:
# What does this PR do ?

Adds GitHub CI tests that check pipeline (PP) and tensor (TP) parallelism actually compute the *right numbers*, using a small stand-in for Gemma4 31B. Parallelism tests din't exist for VLMs before this.

# Why

Our PP tests today only check that training doesn't crash only for LLMs. When the base container was bumped (#2983), PP broke for real models in nemo-ci but GitHub CI stayed green — every PP test used a plain HuggingFace Mixtral at `tp_size=1`, and no VLM was tested under PP at all.

**What actually broke.** `_precompute_stage_shapes` installs static stage metadata so PyTorch can skip its serial shape-inference chain. Commit `00f40419` of PR #2983 made that a no-op on PyTorch 2.13, so stages fell back to dynamic shape inference. The generic dense path still runs that way, so the existing tests passed as smoke tests while the real `gemma4_31b_tp4_pp2` / `tp4_pp4` and `mistral3p5` recipes broke in nemo-ci.

**Why a smoke test can't catch this.** Exit code 0 only proves nothing crashed. Sharding a model changes *how* a step is computed, never *what* it computes, so a correct run must produce the same loss and gradient-norm trajectory as a single-rank run on the same data. Wrong stage metadata, a gradient synced over the wrong group, a mis-sharded CP sequence, or a missing TP reduction all leave the run alive and just change the numbers — which is what these tests compare.

Loss and gradient norm are checked separately because they fail independently: loss is reduced before the gradient all-reduce, so a broken gradient sync leaves the loss curve intact while the norm drifts.

# How it works

1. Builds a tiny random Gemma4 at test time — same shape and settings as the real 31B recipe, just small. Nothing to download or upload to S3.
2. Trains it twice with the same seed and same data: once on 1 GPU, once on 2 GPUs with PP=2 (and separately TP=2).
3. Fails if the loss or the gradient norm from the two runs disagree.

# Changelog

- New `tests/functional_tests/parallelism/` — Gemma4 PP=2 and TP=2 parity tests
- New `L2_Parallelism` CI job (2 GPUs)
- The 3 existing PP tests now fail if PP silently falls back to dynamic shape inference
- 11 new CPU unit tests covering stage metadata at pp=2/4/8

# Verified

Ran on 2×H100 in container `26.08.rc5` (torch 2.13):

- Both GPU tests pass
- Revert PR #3290 → the new log guard catches it
- Revert PR #3344 → the Gemma4 PP test catches it (stage 0 crashes with the image tensors missing)
- The new unit tests fail on a #3290 revert, pass on main

# Series

One PR per model family, so each stays reviewable on its own:

| # | Model | Adds |
|---|---|---|
| 1/4 | gemma4 31B (this PR) | PP=2, TP=2 on a VLM |
| 2/4 | deepseek_v4 | EP and CP axes |
| 3/4 | qwen3_5_moe | EP and CP on a MoE with its own PP tensor contract |
| 4/4 | mistral3p5 | second VLM; its real recipe runs TP=8 x PP=8 |

# Known limits

- Runners have 2 GPUs, so one axis at a time. No `tp2×pp2`.

**Pre checks**:


thomasdhc (dismissed):

thomasdhc (dismissed):

tests/functional_tests/hf_transformer_finetune/L2_HF_Transformer_w_PP_Dynamic_SeqLen_SFT.sh
```diff
@@ -51,15 +51,31 @@ COMMON_ARGS=(
     --distributed.pipeline.scale_grads_in_schedule false
 )
 
+LOG_FILE=$(mktemp)
+trap 'rm -f "$LOG_FILE"' EXIT
+
+# Guard against bug in commit 00f40419 (PR #2983). Checked after each schedule
+# because the metadata path differs between PipelineScheduleSingle and ...Multi.
+assert_static_pp_metadata() {
```
akoumpa, line 62: Please make this test fail on the accumulated-gradient loss fixed by #3530. It has the right trigger (`global_batch_size=32`, `local_batch_size=8`, `dp_size=1` gives four outer accumulation calls, with sequence lengths varying between batches), but the only new assertion rejects dynamic metadata inference. Reverting `_preserve_grads_across_stage_reinit` would keep static mode and exit successfully while `_post_metadata_inference_cleanup` silently discards earlier gradients. Add an independent non-PP/reference run over the same initialized weights and data, then compare the full per-parameter accumulated gradients (and preferably parameters after one optimizer step) after at least one window where `seq_len` changes between outer batches. Loss/no-crash is insufficient because loss is computed before the gradient wipe; an aggregate norm alone can also hide parameter-level errors.

akoumpa (changes requested):
The dynamic-sequence test still needs a numerical gradient regression before this is sufficient.

tests/functional_tests/hf_transformer_finetune/L2_HF_Transformer_w_PP_Dynamic_SeqLen_SFT.sh
athitten: Addressed in https://github.com/NVIDIA-NeMo/Automodel/commit/1685f4f8b8e876d86a30756de745ff7a91f68f23

akoumpa (approved):

PR #3540, opened 2026-08-13
fix(fsdp): uniform reduce dtype and EP-local expert gradients

akoumpa:
# What

Makes the shipped `glm_5.2_tulu3_32k_tilelang_cp8` recipe train. It hit three defects stacked in one code path — FSDP2's post-backward reduce — each masking the next:

1. `AssertionError: FSDP reduce-scatter expects uniform gradient dtype but got {torch.bfloat16, torch.float32}` — **NVBug 6599894**
2. `Tried to allocate 36.00 GiB` in `foreach_reduce`'s `reduce_scatter_input` — **AMINT-274**
3. `chunk_cat: mixed Tensor and DTensor` — **AMINT-273**

All three are fixed here. The recipe now runs to completion at its full 512-GPU configuration (see CI evidence).

# Why

### 1. Mixed gradient dtype

`FSDPParamGroup.post_backward` builds the reduce-scatter input from two slots that hold different dtypes:

| source | dtype | when |
| --- | --- | --- |
| `unsharded_accumulated_grad` | `reduce_dtype` (fp32) | parameter received a gradient during an earlier no-sync micro-batch; `FSDPParam.to_accumulated_grad_if_needed` upcast it |
| `unsharded_param.grad` | `param_dtype` (bf16) | parameter's gradient joined only at the reducing post-backward |

`foreach_reduce` requires one dtype for the whole group, so a group holding both aborts.

Context parallelism makes this reachable: `configure_fsdp_unused_param_reduction` installs a zero fill for locally unused parameters, and that zero is a `param_dtype` gradient sitting next to peers that already hold `reduce_dtype` accumulations. Any `cp_size > 1` recipe with gradient accumulation can land there.

This is not a model or dtype-pinning problem. The fp32 gradients are FSDP2's normal accumulation; the bf16 one is the intruder.

### 2 & 3. Unsharded `DTensor` zeros reaching the reduce

The remaining two failures are the same root cause. On this PyTorch build, the *public* unused-parameter API — `FSDPModule.set_reduce_scatter_unused_params(True)` — appends `zeros_like(unsharded_param)` where the unsharded parameter is still a **`DTensor`**, and hands it to `foreach_reduce` directly instead of routing it through `FSDPParam._get_grad_inner_tensor`.

Two consequences, and which one you get depends only on group composition:

* `fsdp.chunk_cat` rejects a list mixing `Tensor` and `DTensor` → **AMINT-273**.
* When the group is uniformly `DTensor`, nothing rejects it, and FSDP sizes the copy-in buffer from the `DTensor`'s **global** numel instead of the EP-local shard → **AMINT-274**.

The 36 GiB is uniquely identified by arithmetic. GLM-5.2: hidden 6144, moe_inter 2048, 256 routed experts, gated:

```
256×6144×4096 + 256×2048×6144 = 9,663,676,416 elem × 4 B = 36.00 GiB
```

That is exactly **one MoE layer's global, un-EP-split expert gradient**. At the recipe's `ep_size=64` the local shard is 4 experts = 0.56 GiB — a 64× inflation. Instrumentation on [job 396644320](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396644320) confirmed it directly: the failing group is `layers.76…mlp.experts.{gate_and_up_projs (512,3072,4096), down_projs (512,1024,6144)}`, all 256 logged bf16 entries are `DTensor`, while every *passing* group in the same run is a plain `Tensor`.

# What changed

Three commits, each addressing one defect.

**1. `patch_fsdp_uniform_reduce_dtype`** widens the minority gradients inside `foreach_reduce`, the last point before the reduction:

```python
dtypes = {grad.dtype for grad in unsharded_grads}
if len(dtypes) > 1 and all(dtype.is_floating_point for dtype in dtypes):
    unsharded_grads[:] = [g if g.dtype is target else g.to(target) for g in unsharded_grads]
```

FSDP2's own bookkeeping (`unsharded_param.grad` / `unsharded_accumulated_grad`) is left exactly as upstream leaves it, and `foreach_reduce` copies these gradients into a `reduce_dtype` buffer immediately afterwards, so widening first changes no value. The in-place `[:]` assignment is deliberate: `foreach_reduce` frees gradients by clearing that list, and replacing the list object would defeat it. Uniform groups pass straight through, so the upstream assertion still fires for genuinely inconsistent gradients such as fp8 weights that fail to produce higher-precision ones.

**2. Always install the unused-parameter shim**, rather than only as a fallback when the public API is absent. The shim assigns the zero to `param.grad`, so upstream's *normal* gradient path runs and unwraps the EP-local tensor via `_get_grad_inner_tensor`. A `reduce_scatter_unused_params` guard keeps the process-global wrapper from changing FSDP units that were never configured for unused-param reduction.

**3. Localize any residual `DTensor`** in the reduce list, covering paths that reach `foreach_reduce` without going through the shim.

An earlier revision of this PR instead re-slotted gradients via `to_accumulated_grad_if_needed()` in `post_backward`. That mutates FSDP2's bookkeeping and regressed two checkpoint-robustness tests. The current design does not touch those slots.

# CI evidence

All runs on `eos` / `x86`, `release` scope, via `nemo-ci` `main-mirror`.

### The recipe now trains — scale-matched before/after

Same recipe, same cluster, differing only by commit:

| Config | before (`1065b1d26`, defect 3 present) | after (`f22d96361`, this PR head) |
| --- | --- | --- |
| **512 GPU** — `cp8 / ep64 / pp4` (the NVBug configuration) | OOM **36.00 GiB**, no step completed — [396812951](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396812951), [396817606](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396817606) | ✅ **step 49, loss 0.5900, 2 epochs** — [396937806](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396937806) |
| **256 GPU** — `cp8 / ep32 / pp4` | `chunk_cat` mixed `DTensor`, no step completed — [396902549](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396902549), [396926560](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396926560) | ✅ **step 1/2, loss 0.7167** — [396944715](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396944715) |

The 512-GPU passing run shows `mem 25.26 → 35.16 GiB` and no OOM or assertion anywhere in its log.

For reference, the un-patched baseline fails the same way: the 36 GiB OOM reproduces on plain `main` ([job 396279513](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396279513)) and on an earlier `main` from before this work began ([job 394555501](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/394555501)).

### Earlier assertion-only verification (defect 1)

| Pipeline | Ref | Result |
| --- | --- | --- |
| [62399126](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62399126) | fix (initial) | `uniform gradient dtype` assertion: **0 occurrences** (QA saw it on 36 ranks) |
| [62431917](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62431917) | fix (scope-reduced) | assertion still absent |
| [62569328](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62569328) | fix (current design) | assertion absent; `gpt_oss_20b` and `qwen3_moe_30b_te_deepep` both pass |

### Validated on r0.6.0 and on current `main`

The same 512-GPU recipe passes on every variant of the change, with final loss agreeing to four decimals:

| branch | head | 512-GPU result |
| --- | --- | --- |
| this PR | `f22d96361` | [396937806](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396937806) loss 0.5900 · [397299768](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/397299768) loss 0.5898 |
| `…-r060` (r0.6.0 backport) | `b8f6282c` | [397401915](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/397401915) loss 0.5898 |
| `…-rebased` (onto current `main`) | `dbbb2db7` | [397429251](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/397429251) loss 0.5897 |

All 50 steps, `cp8/ep64/pp4`, zero OOM / assertion / `chunk_cat` occurrences.

### `gpt_oss_20b` — resolved: pre-existing flakiness, not caused by this PR

`gpt_oss_20b` intermittently fails its checkpoint-robustness *Phase 4 HF reload parity* check against the recipe's hand-tuned `hf_kl_threshold: 1e-1`. Two things were needed to settle it, because an early 6-sample comparison looked like a real regression.

**1. The wrapper provably never fires for this recipe.** An instrumented build counted every entry into the patched `foreach_reduce` and every branch taken. Across 66–73 observations on all 8 ranks of two independent runs ([397799636](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/397799636), [397824828](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/397824828)), every line reads:

```
WIDENPROBE calls=100 localized=0 widened=0 dtypes_seen=['torch.float32'] this_call_dtensors=0
```

Zero `DTensor` localizations, zero widenings, every reduce group uniformly `float32`. The patch is a strict pass-through here — as the topology predicts (`WORLD_SIZE=8, ep_size=8, cp=pp=tp=1` ⇒ `ep_shard == 1`, so `configure_fsdp_unused_param_reduction` is skipped under `if cp_enabled:` and experts land in `ignored_params`).

**2. Unmodified `main` fails the same check.** Extending the paired comparison:

| arm | measured Phase-4 max KL (×1e-2) |
| --- | --- |
| current `main` (8 runs) | 2.11, 3.06, 4.04, 4.49, 5.97, 8.50, **11.00** ❌, **23.87** ❌ |
| with this change (10 runs) | 0.095, 4.77, 5.80, 9.57, **10.4** ❌, **13.5** ❌, **15.0** ❌, **20.1** ❌, **29.3** ❌, **32.0** ❌ |

Both arms straddle the gate, and the baseline reaches 2.4e-01 — inside the range this branch produces. The metric spans roughly 1e-3 to 3e-1 run-to-run on identical code, so a fixed `1e-1` constant sits in the middle of the noise band.

Phase 3 (`Automodel-from-consolidated max KL`) is **exactly `0.000000e+00`** in every run of both arms, so the deterministic reload path is bit-identical; only the HF-reload path moves.

This is the known class of problem with the hand-tuned `ci.checkpoint_robustness` constants and is out of scope here; the fix belongs in the metric (gate a scale-free panel against a per-recipe baseline band) rather than in this PR. An earlier revision of this description first dismissed these failures as noise on n=3, then over-corrected and called them a regression on n=6; both readings were premature.

# Tests

* `tests/functional_tests/training/test_cp_flce_fsdp_correctness.py` — 2-rank regression. Reverting the change reproduces the exact bug string.
* `tests/unit_tests/distributed/test_parallelizer_utils.py` — widening a mixed group, uniform groups untouched, non-float mixtures left for the upstream assertion, `DTensor` localization, patch idempotency, and install ordering relative to the zero fill. 34 pass.
* Local: 1306 unit tests pass; DTensor/expert-parallel probes confirm reduce groups stay EP-local.

# Known limitation (pre-existing, unchanged)

A unit skipped *entirely* on a rank still cannot be filled: FSDP2 creates `_unsharded_param` lazily in the forward pre-hook, so there is no tensor to shape a zero from, and no local signal that peers used it. That applies equally to the non-pipeline path today.

### DIFF
--- a/nemo_automodel/shared/torch_patches.py
+++ b/nemo_automodel/shared/torch_patches.py
@@ -50,7 +50,7 @@ def patch_fsdp_uniform_reduce_dtype() -> None:
 
     Gradient accumulation leaves a group holding ``reduce_dtype`` accumulations
     for the parameters used so far, while any parameter whose gradient joins
-    later -- a locally unused parameter zero-filled by
+    later -- a locally unused parameter zero-filled by PyTorch's public API or
     :func:`patch_fsdp_unused_param_reduction`, or one whose gradient lands after
     its group's post-backward already ran -- contributes ``param_dtype``.
     ``foreach_reduce`` then aborts with ``FSDP reduce-scatter expects uniform
@@ -108,16 +108,13 @@ def patch_fsdp_uniform_reduce_dtype() -> None:
 
 
 def patch_fsdp_unused_param_reduction() -> None:
-    """Make FSDP2 unused-parameter reduction safe across PyTorch versions.
+    """Backport FSDP2 unused-parameter reduction when the public API is absent.
 
     The patch is process-global and idempotent. It only fills a missing local
     gradient with zeros immediately before FSDP2 post-backward reduction, so
     ranks that skipped a parameter still participate in the same collective as
-    ranks that used it. This backports the behavior when
-    ``FSDPModule.set_reduce_scatter_unused_params`` is absent. It also works
-    around earlier public implementations that passed ``zeros_like`` of an
-    unsharded ``DTensor`` directly to ``foreach_reduce``: assigning the zero to
-    ``param.grad`` makes upstream use its normal local-tensor extraction path.
+    ranks that used it. Callers must first prefer the public
+    ``FSDPModule.set_reduce_scatter_unused_params`` API.
 
     Raises:
         RuntimeError: If the installed PyTorch exposes neither the public API
@@ -138,12 +135,7 @@ def patch_fsdp_unused_param_reduction() -> None:
         return
 
     def _post_backward_with_unused_param_reduction(self, *args, **kwargs):
-        # Newer versions expose this flag on the parameter group. Respect it so
-        # installing the process-global compatibility wrapper does not change
-        # unconfigured FSDP units. Versions predating the public API have no
-        # flag and are configured by installing this wrapper itself.
-        reduce_unused = getattr(self, "reduce_scatter_unused_params", True)
-        if self.reduce_grads and reduce_unused and self._training_state == TrainingState.PRE_BACKWARD:
+        if self.reduce_grads and self._training_state == TrainingState.PRE_BACKWARD:
             for fsdp_param in self.fsdp_params:
                 if not hasattr(fsdp_param, "_unsharded_param"):
                     continue
@@ -152,10 +144,10 @@ def patch_fsdp_unused_param_reduction() -> None:
                 param = fsdp_param.unsharded_param
                 if param.requires_grad and param.grad is None:
                     # ``zeros_like`` on the *unsharded* parameter is the dtype
-                    # autograd would have produced under any precision policy.
-                    # For a DTensor this still allocates only local storage;
-                    # upstream then unwraps that local tensor through the same
-                    # path as a real gradient before calling ``foreach_reduce``.
+                    # autograd would have produced under any precision policy
+                    # (bf16 compute over fp32 storage, an fp32-pinned unit, or no
+                    # casting at all). ``_align_accumulated_grad_dtype`` then
+                    # promotes it to ``reduce_dtype`` if the group is accumulating.
                     param.grad = torch.zeros_like(param, memory_format=torch.preserve_format)
         return original_post_backward(self, *args, **kwargs)
 
--- a/nemo_automodel/shared/torch_patches.py
+++ b/nemo_automodel/shared/torch_patches.py
@@ -46,7 +46,7 @@ def _widest_float_dtype(dtypes: Iterable[Any]) -> Any:
 
 
 def patch_fsdp_uniform_reduce_dtype() -> None:
-    """Give every FSDP2 reduce-scatter group a single gradient dtype.
+    """Give every FSDP2 reduce-scatter group local gradients of one dtype.
 
     Gradient accumulation leaves a group holding ``reduce_dtype`` accumulations
     for the parameters used so far, while any parameter whose gradient joins
@@ -56,12 +56,14 @@ def patch_fsdp_uniform_reduce_dtype() -> None:
     ``foreach_reduce`` then aborts with ``FSDP reduce-scatter expects uniform
     gradient dtype``.
 
-    Widen the minority gradients at the last possible moment, inside
+    Normalize and widen gradients at the last possible moment, inside
     ``foreach_reduce`` itself. That placement matters:
 
-    * the gradients are already plain local tensors here (``FSDPParam``'s
-      ``_get_grad_inner_tensor`` has unwrapped any ``DTensor``), so nothing has
-      to reason about distributed wrappers;
+    * ``FSDPParam`` normally unwraps gradients through
+      ``_get_grad_inner_tensor``. PyTorch versions whose public unused-parameter
+      API appends ``zeros_like(unsharded_param)`` directly can still leave a
+      ``DTensor`` in this list, so unwrap that residual value before sizing the
+      reduce-scatter buffer;
     * FSDP2's own bookkeeping (``unsharded_param.grad`` /
       ``unsharded_accumulated_grad``) is left exactly as upstream leaves it, so
       no later reader of that state sees anything unusual;
@@ -83,6 +85,15 @@ def patch_fsdp_uniform_reduce_dtype() -> None:
         return
 
     def foreach_reduce_uniform_dtype(fsdp_params, unsharded_grads, *args, **kwargs):
+        from torch.distributed.tensor import DTensor
+
+        # PyTorch 2.13a0's unused-parameter branch can append a DTensor zero
+        # directly, while gradients from used parameters are already local
+        # tensors. Besides making ``fsdp.chunk_cat`` reject the mixed list, the
+        # DTensor's global numel makes FSDP size a global-shape staging buffer.
+        # Current PyTorch routes the zero through ``_get_grad_inner_tensor``;
+        # localizing here is the equivalent compatibility path for that build.
+        unsharded_grads[:] = [grad.to_local() if isinstance(grad, DTensor) else grad for grad in unsharded_grads]
         dtypes = {grad.dtype for grad in unsharded_grads}
         if len(dtypes) > 1 and all(dtype.is_floating_point for dtype in dtypes):
             target = _widest_float_dtype(dtypes)
--- a/nemo_automodel/shared/torch_patches.py
+++ b/nemo_automodel/shared/torch_patches.py
@@ -22,10 +22,78 @@ that already depend on torch (training / distributed / dataloading).
 from __future__ import annotations
 
 import logging
+from collections.abc import Iterable
+from typing import Any
 
 _logger = logging.getLogger(__name__)
 
 _TORCH_PATCHES_APPLIED = False
+
+
+def _widest_float_dtype(dtypes: Iterable[Any]) -> Any:
+    """Return the float dtype among ``dtypes`` that every other one converts into losslessly.
+
+    Args:
+        dtypes: Gradient dtypes from a single reduce-scatter group.
+
+    Returns:
+        The dtype with the largest element size; ties resolve to float32 over
+        the 2-byte float types.
+    """
+    import torch
+
+    return max(dtypes, key=lambda dtype: (torch.finfo(dtype).bits, dtype is torch.float32))
+
+
+def patch_fsdp_uniform_reduce_dtype() -> None:
+    """Give every FSDP2 reduce-scatter group a single gradient dtype.
+
+    Gradient accumulation leaves a group holding ``reduce_dtype`` accumulations
+    for the parameters used so far, while any parameter whose gradient joins
+    later -- a locally unused parameter zero-filled by
+    :func:`patch_fsdp_unused_param_reduction`, or one whose gradient lands after
+    its group's post-backward already ran -- contributes ``param_dtype``.
+    ``foreach_reduce`` then aborts with ``FSDP reduce-scatter expects uniform
+    gradient dtype``.
+
+    Widen the minority gradients at the last possible moment, inside
+    ``foreach_reduce`` itself. That placement matters:
+
+    * the gradients are already plain local tensors here (``FSDPParam``'s
+      ``_get_grad_inner_tensor`` has unwrapped any ``DTensor``), so nothing has
+      to reason about distributed wrappers;
+    * FSDP2's own bookkeeping (``unsharded_param.grad`` /
+      ``unsharded_accumulated_grad``) is left exactly as upstream leaves it, so
+      no later reader of that state sees anything unusual;
+    * ``foreach_reduce`` immediately copies these gradients into a
+      ``reduce_dtype`` buffer anyway, so widening first changes no value.
+
+    Uniform groups are passed straight through, so the upstream assertion still
+    fires for genuinely inconsistent gradients such as fp8 weights that fail to
+    produce higher-precision ones. The patch is process-global and idempotent.
+    """
+    try:
+        import torch.distributed.fsdp._fully_shard._fsdp_collectives as collectives
+        import torch.distributed.fsdp._fully_shard._fsdp_param_group as param_group
+    except ImportError:
+        return
+
+    original_foreach_reduce = collectives.foreach_reduce
+    if getattr(original_foreach_reduce, "_automodel_uniform_reduce_dtype", False):
+        return
+
+    def foreach_reduce_uniform_dtype(fsdp_params, unsharded_grads, *args, **kwargs):
+        dtypes = {grad.dtype for grad in unsharded_grads}
+        if len(dtypes) > 1 and all(dtype.is_floating_point for dtype in dtypes):
+            target = _widest_float_dtype(dtypes)
+            # Mutate in place: ``foreach_reduce`` frees the gradients by clearing
+            # this list, and that must still release the caller's references.
+            unsharded_grads[:] = [grad if grad.dtype is target else grad.to(target) for grad in unsharded_grads]
+        return original_foreach_reduce(fsdp_params, unsharded_grads, *args, **kwargs)
+
+    foreach_reduce_uniform_dtype._automodel_uniform_reduce_dtype = True
+    collectives.foreach_reduce = foreach_reduce_uniform_dtype
+    param_group.foreach_reduce = foreach_reduce_uniform_dtype
 
 
 def patch_fsdp_unused_param_reduction() -> None:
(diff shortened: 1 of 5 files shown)


akoumpa:
AMINT-274 root cause and fix update (head `1065b1d26203d163d9b52b08007aa70930dce7e1`):

- The failing image uses PyTorch `2.13.0a0+8145d630e8.nv26.06`. In that revision, `FSDPParamGroup.post_backward()` handles an unused parameter by appending `torch.zeros_like(fsdp_param.unsharded_param)` directly to `foreach_reduce`.
- For EP parameters that object is a DTensor with the global expert shape. The GLM-5.2 gate/up + down group is 9,663,676,416 fp32 elements = exactly the reported 36 GiB copy-in allocation.
- Real gradients use `unsharded_grad_data`, which unwraps the EP-local tensor. Newer PyTorch similarly fixed the unused path with `unsharded_zero_grad_data`.
- This commit installs Automodel's existing zero-fill compatibility wrapper even when the affected public API exists. It puts the zero in `param.grad` first, so upstream follows its normal DTensor-to-local extraction path. Unconfigured FSDP units remain untouched.

Validation:
- Focused regression: `33 passed`.
- Distributed + MoE unit suites: `1240 passed, 53 skipped`.
- Exact failing production image, 8xH100:
  - unpatched control Slurm `5847977`: reproduced mixed `Tensor`/`DTensor` at `fsdp.chunk_cat`;
  - patched Slurm `5847992`: exit 0; used and unused expert inputs were both plain EP-local tensors (`(2,32,32)` and `(2,16,32)`).
- Exact-head image/pipeline: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62634622
- Singleton-free exact 64-node recipe rerun: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396817606 (Slurm `5848000`, queued on EOS priority with no dependency).

No existing CI or Slurm work was canceled.

akoumpa:
Exact-head artifact check: Slurm `5848034` ran the same 8xH100 unused-EP-gradient probe directly from `main-mirror/automodel:pipe.62634622` with no code overlays. It exited 0 on PyTorch `2.13.0a0+8145d630e8.nv26.06`; both used and unused reduce inputs were plain EP-local `Tensor`s. This confirms the image used by full leaf job `396817606` contains the fix.

akoumpa:
Validation correction/update:

- The one-node artifact probe on `1065b1d26` was insufficient for the real EP-shard2 path.
- Exact-head 32-node smoke [GitLab job 396902549](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396902549), Slurm `5848660`, reached the first backward with zero 36 GiB allocations, then failed because `foreach_reduce` still received a mixed local `Tensor` / residual unused-gradient `DTensor` list.
- Commit `f22d96361` now localizes every residual `DTensor` at the shared `foreach_reduce` boundary before dtype alignment and buffer sizing. This mirrors current PyTorch's `unsharded_zero_grad_data -> _get_grad_inner_tensor` behavior while covering the affected 2.13a0 inline branch.
- Added a regression that supplies a local Tensor plus a global-shape DTensor and verifies both reach the reduce as local tensors/local numel.
- Local validation: `1241 passed, 53 skipped`; Ruff/format/diff checks clean.

Fresh exact-SHA image build + full recipe: [nemo-ci pipeline 62650774](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62650774). I am not treating this as resolved until that artifact passes the real recipe.

akoumpa:
Corrected-head distributed validation update (`f22d9636151b46be363a290f0df3f238151fff60`):

- Exact built artifact: `main-mirror/automodel:pipe.62650774` (no source overlays)
- 8-GPU real-DTensor artifact probe: Slurm `5849062`, `COMPLETED 0:0`; used and unused expert gradients both reached FSDP as plain EP-local tensors
- 256-GPU EP-shard2 / multi-microbatch smoke: [GitLab job 396944715](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396944715), leaf [pipeline 62652272](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62652272), Slurm `5848905`, `COMPLETED 0:0`
  - topology: world 256, PP4 / CP8 / EP32, local/global batch 4/32, seq 4096
  - both requested steps completed: loss `1.0061 -> 0.7167`, memory `25.26/35.16 GiB`
  - zero 36 GiB allocations, mixed Tensor/DTensor errors, or training tracebacks
- Old-head 512-GPU control (`1065b1d2`): Slurm `5848000`, `FAILED 15:0` at 25:38 with the original 36.00 GiB global-shape allocation, confirming the A/B boundary.

The exact 512-GPU GLM-5.2 recipe is still queued as Slurm `5848881` under [GitLab job 396937806](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396937806). I am keeping it alive through terminal status before calling the full validation complete.

akoumpa:
Full AMINT-274 validation is complete on corrected head `f22d9636151b46be363a290f0df3f238151fff60`.

Exact 512-GPU recipe:

- [GitLab job 396937806](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396937806)
- root / child / leaf pipelines: [62650774](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62650774) / [62651476](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62651476) / [62651643](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/62651643), all `success`
- Slurm `5848881`: `COMPLETED 0:0` in 43:11
- identity/topology: artifact `pipe.62650774`, Automodel `0.5.0+f22d96361`, world 512, PP4 / CP8 / EP64, local/global batch 4/64, 32,768 sequence, 50 steps
- all steps 0–49 completed; step 49 loss `0.5900`, memory `37.33 GiB`, throughput `95,665.67` tokens/s
- zero 36 GiB allocations, mixed Tensor/DTensor exceptions, training tracebacks, or runtime errors

A/B reference: old-head Slurm `5848000` failed at 25:38 with the original 36.00 GiB global-shape allocation; corrected head crossed that boundary and completed all 50 steps. The 256-GPU EP-shard2 smoke and 8-GPU real-DTensor artifact probe also remain terminal-success as recorded above.

nemo_automodel/shared/torch_patches.py
```diff
@@ -56,14 +138,24 @@ def patch_fsdp_unused_param_reduction() -> None:
         return
 
     def _post_backward_with_unused_param_reduction(self, *args, **kwargs):
-        if self.reduce_grads and self._training_state == TrainingState.PRE_BACKWARD:
+        # Newer versions expose this flag on the parameter group. Respect it so
+        # installing the process-global compatibility wrapper does not change
+        # unconfigured FSDP units. Versions predating the public API have no
+        # flag and are configured by installing this wrapper itself.
+        reduce_unused = getattr(self, "reduce_scatter_unused_params", True)
+        if self.reduce_grads and reduce_unused and self._training_state == TrainingState.PRE_BACKWARD:
```
HuiyingLi: ## Patch 2 (`fix(fsdp): reduce unused expert gradients from local tensors`) appears to be inert under pipeline parallelism

The zero-fill shim installed by `patch_fsdp_unused_param_reduction` is gated on:

```python
if self.reduce_grads and reduce_unused and self._training_state == TrainingState.PRE_BACKWARD:
```

Under PP, `moe/fsdp_mixin` drives the reduction by hand, outside autograd — [`_run_post_backward_hooks`](https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/components/moe/fsdp_mixin.py#L114) calls `state._fsdp_param_group.post_backward()` directly. By then the training state has returned to `IDLE`, so the two halves of the gate land on different calls and it never opens.

### Measurement

Instrumented the shim's gate (log the three operands on every call). Qwen3-MoE, 8×H100, `ep_shard=2` in both arms, PR head `f22d96361`, torch `2.13.0a0+8145d630e8.nv26.06`:

| arm | `gate=True` | breakdown of all calls |
| --- | --- | --- |
| `cp=2, ep=4` (no PP) | **264** | 264 × `reduce_grads=True, state=PRE_BACKWARD` → open<br>264 × `reduce_grads=False` → closed (no-sync micro-batches) |
| `pp=2, cp=2, ep=2` | **0 / 792** | 264 × `reduce_grads=True, state=IDLE` → **closed**<br>528 × `reduce_grads=False` → closed |

`reduce_unused=True` in both arms, so `configure_fsdp_unused_param_reduction` did take effect — the public flag is on, only the shim's own gate stays shut.

### Why this matters for this PR

`glm_5.2_tulu3_32k_tilelang_cp8` runs `pp4`. If the gate is shut there too, the zero fill is still being produced by the PyTorch build's public API — the one that appends `zeros_like(unsharded_param)` as a global-shaped DTensor — which is precisely what patch 2 was meant to bypass.

That matches the CI history in this PR: after patch 2 (`1065b1d26`), [job 396902549](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/396902549) still failed with `foreach_reduce` receiving a mixed `Tensor` / `DTensor` list, and patch 3 was added in response.

**Implication:** on the GLM path, patch 3 is likely doing all the work and patch 2 is a no-op. Worth confirming before deciding whether patch 2 should be kept, dropped, or have its gate fixed.

HuiyingLi (approved):

nemo_automodel/shared/torch_patches.py
akoumpa: Confirmed — your PP-state analysis is correct.

I reverted the always-on Patch 2 path in `da17ffb15`: Automodel now prefers `FSDPModule.set_reduce_scatter_unused_params` and installs the private shim only as a legacy fallback. A focused regression test asserts that the public-API path does not install the shim.

The validation history explains why Patch 2 looked necessary:

- The original evidence was an 8-GPU, no-PP probe: Slurm `5847977` reproduced mixed Tensor/DTensor, `5847992` passed with the Patch 2 overlay, and `5848034` passed the Patch 2 image.
- The representative PP jobs then disproved it: 512-GPU jobs `396812951` / `396817606` still hit the 36 GiB OOM, and 256-GPU jobs `396902549` / `396926560` still hit mixed Tensor/DTensor.
- Patch 3 fixed those PP paths: historical current-design jobs `396937806` (512 GPU) and `396944715` (256 GPU) passed.

Fresh evidence after removing Patch 2:

- Local focused suites: `63 passed`; 2-GPU CP/FSDP correctness: pass.
- Original PyTorch 2.13 image with reverted source overlay: Slurm `5854541`, exit 0; both used and intentionally unused expert reduce inputs were plain EP-local tensors.
- Exact pushed image `pipe.62777428` from `da17ffb15`: build job `397920296` passed; artifact probe Slurm `5854625` passed with the same EP-local tensor evidence.
- Exact-head scale replays are live but still waiting for an `eos-nemo-ci` runner: [512 GPU job 397953147](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/397953147) and [256 GPU job 397953795](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/397953795). Neither has reached Slurm yet, so I am not counting them as passes.

This leaves Patch 3 as the production-path fix and keeps Patch 2 only for PyTorch versions that lack the public API.

PR #3543, opened 2026-08-13
fix(mtp): support context-parallel caller inputs

Slyne:
# What does this PR do?

Adds opt-in interfaces that allow callers to prepare MTP future-token inputs and targets in global sequence order before context-parallel sharding.

Without these interfaces, MTP rolls rank-local token IDs, position IDs, and labels after CP sharding, producing incorrect values around CP shard and packed-sequence boundaries.

This is an enabling interface PR. Native `train_ft` integration is provided by the stacked follow-up #3544.

# Changelog

- Allow `calculate_mtp_loss` to consume authoritative precomputed per-depth targets.
  - Targets must match the local label shape and CP token layout.
  - Target count must match the number of MTP outputs.
  - Precomputed targets are mutually exclusive with local `cu_seqlens`/`seq_idx` rolling because rank-local metadata cannot repair a global CP shift.
- Allow `MTPModule` to consume precomputed per-depth position IDs instead of rolling rank-local positions.
  - Validate their shape even when base `position_ids` are absent.
- Expose precomputed MTP position IDs through `NemotronHForCausalLM`.
  - Keep SDPA/flex MTP tensors in BSHD when a packed batch carries `qkv_format="thd"`; only TE uses the THD squeeze.
- Configure Nemotron MTP attention and Mamba layers with the same CP collectives as the backbone layers.
  - Fail explicitly if MTP is enabled but its layers cannot be wired.
  - Resolve CP process-group ranks once rather than once per attention layer.
- Add a shared packed-sequence shift helper and use it in the hybrid functional test.
- Extend the hybrid Nemotron CP functional test with packed-THD MTP parity coverage for exact targets, logits, hidden states, loss, gradients, and an optimizer update.

# Scope

This PR intentionally does not integrate MTP preprocessing into the native Automodel `train_ft` recipe.

Callers must shift and boundary-mask MTP inputs and targets in global sequence order, apply the same CP partition used for the model inputs, and pass the resulting local tensors through the new interfaces.

The new arguments are optional, so existing non-CP MTP behavior remains unchanged. No additional all-gather is introduced.

# Validation

- [x] Focused unit tests: 87 passed, 2 skipped
- [x] Ruff formatting and lint
- [x] Python compilation and `git diff --check`
- [x] Two-GPU IAD packed-THD MTP parity test
  - exact global target reconstruction
  - logits, hidden-state, loss, and gradient parity
  - optimizer-step parity

**Pre checks**:

# Additional Information

- Follow-up: #3544 adds native `train_ft` MTP + CP integration.
- MTP + CP + PP transport remains a separate follow-up.
- This PR is independent of the per-depth MTP metric API in #3525.

### DIFF
--- a/nemo_automodel/components/distributed/parallelizer_utils.py
+++ b/nemo_automodel/components/distributed/parallelizer_utils.py
@@ -37,7 +37,9 @@ UniformSubtreeItem = Union[Tuple[nn.Module, torch.dtype], Tuple[str, nn.Module, 
 
 def reject_unsupported_mtp_cp_pp(model: nn.Module) -> None:
     """Reject MTP+CP on every trimmed pipeline stage before CP collectives."""
-    supports = model.supports
+    supports = getattr(model, "supports", None)
+    if supports is None:
+        return
     is_pp_stage_fn = getattr(model, "_is_pipeline_parallel_stage", None)
     if supports.mtp_enabled and not supports.supports_mtp_cp_pp and callable(is_pp_stage_fn) and is_pp_stage_fn():
         raise NotImplementedError(
--- a/nemo_automodel/components/distributed/parallelizer_utils.py
+++ b/nemo_automodel/components/distributed/parallelizer_utils.py
@@ -35,6 +35,14 @@ from nemo_automodel.shared.torch_patches import (
 UniformSubtreeItem = Union[Tuple[nn.Module, torch.dtype], Tuple[str, nn.Module, torch.dtype]]
 
 
+def reject_unsupported_mtp_cp(model: nn.Module) -> None:
+    """Reject enabled MTP when the model has not declared CP support."""
+    supports = getattr(model, "supports", None)
+    mtp_enabled = bool(getattr(supports, "mtp_enabled", getattr(getattr(model, "mtp_config", None), "enabled", False)))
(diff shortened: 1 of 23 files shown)


nemo_automodel/components/distributed/parallelizer_utils.py
```diff
@@ -32,6 +32,16 @@
 UniformSubtreeItem = Union[Tuple[nn.Module, torch.dtype], Tuple[str, nn.Module, torch.dtype]]
 
 
+def reject_unsupported_mtp_cp_pp(model: nn.Module) -> None:
+    """Reject MTP+CP on every trimmed pipeline stage before CP collectives."""
+    mtp_enabled = bool(getattr(getattr(model, "mtp_config", None), "enabled", False))
```
HuiyingLi: use capability if possible, wdyt

Slyne: Done. The guard now uses the explicit `model.supports` capability contract (`mtp_enabled` and `supports_mtp_cp_pp`) instead of inferring model structure. It also safely no-ops when capabilities have not been attached, with direct regression coverage for a model without `.supports`.

HuiyingLi (approved):

Slyne:
Great thanks to Huiying for helping out!

PR #3569, opened 2026-08-18
fix(perf): honor configured peak in MFU logging

yuhezhang-ai:
## Summary

- require every low-level `calculate_mfu` caller to provide an explicit reference peak instead of inheriting the ambiguous `1979` default
- make `AutoMFU` detect the current device by default and use its dense-BF16 peak from the device table
- allow `mfu.device` and `mfu.peak_tflops` recipe overrides when hardware detection or the dense-BF16 convention is not appropriate
- pass AutoMFU's selected peak through the LLM, sequence-classification, and diffusion-LM training recipes
- add Qwen3.5 MoE FLOP-accounting coverage for activated and shared experts

Before this change, the standard recipes constructed AutoMFU with an H100 dense-BF16 peak of 989 TFLOP/s but discarded it and called `calculate_mfu` with the unrelated 1979 TFLOP/s default. That under-reported dense-BF16 MFU by about 2x. Hardware detection now covers the ordinary BF16 path without YAML; FP8, structured sparsity, and unknown devices can provide an explicit peak.

Tracking: [AM-659](https://linear.app/nvidia/issue/AM-659/tracking-large-scale-moe-vlm-training-improvements)

## Tests

- `pytest tests/unit_tests/utils/test_flops_utils.py tests/unit_tests/utils/test_flops_utils_new_models.py -q` (66 passed)
- `pytest tests/unit_tests/recipes/test_train_ft.py -q` (88 passed, 7 skipped)
- `ruff check` and `ruff format --check` on changed Python files
- `git diff --check`

### DIFF
--- a/docs/guides/configuration.mdx
+++ b/docs/guides/configuration.mdx
@@ -1,6 +1,6 @@
 ---
 title: "YAML Configuration"
-description: ""
+description: "Configure NeMo AutoModel recipes with YAML, including target instantiation, distributed settings, prewarm options, and environment variable interpolation."
 position: 4
 ---
 NeMo AutoModel recipes use YAML configs. The YAML parser creates a `ConfigNode` that performs the following actions:
@@ -113,7 +113,9 @@ mfu:
 `peak_tflops` takes precedence over `device`. A device name alone selects an
 entry from AutoMFU's table; omitting both fields uses the current CUDA device.
 Hardware detection cannot determine the effective training precision, so FP8
-and structured-sparse runs must provide the appropriate peak explicitly.
+and structured-sparse runs must provide the appropriate peak explicitly. An
+unknown device emits a warning and reports 0% MFU until you configure a
+supported `device` value or an explicit `peak_tflops` value.
 
 ## Repair Near-Zero Input Embeddings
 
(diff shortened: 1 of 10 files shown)


yuhezhang-ai:
/ok to test [f82c406](https://github.com/NVIDIA-NeMo/Automodel/pull/3569/commits/f82c40687c87dab3a2569955652fb3d7d5aa3fc6)

docs/guides/configuration.mdx
jgerh:
```suggestion
description: "Configure NeMo AutoModel recipes with YAML, including target instantiation, distributed settings, prewarm options, and environment variable interpolation."
```

jgerh:
```suggestion
- `root` (default): The always-run outer FSDP root owns frozen multimodal parameters. This keeps collective ordering aligned when some ranks execute a modality branch and others skip it.
- `per_layer`: Layers inside the frozen multimodal module get their normal FSDP units. This can reduce peak unshard memory, but it is an expert option: every rank in the FSDP group must execute or skip those units the same number of times and in the same order on every microbatch.
- `replicate`: Frozen multimodal parameters are excluded from FSDP and copied on every rank. This removes their collectives at the cost of higher per-rank parameter memory.
```

jgerh (commented):
Completed tech pubs review of docs/guides/configuration.mdx. Added frontmatter description and capitalized definition-list items in frozen FSDP policy list.

akoumpa (approved):

PR #3570, opened 2026-08-18
fix(mtp): generalize context parallelism across MTP models

HuiyingLi:
## What changed

- Add common preparation of per-depth MTP token, position, label, and validity streams before context-parallel sharding.
- Use explicit model capabilities and model-owned position policy instead of assuming one model-specific MTP hook/layout.
- Wire MTP+CP through DeepSeek V4, Qwen3.5, MiniMax M3, and Step3.7, with cross-rank parity and recipe coverage.

## Why

The existing recipe support handled MTP+CP around a specific input convention. Other MTP models can use different position semantics or embedding paths, so they need one common global-input contract while retaining model-owned preparation where their layout differs. The CP sharder then performs the actual local tensor placement.

This PR intentionally keeps MiniMax's missing-MTP-checkpoint compatibility out of scope.

## Impact

Models explicitly advertising MTP+CP support can use the shared recipe path without duplicating CP index calculations. Models without that capability keep their existing behavior and fail early if MTP+CP is requested.

## Validation

- Added unit and distributed parity coverage for common MTP rolling, recipe preparation, and supported model layouts.
- Static compilation and diff checks pass in the current checkout.
- Full pytest was not rerun in the current shell because PyTorch is unavailable there.

### 50-step MTP+CP parity

- **DeepSeek V4:** CP1 and CP2 both completed 50 steps on 16 nodes / EP128 / PP1 / GBS256 / LBS1 with sequence length 1024. Every step used identical `num_label_tokens`; the mean absolute loss difference was 0.000164 and the maximum was 0.0006. [CP1 pipeline](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63540333), [CP2 pipeline](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63540337). The parity probe used sequence length 1024 because the full 16K recipe exceeded H100 memory.
- **Step3.7:** CP1 and CP8 both completed 50 steps on 18 nodes / EP144 / PP1 / GBS144 / LBS1 with sequence length 2048. Every step used identical `num_label_tokens`; the mean absolute loss difference was 0.0114 and the maximum was 0.0469 (2.2% maximum relative difference). [CP1 pipeline](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63577382), [CP8 pipeline](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63577385).
- **Qwen3.5 dense:** CP1 and CP2 both completed 50 steps on `Qwen/Qwen3.5-4B` with one MTP depth, sequence length 1024, GBS2, LBS1, and LR0. Raw MTP loss was isolated with paired MTP-on (weight 1.0) and MTP-off runs; a direct `calculate_mtp_loss` probe confirmed that the subtraction gives the returned raw MTP loss. Every step used identical `num_label_tokens` and `num_tokens_per_step`. Mean raw MTP loss was 2.361566 for CP1 and 2.361717 for CP2 (0.0064% difference); the mean per-step relative difference was 0.42% and the maximum was 1.41%, within the base-loss CP variation. [CP1 MTP-on pipeline](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63614506), [CP1 MTP-off pipeline](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63614515), [CP2 MTP-on pipeline](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63614505), [CP2 MTP-off pipeline](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63614507).


nemo_automodel/components/loss/mtp.py
```diff
@@ -291,6 +294,11 @@ def calculate_mtp_loss(
             lm_head = _get_lm_head_module(model)
             if lm_head is None:
                 raise ValueError("lm_head module not found in model")
+            lm_head_weight = getattr(lm_head, "weight", None)
+            if isinstance(lm_head_weight, torch.Tensor):
+                # Match the projection's parameter dtype instead of relying on
+                # linear kernels to accept mixed input/weight dtypes.
+                mtp_output = mtp_output.to(dtype=lm_head_weight.dtype)
```
akoumpa: @HuiyingLi there's a similar case for AC IIRC, if the lm_head is fsdp wrapped then accessing .dtype will give you the master-copy dtype, not the compute dtype

nemo_automodel/components/models/common/mtp/mtp.py
```diff
@@ -309,18 +308,20 @@ def prepare_mtp_context_parallel_inputs(
     position_seq_dim = 2 if position_ids.dim() == 3 else 1
     seq_idx = _packed_seq_ids_from_batch(batch, input_ids=input_ids)
     depths = tuple(range(1, num_depths + 1))
+    per_depth_position_ids = tuple(
+        shift_packed_tensor(
```
akoumpa, line 312: 🤔 if we have packed sequences, will there be any cross-sequence contamination here? @HuiyingLi , can we add a unit test for `shift_packed_tensor` for this case?

HuiyingLi: Yes — this is already covered by [`test_shift_packed_tensor_masks_trailing_and_cross_sequence_positions`](https://github.com/NVIDIA-NeMo/Automodel/blob/f64478c3373ef52740411952fdd68fbdeb91b485/tests/unit_tests/models/common/test_mtp_rolling.py#L318-L326), which checks depths 1 and 2 across packed-sequence boundaries. This PR also covers packed multi-axis position IDs in [`test_forward_backward_step_shards_global_vlm_mtp_inputs_with_mrope`](https://github.com/NVIDIA-NeMo/Automodel/blob/f64478c3373ef52740411952fdd68fbdeb91b485/tests/unit_tests/recipes/test_finetune_vlm_helpers.py#L536-L650).

nemo_automodel/components/loss/mtp.py
```diff
@@ -278,6 +280,9 @@ def calculate_mtp_loss(
                 num_label_tokens=num_label_tokens,
             )
         elif isinstance(loss_fn, FusedLinearCrossEntropy):
+            # Some models (notably DeepSeek V4) deliberately keep the shared
+            # LM head in fp32 while decoder/MTP activations are bf16.
+            mtp_output = mtp_output.to(dtype=lm_weight.dtype)
```
akoumpa, line 285: **Blocking.** This cast makes *both* operands fp32, and `cut_cross_entropy` hard-asserts otherwise in the backward:

```python
# cut_cross_entropy/cce_backward.py:340-347
assert e.dtype in (torch.float16, torch.bfloat16), "Backwards requires embeddings to be bf16 or fp16"
assert c.dtype in (torch.float16, torch.bfloat16), "Backwards requires classifier to be bf16 or fp16"
```

So in exactly the scenario the comment describes (fp32 head, bf16 MTP activations), the forward now *succeeds* and the backward dies. End-to-end through `calculate_mtp_loss`:

```
A fp32 head / bf16 MTP hidden  ->  FORWARD OK loss=0.634765   BACKWARD FAILED: AssertionError
B bf16 head / fp32 MTP hidden  ->  FORWARD OK                 BACKWARD OK
C bf16 / bf16 (status quo)     ->  OK / OK  (bit-identical, the .to() is an identity no-op)
```

Under `torch.no_grad()` (validation-only) it now passes silently, which is worse than the old loud failure. The direction this actually fixes is **B** — fp32 hidden against a *bf16* head — not the one named in the comment.

Two more things worth considering:
1. Both recipes call the **main** `calculate_loss` with the same `shared_lm_weight` and **no** cast, *before* `calculate_mtp_loss` (`train_ft.py:1147-1155`). So any config where this cast would matter dies on the main loss first — the reconciliation probably belongs in `calculate_loss`/`FusedLinearCrossEntropy.forward`, not in one branch of one caller.
2. There is no test for this line: deleting it gives `500 passed`.

Suggestion: raise a clear error when `lm_weight.dtype not in (float16, bfloat16)` rather than upcasting into an assert.

nemo_automodel/components/loss/mtp.py
```diff
@@ -291,6 +296,18 @@ def calculate_mtp_loss(
             lm_head = _get_lm_head_module(model)
             if lm_head is None:
                 raise ValueError("lm_head module not found in model")
+            compute_dtype = None
+            if isinstance(lm_head, FSDPModule):
+                # A resharded head exposes its persistent master-copy dtype
+                # until the FSDP pre-forward hook materializes compute weights.
+                compute_dtype = lm_head._get_fsdp_state()._mp_policy.param_dtype
```
akoumpa, line 303: Nit (non-blocking): these six lines are byte-for-byte identical to `nemo_automodel/components/models/common/utils.py:1022-1028`, including the `from torch.distributed.fsdp import FSDPModule` import. Worth extracting a shared `lm_head_compute_dtype(lm_head)` helper — the import-linter contract doesn't forbid it (`components.models` isn't in the `modules` list) and line 31 of this file already imports from `components.models.common.mtp`.

Two smaller notes on the heuristic itself:
- `MixedPrecisionPolicy.cast_forward_inputs` defaults to `True`, and `FSDPState._pre_forward` already casts forward inputs to `param_dtype`. Every policy this repo constructs sets it `True`, so this branch is a no-op in practice — which is also why the new test has to build `cast_forward_inputs=False` by hand, and why that test still **passes on the pre-PR tree**.
- `_mp_policy.param_dtype` is given precedence over the live weight dtype, but it's wrong for params passed to `fully_shard(..., ignored_params=...)` (policy says bf16, weight is fp32 -> this downcasts and the matmul raises). Checking the weight first and using the policy only as a fallback would be strictly safer.

nemo_automodel/components/models/deepseek_v4/mtp.py
```diff
@@ -233,27 +233,67 @@ def forward(
...
+                    cur_input_ids = input_ids_per_depth[depth]
+                else:
+                    cur_input_ids = roll_tensor(cur_input_ids, shifts=-1, dim=-1)
                 decoder_input = embed_fn(cur_input_ids)
             else:
                 decoder_input = embed_inputs[depth]
             kwargs = dict(block_kwargs)
-            if position_ids is not None:
-                kwargs["position_ids"] = position_ids
+            depth_position_ids = position_ids_per_depth[depth] if position_ids_per_depth is not None else position_ids
```
akoumpa, line 294: **Blocking.** The per-depth *shifted* positions flow from here into the TileLang sparse-window index builder, where position **values** are used as **absolute indices** into the gathered KV:

`mtp.py:294` -> `layers.py:1281-1296` -> `layers.py:1426` -> `optimized_kernels.py:229-231`

```python
k_pos = (q_pos.unsqueeze(-1) - window_size + 1).clamp(min=0) + offsets
```

`_query_positions()` returns `position_ids` verbatim (`layers.py:424-426`), and since `packed_seq_ids` is never in `mtp_kwargs`, `sparse_query_positions = query_positions` (`layers.py:1296`). The KV is still slot-indexed, so the window becomes `[i+d-W+1, i+d]` instead of `[i-W+1, i]`. With `W=5, depth=1`:

```
q slot  4: non-CP keys [0,1,2,3,4]     CP keys [1,2,3,4]
q slot 10: non-CP keys [6,7,8,9,10]    CP keys [7,8,9,10]
q slot 11: non-CP keys [7,8,9,10,11]   CP keys []          <-- empty
```

Causality isn't violated (the additive mask still applies), but the window shrinks by `depth`. `sliding_window` defaults to 128. End-to-end with `FORCE_SPARSE=1`: `mtp_d2 cp1=5.298528 cp2=5.299803, rel=2.4e-4`, growing with depth.

What makes this hard to catch: `_transformers/capabilities.py:265-266` returns `supports_cp = (backend_attn == "tilelang")` for DSV4, so **tilelang is the only backend for which DSV4 CP is even allowed** — but `test_mtp_cp_training.py:41,84` builds it with `attn="sdpa"`, which that gate refuses for CP. The automated evidence never reaches `build_dsv4_sparse_topk_indices`, and the quoted `1.64e-4 / 6e-4` is the same order of magnitude as this bug.

nemo_automodel/components/models/deepseek_v4/model.py
```diff
@@ -695,6 +699,7 @@ class ModelCapabilities:
         supports_pp: bool = True
         supports_ep: bool = True
         supports_thd: bool = True
+        supports_mtp_cp: bool = True
```
akoumpa, line 702: **Blocking (see review body item 6).** Flipping this to `True` is what un-gates MTP+CP for DSV4 — on base the flag was absent, so `reject_unsupported_mtp_cp` (`moe/parallelizer.py:1047`) hard-rejected the combination. Two paths become reachable that aren't correct yet:

1. **Packed data**: the MTP head always uses `build_dsv4_cp_causal_padding_mask`, while the backbone branches to `build_dsv4_cp_packed_causal_padding_mask` at `model.py:551-563`. The non-packed builder compares per-document position *values* against global key *indices* (`cp.py:203-205`), so with docs `[0..4]`/`[5..11]` every doc-2+ MTP query attends only document-1 keys. DSV4 declares `supports_thd=True`, so this is reachable.
2. **tilelang** — see my comment on `mtp.py:294`.

Also worth a guard: CP-on and CP-off currently optimize different objectives on packed data (`roll_tensor` crosses document boundaries, `shift_packed_tensor` doesn't) — measured `rel = 1.4e-2` on a 24+40 two-document batch. Re-running CP1 *with* the per-depth prep collapses it to `4e-4`, which isolates the cause.

Would it be worth scoping this flag to the non-packed + tilelang-aware case for now, or adding an explicit reject for `packed_sequence_size > 0`?

nemo_automodel/components/models/qwen3_5/model.py
```diff
@@ -1295,6 +1348,20 @@ def forward(
...
+            if self.mtp is not None and self.training:
+                if (
+                    mtp_per_depth_input_ids is None
+                    or mtp_per_depth_position_ids is None
+                    or mtp_per_depth_valid_masks is None
+                ):
+                    raise ValueError(
+                        "Context-parallel Qwen3.5 MTP requires precomputed per-depth inputs, positions, and masks"
+                    )
+                mtp_embed_inputs = tuple(
```
akoumpa, line 1360: **Blocking.** This builds the CP MTP embeddings from the **vision-spliced** `inputs_embeds`, but the CP1 reference path at `:1488-1496` takes the `input_ids` + `embed_fn` branch and **re-embeds the image placeholder IDs**. The two are not the same objective.

`Qwen3_5MoeForConditionalGeneration` already carries the fix for this at `qwen3_5_moe/model.py:1657-1683`, and its own comment says it out loud: *"Re-embedding media placeholder IDs would silently discard the continuous image/video features on the CP1 reference."* That block wasn't ported here.

Measured on a tiny Qwen3.5-VL (4 layers, h=64, 1 depth), batch = 1x64 tokens with 8 image placeholders + `pixel_values`, cp=2:

```
MTP_BRANCH cp1 ['input_ids', 'position_ids']                 <- re-embeds placeholders
MTP_BRANCH cp2 ['embed_inputs', 'position_ids_per_depth']    <- fused vision embeds
base LM loss:  matches to 1e-5   (backbone CP is fine)
raw MTP loss:  0.0605 absolute delta (1.2% rel)
MTP gradient:  max|cp1-cp2| = 0.00099182  vs |grad| = 0.00111389  -> 89% of its own magnitude
```

Two controls isolate it: feeding CP1 the correct mRoPE positions but still re-embedding **still fails identically**, while forcing CP1 to pre-embed+splice **passes** (`2.9e-6`). So it's the embedding source, not the positions.

`_make_batch` in the functional test is text-only, so this never shows up there. Real VLM batches are frequently 50-90% image tokens, where it scales up from the 12.5% used here.

nemo_automodel/components/models/qwen3_5/model.py
```diff
@@ -828,6 +868,7 @@ class ModelCapabilities:
         supports_pp: bool = True
         supports_ep: bool = False
         supports_thd: bool = False
+        supports_mtp_cp: bool = True
```
akoumpa, line 871: **Blocking.** This capability lands on `Qwen3_5ForConditionalGeneration`, but the new MTP+CP plumbing at `:726-727` and `:773-784` is in **`Qwen3_5ForCausalLM`** — which never declares `supports_mtp_cp` (defaults `False` via `model_capabilities.py:88`) and never gained `prepare_mtp_inputs_for_cp`. Replaying the recipe gate at `train_ft.py:1029-1037` with `cp_size=2`:

```
mtp_cp_enabled = True
RECIPE RAISES: NotImplementedError Qwen3_5ForCausalLM declares supports_mtp_cp=False;
               MTP target preparation for context parallelism is unavailable
--- with the flag forced True ---
THEN RAISES: AttributeError 'Qwen3_5ForCausalLM' object has no attribute 'prepare_mtp_inputs_for_cp'
```

Coverage agrees: `:774` (the happy-path `self.mtp(...)` call) is executed by no test anywhere — the functional test's `_qwen3_5_model()` builds the *VL* class. Which means the description's "Qwen3.5 dense: CP1 vs CP2 raw MTP loss 2.361566 vs 2.361717" didn't exercise the class that received the guard.

Either wire the dense class up properly (`supports_cp`, `supports_mtp_cp`, `prepare_mtp_inputs_for_cp`) or drop the dead branch at `:773-784`.

Note `Qwen3_5ForCausalLM` also lacks `_is_pipeline_parallel_stage`, and `Qwen3_5ParallelizationStrategy` (`parallelizer.py:617`) never calls either reject helper — so `supports_mtp_cp_pp=False` isn't actually enforced for this family either.

nemo_automodel/components/models/qwen3_5/model.py
```diff
@@ -742,7 +770,19 @@ def forward(
...
+                mtp_per_depth_h = self.mtp(
+                    hidden_states,
+                    input_ids_per_depth=mtp_per_depth_input_ids,
+                    embed_fn=self.model.embed_tokens,
+                    position_ids_per_depth=mtp_per_depth_position_ids,
+                    attention_mask=causal_mask,
+                    rotary_emb=self.model.rotary_emb,
+                    **kwargs,
+                )
+            elif int(kwargs.get("cp_size", 1) or 1) > 1:
```
akoumpa, line 783: This guard can't fire on the path it's meant to protect. `cp_size` is injected into the batch **only** by `_shard_thd_chunk_for_te` (`context_parallel/utils.py:800` and `:860`); neither `shard_batch_load_balanced` nor `shard_batch_aux_only` (`sharder.py:811-1022`) sets it. So on the default BSHD round-robin CP path `kwargs.get("cp_size", 1)` is always `1`.

Today that's moot because the recipe blocks first (see my comment at `:871`), but it isn't the safety net it looks like: the moment the capability flag is flipped without the recipe wiring, a BSHD CP run falls through to `elif input_ids is None` / `else` and does a **rank-local roll on a round-robin shard** — silently wrong, no error.

Reading `self.cp_mesh` (as the VL class does at `:1331` and `:1475`, and as step3p7 does) would actually fire. Same applies to the identical guard at `qwen3_5_moe/model.py:1023`.

nemo_automodel/components/models/step3p7/model.py
```diff
@@ -179,15 +185,43 @@ def _process_image_input(
...
+        patch_projector_inputs = None
         if patch_pixel_values is not None and patch_pixel_values.numel() > 0:
             patch_pixel_values = patch_pixel_values.reshape(-1, *patch_pixel_values.shape[-3:]).to(
                 device=vision_device,
                 dtype=vision_dtype,
             )
-            patch_image_features = self._process_image_features(self.vision_model(patch_pixel_values))
+            patch_projector_inputs = self._downsample_image_features(self.vision_model(patch_pixel_values))
+
+        image_projector_inputs = self._downsample_image_features(self.vision_model(pixel_values))
```
akoumpa, line 196: **Blocking.** The projector merge is correct and I confirmed it fixes half the problem — but `self.vision_model` is **still** invoked a data-dependent number of times, which is the same desync the comment above describes. Measured on a real `Step3p7Model`:

```
crops=True   vit_large_projector=1  vision_model=2  vit_downsampler1=2
crops=False  vit_large_projector=1  vision_model=1  vit_downsampler1=1
```

This is data-driven, not hypothetical: `processing_step3.py:224-228` returns `img, [], None` when `determine_window_size(...) == 0`, and `:454` omits `patch_pixel_values` entirely — MedPix images vary in size, so DP ranks really do get 0 vs N crops.

Reproduced the hang on 2 GPUs (NCCL, rank0 has crops / rank1 doesn't), with symmetric-batch controls to rule out coincidence:

| tree / config | result |
|---|---|
| base (pre-PR) | **HANG** (exit 143 @240s, exit 124 @100s) |
| base, symmetric batches | OK |
| **this PR** | **OK** — the projector half is proven |
| this PR + `frozen_sharding: per_layer` | **HANG** (exit 124 @120s) |
| this PR + `per_layer`, symmetric | OK |

The shipped recipe dodges it only because `freeze_vision_tower: true` + default `frozen_sharding: root` puts the frozen tower in the root FSDP unit. It stays live for `freeze_vision_tower: false` and for `per_layer` — a mode `moe/parallelizer.py:720-726` warns about in these exact terms. (`vit_large_projector` escapes `freeze_vision_tower` because `_freeze_module_by_attribute_and_patterns` matches only `["vision","visual","image_encoder"]` in the module *name* — `model_utils.py:384`.)

Suggestion: call `self.vision_model` unconditionally with a zero-row crop batch when there are no crops, so the call count is rank-uniform too. Also worth noting a rank whose microbatch has *no images at all* calls the projector 0x — MedPix is all-image today, but any mixed text/image dataset reopens it.

nemo_automodel/components/models/step3p7/mtp.py
```diff
@@ -265,8 +294,8 @@ def forward(
             hidden_states, logits = block(
                 hidden_states,
                 embed_input=decoder_input,
-                freqs_cis=freqs_cis,
-                position_ids=position_ids,
+                freqs_cis=freqs_cis_per_depth[depth],
```
akoumpa, line 297: Non-blocking, but worth knowing: `freqs_cis_per_depth` is inert in both YAMLs this PR ships. `step3p5/layers.py:299-304` recomputes it from `position_ids` and **overwrites the argument**:

```python
if position_ids is not None and not self.backend.rope_fusion:
    cos, sin = self.rotary_emb(x, position_ids)
    rotary_half_dim = cos.shape[-1] // 2
    freqs_cis = torch.cat((cos[..., :rotary_half_dim], sin[..., :rotary_half_dim]), dim=-1)
```

Both new recipes set `rope_fusion: false` (`deepseek_v4_flash_mtp_cp_tulu3.yaml:82`, `step3p7_medpix_200b_mtp_cp8_ep144pp1.yaml:42`), and `position_ids_per_depth[depth]` is always passed alongside. Forcing `freqs_cis_per_depth[0]` for every depth produces **bit-identical** losses.

So the per-depth RoPE fix genuinely works — via `position_ids_per_depth` — but the `freqs_cis_per_depth` plumbing only matters under `rope_fusion: true`, which has no test. The two new assertions about it (`test_model.py:401`) compare against a **monkeypatched** `position_ids_to_freqs_cis` using the same formula the mock returns, so they'd pass either way.

Either drop the parameter and derive per-depth RoPE from the positions, or add a `rope_fusion: true` case that actually pins it.

tests/functional_tests/models/test_mtp_cp_training.py
```diff
@@ -0,0 +1,360 @@
...
+    local_values = torch.stack([base_loss, mtp_loss.loss, total_loss, *mtp_loss.per_depth_losses]).detach()
+    dist.all_reduce(local_values, group=cp_mesh.get_group())
+    return local_values, grad.detach()
+
+
+def _run_model(name: str, device: torch.device) -> dict[str, list[float]]:
+    cp1_values, cp1_grad = _run_cp1(name, device)
+    cp2_values, cp2_grad = _run_cp2(name, device)
+
+    torch.testing.assert_close(cp2_values, cp1_values, rtol=5e-3, atol=5e-3)
```
akoumpa, line 302: This tolerance is too loose to detect a completely broken MTP input stream. The models are freshly initialised, so every loss sits at `ln(vocab) = ln(128) = 4.852`; `rtol=5e-3` on ~5.0 is about 0.03 of absolute slack.

Replacing **every** MTP future-token embedding with zeros on the CP path:

```
cp1 per-depth 4.962883   cp2 per-depth 4.973676   -> 0.22%  -> PASSES
```

I injected 18 deliberate defects; **12 went uncaught**, most of them for this reason — wrong-depth `input_ids_per_depth[0]`, dropping the `~` in `_mask_mtp_embed_inputs`, deleting the mask call entirely, deleting the step3p7 `masked_fill` (bit-identical), and ignoring per-depth positions in DSV4 (delta `2e-5`). Three `RuntimeError` probes planted at the top of all three CP+MTP branches at once produced **zero hits** across the full 12,458-test unit suite.

`_make_batch` also uses one unpacked 64-token document, so exactly **one** token in the whole batch is ever invalid (instrumented: `n_invalid=0` on rank 0, `n_invalid=1` on rank 1) — the valid-mask logic is a statistical no-op here.

Concrete suggestions, roughly in value order:
1. Move this file to `tests/functional_tests/context_parallel/` with an `L2_*.sh` — as-is it is **never collected by CI** (`models` is not in the `cicd-main.yml` test-folder matrix), and the sibling `run_qwen3_5_moe_mtp_cp_parity.py` already lives there.
2. Anchor on correctness, not just the scalar loss: gather CP2's per-depth logits and compare element-wise against CP1, at `rtol=1e-5`.
3. `num_nextn_predict_layers=1` at `:82` makes `input_ids_per_depth[depth]` identical to `[0]`. Bumping it to `2` makes the wrong-depth mutation fail immediately (`Mismatched elements: 2 / 5`).
4. Add a packed 2-document batch and a `pixel_values` batch for the two VLMs — `vision_config` currently uses `"layers": 0`, so none of the vision-side changes in this PR are exercised at all.
5. `_step_model` uses `attn="te"` and `apply_cp` imports TE unconditionally, so on a TE-less runner this **errors rather than skips** — `pytest.importorskip("transformer_engine")` would help.
6. The subprocess at `:333-342` passes no `env=`/`cwd=`/`PYTHONPATH`, so `torch.distributed.run` can import a different `nemo_automodel` than the checkout under test (it picked up a stale `site-packages` copy here).

akoumpa (changes requested):
Deep review — I checked this out against the merge base in six isolated worktrees and **reproduced** each finding on 2 GPUs rather than reading for it. Requesting changes on the blocking items; happy to pair on any of them.

### First, what holds up

The CP1==CP2 design is correct, and I want that on the record before the list:

- I re-ran Step3.7 parity with a **real** 1-layer ViT and an image+crops batch (the shipped test stubs vision out with `"layers": 0`): `rel <= 8.9e-5`, MTP-grad delta `4.5e-6`. DSV4 and Qwen text parity reproduce too.
- `num_label_tokens` is genuinely CP-invariant — computed pre-shard, reduced with `include_cp=False`. Verified numerically across 4 configs.
- The `common/mtp/mtp.py` refactor is byte-identical across 6 layouts incl. 3-axis mRoPE and packed.
- `roll_tensor` zero-fills rather than wrapping, so the cp=1 Step3.7 position change is confined exactly to the `ignore_index` tail — loss-neutral. The embed masking is likewise defensive, not load-bearing (flipping the fill `0` -> `100.0` leaves losses bit-identical).
- No regressions: base `6 failed / 1931 passed` vs head `6 failed / 1937 passed`, identical failure sets, and the test diff is pure addition — nothing weakened.

### Blocking

**1. The linked parity runs exercised code that isn't in this PR.** All eight pipelines in the description mounted Automodel SHA `24d9984e54e6ea1c28958f8f09eef8b53246f67b` ("debug: instrument Step3.7 distributed hang"), which is **not an ancestor of `f64478c33`**:

```
$ git merge-base --is-ancestor 24d9984e5 f64478c33 ; echo $?
1
$ git diff --stat 24d9984e5 f64478c33 -- nemo_automodel/components/{loss,models}
 11 files changed, 169 insertions(+), 127 deletions(-)
```

That delta contains `step3p7/mtp.py` per-depth RoPE (`fd9e827af`, ~8h after the Step3.7 runs), the `loss/mtp.py` dtype cast (`f64478c33`, two days after), the vision-projector merge, and `deepseek_v4/mtp.py`. So the Step3.7 CP1-vs-CP8 numbers measured the *pre-fix* collapsed-position semantics. Could you re-run at least the DSV4 and Step3.7 parity at the head SHA? (Minor: the Step3.7 CP1 job `404202769` is red — benign, GitLab artifact-upload 500s with `SLURM_EXITCODE=0` — but it reads as passing evidence.)

**2. `loss/mtp.py:285` turns a forward failure into a backward assert.** See inline.

**3. Qwen3.5-VL multimodal MTP diverges CP1 vs CP>1** — MTP loss 1.2%, MTP gradient 89% of its own magnitude. See inline.

**4. Packed VLM + CP + MTP hard-crashes, and this PR is what makes it reachable.** `_packed_seq_ids` (emitted by `vlm/collate_fns.py:1618`) is never sharded nor popped by `shard_batch_aux_only`, so `_mtp_block_causal_mask` sizes `[B,1,S,S]` off the **global** `S` against a local shard:

```
RuntimeError: The expanded size of the tensor (64) must match the existing size (128) at
non-singleton dimension 2.  Target sizes: [1, 4, 64, 64].  Tensor sizes: [1, 1, 128, 64]
```

CP1 on the identical batch completes. Pre-PR `supports_mtp_cp` defaulted `False` so `recipes/vlm/finetune.py:889` refused; this PR flips it at `:871`.

**5. DeepSeek V4 + tilelang: shifted per-depth positions are consumed as KV indices.** See inline.

**6. DSV4 MTP uses the non-packed CP mask on packed data.** Under CP the MTP head always calls `build_dsv4_cp_causal_padding_mask` while the backbone branches to the packed builder (`model.py:551-563`). With docs `[0..4]` and `[5..11]`: `q=5` allows keys `[0]` (correct `[5]`), `q=11` allows `[0..6]` (correct `[5..11]`) — every doc-2+ MTP query attends **only document-1 keys**. `mtp_kwargs` also never forwards `packed_seq_ids`/`seq_lens`/`qkv_format`, so the blocks can't self-correct. Separately, CP-on and CP-off optimize different objectives on packed data (`roll_tensor` crosses boundaries, `shift_packed_tensor` doesn't): `rel = 1.4e-2`.

**7. The Step3.7 FSDP-uniformity fix is half done.** See inline.

### High

**8. The dense half of the Qwen3.5 diff is unreachable.** See inline at `:871`.

**9. `tests/functional_tests/models/test_mtp_cp_training.py` never runs in CI.** `tests/run_test.sh:61` builds `tests/functional_tests/$TEST_FOLDER` from the `cicd-main.yml` matrix. The complete folder list across x86 + gb200 is `context_parallel, data, diffusion, dllm, hf_dcp, hf_peft, hf_transformer, hf_transformer_finetune, hf_transformer_llm, hf_transformer_vlm, llm_pretrain_and_kd, moe, parallelism, retrieval, speculative` — **`models` is not among them**, and nothing under a collected folder references it. The sibling `run_qwen3_5_moe_mtp_cp_parity.py` already lives in `functional_tests/context_parallel/`, which *is* collected. Moving it there plus an `L2_*.sh` (with `export PYTHONPATH=$(pwd)`) fixes this and the `sys.path` issue below in one go.

**10. The two recipes implement different MTP-CP contracts.** `train_ft.py:1051-1060` vs `vlm/finetune.py:901-913`:

| key | LLM recipe | VLM recipe |
|---|---|---|
| `mtp_per_depth_position_ids` | `seq_dim=1` **hardcoded** | `seq_dim=mtp_cp_inputs.position_ids_seq_dim` |
| `mtp_per_depth_valid_masks` | **computed, then discarded** | sharded, `fill=False` |

The hardcoded `seq_dim=1` normally trips the sharder's length check, but when `B == S` it silently shards the **batch axis** instead of the sequence axis (`seq_dim=1 -> (3,4,8)` vs `seq_dim=2 -> (3,8,4)`). And any future LLM-recipe model adopting the Step3.7/Qwen-VL mask convention will get `None`.

**11. Packed boundary masking is applied at CP>1 but not CP=1.** `train_ft.py:1180` / `finetune.py:1010` pass `cu_seqlens=None` when per-depth targets exist; at CP>1 boundaries come from `seq_lens_padded`, at CP=1 only from `cu_seqlens` — which `packed_sequence_thd_collater` never emits. Result on a 2-doc/12-token batch: `cp1 6.3877` (unmasked, wrong) vs `cp2 6.3318` (masked, right). Pre-existing wiring, but this PR makes it live for three more families and it contaminates any packed CP-parity measurement.

### Medium

- **12 of 18 mutations I injected went uncaught.** Zeroing *every* MTP future-token embedding on the CP path still passes (`4.962883 -> 4.973676`, 0.22%, under `rtol=5e-3`). Three `RuntimeError` probes planted at the top of all three CP+MTP branches simultaneously: full 12,458-test unit suite, **zero hits**. Patch coverage is 62% (41/108 lines); `Step3p5MTPModule.forward` — the function this PR rewrites — is 0% in the unit suite.
- **DSV4's test config uses `num_nextn_predict_layers=1`**, which makes `input_ids_per_depth[depth]` identical to `[0]`. Bumping that one literal to `2` makes the wrong-depth mutation fail immediately (`Mismatched elements: 2 / 5`). The shipped YAML also uses 1.
- **The `cp_size > 1` guards are dead on the generic CP path** — `cp_size` only enters the batch via `_shard_thd_chunk_for_te` (`context_parallel/utils.py:800,860`); neither round-robin sharder sets it. Reading `self.cp_mesh` (as the VL class does at `:1331`) would actually fire.
- `mtp_per_depth_input_ids` is required-then-never-read by Step3.7 (`model.py:598,710-716`) and by Qwen-VL (`:1281,1353`). The recipes shard `num_depths` extra `[B,S]` int64 tensors per microbatch for nothing, and the error text advertises a contract the code doesn't honour.
- A `cu_seqlens`-only tail-padded batch raises `cu_seqlens cannot describe a padded materialized MTP token layout ... provide cu_seqlens_padded` under CP but trains at CP=1 — and no LLM collator in the repo emits `cu_seqlens_padded`.
- All 15 new `ValueError` guards have zero test coverage (`grep` over `tests/` for each message returns 0).
- `test_cp_forward_suspends_ring_dispatcher_for_vision` only asserts *that the CM was entered*; replacing the body with `pass` and moving the tower outside the suspension passes all 48 step3p7 tests.

### Cleanup

- `_mask_mtp_embed_inputs` (`qwen3_5/model.py:193-211`) is byte-for-byte identical to `qwen3_5_moe/model.py:347-375`, and `loss/mtp.py:299-308` is identical to `models/common/utils.py:1022-1028`. Both belong in `components/models/common/mtp/` — the import-linter contract doesn't forbid it, and `loss/mtp.py:31` already imports from there. Step3.7 open-codes the same masking at `:723` **without** the shape check.
- The description still says "Wire MTP+CP through ... MiniMax M3", but `c3f647377` dropped it and no minimax file is touched.
- Neither new recipe is in `nightly_recipes.yml`, so only the release/performance rglob scopes pick them up — an 18-node job nobody watches nightly.
- `freeze_config.freeze_embeddings: true` in the Step3.7 recipe is a **no-op** (`apply_parameter_freezing`, `utils/model_utils.py:377-401`, never reads that key; `embed_tokens.requires_grad` stays `True`). Worth fixing since the projector comment blames the hang on ranks all-gathering `embed_tokens`.
- The DSV4 YAML header says "CP is NOT compatible with packed sequences", which contradicts the packed branch at `deepseek_v4/cp.py:456-508`. Given items 4/6 above, "unsafe" is more accurate than "unsupported" — or add a hard guard.

nemo_automodel/components/models/step3p7/model.py
```diff
@@ -146,7 +150,7 @@ def _vision_dtype_device(self) -> tuple[torch.dtype, torch.device]:
             dtype = get_dtype(getattr(self.config.text_config, "torch_dtype", "bfloat16"), torch.bfloat16)
             return dtype, torch.device(f"cuda:{torch.cuda.current_device()}" if torch.cuda.is_available() else "cpu")
 
-    def _process_image_features(self, image_features: torch.Tensor) -> torch.Tensor:
+    def _downsample_image_features(self, image_features: torch.Tensor) -> torch.Tensor:
```
akoumpa, line 153: Follow-up / correction to my comment on `:196`, and a caveat before you act on it.

This extracted method calls `self.vision_model.vit_downsampler1/2` from **outside** `vision_model.forward` (`:159-160`). When `vision_model` is its own FSDP unit with `reshard_after_forward=True`, those conv weights are resharded DTensors at this point:

```
RuntimeError: aten.convolution.default: got mixed torch.Tensor and DTensor, need to convert
all torch.Tensor to DTensor before calling distributed operators!
```

The tower gets its own unit whenever it is **trainable** (`"vision_model"` is in `MULTIMODAL_TOWER_NAMES`, `shared/multimodal_fsdp.py:32`), and `step3p7_medpix_200b_mtp_cp8_ep144pp1.yaml:69` sets `moe.reshard_after_forward: true`. Reproduces on this branch and on the merge base, so it is pre-existing — but the refactor moved this code, and it interacts directly with my other comment.

Two practical consequences:

1. `freeze_vision_tower: false` is currently unusable for Step3.7 under this recipe's sharding. The shipped config only avoids it via `freeze_vision_tower: true` + default `frozen_sharding: root`.
2. **This is the configuration you'd naturally reach for to verify the `vision_model` call-count fix I suggested on `:196`** — so you'd hit this error and conclude the fix didn't work. Worth folding the downsampler calls into `vision_model.forward` (or into a method on the tower) at the same time, so the whole vision path is one FSDP-visible call.

Not blocking on its own — flagging it so the two don't collide.

akoumpa (commented):
Follow-up to my review above — one correction to my own advice, three things I under-weighted on the test side, and an ordered "minimum to unblock" list since the blocking items are spread across ten threads.

### Correction to my own suggestion

My comment on `model.py:196` says to make the `vision_model` call rank-uniform. Before you try that with `freeze_vision_tower: false`, see the inline note below — you'll hit a DTensor error first, for an unrelated reason. Sorry for sending you at that in the wrong order.

### Three things I under-weighted on tests

**1. ~120 of the ~700 new test lines don't cover this PR.** `git diff 74bce7ea..f64478c3 -- nemo_automodel/recipes/` is **empty**, so `tests/unit_tests/recipes/test_finetune_vlm_helpers.py` (+120) is exercising `vlm/finetune.py`, which this PR doesn't change. It's a good test — it just belongs to whichever PR added that recipe block. Together with the 62% patch coverage and `step3p7/mtp.py` at 0% in the unit suite, the effective new coverage of the ~230 changed source lines is a lot thinner than the line count suggests.

**2. The functional test replicates the *wrong* recipe for two of its three models — which is why the dead dense-class path shipped green.** `_run_cp2` (`:244-268`) hand-copies the **VLM** recipe's wiring: `ContextParallelSharder` built before `prepare_mtp_inputs_for_cp`, `seq_dim=mtp_inputs.position_ids_seq_dim`, and `mtp_per_depth_valid_masks`. It then applies that uniformly to `deepseek_v4`, `step3p7` and `qwen3_5`. But DSV4 and Qwen-dense are driven by the **LLM** recipe in production, which does **none** of those three (`train_ft.py:1051-1060` hardcodes `seq_dim=1` and never sets `valid_masks` — see item 10 in my review).

It also calls `apply_cp(model, cp_mesh)` directly rather than `parallelize_model`, so it never runs `reject_unsupported_mtp_cp` / `reject_unsupported_mtp_cp_pp`. That's structural: the test *cannot* catch a wrong capability declaration, which is exactly how `Qwen3_5ForCausalLM` ended up with plumbing but no `supports_mtp_cp`. Driving parity through `recipe._forward_backward_step` — as `tests/unit_tests/recipes/test_train_ft.py:2605` already does with a fake sharder — would close both gaps and force the two recipes' contracts to converge.

**3. The parity numbers have no control.** Step3.7's max 2.2% relative and Qwen's max 1.41% are described as "within the base-loss CP variation", but that variation is never measured. A single CP1 rerun at a different seed would settle whether those are noise or signal — and it's cheap next to the 16/18-node parity jobs themselves. (DSV4's `1.6e-4 / 6e-4` is tight enough to speak for itself.)

### Minimum to unblock, in order

1. **Re-run parity at `f64478c33`** for DSV4 and Step3.7, with a CP1-vs-CP1 control. Everything below is easier to judge once the numbers describe the shipped code.
2. **`loss/mtp.py:285`** — raise when `lm_weight.dtype not in (float16, bfloat16)` instead of upcasting into the `cce_backward` assert. One line. (Or move the reconciliation into `calculate_loss` so the main and MTP paths share one policy.)
3. **Qwen3.5** — port `qwen3_5_moe/model.py:1657-1683` for the CP1 multimodal reference; and either wire up `Qwen3_5ForCausalLM` properly or delete the dead branch at `:773-784`.
4. **Scope the two new capability flags to what's actually validated.** `supports_mtp_cp=True` on DSV4 currently un-gates both tilelang and packed; on Qwen-VL it un-gates packed. An explicit reject for `packed_sequence_size > 0` would buy time without blocking the feature.
5. **Step3.7** — make the `vision_model` call rank-uniform (after reading the inline below).
6. **Tests** — move the file to `functional_tests/context_parallel/` with an `L2_*.sh`; bump DSV4 to `num_nextn_predict_layers=2`; tighten to `rtol=1e-5` with an element-wise per-depth logits anchor rather than the scalar loss.

The rest of my review — guard ordering at `deepseek_v4/model.py:980`, bare `seq_lens` not handled in `_packed_seq_ids_from_batch`, `reject_unsupported_mtp_cp` missing from the generic parallelizer path, and the recipe knobs — is fine as follow-up work and shouldn't hold this up. Two of those are worth a tracking issue though: the bare-`seq_lens` case is *silent* (the `cu_seqlens` variant at least raises), and `deepseek_v4_flash_mtp_cp_tulu3.yaml` sets `val_every_steps: 500` with `max_steps: 100`, so validation never actually runs in that recipe.

Happy to take any of 2/3/5 as a PR against your branch if that's easier than round-tripping.

Issue #3107, opened 2026-07-16
refactor(packing): replace global HF monkeypatch with typed metadata


yuhezhang-ai:
## Context

AutoModel NEAT packing currently encodes both document boundaries and padding in an integer `attention_mask` such as `[1, 1, 2, 2, 0]`. The FlashAttention path then globally monkeypatches private Transformers functions, including `_get_unpad_data` and model-local `create_causal_mask` references, to recover per-document `cu_seqlens`.

PR NVIDIA-NeMo/Automodel#2996 hardens this legacy contract for Transformers 5.x and makes incompatible private APIs fail closed. That is the right compatibility fix for the current design, but it does not remove the process-wide private-API coupling.

Pinned Transformers exposes typed FlashAttention packing kwargs (`cu_seq_lens_q`, `cu_seq_lens_k`, `max_length_q`, and `max_length_k`). AutoModel should move packing metadata through a supported Transformers path or a typed AutoModel/HF adapter.

Related: NVIDIA-NeMo/Automodel#2996, NVIDIA-NeMo/Automodel#2985.

## Proposed scope

* Introduce typed packed-sequence metadata produced by the LLM and VLM NEAT collaters.
* Separate document IDs from the normal padding `attention_mask` contract.
* Thread explicit cumulative sequence lengths and maximum lengths through supported HF FlashAttention kwargs or a typed attention adapter.
* Preserve per-document IDs for AutoModel consumers such as SqrtCrossEntropy, Qwen3.5 MTP, and hybrid/linear-attention paths without overloading the HF padding mask.
* Define backend capability checks explicitly: support FA2 first; enable FA3 only when AutoModel can select and validate it.
* Remove global mutation of private Transformers functions once parity coverage is in place.

## Acceptance criteria

* No global monkeypatch of Transformers `_preprocess_mask_arguments`, `_get_unpad_data`, or model-module `create_causal_mask` for NEAT packing.
* No cross-document attention for padded and unpadded packs.
* Model-level parity coverage for LLM and VLM packing, including batch size greater than one, uneven padding, image/video inputs, and Qwen mRoPE position IDs.
* Existing SDPA/eager block-causal packing behavior remains unchanged.
* Loss and auxiliary model paths that consume packed document IDs continue to receive equivalent typed metadata.
* Failure is explicit when a model or attention backend cannot consume the typed packing contract.

saisharan0103:
[...]

PR #3577, opened 2026-08-18
refactor(packing): replace global HF monkeypatch with typed metadata

beccohov:
# What does this PR do ?

Replaces the process-wide monkeypatching that NEAT packing used to route packed sequences through FlashAttention with typed metadata that the collaters build and pass through the public HF kwargs.

# Changelog

* Add `PackedSeqParams` and `packed_seq_params_from_doc_ids` (`datasets/packed_seq.py`). The collater derives `cu_seqlens`/`max_seqlen` from the document map and emits them as `FlashAttentionKwargs` (`cu_seq_lens_q/k`, `max_length_q/k`). The `cu_seqlens` span the full flattened batch, including padding, with a forced boundary at every row so a row-filling document cannot merge with the next row.
* LLM and VLM NEAT collaters stop emitting an indexed attention mask on the flash path. Without a mask HF takes its varlen-kwargs branch instead of unpadding a binary one. The per-document map is still exposed as `_packed_seq_ids` for the loss and context-parallel consumers.
* Remove `configure_packing` and the three patch sites (`_get_unpad_data`, `_preprocess_mask_arguments`, per-module `create_causal_mask`). In their place `validate_flash_packing_support` fails early when the installed Transformers build or the model cannot accept the typed contract.
* SDPA/eager keep the existing 4D block-causal path unchanged.
* Tests cover the metadata, both collater backends, the validation failure modes, and forward-pass parity showing no cross-document leakage: block-causal on CPU, plus a flash-varlen check on GPU for padded and unpadded packs.

**Pre checks**:

- [ ] Did you add or update any necessary documentation? No user-facing behavior or config changed.

# Additional Information

* Related to NVIDIA-NeMo/Automodel#3107


beccohov:
Hey, @yuhezhang-ai
as promised, here is the PR. I believe I understood your idea correctly!
I run tests locally and it works correctly.

yuhezhang-ai (commented):
Thanks for taking this on — replacing the process-wide monkeypatches with explicit packed-sequence metadata is a much cleaner direction. I had two follow-ups that would make this safer to rely on broadly:

1. In `_validate_model_consumes_packed_contract`, could we require either `**kwargs`, `_packed_seq_ids`, or **all four** varlen kwargs? The current set intersection accepts a forward that exposes only one of the four. `filter_forward_kwargs` can then drop the remaining cumulative-length fields, while Transformers enters its varlen path only when all four are present. Using `set(_PACKED_VARLEN_KWARGS).issubset(params)` for that branch would keep this check fail-closed.

2. Could we narrow the FlashAttention availability probe so unexpected model/CUDA execution failures are not converted into skips? The real GPU parity test currently covers Llama with a single packed row. It would also be valuable to exercise batch size greater than one with uneven row padding and the actual Qwen3-VL packed FlashAttention path. The release matrix should auto-discover the Qwen3-VL config; we would also be happy to help run the relevant scoped CI job after the PR is updated.

The focused CPU tests passed locally; the FlashAttention cases skipped in the CPU-only environment. Thanks again!

beccohov:
Thanks, @yuhezhang-ai, I believe I've addressed all of your suggestions:
1. Contract check -- a forward that exposes only some of the four no longer slips through.
2. FA availability. Instead of narrowing the probe, I dropped the forward-based probe entirely and gate on `importlib.util.find_spec("flash_attn")` + CUDA, matching `test_eagle3_packing_fa2_parity.py`. There's no probe forward left, so nothing can turn a real model/CUDA failure into a skip.
3. Coverage. Added batch > 1 with uneven per-row padding, and a Qwen3-VL packed FlashAttention parity check that asserts per-document logits match the isolated forward. Left the full-size Qwen3-VL config to the release matrix, as you offered.

I think we need to run CI

yuhezhang-ai:
/ok to test [9fb600e](https://github.com/NVIDIA-NeMo/Automodel/pull/3577/commits/9fb600e39946c8ea57bd8e85ae1d2d3e10e71f76)

yuhezhang-ai:
Thanks, @beccohov. I ran an internal scoped release ci test job which exposed one remaining multimodal edge case: the packed text `cu_seq_lens_q/k` and `max_length_q/k` values were also reaching Qwen3-VL's vision tower, which supplies its own values and failed with a duplicate-keyword error.

I pushed a small follow-up directly to this PR branch in [`1cdf4a0`](https://github.com/NVIDIA-NeMo/Automodel/pull/3577/commits/1cdf4a07e3c4946a5ac161a675f58e11f0798f0c):

- Keep packed text attention metadata on Qwen3-VL's language-model path when media is present.
- Replace the text-only Qwen3-VL GPU check with a functional regression using synthetic image patch tensors through the real tiny vision tower.

Validation is now green:

- 208 focused unit tests passed, with 7 expected GPU skips; Ruff and diff checks passed.
- The new one-GPU packed-image functional test passed.
- The exact scoped [`qwen3_vl_4b_neat_packing` release job](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/403988660) completed all 50 steps successfully. The [parent pipeline](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63549710) and both child pipelines passed.

This addresses the remaining issue from my side. Thanks again!

yuhezhang-ai:
/ok to test [1cdf4a0](https://github.com/NVIDIA-NeMo/Automodel/pull/3577/commits/1cdf4a07e3c4946a5ac161a675f58e11f0798f0c)

yuhezhang-ai (dismissed):

beccohov:
Hey, 

I fixed small test changes due to updated main code. The L0 failure was a stale test. `test_flash_attention_2_returns_2d_mask` in `test_collate_fns.py` still asserted the old contract, but dropping `attention_mask` on the flash path is the whole point of the change — that's what routes HF to `flash_attn_varlen_func` instead of the unpad path.
Updated it to check the varlen kwargs instead; the actual cu_seqlens boundaries are already covered by `test_flash_emits_varlen_kwargs_and_no_mask` in `test_neat_packing_vlm.py`, so I didn't duplicate them.

Could you kick off CI?
cc @yuhezhang-ai @akoumpa

yuhezhang-ai:
/ok to test [9e4f476](https://github.com/NVIDIA-NeMo/Automodel/pull/3577/commits/9e4f476b0025b29528f73e0a9dc69b670b2cc49b)

PR #3579, opened 2026-08-18
feat(diffusion): periodic validation loss during training

### LINKED ISSUE #3244: Support periodic validation loss evaluation for diffusion training
**Problem**

The diffusion training recipe currently does not support computing validation loss on an eval dataset every N training steps.

To check eval loss during training, users need to save a checkpoint, load it separately, and manually run the eval dataset through the model.

**Requested behavior**

Support configuring an eval dataset and computing its validation loss periodically during diffusion training, for example through `step_scheduler.val_every_steps`.

**Additional context**

The shared step scheduler already has `val_every_steps`, but the diffusion training recipe currently only builds the training dataloader and does not run validation.

beccohov:
# What does this PR do ?

Adds periodic validation loss to the diffusion training recipe, so checking eval loss no longer means saving a checkpoint and running the eval set through the model by hand.

# Changelog

* New optional `data.validation_dataloader` block, resolved by the same `resolve_diffusion_dataloader` as the training loader (`_typed_config.py`), so it takes the same four builders and the same field validation. One loader, matching the training side, rather than the dict of named loaders the LLM recipe has.
* The recipe builds it in `setup()` with the same dp sharding and batch size as the training loader, and raises if it comes out empty. Half-configured setups get a warning instead of silence: a loader without `val_every_steps` validates on checkpoint steps only, `val_every_steps` without a loader is skipped.
* `_run_validation_epoch` scores the held-out set with the training flow-matching objective in `eval()` under `torch.no_grad()`, restores train mode in a `finally`, and reduces `[loss_sum, batch_count]` in one all-reduce over the dp group, so the mean does not depend on how batches are spread across ranks. CP is excluded from that reduction since every peer computes the same full-sequence loss.
* The pass runs inside `ScopedRNG(seed=self.seed)`. Without it the sampled timesteps dominate the curve and validation would consume the training RNG stream, changing the run. FP8 autocast is deliberately not applied: delayed-scaling amax history is training state and should not be updated from held-out batches.
* The loop calls it on `step_scheduler.is_val_step`, before the checkpoint branch, and logs `val_loss` to the console and to W&B.
* Example config (`wan2_2_t2v_flow.yaml`) and the diffusion fine-tuning guide.

**Pre checks**:

# Additional Information

* Closes NVIDIA-NeMo/Automodel#3244
* Unit tests cover the averaging in eval mode, train mode restored after a failing batch, repeatable sampling with the training RNG left untouched, the dp reduction, and that the loop validates only when a loader is configured.
* Ran it for real on 2x H200 with FSDP2 (`dp_size: 2`) against a small Wan transformer: validation fired at the configured steps and the run finished clean. Also ran the same training twice, with and without validation enabled, and the training losses and final weights come out bit-identical, which is the property the fixed seed and RNG restore are there for.
* Not exercised: `cp_size > 1`, `tp_size > 1`, multi-node, and the LoRA and FP8 paths.

### DIFF
--- a/docs/guides/diffusion/finetune.mdx
+++ b/docs/guides/diffusion/finetune.mdx
@@ -1,6 +1,6 @@
 ---
 title: "Diffusion Model Fine-Tuning with NeMo AutoModel"
-description: ""
+description: "Fine-tune text-to-image and text-to-video diffusion models with NeMo AutoModel using flow matching, from dataset prep through multi-node training and inference."
 position: 16
 ---
 ## Introduction
@@ -46,7 +46,7 @@ All models use FSDP2 for distributed training and flow matching for loss computa
 ## Install NeMo AutoModel
 
 Diffusion fine-tuning needs the `diffusion` extra plus `diffusion-media` (OpenCV
-Diffusion fine-tuning requires the `diffusion` and `diffusion-media` extras. The `diffusion-media` extra provides OpenCV for image and video preprocessing and imageio-ffmpeg for T2V export:
+Diffusion fine-tuning requires the `diffusion` and `diffusion-media` extras. The `diffusion-media` extra provides OpenCV for image and video preprocessing, and `imageio-ffmpeg` for text-to-video export:
 
 ```bash
 uv venv
@@ -80,7 +80,7 @@ For the full set of installation methods, see the [installation guide](/get-star
 ## Prepare Your Dataset
 
 Diffusion models operate in latent space — a compressed representation of visual data — rather than directly on raw images or videos. To avoid re-encoding data on every training step, the preprocessing
-  pipeline encodes all inputs ahead of time and saves them as `.meta` or `.pt` cache files.
+Diffusion models operate in latent space (a compressed representation of visual data) rather than directly on raw images or videos. To avoid re-encoding data on every training step, the preprocessing pipeline encodes all inputs ahead of time and saves them as `.meta` or `.pt` cache files.
 
 Each cache file contains:
 
@@ -142,7 +142,8 @@ Fine-tuning is driven by two components:
 1. A recipe script (e.g., [`train.py`](https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/recipes/diffusion/train.py)) — the Python entry point that orchestrates the training loop: loading the model, building the dataloader, running forward/backward passes, computing the flow matching loss, checkpointing, and logging.
 2. A YAML configuration file — a text file in YAML format that specifies all settings the recipe uses: which model to fine-tune, where the data lives, optimizer hyperparameters, parallelism strategy, etc.
 1. A recipe script, for example [`train.py`](https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/recipes/diffusion/train.py), is the Python entry point that orchestrates the training loop: loading the model, building the data loader, running forward and backward passes, computing the flow matching loss, checkpointing, and logging.
-2. A YAML configuration file is a text file in YAML format that specifies all settings the recipe uses: which model to fine-tune, where the data lives, optimizer hyperparameters, parallelism strategy, and other settings. You customize training by editing this file rather than modifying code, allowing you to scale seamlessly from 1 to hundreds of GPUs.
+1. A recipe script, such as [`train.py`](https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/recipes/diffusion/train.py), orchestrates the training loop by loading the model, building the data loader, running forward and backward passes, computing the flow matching loss, checkpointing, and logging.
+2. A YAML configuration file is a text file in YAML format that specifies all settings the recipe uses: which model to fine-tune, where the data lives, optimizer hyperparameters, parallelism strategy, and other settings. You customize training by editing this file rather than modifying code, allowing you to scale seamlessly from one to hundreds of GPUs.
 
 The following annotated [`wan2_1_t2v_flow.yaml`](https://github.com/NVIDIA-NeMo/Automodel/blob/main/examples/diffusion/finetune/wan2_1_t2v_flow.yaml) config explains each section:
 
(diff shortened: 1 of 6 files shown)


beccohov:
Hey @HuiyingLi,
I created PR as we agreed in #3244

HuiyingLi:
Thank you! Taking a look

docs/guides/diffusion/finetune.mdx
jgerh:
```suggestion
description: "Fine-tune text-to-image and text-to-video diffusion models with NeMo AutoModel using flow matching, from dataset prep through multi-node training and inference."
```

jgerh:
```suggestion
Diffusion fine-tuning requires the `diffusion` and `diffusion-media` extras. The `diffusion-media` extra provides OpenCV for image and video preprocessing, and `imageio-ffmpeg` for text-to-video export:
```

jgerh:
```suggestion
Diffusion models operate in latent space (a compressed representation of visual data) rather than directly on raw images or videos. To avoid re-encoding data on every training step, the preprocessing pipeline encodes all inputs ahead of time and saves them as `.meta` or `.pt` cache files.
```

jgerh:
```suggestion
1. A recipe script, such as [`train.py`](https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/recipes/diffusion/train.py), orchestrates the training loop by loading the model, building the data loader, running forward and backward passes, computing the flow matching loss, checkpointing, and logging.
2. A YAML configuration file is a text file in YAML format that specifies all settings the recipe uses: which model to fine-tune, where the data lives, optimizer hyperparameters, parallelism strategy, and other settings. You customize training by editing this file rather than modifying code, allowing you to scale seamlessly from one to hundreds of GPUs.
```

docs/guides/diffusion/finetune.mdx
```diff
@@ -245,8 +245,42 @@ checkpoint:
...
+```
+
+The validation loader takes the same builders and fields as the training one and is sharded across
+data-parallel ranks the same way. Validation scores the held-out set with the training flow-matching
+objective in `eval()` mode under `torch.no_grad()`, and logs `val_loss` to the console and to
+Weights & Biases, on the same scale as `train_loss`.
+
+Timesteps and noise are drawn under a fixed seed, and the training RNG state is restored afterwards,
+so the curve reflects the model rather than the sampling and enabling validation does not change the
+training trajectory.
```
jgerh:
```suggestion
Timesteps and noise are drawn under a fixed seed, and the training RNG state is restored afterward. The curve therefore reflects the model rather than the sampling. Enabling validation does not change the training trajectory.
```

docs/guides/diffusion/finetune.mdx
jgerh:
```suggestion
For example, to train on two nodes with eight GPUs each (16 GPUs total):
```

jgerh:
```suggestion
| **Video generation on limited hardware** | [Wan 2.1 T2V 1.3B](#wan-21-t2v-13b) | Smallest model (1.3B params): fast iteration, fits on a single A100 40 GB |
```

jgerh (commented):
Completed tech pubs review of docs/guides/diffusion/finetune.mdx and provided a few copyedits and suggested text revisions.

beccohov:
Thanks, @HuiyingLi 
I'll wait till the tests finish and then probably I should update the docs according to what @jgerh suggested?

HuiyingLi:
YEah please apply the suggestions, thanks!

beccohov:
@HuiyingLi I applied the suggested changes to docs. Now we need "ok" to run tests

HuiyingLi:
@zyzhou5  could you please take a look too, thanks!

HuiyingLi:
@beccohov  could you please fix the ci error? Thank you so much!

beccohov:
Hey, @HuiyingLi 
I believe I addressed the issue in test - that was due to main update. Please launch the CI again. Thanks!

@jgerh I also fixed issue in duplicated documentations lines

examples/diffusion/finetune/wan2_2_t2v_flow.yaml
```diff
@@ -42,6 +44,16 @@ data:
     shuffle: true
     drop_last: false
     num_workers: 2
+  # Held-out cache for the validation loss. Shuffle off keeps the loss comparable across steps.
+  validation_dataloader:
+    _target_: nemo_automodel.components.datasets.diffusion.build_video_multiresolution_dataloader
+    cache_dir: PATH_TO_YOUR_WAN22_VAL_DATA
```
zyzhou5: We need to either comment this out or add an override in the ci launcher

nemo_automodel/recipes/diffusion/train.py
```diff
@@ -897,6 +925,59 @@ def _transformer_engine_fp8_context(self) -> Any:
...
+
+        Returns:
+            Mean loss over the validation batches of the data-parallel group, comparable to the
+            logged ``train_loss``.
+        """
+        self.model.eval()
+        local_loss_sum = 0.0
+        local_num_batches = 0
+        try:
+            with ScopedRNG(seed=self.seed, ranked=False), torch.no_grad(), self._autocast_context():
```
zyzhou5: We should use different seeds across DP-ranks to align with the training

zyzhou5 (commented):
Overall LGTM.

nemo_automodel/recipes/diffusion/train.py
beccohov: Agree!
One nuance on how: `ranked=True` adds the global rank, so with` cp_size > 1` it would decorrelate CP peers that are supposed to draw identical noise for the same batch. I'll seed with `self.seed + self._get_dp_rank()` instead — same thing setup() already does for the CP case - so DP ranks are decorrelated, CP peers stay aligned, and validation still draws the same values on every evaluation.

examples/diffusion/finetune/wan2_2_t2v_flow.yaml
beccohov: Maybe should I completely remove the val-related sections/args from the config? It's covered in docs now, so probably makes no sense to keep these dead lines here?

beccohov: All in all, I came up with idea that since this config is presented in docs anyways, we don't need to put `validation_dataloader` section in this config at all. I decided to keep it short and docs explain how to use the `validation_dataloader` just in case

beccohov:
Okay, I had to push again because test coverage was lower than 70%. I also took into account suggestions  @zyzhou5 (however, I decided to remove val section from guide config because guide explains how to use it anyways).
@HuiyingLi sorry, one more time for CI 😄

zyzhou5:
/ok to test [011d75b](https://github.com/NVIDIA-NeMo/Automodel/pull/3579/commits/011d75b14a29f0e02a79324f9fd9916693739a66)

HuiyingLi (approved):

Issue #3576, opened 2026-08-18
[Tracking] checkpoint I/O architecture, correctness, performance, and memory


yuhezhang-ai:
## Summary

Track the multi-PR effort to make checkpoint I/O correct, fast, and memory-bounded across model families, checkpoint layouts, distributed configurations, and hardware memory architectures.

This roadmap covers more than base-model load latency. It includes:

* Base-model loading and training resume.
* MoE tensor merging and other layout conversions.
* FP8, MXFP4, scale handling, and other precision transforms.
* Discrete CPU/GPU memory and unified-memory systems such as DGX Spark.
* Checkpoint save and consolidated export time and memory.
* End-to-end correctness and observability.

This is broader than the earlier recipe/thread-budget mitigation in NVIDIA-NeMo/Automodel#3160 and NVIDIA-NeMo/Automodel#3370. The current loading PRs are concrete steps toward the architecture described below, not the complete architecture.

## Current implementation stack

Review and merge the loading PRs in this order. Each PR is based on the PR immediately above it, so its GitHub diff contains only that step.

```text
main
  └─ #3580  grouped-MoE load overhead
       └─ #3610  low-memory DCP for custom MoE adapters
            └─ #3616  ordinary HF/external safetensors
                 └─ #3619  bounded FP8 conversion
                      └─ #3623  bounded GPT-OSS MXFP4 conversion
```

- [x] **Establish the adapter-owned direct-load safety contract** — NVIDIA-NeMo/Automodel#3574 (merged)
  * Distinguish destinations backed by final model weights from destinations that allocate transformed tensors.
  * Keep allocating and dequantizing cases on a safe fallback until a bounded implementation exists.

- [ ] **Reduce grouped-MoE load overhead** — NVIDIA-NeMo/Automodel#3580 (ready for review)
  * Fill final grouped expert tensors directly, avoid repeated full-heap GC, and reuse safe TE/MoK checkpoint views.
  * Make Gemma4 EP ranks read only their local experts instead of expanding every rank to the global expert tensor.
  * Headline result: Gemma4 26B EP8 model load **451.36 s → 26.60 s (17.0x)**; Qwen3 30B TE retains the lower-memory destination path.

- [ ] **Avoid complete CPU checkpoints for safe custom MoE adapters** — NVIDIA-NeMo/Automodel#3610 (draft, stacked on #3580)
[...]

PR #3574, opened 2026-08-18
refactor(checkpoint): gate direct loads by adapter capability

yuhezhang-ai:
# What does this PR do ?

Makes single-device direct checkpoint loading explicit and safe through an adapter-owned capability contract.

Current `main` already sends single-GPU custom models through DCP after PR NVIDIA-NeMo/Automodel#3533. This PR adds the missing ownership contract: audited adapters whose load destinations alias final model storage remain direct, while allocating or quantized adapters return to the conservative materializing fallback. It also opens each indexed fallback shard once per file, adds phase-level timing, and expands cross-model routing tests.

This distinction matters for performance attribution: Nemotron and the newly marked alias-safe adapters are not faster than current `main`; their opt-in prevents the safety gate from slowing them back to the fallback. The meaningful incremental change over current `main` is safe route selection and fallback reader efficiency.

# Performance evidence

The same Nemotron release recipe demonstrates why the direct route matters:

| Exact job | AutoModel ref | Checkpoint load |
| -- | -- | -- |
| [August 15 release](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/398442222>) | `448bf5061` | 58.82 GB in 165.45 s |
| [PR head](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/402313157>) | `55a83af71` | 58.82 GB in 30.36 s |

That is a 5.45x historical speedup and an 81.7% load-time reduction versus the [pre-NVIDIA-NeMo/Automodel#3533](<https://github.com/pre-NVIDIA-NeMo/Automodel/issues/3533>) release path. It is not an incremental speedup over current `main`: PR NVIDIA-NeMo/Automodel#3533 (`060cc495a`) already restored single-GPU custom-model DCP loading before this branch's final head.

Recent release jobs also show that the multi-rank opt-in representatives were already direct and remain performance-neutral. Using the same version-to-model-summary timestamp boundary where the older direct path lacked a phase timer: Qwen3 30B MoE was 24 s versus 26 s, Gemma4 2B was 13 s versus 13 s, Gemma4 26B MoE was 452 s versus 444 s, and Qwen3-VL 30B MoE was 35 s versus 36 s. ERNIE's exact full-state timer was 7.18 s versus 7.12 s. These are parity results, not new speedup claims.

# Changelog

* Add an adapter-owned capability for direct, write-through checkpoint loading.
* Keep audited alias-safe adapters on the direct route after applying the safety gate.
[...]

PR #3580, opened 2026-08-18
perf(checkpoint): reduce allocating grouped MoE load overhead

yuhezhang-ai:
## What this PR changes

This PR improves three grouped-MoE checkpoint paths left after NVIDIA-NeMo/Automodel#3574:

1. **MoE reconstruction:** write each expert directly into the final grouped tensor instead of building one tensor per expert and stacking them afterward. Use generation-0 garbage collection instead of repeatedly scanning the full Python heap.
2. **Transformer Engine and MoK destinations:** return the checkpoint-layout view that DCP should fill. TE's views are contiguous; some MoK views are not, but DCP supports both. This avoids allocating a second set of expert tensors and avoids full GC or `torch.cuda.empty_cache()` while those views remain in use.
3. **Gemma4 grouped checkpoints:** load expert tensors through transposed views of the model weights. With expert parallelism, each rank reads only its own experts instead of materializing all experts on every rank.

The loader also reports destination preparation, storage read, adapter conversion, and model installation separately.

This PR optimizes the existing load pipeline. General sequential checkpoint streaming remains follow-up work in NVIDIA-NeMo/Automodel#3576.

## Why the model paths differ

Qwen TE and Gemma4 both transform expert weights, but their checkpoint layouts are different:

* **Qwen TE:** the HF checkpoint has separate keys for each expert. TE exposes a temporary grouped stack. The adapter turns that stack back into per-expert HF tensors before DCP reads them, then combines the loaded values into TE parameters.
* **Gemma4:** the HF checkpoint already stores all 128 experts in two grouped tensors per layer. The model stores transposed grouped tensors and also absorbs a small per-expert scale into the down projection.

For Gemma4 EP8, the old path expanded every rank's 16 experts into full 128-expert CPU tensors, read the full grouped checkpoint, and then sliced back to 16 experts. The new path gives DCP transposed, expert-sharded views of the model weights, so each rank reads only its own 16-expert slice. The small scale vector is applied in place after the read.

### Gemma4 26B EP8

Both runs use the same 8×H100 `gemma4_26b_a4b_moe_peft` release recipe.

| Load phase | Before this optimization | Current | Reduction |
| -- | -- | -- | -- |
| Destination preparation | 165.90 s | **0.07 s** | **99.96%** |
| Checkpoint storage read | 275.96 s | **26.41 s** | **90.4%** |
| Adapter conversion | 9.37 s | **0.01 s** | **99.9%** |
| Model installation | 0.13 s | 0.12 s | — |
| **Total model load** | **451.36 s** | **26.60 s** | **94.1% / 17.0x** |

* Before: [job 405235261](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/405235261>)
* Current: [job 405305252](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/405305252>), [pipeline 63712014](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63712014>)

A matched local 8-H100 A/B measured 190.44 s before versus 24.46 s on the first optimized read and 13.87 s warm. Per-rank peak process RSS fell from 86.85 GiB to 7.41 GiB. Rank-by-rank expert probes matched, every rank owned 16 experts, and no meta parameters remained.

For single-device Gemma4, matched warm runs reduced peak process RSS from 64.30 GiB to 47.62 GiB (16.68 GiB / 25.9%). Model-load time itself was approximately neutral; EP is the large speed improvement.

### Qwen3 30B TE: corrected speed and memory explanation

The earlier explanation was incomplete. TE creates its virtual grouped tensor with `torch.stack(...).transpose(...)`. After the adapter slices and transposes an expert back to HF layout, that tensor is already contiguous. Therefore the old `.contiguous()` call reused the TE stack; it did not allocate a copy.

The large destination-setup speedup comes from avoiding repeated full `gc.collect()` calls. Full GC took roughly 19–23 seconds across the 46 expert projections. Generation-0 collection or no collection takes less than one second. `torch.cuda.empty_cache()` was not responsible for the old path's lower peak.

The initially proposed blank destinations were fast but always allocated a second set of expert tensors. The final implementation instead reuses TE's already-contiguous checkpoint views:

| Code state | Load-only elapsed | Final CUDA allocation / rank | Peak CUDA allocation / rank | What it shows |
| -- | -- | -- | -- | -- |
| Before this PR: `cat` then `stack` | 59.25 s | 7.109 GiB | 14.093 GiB | Original reconstruction and full GC |
| Direct fill + `from_hf` generation-0 GC | 35.43 s | 7.109 GiB | **13.999 GiB** | Direct fill saves 96 MiB/rank; cleanup provides most of the time gain |
| Rejected blank TE destinations | 15.21 s | 7.109 GiB | 20.609 GiB | Fast setup, but duplicates 6.61 GiB/rank |
| **Current: reuse TE views** | **20.36 s** | 7.109 GiB | **13.999 GiB** | Keeps the fast setup without the memory regression |

The first three rows are the matched EP8 job `16206220`. The final row is job `16219024`, using the same cached Qwen3-30B checkpoint, EP8 topology, staged image, and load-only harness. Storage timing varies with filesystem state, so CUDA peak is the controlled comparison. In the final run, the loader's own phase timer reported 18.26 s total: 0.40 s destination setup, 17.42 s storage read, 0.31 s adapter conversion, and 0.13 s installation.

All four states completed with zero meta parameters and matching rank-local expert fingerprints. The current path reduces the rejected implementation's peak by 6.610 GiB/rank (32.1%) and returns exactly to the direct-fill peak.

### Nemotron Nano V3 single-device fallback

Matched one-GPU runs used the same 58.82 GB checkpoint, recipe, image, and one-step workload.

| Load phase | Before this PR | Current | Reduction |
| -- | -- | -- | -- |
| Adapter conversion | 246.81 s | **146.69 s** | **40.6%** |
| Total model load | 265.78 s | **162.65 s** | **38.8%** |

The time reduction comes from replacing 46 full Python-heap scans—two expert projections across 23 MoE layers—with generation-0 collection. Direct fill separately removes one final-size reconstruction scratch buffer. The run completed successfully with the same finite first-step metrics as the parent. This older job did not record a clean load-only CUDA peak; that measurement is tracked with the sequential-loading follow-up.

## Correctness boundaries

* **Pretraining from config is unchanged.** Without a checkpoint, `load_model()` is not called and model initialization remains unchanged.
* **Checkpoint views are load-only.** DCP fully overwrites the returned views. Save/export conversion still creates contiguous tensors that preserve their current values.
* **No model-sized blank destinations are created.** DCP can fill contiguous and non-contiguous views, so both TE and MoK reuse their existing checkpoint-layout storage.
* **Partial loads preserve initialization.** Missing allowed keys are removed before DCP runs, and incomplete expert groups fail validation.
* **Quantized conversion is unchanged.** Quantized paths keep value-preserving conversion and the existing rebuild behavior.
* **Gemma4 EP direct load is topology-gated.** It requires materialized expert DTensors sharded only on the expert axis across EP. Unsupported sharding and quantized conversion use the existing fallback.

## Validation

* New TE storage test proves checkpoint-layout outputs reuse TE's temporary stack without allocating a second buffer.
* New MoK storage test proves its non-contiguous checkpoint-layout outputs reuse the existing grouped storage.
* New CUDA allocator regression test limits TE destination construction to 70 MiB for a 64 MiB stack; the rejected blank path would require 128 MiB.
* New CUDA allocator regression test limits direct fill to 40 MiB for 16 MiB of inputs plus one 16 MiB output; the old per-expert concatenations would raise the peak to 48 MiB.
* Focused GPU tests: **2 passed** in Slurm job `16219024`.
* MoE mixin CPU suite: **51 passed, 4 GPU-only skipped, 1 unrelated Gloo test deselected**.
* Qwen3 30B TE load-only EP8: job `16219024`, peak 13.999 GiB/rank, zero meta parameters, matching expert fingerprints.
* Qwen3 30B MoE checkpoint robustness: [pipeline 63370169](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63370169>).
* Gemma4 26B EP 50-step run: [job 405305252](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/405305252>).
* Gemma4 full checkpoint robustness: [job 405377241](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/405377241>), [pipeline 63721419](<https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63721419>) — base run plus all four save/reload phases passed.
* Ruff format/check and `git diff --check` pass.

Builds on NVIDIA-NeMo/Automodel#3574. Part of NVIDIA-NeMo/Automodel#3576.

### DIFF
--- a/nemo_automodel/components/moe/state_dict_mixin.py
+++ b/nemo_automodel/components/moe/state_dict_mixin.py
@@ -598,12 +598,23 @@ class MoESplitExpertsStateDictMixin:
             Creates gate_and_up_projs [n_experts, dim, inter_dim] and transposed down_projs tensors.
 
         Args:
+            hf_state_dict: State mapping consumed by this method. Per-expert gate and up tensors have shape
+                [expert_hidden, hidden], while down tensors have shape [hidden, expert_hidden]. DTensor values
+                use the same global layouts and are localized before merging.
+            device_mesh: Optional device mesh whose expert-parallel dimension selects the local experts. The
+                returned grouped expert tensors use the placements created by ``create_dtensor_from_local``.
             reset_view_loaded_keys: Clear the in-place (strided-view) loaded-key record at the
                 start of this call. A single ``from_hf`` may invoke this method more than once
                 (e.g. backbone then MTP merge); the later call(s) pass ``False`` so the view-loaded
                 keys accumulate across one logical load. Resetting here (rather than in the loader)
                 keeps the whole view-key lifecycle inside the adapter and ensures each load starts
                 clean (no leak from a prior load such as an init-time partial load).
+
+        Returns:
+            Native state mapping. Gated input projections have shape
+            [local_experts, hidden, 2 * expert_hidden], non-gated input projections have shape
+            [local_experts, hidden, expert_hidden], and down projections have shape
+            [local_experts, expert_hidden, hidden].
         """
         if reset_view_loaded_keys:
             self._view_loaded_native_keys = set()
@@ -696,7 +707,7 @@ class MoESplitExpertsStateDictMixin:
 
                     if all_complete:
                         expert_ids = sorted(expert_weights_by_layer[layer_num][native_key].keys())
-                        tensors = []
+                        expert_parts = []
                         for expert_id in expert_ids:
                             expert_data = expert_weights_by_layer[layer_num][native_key][expert_id]
 
@@ -709,29 +720,26 @@ class MoESplitExpertsStateDictMixin:
                                     up_weight = up_weight.to_local()
                                 gate_t = gate_weight.transpose(0, 1)
                                 up_t = up_weight.transpose(0, 1)
-                                tensors.append(torch.cat([gate_t, up_t], dim=-1))
+                                expert_parts.append((gate_t, up_t))
                             else:
                                 up_weight = expert_data
                                 if is_dtensor(up_weight):
                                     up_weight = up_weight.to_local()
-                                tensors.append(up_weight.transpose(0, 1))
+                                expert_parts.append((up_weight.transpose(0, 1),))
 
-                        stacked = torch.stack(tensors, dim=0).to(self.dtype)
-                        state_dict[native_key] = create_dtensor_from_local(stacked, device_mesh, rank)
+                        merged = self._direct_fill_grouped_expert_tensor(expert_parts)
+                        state_dict[native_key] = create_dtensor_from_local(merged, device_mesh, rank)
+                        merged_on_cuda = merged.is_cuda
 
-                        # Aggressively release intermediates so the per-layer
-                        # transient does not pile on top of the model's
-                        # already-materialized GPU DTensors. Without this,
-                        # ``tensors``/``stacked`` and the per-expert dict
-                        # entries hang around until Python's refcount GC
-                        # eventually runs — too late under tight GPU budgets
-                        # (e.g. a large MoE on 2 nodes / 8 GPUs).
-                        del tensors, stacked
+                        # Release the per-expert sources before processing the next projection or layer so they
+                        # do not accumulate alongside the grouped output. Only trim the CUDA allocator when the
+                        # merge itself used CUDA storage; the single-device fallback merges on the host.
+                        del expert_parts, merged
                         del expert_weights_by_layer[layer_num][native_key]
                         if not expert_weights_by_layer[layer_num]:
                             del expert_weights_by_layer[layer_num]
-                        gc.collect()
-                        if torch.cuda.is_available():
+                        gc.collect(0)
+                        if merged_on_cuda:
                             torch.cuda.empty_cache()
 
                 else:  # down_proj
@@ -740,7 +748,7 @@ class MoESplitExpertsStateDictMixin:
                     if len(expert_weights_by_layer[layer_num][native_key]) == expected_experts_per_rank:
                         expert_ids = sorted(expert_weights_by_layer[layer_num][native_key].keys())
 
-                        ordered = []
+                        expert_parts = []
                         for expert_id in expert_ids:
                             down_weight = expert_weights_by_layer[layer_num][native_key][expert_id]  # [dim, inter_dim]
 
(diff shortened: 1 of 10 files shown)


yuhezhang-ai:
/ok to test [74f3539](https://github.com/NVIDIA-NeMo/Automodel/pull/3580/commits/74f3539fc6386c6c234d5372df407af592888b7e)

yuhezhang-ai:
/ok to test [dfe3ea5](https://github.com/NVIDIA-NeMo/Automodel/pull/3580/commits/dfe3ea507cd103004ac9dc82e0a654e5d5d2fe66)

nemo_automodel/components/moe/state_dict_mixin.py
```diff
@@ -709,29 +720,26 @@ def _from_hf_w_merged_experts(
...
-
-                        # Aggressively release intermediates so the per-layer
-                        # transient does not pile on top of the model's
-                        # already-materialized GPU DTensors. Without this,
-                        # ``tensors``/``stacked`` and the per-expert dict
-                        # entries hang around until Python's refcount GC
-                        # eventually runs — too late under tight GPU budgets
-                        # (e.g. a large MoE on 2 nodes / 8 GPUs).
-                        del tensors, stacked
+                                expert_parts.append((up_weight.transpose(0, 1),))
```
akoumpa, line 727: Hi @yuhezhang-ai , i'm concerned about this part, in particular, whether it would cause any OOMs on MoE that are tight in VRAM. Do you have any memory profiles to share showing before/after? I just want to make sure since this path is used broadly across MoEs.

yuhezhang-ai: The new buffer `expert_parts` replaces the old stacked buffer (`tensors`, `stacked`), while removing per-expert concatenations, so the allocation structure is strictly equal or better. We have strong structural and host-cpu memory evidence, but I'll provide a representative before/after CUDA peak measurement to fully answer the VRAM question.

akoumpa: Thanks, @yuhezhang-ai! The before/after CUDA peak measurement would address my immediate concern. Could you also add a focused VRAM regression test for this adapter path, so we catch any future increase in peak CUDA memory automatically? The existing torch_memory_limit fixture may be reusable here.

yuhezhang-ai: Thanks for flagging this, Alex — you were right to ask for the memory measurement. I ran the Qwen3-MoE EP8 test and found a real regression: peak GPU memory increased from 13.999 GiB to 20.609 GiB per rank.

The regression is not caused by the direct-fill code in this thread. It comes from the later empty-destination change. For TE, the old .contiguous() call was already a no-op because the converted checkpoint views were contiguous and reused TE’s temporary stack. Replacing those views with torch.empty_like() allocated a second full set of expert tensors, adding about 6.61 GiB per rank.

I fixed that by reusing TE’s existing contiguous views. The peak is now back to 13.999 GiB per rank, with the same final model memory and matching loaded weights.

The main loading-speed improvement also comes from changing repeated full gc.collect() calls to generation-0 collection or no collection in both destination preparation and expert reconstruction. Direct fill itself has a smaller but real memory benefit: 14.093 GiB to 13.999 GiB, or 96 MiB per rank in EP8. I would keep it for that memory saving, while simplifying its explanation and implementation where possible.

I also added focused GPU-memory regression tests so both the TE duplication and the direct-fill temporary can’t be reintroduced unnoticed.

yuhezhang-ai:
/ok to test [8949e6b](https://github.com/NVIDIA-NeMo/Automodel/pull/3580/commits/8949e6bb2e6891f21d767f9b871ee2cae1d024d1)

akoumpa (approved):

PR #3007, opened 2026-07-10
feat: Support preemption checkpointing

edjson:
# What does this PR do ?

Adds support for saving a checkpoint when a job is about to be killed. Lets users pick which signal to listen for in the YAML config. When the signal arrives, all ranks save a checkpoint together at the next training step and shut down cleanly.

# Changelog

* Make signal handler watch for multiple signals
* Allow signals to be written as names in YAML ("SIGUSR1"), not just Python objects.
* Fix `slurm.sub` so the signal handling actually works: run `srun` in the background, set up the signal trap before launching, add the `--signal=B:USR1@300` line, and report the job's real exit code.

# Before your PR is "Ready for review"

**Pre checks**:

- [X] Make sure you read and followed [Contributor guidelines](<https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md>)
- [X] Did you write any new necessary tests?
- [X] Did you add or update any necessary documentation?

If you haven't finished some of the above items you can still open "Draft" PR.

# Additional Information

* Related to NVIDIA-NeMo/Automodel#2972

PR #3583, opened 2026-08-19
fix(retrieval): do not lose metric records on preemption or error

sahel-sh:
# What does this PR do ?

The bi-encoder recipe only closed its metric loggers on the success path, and records are buffered until close(), so a run that was preempted or raised lost every metric it had produced.

# Changelog

- close() both loggers in the finally, so an exception cannot skip them
- poll StepScheduler.sigterm_received per step; both of its iterators already stop on the flag, but nothing polled it inside the step loop, so it was only checked once per epoch
- size the training buffer from ckpt_every_steps, capped at DEFAULT_BUFFER_SIZE
- flush both loggers: draining the buffer only reaches the file object, whose own buffer dies with the process

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo_automodel/recipes/retrieval/train_bi_encoder.py
+++ b/nemo_automodel/recipes/retrieval/train_bi_encoder.py
@@ -448,8 +448,11 @@ class TrainBiEncoderRecipe(BaseRecipe):
                             train_loss=train_log_data.metrics["loss"],
                             val_loss=val_loss,
                         )
-                        # Flush AFTER the checkpoint lands, so a checkpoint is never on disk
-                        # without the metrics describing the steps it covers. Buffer size alone
+                        # Flush once the checkpoint save returns, so the metrics describing the
+                        # steps it covers are durable by the time the loop moves on rather than
+                        # waiting for the buffer to fill. This is a post-save boundary, not an
+                        # atomic one: a crash between the save completing and this flush leaves
+                        # the checkpoint on disk with those records still buffered. Buffer size alone
                         # does not give this: it counts records, and the count only lines up
                         # with checkpoint steps while training runs from step 0 with periodic
                         # checkpoints. Resuming at a step that is not a multiple of
--- a/nemo_automodel/recipes/retrieval/train_bi_encoder.py
+++ b/nemo_automodel/recipes/retrieval/train_bi_encoder.py
@@ -388,10 +388,13 @@ class TrainBiEncoderRecipe(BaseRecipe):
         )
         self._log_model_and_optimizer_details(self.model_parts, self.optimizer, self.lr_scheduler)
 
-        # Train metrics reach disk at least as often as the checkpoint they accompany,
-        # capped at the default so a long checkpoint interval cannot buffer unboundedly.
-        # flush because a buffered write only reaches the file object, whose own buffer is
-        # lost with the process. Validation is rare enough to write every point.
+        # buffer_size bounds how many records can be pending, so a hard kill that runs no
+        # cleanup loses at most one checkpoint interval of train metrics; it is NOT what keeps
+        # metrics in step with checkpoints, since it counts records rather than watching
+        # checkpoint boundaries. The explicit flush after each successful checkpoint is what
+        # provides that. flush=True because a buffered write only reaches the file object,
+        # whose own buffer dies with the process. Validation is rare enough to write every
+        # point.
         train_logger_kwargs = {"flush": True}
         if self.step_scheduler.ckpt_every_steps > 0:
             train_logger_kwargs["buffer_size"] = min(DEFAULT_BUFFER_SIZE, self.step_scheduler.ckpt_every_steps)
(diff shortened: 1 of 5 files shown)


nemo_automodel/recipes/retrieval/train_bi_encoder.py
```diff
@@ -432,12 +446,21 @@ def run_train_validation_loop(self):
                             val_loss=val_loss,
                         )
                     self._maybe_collect_garbage()
+
+                    # Poll per step; the StepScheduler iterators stop on the flag.
+                    if self.step_scheduler.sigterm_received:
```
oliverholworthy, line 468: @yuhezhang-ai I mentioned earlier about this property performing a side-effect, but upon closer inspection it only triggers the side-effectful signals_received check at most once per step since #3007 

This additional check here is necessary because the preceding is_ckpt_step property can short-circuit before evaluating sigterm_received when the step is already a periodic, epoch-boundary, or final checkpoint. In those cases, the checkpoint is saved but the signal flag would not be updated, so training could continue.

This reveals a potential StepScheduler design improvement: signal polling could happen explicitly once at a defined step boundary, while sigterm_received and is_ckpt_step become side-effect-free cached-state reads. That would make the behavior less dependent on property access order and boolean short-circuiting. Not blocking for this PR, but maybe worth looking into whether this could be clearer.

yuhezhang-ai (commented):
Thanks for fixing metric durability on exceptional and preempted exits. @sahel-sh A few follow-ups:

1. Please run Ruff formatting. The current lint job reports that `nemo_automodel/recipes/retrieval/train_bi_encoder.py` would be reformatted: https://github.com/NVIDIA-NeMo/Automodel/actions/runs/32207675344/job/95933955404

2. Please add focused CPU regression coverage for both critical paths introduced here:
   - a signal arriving on a scheduled checkpoint step, proving the explicit post-checkpoint poll detects it and stops the loop; and
   - buffered train/validation metrics being persisted when the loop raises and cleanup runs from `finally`.

3. Please revisit the checkpoint-alignment claim around `buffer_size=min(DEFAULT_BUFFER_SIZE, ckpt_every_steps)`. Record-count buffering is not necessarily checkpoint-aligned after resuming at a nonzero step or when checkpoints occur at epoch boundaries, so a checkpoint may be persisted while its metrics remain buffered. Either flush explicitly after successful checkpoint boundaries, or narrow/correct the guarantee and test the intended behavior.

The broader `StepScheduler` explicit-poll/state-query redesign can remain a follow-up; I am not asking for that refactor in this PR.

nemo_automodel/recipes/retrieval/train_bi_encoder.py
yuhezhang-ai: Thanks, this makes sense. The per-step cache means line 451 does not add a second collective on ordinary steps, while the explicit access covers checkpoint steps where `is_ckpt_step` short-circuits before polling the signal.

I agree this is not blocking for this PR. We should have a follow-up to make signal polling an explicit `StepScheduler` operation with state-only properties.

sahel-sh:
> Thanks for fixing metric durability on exceptional and preempted exits. @sahel-sh A few follow-ups:
> 
> 1. Please run Ruff formatting. The current lint job reports that `nemo_automodel/recipes/retrieval/train_bi_encoder.py` would be reformatted: https://github.com/NVIDIA-NeMo/Automodel/actions/runs/32207675344/job/95933955404
> 2. Please add focused CPU regression coverage for both critical paths introduced here:
>    
>    * a signal arriving on a scheduled checkpoint step, proving the explicit post-checkpoint poll detects it and stops the loop; and
>    * buffered train/validation metrics being persisted when the loop raises and cleanup runs from `finally`.
> 3. Please revisit the checkpoint-alignment claim around `buffer_size=min(DEFAULT_BUFFER_SIZE, ckpt_every_steps)`. Record-count buffering is not necessarily checkpoint-aligned after resuming at a nonzero step or when checkpoints occur at epoch boundaries, so a checkpoint may be persisted while its metrics remain buffered. Either flush explicitly after successful checkpoint boundaries, or narrow/correct the guarantee and test the intended behavior.
> 
> The broader `StepScheduler` explicit-poll/state-query redesign can remain a follow-up; I am not asking for that refactor in this PR.

@yuhezhang-ai thanks for the review, and apologies for my delay: 1 and 2 are done, for 3, I made sure flush happens at successful checkpoint boundaries per your suggestion

yuhezhang-ai (commented):
Thanks for the thoughtful follow-up and for addressing the earlier feedback. Ruff is clean now, and the three focused CPU tests pass locally. I found one blocking runtime issue in the new flush path, plus a test gap that currently hides it:

1. `MetricLogger.__init__` already assigns `self.flush = flush`, so the new `flush()` method is shadowed by that per-instance boolean. On the current head, `type(logger.flush)` is `bool`, and calling it raises `TypeError: 'bool' object is not callable`. Consequently, the first checkpoint will fail at `self.metric_logger_train.flush()`. Could you please rename the stored flag (for example, `_fsync_on_write`) or give the buffer operation a distinct name, and add a small test using the real `MetricLogger`?

2. `test_checkpoint_flushes_metrics_when_not_buffer_aligned` does not currently prove the checkpoint flush occurred: `at_save` is sampled before the flush, while the final persisted count is satisfied by `close()` in the `finally`. The test would still pass if the recipe's explicit checkpoint flush calls were removed. Please observe the real file after the checkpoint flush but before final close, or otherwise assert the intermediate durable state.

3. The comment saying that flushing *after* checkpoint save means a checkpoint is “never on disk without metrics” is stronger than the implementation guarantees: there is still a window between save completion and the subsequent flush. Please either flush before checkpoint publication if that strict invariant is intended, or narrow the wording to the actual post-save boundary guarantee.

Finally, could you please rebase the branch onto current `main` with DCO sign-off on every resulting commit? The current head is `bd4f63b7`, and its DCO check is `ACTION_REQUIRED` because the latest commit has no `Signed-off-by` trailer.

Thank you again—the overall direction looks good once these points are addressed.

yuhezhang-ai:
/ok to test [90077d6](https://github.com/NVIDIA-NeMo/Automodel/pull/3583/commits/90077d6202520c505c39892be79e196770d59b2a)

yuhezhang-ai:
/ok to test [399d0e6](https://github.com/NVIDIA-NeMo/Automodel/pull/3583/commits/399d0e6702824a0f909a60d0722ba790515a4cbe)

yuhezhang-ai (dismissed):

yuhezhang-ai:
/ok to test [85cb9c2](https://github.com/NVIDIA-NeMo/Automodel/pull/3583/commits/85cb9c2456ee86a74c16a05db03860154d31d25c)

yuhezhang-ai (approved):

PR #3589, opened 2026-08-19
fix(diffusion): apply compute_dtype via autocast when FSDP2 skips 

### LINKED ISSUE #3578: Diffusion training crashes with split torch_dtype/compute_dtype when the FSDP mesh is size 1
Ran the diffusion finetune recipe on a single GPU and it dies on the first training step whenever `model.torch_dtype` and `model.compute_dtype` differ:

```
File "diffusers/models/transformers/transformer_wan.py", line 670, in forward
    hidden_states = self.patch_embedding(hidden_states)
RuntimeError: Input type (c10::BFloat16) and bias type (float) should be the same
```

The exact same config on 2 GPUs with `dp_size: 2` trains fine, so it is the unsharded path that breaks.

Why it happens: `FSDP2Manager.parallelize` returns early when the world size or the FSDP mesh size is 1 (`nemo_automodel/components/distributed/fsdp2.py:126`), so `fsdp2_strategy_parallelize` never runs and the `MixedPrecisionPolicy(param_dtype=compute_dtype)` built in `_build_diffusion_parallel_manager_args` (`nemo_automodel/recipes/diffusion/train.py:202`) is never applied. Weights stay in `torch_dtype` while `FlowMatchingPipeline.step` casts the latents to `compute_dtype` (`nemo_automodel/components/flow_matching/pipeline.py:465`), so the first Conv3d gets bf16 activations against fp32 weights.

There is already a guard for this situation: `_validate_precision_configuration` (`nemo_automodel/recipes/diffusion/train.py:88`) rejects split storage/compute dtypes for DDP and PEFT/LoRA, for exactly the same reason. It just does not cover the case where FSDP is configured but effectively disabled.

Repro:

```yaml
model:
  pretrained_model_name_or_path: pruna-test/wan-t2v-tiny-random
  mode: finetune
  torch_dtype: float32
  compute_dtype: bfloat16
fsdp:
  dp_size: 1
flow_matching:
  adapter_type: "simple"
```

```bash
torchrun --nproc_per_node=1 examples/diffusion/finetune/finetune.py -c repro.yaml
```

Setting `compute_dtype: float32` makes the same run pass, and so does `--nproc_per_node=2` with `dp_size: 2`.

Expected: either the parameters get cast on the unsharded path and training runs, or setup fails early with the same actionable message `_validate_precision_configuration` already produces for DDP/PEFT. A dtype mismatch deep inside the model forward is a bad way to find out.

One caveat: I only reproduced the `world_size == 1` case. The `device_mesh.size() == 1` half of that condition (say `world_size=2, tp_size=2, dp_size=1`) looks like it hits the same early return, but I have not run it.

Environment: torch 2.10.0+cu130, diffusers 0.39.0, 1x A100, reproduced on main @ 69097120.

I would like to fix this one myself. My preference is to extend `_validate_precision_configuration` to also cover the mesh-size-1 case, so the run fails at setup with a clear message, unless you would rather have the model cast to the compute dtype on that path instead. Happy to go either way.

zyzhou5:
# What does this PR do ?

Split model.torch_dtype/model.compute_dtype relies on FSDP2's MixedPrecisionPolicy to cast parameters to the compute dtype. On a single-rank world or mesh, FSDP2Manager.parallelize skips sharding, so that cast never happens: parameters stay in torch_dtype while the flow-matching step still casts its inputs to compute_dtype. The forward pass then mixes dtypes and fails with

  RuntimeError: Input type (c10::BFloat16) and bias type (float) should
  be the same

on the transformer's fp32 patch embedding. The same config works on 2+ GPUs, and five shipped diffusion example configs use this split (torch_dtype: float32 with compute_dtype: bfloat16).

Wrap the forward pass in torch.autocast on that path, following the autocast approach the retrieval and dLLM recipes already use for FSDP2Config.autocast_dtype. Resident parameters and their gradients stay in torch_dtype, so optimizer state is unchanged. Numerics are close to but not identical to the sharded path, since autocast keeps reductions such as normalization in fp32 rather than casting every parameter.

Extract the skip condition into fsdp2_sharding_enabled() so the recipe checks the same predicate FSDP2Manager acts on instead of duplicating it.

LLM/VLM recipes drop mp_policy at world size 1 the same way, but degrade silently to fp32 compute rather than failing; that is left for a separate change.

Fixes #3578

### DIFF
--- a/nemo_automodel/components/distributed/fsdp2.py
+++ b/nemo_automodel/components/distributed/fsdp2.py
@@ -35,6 +35,25 @@ from nemo_automodel.components.distributed.parallelizer import (
 logger = logging.getLogger(__name__)
 
 
+def fsdp2_sharding_enabled(device_mesh: DeviceMesh) -> bool:
+    """Report whether :meth:`FSDP2Manager.parallelize` shards the model for this mesh.
+
+    Parallelization is skipped on a single-rank world or a single-element mesh, which
+    also skips every side effect of ``fully_shard`` — most importantly the
+    ``MixedPrecisionPolicy`` cast of parameters to the compute dtype. Callers that
+    depend on that cast must check this instead of assuming FSDP2 is active.
+
+    Args:
+        device_mesh: Device mesh the ``FSDP2Manager`` was constructed with.
+
+    Returns:
+        True when ``fully_shard`` is applied, False when parallelization is skipped.
+    """
+    if get_world_size_safe() == 1 or device_mesh.size() == 1:
+        return False
+    return True
+
+
 def _patch_is_packed_sequence_for_training() -> None:
     """Eliminate CPU-GPU sync from flash attention for standard (non-packed) training.
 
@@ -123,7 +142,7 @@ class FSDP2Manager:
         Returns:
             The parallelized model.
         """
-        if get_world_size_safe() == 1 or self.device_mesh.size() == 1:
+        if not fsdp2_sharding_enabled(self.device_mesh):
             logger.info("World size or FSDP mesh size is 1, skipping parallelization.")
             if self.activation_checkpointing:
                 if is_selective_activation_checkpointing(self.activation_checkpointing):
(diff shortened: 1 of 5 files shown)


zyzhou5:
/ok to test [7add837](https://github.com/NVIDIA-NeMo/Automodel/commit/7add83755e4f7600df6b1e624776b68f7853ff5a)

pthombre (approved):

PR #3591, opened 2026-08-19
refactor(checkpoint): StateDictAdapter method target-module rename

stanley1208:
@HuiyingLi    follow up to #3435.

made the rename hook a real method on `StateDictAdapter` instead of a getattr. so now the base class has the method, it just returns the name unchanged, and kimi_k3 overrides it with its own renames.

some older adapters (llama, qwen2/3, kimivl) don't inherit from `StateDictAdapter`, so I added an isinstance check to keep their saves working like before.


HuiyingLi:
Thank you I appreciate it!

akoumpa (approved):

PR #3606, opened 2026-08-21
feat(moe): add MoK MXFP8 expert support

HuiyingLi:
## Summary

Add MXFP8 routed-expert compute to the Mixture-of-Kittens backend.

- Add `backend.mok.precision: bf16 | mxfp8` with BF16 as the default.
- Quantize each routed Gate, Up, and Down weight into MoK's normal and transposed MXFP8 layouts.
- Reuse quantized weights across activation-checkpoint recomputation and microbatches, then invalidate them immediately after `optimizer.step()` updates the BF16 master weights.
- Keep the existing shared-expert and BF16 paths unchanged.
- Reject `ep_shard > 1` with a clear error. Expert-FSDP support is intentionally outside this PR.

## Performance

Both comparisons use full pretrained weights, FineWeb data, 8K sequence length, global batch size 64, world size 64, EP64, PP1, 20 steps with 5 warmup steps, and fake balanced routing disabled. Speedup is normalized to the HybridEP BF16 step time for the same model. Lower step time and higher speedup are better.

### DeepSeek V4 Flash — TileLang attention

```mermaid
%%{init: {"xyChart": {"width": 520, "height": 300}}}%%
xychart-beta
    title "DeepSeek V4 Flash 8K speedup vs HybridEP BF16"
    x-axis ["HEP BF16", "HEP MXFP8", "MoK BF16", "MoK MXFP8"]
    y-axis "Speedup (x)" 0.9 --> 1.3
    bar [1.000, 1.066, 1.136, 1.234]
```

| Backend | Step time | Speedup vs HEP BF16 | W&B |
|---|---:|---:|---|
| HybridEP BF16 (`torch_mm`) | 9.1640 s | 1.000x | [run](https://wandb.ai/Nemo-automodel/mok/runs/q7nqqwea) |
| HybridEP MXFP8 (TE) | 8.5959 s | 1.066x | [run](https://wandb.ai/Nemo-automodel/mok/runs/9nwcfbsq) |
| MoK BF16 | 8.0644 s | 1.136x | [run](https://wandb.ai/Nemo-automodel/mok/runs/kky9rxda) |
| **MoK MXFP8** | **7.4245 s** | **1.234x** | [run](https://wandb.ai/Nemo-automodel/mok/runs/apfgbb1l) |

MoK MXFP8 is 1.086x faster than MoK BF16 and 1.158x faster than HybridEP MXFP8.

### GLM-5.2 — cuDNN attention

```mermaid
%%{init: {"xyChart": {"width": 520, "height": 300}}}%%
xychart-beta
    title "GLM-5.2 8K speedup vs HybridEP BF16"
    x-axis ["HEP BF16", "HEP MXFP8", "MoK BF16", "MoK MXFP8"]
    y-axis "Speedup (x)" 0.9 --> 1.3
    bar [1.000, 1.111, 1.174, 1.295]
```

| Backend | Step time | Speedup vs HEP BF16 | W&B |
|---|---:|---:|---|
| HybridEP BF16 (`torch_mm`) | 6.2183 s | 1.000x | [run](https://wandb.ai/Nemo-automodel/mok/runs/89de7th8) |
| HybridEP MXFP8 (TE) | 5.5979 s | 1.111x | [run](https://wandb.ai/Nemo-automodel/mok/runs/xu4ka137) |
| MoK BF16 | 5.2987 s | 1.174x | [run](https://wandb.ai/Nemo-automodel/mok/runs/lk2bbk6o) |
| **MoK MXFP8** | **4.8013 s** | **1.295x** | [run](https://wandb.ai/Nemo-automodel/mok/runs/mmd4uzr1) |

MoK MXFP8 is 1.104x faster than MoK BF16 and 1.166x faster than HybridEP MXFP8.

## 100-step convergence parity

The full pretrained models were trained on the same deterministic FineWeb shard with 8K sequence length, global batch size 64, world size 64, EP64, PP1, activation checkpointing enabled, and fake balanced routing disabled. Each comparison uses the same initial checkpoint and data order. Step time excludes the first five warmup steps.

| Model | HybridEP BF16 loss | MoK MXFP8 loss | 100-step MAE | Max gap | Pearson r | HEP / MoK step time | W&B overlay |
|---|---:|---:|---:|---:|---:|---:|---|
| DeepSeek V4 Flash | 9.1327 → 6.4058 | 9.1295 → 6.4084 | 0.00109 | 0.0060 | 0.999996 | 8.8030 / 7.3304 s | [HEP](https://wandb.ai/Nemo-automodel/mok/runs/g6zk6mg7) / [MoK](https://wandb.ai/Nemo-automodel/mok/runs/o7zi6mm9) |
| GLM-5.2 | 1.9717 → 0.0191 | 1.9725 → 0.0221 | 0.02125 | 0.0511 | 0.999693 | 6.5405 / 4.8640 s | [HEP](https://wandb.ai/Nemo-automodel/mok/runs/k279zt0s) / [MoK](https://wandb.ai/Nemo-automodel/mok/runs/4ymv29ic) |

## Validation

- `218 passed`: focused MoK, MoE parallelizer, and training-recipe unit tests.
- Four full-model, 100-step HybridEP BF16 versus MoK MXFP8 convergence runs completed successfully.
- Ruff checks and formatting checks pass for every changed file.
- `git diff --check` passes.

- [CICD NeMo run 33983262747](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/33983262747) passed on `95205e446`: CPU `12518 passed`, GPU 1/2 `6850 passed`, GPU 2/2 `6774 passed`.


examples/llm_pretrain/glm/glm_5.2_mok_mxfp8_mock_8k.yaml
```diff
@@ -0,0 +1,137 @@
...
+checkpoint:
+  enabled: false
+  dequantize_base_checkpoint: true
+
+loss_fn:
+  _target_: nemo_automodel.components.loss.masked_ce.MaskedCrossEntropy
+  fp32_upcast: true
+
+dataset:
+  _target_: nemo_automodel.components.datasets.llm.mock_iterable_dataset.MockIterableDataset
```
akoumpa: @HuiyingLi can we use an actual dataset instead of mock? let's leave the mock for ci.

examples/llm_pretrain/deepseek_v4/deepseek_v4_flash_mok_mxfp8_mock_8k.yaml
```diff
@@ -0,0 +1,139 @@
...
+checkpoint:
+  enabled: false
+  dequantize_base_checkpoint: true
+
+loss_fn:
+  _target_: nemo_automodel.components.loss.masked_ce.MaskedCrossEntropy
+  fp32_upcast: true
+
+dataset:
+  _target_: nemo_automodel.components.datasets.llm.mock_iterable_dataset.MockIterableDataset
```
akoumpa: same as glm

nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -1253,6 +1257,11 @@ def _run_train_optim_step(self, batches, max_grad_norm: float | None = None):
             opt.step()
             opt.zero_grad()
 
+        if mok_mxfp8_runtimes := getattr(self, "_mok_mxfp8_runtimes", ()):
+            from nemo_automodel.components.moe.mok_experts import clear_mok_mxfp8_optimizer_step_cache
+
+            clear_mok_mxfp8_optimizer_step_cache(mok_mxfp8_runtimes)
```
akoumpa: @HuiyingLi do we need to run this before or after opt.step? Also any way to avoid the getattr?

nemo_automodel/components/moe/mok_experts.py
```diff
@@ -261,30 +345,43 @@ def backward(
...
+                backward_routed_up_weights,
+                backward_routed_down_weights,
+                self.swiglu_limit,
+            )
+        finally:
+            if not self._retain_mxfp8_cache_until_optimizer_step:
+                self._clear_mxfp8_forward_cache()
 
 
 class _MoKAutogradFunction(torch.autograd.Function):
```
akoumpa, line 412: @HuiyingLi do we cover `_MoKAutogradFunction` in the functional tests?

nemo_automodel/components/moe/mok_experts.py
```diff
@@ -135,6 +157,53 @@ def __init__(self, mok_config: MoKBackendConfig, *, swiglu_limit: float) -> None
...
+        )
+
+    @staticmethod
+    def _weight_cache_key(weight: torch.Tensor) -> tuple[object, ...]:
+        """Identify one materialized routed weight without retaining the BF16 tensor."""
+        try:
+            version: int | None = weight._version
+        except RuntimeError:  # Inference tensors do not expose a version counter.
+            version = None
+        return (weight.device, weight.data_ptr(), weight.storage_offset(), weight.shape, weight.stride(), version)
```
akoumpa: This cache key can't detect an optimizer update for EP expert weights, so it never invalidates. Repro below uses the real `_weight_cache_key` / `_get_mxfp8_weights` from this branch — only mok's CUDA quantizer is faked, as a call counter, so the cache logic under test is untouched.

<details>
<summary><code>prove_stale_cache.py</code> (single process, 1 GPU)</summary>

```python
import os, sys, torch, torch.nn as nn, torch.distributed as dist
sys.path.insert(0, "<path-to-this-branch>")
os.environ.setdefault("MASTER_ADDR", "127.0.0.1"); os.environ.setdefault("MASTER_PORT", "29777")
dist.init_process_group("nccl", rank=0, world_size=1); torch.cuda.set_device(0)

from torch.distributed.device_mesh import init_device_mesh
from torch.distributed.tensor import distribute_tensor, Shard
from nemo_automodel.components.models.common.utils import MoKBackendConfig
import nemo_automodel.components.moe.mok_experts as mok
from nemo_automodel.components.moe.mok_experts import _MoKRuntime, _local_tensor

calls = 0                                  # fake ONLY the CUDA quantizer, to count real invocations
def fake_quantize(w):
    global calls; calls += 1
    z = torch.zeros(1, device=w.device)
    return (w.to(torch.float8_e4m3fn), z, w.transpose(-1, -2).contiguous().to(torch.float8_e4m3fn), z)
mok._mxfp8_weight_both = fake_quantize

mesh = init_device_mesh("cuda", (1,))
def make_param():                          # [E, hidden, inter], EP-sharded on dim 0
    return nn.Parameter(distribute_tensor(                       # == ExpertParallel._partition_fn
        torch.randn(4, 256, 256, device="cuda", dtype=torch.bfloat16), mesh, [Shard(0)]))
gate, up, down = make_param(), make_param(), make_param()

runtime = _MoKRuntime(MoKBackendConfig(precision="mxfp8"), swiglu_limit=0.0)
runtime._retain_mxfp8_cache_until_optimizer_step = True          # what train_ft.setup() does

opt = torch.optim.AdamW([gate, up, down], lr=1.0, betas=[0.9, 0.95], eps=1e-8,
                        weight_decay=0.1, foreach=False)         # the shipped YAML config

locals_ = lambda: (_local_tensor(gate), _local_tensor(up), _local_tensor(down))  # what forward() passes
snapshot = _local_tensor(gate).clone()

first = runtime._get_mxfp8_weights(*locals_()); key0 = runtime._mxfp8_cache_key   # step-0 forward
for _ in range(3):
    sum(_local_tensor(p).float().pow(2).sum() for p in (gate, up, down)).backward()
    opt.step(); opt.zero_grad()
later = runtime._get_mxfp8_weights(*locals_()); key1 = runtime._mxfp8_cache_key   # next forward

drift = (_local_tensor(gate) - snapshot).abs().max().item()
print(f"BF16 master weights changed by 3 AdamW steps : {drift > 0}  (max|delta| = {drift:.4f})")
print(f"_weight_cache_key identical across the steps : {key0 == key1}")
print(f"cache returned the SAME stale fp8 objects    : {first is later}")
print(f"mxfp8_quantize invocations (3 weights x N)   : {calls}")
runtime._clear_mxfp8_forward_cache()                             # what clear_mok_...() does
runtime._get_mxfp8_weights(*locals_())
print(f"...after the explicit clear                  : {calls}")
dist.destroy_process_group()
```
</details>

Output:

```
BF16 master weights changed by 3 AdamW steps : True  (max|delta| = 4.0625)
_weight_cache_key identical across the steps : True
cache returned the SAME stale fp8 objects    : True
mxfp8_quantize invocations (3 weights x N)   : 3      <- quantized once, at step 0
...after the explicit clear                  : 6      <- only the explicit clear re-quantizes
```

The weights moved by 4.06 and the key never noticed. Three swaps make the mechanism explicit:

| swap | key detects change |
|---|---|
| as above (`DTensor`, `foreach=False`) | **no** |
| `foreach=True` | yes |
| `fused=True` | **no** |
| `make_param()` → `nn.Parameter(torch.randn(...))` (plain tensor) | yes |

That last row is why CI is green: `test_mok_experts.py` builds plain tensors, where `_version` bumps normally (`0 → 2 → 4`). Under DTensor the optimizer math runs through `__torch_dispatch__`, so the bump lands on the wrapper's version counter and `to_local()`'s stays at `0` forever — and `data_ptr`/`shape`/`stride` are stable because experts are `ignored_params` and never FSDP-wrapped at `ep_shard=1`. Every component of the key is constant.

Two consequences:

1. The docstring just above (*"Its key includes storage identity and tensor version so a rematerialized or modified weight cannot reuse stale quantized data"*) doesn't hold for the configuration this PR ships. Whether the guard works at all is decided by an unrelated optimizer implementation flag, and both new YAMLs set `foreach: false`.
2. `clear_mok_mxfp8_optimizer_step_cache` is therefore load-bearing rather than belt-and-braces. `train_ft.py:657` enables retention in the **base** `setup()` (disabling the `finally:` clear at line 383), but the compensating clear exists only at `train_ft.py:1261` and `benchmark.py:379`. These three inherit that `setup()` and override `_run_train_optim_step` wholesale, so they'd quantize once at step 0 and reuse forever — no crash, no NaN, just frozen routed experts while everything else trains:
   - `recipes/llm/kd.py:833`
   - `recipes/llm/kd.py:931` (PP variant)
   - `recipes/dllm/train_ft.py:557`

   Nothing shipped hits this today, but `DiffusionGemmaSFTRecipe` already runs `ep_size: 8`, so it's one config line away.

Suggestion: replace the `data_ptr`/`_version` key with a generation counter owned by the module and bumped on invalidation, and install the invalidation via `optimizer.register_step_post_hook` instead of hand-copying it to each `opt.step()` site — that removes the whole class of miss rather than patching sites one at a time (cf. b1e004e2a, which patched `benchmark.py` after the fact).

examples/llm_pretrain/glm/glm_5.2_mok_mxfp8_fineweb_8k.yaml
```diff
@@ -0,0 +1,137 @@
...
+wandb:
+  enable: false
+  entity: ${oc.env:WANDB_ENTITY,Nemo-automodel}
+  project: ${oc.env:WANDB_PROJECT,mok}
+  name: glm-5.2-mok-mxfp8-mock-8k
+  group: mxfp8-core-convergence100-glm52-mock-8k
+  mode: online
+  dir: ${oc.env:WANDB_DIR,training_logs/wandb}
+
+ci:
```
akoumpa, line 135: @HuiyingLi i think this won't be picked up by the test generator, can you take a look and include it wherre appropriately?

examples/llm_pretrain/glm/glm_5.2_mok_mxfp8_mock_8k.yaml
```diff
@@ -0,0 +1,137 @@
...
+  top_k_experts: 0
+
+wandb:
+  enable: false
+  entity: ${oc.env:WANDB_ENTITY,Nemo-automodel}
+  project: ${oc.env:WANDB_PROJECT,mok}
+  name: glm-5.2-mok-mxfp8-mock-8k
+  group: mxfp8-core-convergence100-glm52-mock-8k
+  mode: online
+  dir: ${oc.env:WANDB_DIR,training_logs/wandb}
```
akoumpa: @HuiyingLi does `${oc.env:WANDB_DIR,training_logs/wandb}` work? 😕

akoumpa (approved):

PR #3610, opened 2026-08-21
perf(checkpoint): expand low-memory DCP support across MoE adapters

### LINKED ISSUE #3664: Avoid full CPU checkpoint loads for safe single-GPU MoE adapters
Tracks PR #3610 under #3576.

## Scope

Use ordinary DCP on one GPU when most tensors load into final model weights and any remaining temporary tensors are small. Covers Nemotron Nano V3, Ling, and the shared one-process TE/EP correction.

The final contract review is included in #3610 rather than split into duplicate issues:

- Match checkpoint storage capability to the expert implementation selected by the MoE constructor for all maintained EP dispatchers.
- Register in-place loaded-key bookkeeping only during checkpoint loads, so save/export cannot affect a later `from_hf` call.
- Treat grouped expert MLP bias honestly: `expert_bias=True` does not opt into low-memory DCP and unsupported split/merge conversion fails clearly instead of passing keys through silently.
- Keep router correction-bias handling separate and supported.
- Guard dense Nemotron MTP conversion, forward MTP conversion kwargs, and re-prefix Nemotron Omni view-loaded keys.
- Discover HF key names with shape-only tensors while excluding `_extra_state` metadata.
- Cover Ling with an end-to-end safetensors DCP reader test.

Supported distributed Ling and Nemotron configurations keep their existing rank-local DCP path. The allowed but unshipped `experts="torch"` + EP-dispatcher combination now rebuilds temporary TE storage instead of treating it as final model memory.

## Evidence

Nemotron loader: 157.92 s to 28.64 s; host RSS: 114.86 GiB to 6.04 GiB. Ling loader: 26.22 s to 12.73 s; host RSS: 58.22 GiB to 11.27 GiB.

Local validation at `d047b3a2d`:

- Changed checkpoint and adapter tests: 362 passed, 9 skipped.
- Affected grouped-MoE adapter regression tests: 109 passed, 23 skipped.
- Infrastructure and Gemma4 adapter tests: 55 passed, 30 skipped.
- Ruff and `git diff --check` pass.

## Follow-ups

- #3688 — grouped expert MLP bias split/merge conversion.
- #3689 — centralize MoE runtime/backend capability truth without violating component boundaries.

The broader composite-adapter audit remains a note on #3576 until it has a concrete affected set and acceptance matrix.

## Dependency

#3580 is merged.

yuhezhang-ai:
## Summary

This replaces #3601, which GitHub automatically closed when its head branch was renamed. The old draft had no human reviews or discussion to migrate.

This PR adds a shipped single-GPU Ling-mini-2.0 recipe and lets that configuration use AutoModel's existing DCP checkpoint loader instead of first building the complete checkpoint on CPU. It also corrects the route for distributed TE/EP backend configurations run with one process, where the runtime constructs ordinary grouped experts:

```text
Before:
safetensors -> complete CPU checkpoint -> model-specific conversion -> GPU model

This PR:
safetensors -> DCP -> existing GPU model weights
```

Ling still needs a small exception for its fused attention weights:

```text
fused QKV checkpoint tensor -> temporary fused QKV tensor -> final Q, K, and V model weights
```

There is no new streaming executor, load-group API, or per-adapter iteration protocol. The implementation reuses one standard DCP load and the adapters' existing `to_hf` / `from_hf` conversion.

Adapters opt into this route with one capability: `supports_low_memory_dcp_load`. It means that most checkpoint tensors load directly into the model's existing weight memory, while any remaining temporary tensors are small enough to be safe on one GPU. Adapters that need another model-sized set of tensors keep the CPU fallback.

This was originally stacked on #3580, which is now merged. That work makes the existing MoE conversion path faster and more memory-efficient. This PR changes which single-GPU configurations can skip the full-CPU-checkpoint fallback entirely.

The broad standard-HF/external-model one-GPU case is handled separately by #3616. This PR is the custom-adapter capability contract and the Ling implementation; adapters whose current conversion materializes model-sized tensors remain on the fallback.

Review later exercised the shared MoE adapter contract beyond the two benchmarked configurations. The resulting safeguards and pre-existing adjacent fixes are separated under **Scope added during review** so they are not mistaken for additional performance claims.

## Why the one-process TE/EP fallback can load directly

The Nemotron benchmark intentionally used TE experts and a DeepEP dispatcher with one process, matching a distributed backend configuration collapsed to one GPU. This is a supported fallback topology, but it is not the normal shipped one-GPU Nano V3 configuration: the shipped recipe uses `torch_mm` experts with the `torch` dispatcher and already used DCP before this PR.

With one process, the MoE layer cannot use expert parallelism, so it deliberately constructs ordinary `GroupedExperts`. The adapter previously looked only at the configured backend. It therefore assumed the expert tensors were TE temporary stacks and selected the full CPU fallback, even though the actual one-GPU model has ordinary grouped parameters that DCP can fill directly.

This PR makes the shared MoE adapter check match the MoE constructor:

- TE + EP dispatcher + one process: ordinary grouped parameters; direct load is safe.
- TE + EP dispatcher + multiple processes: TE grouped storage; keep the existing conversion path.
- Non-EP dispatcher: ordinary grouped parameters; direct load is safe.
- MoK: keep its existing special handling.

Nemotron's expert checkpoint tensors are non-contiguous transposed views of the grouped model weights. DCP writes through those views into the model's existing memory.

## Why Ling no longer needs load groups

Ling's expert and ordinary weights already have safe DCP destinations. Only its checkpoint QKV layout differs: the checkpoint stores one fused tensor while the model owns three Q, K, and V parameters.

The normal adapter path already handles this correctly:

1. Create the fused-QKV destinations (about 0.23 GiB total for Ling-mini-2.0).
2. Let DCP load the checkpoint once.
3. Split the loaded fused tensors into the final Q, K, and V weights.

The adapter now declares that this small temporary state is safe, so the loader does not materialize the complete 30.28 GiB checkpoint on CPU.

The discarded prototype loaded one dependency group at a time and reduced Ling peak host RSS further, from 11.27 GiB to 2.75 GiB. It required more than 1,000 lines of generic executor, validation, and adapter-specific planning code. The standard DCP implementation is slightly faster in the matched measurements below, keeps host memory 80.6% below the parent, and removes that framework/API complexity.

## Small shared cleanup

AutoModel asks adapters for their Hugging Face checkpoint key names before sharding. The base implementation previously performed this keys-only operation with real model tensors, so an allocating adapter could accidentally convert real weights just to discover names.

The base adapter now performs key discovery with shape-only meta tensors. This also replaces Gemma4's model-specific version of the same workaround.

### Nemotron Nano V3, one-process TE/DeepEP fallback

The 58.82 GiB checkpoint files were not already in the compute node's RAM cache (`mincore`: 0.0 GiB resident before construction). The run used an H100. The timed loader phase includes DCP planning, storage access, destination writes, and synchronization; it is not a pure filesystem-read timer.

| Code state | Job | Checkpoint loader | Full construction | Peak host RAM (RSS) | Peak GPU allocation |
| --- | --- | ---: | ---: | ---: | ---: |
| #3580 parent: full CPU fallback | `16225281` | 157.92 s | 160.814 s | 114.858 GiB | 58.831 GiB |
| This PR (`1933190d5`) | `16275130` | **28.64 s** | **30.005 s** | **6.037 GiB** | 58.840 GiB |

Compared with the parent, this is:

- **5.51x faster** in the checkpoint loader.
- **5.36x faster** through full model construction.
- **108.82 GiB / 94.7% lower** peak host RSS.
- About **9.3 MiB** of GPU peak above the final 58.831 GiB model.

The run loaded all parameters, left no meta parameters, and reproduced the established pretrained embedding, expert gate/up, expert down, and shared-expert fingerprints.

### Ling-mini-2.0, one GPU

Job `16275108` used the same 30.28 GiB Hugging Face checkpoint files, H100, staged image, and alternating fresh Python processes. The initial treatment process placed the files in the compute node's RAM cache and is excluded from the means.

| Code state | Checkpoint loader, two samples | Mean loader | Mean construction | Peak host RAM (RSS) | Peak GPU allocation |
| --- | --- | ---: | ---: | ---: | ---: |
| #3580 parent (`d6f363166`) | 26.77 s, 25.66 s | 26.22 s | 27.63 s | 58.218 GiB | 32.378 GiB |
| This PR (`1933190d5`) | 12.81 s, 12.65 s | **12.73 s** | **13.66 s** | **11.273 GiB** | 32.378 GiB |

Compared with the parent, this is:

- **2.06x faster** in the checkpoint loader.
- **2.02x faster** through full model construction.
- **46.95 GiB / 80.6% lower** peak host RSS.
- No increase in peak GPU allocation.

Every sample produced identical Q/K/V, expert, and routing-bias probes, the same forward checksum (`176.02343750`) and top token (`670`), finite logits, and no meta parameters.

## Scope

The measurable route change is for single-device, non-quantized custom-model safetensors initialization.

- The new single-GPU Ling recipe added by this PR now skips the full CPU checkpoint described above. The benchmarked Nemotron TE+DeepEP one-process fallback also skips it; the normal shipped one-GPU Nano V3 recipe already used DCP.
- The shared one-process TE/EP correction also applies to MoE adapters whose other tensors already satisfy the low-memory DCP requirement: Qwen3-MoE, Qwen3-Next, Qwen3-Omni-MoE, ERNIE-MoE, GLM-MoE, HY-V3, HY-MT2, and Laguna.
- Supported multi-GPU Ling and Nemotron configurations keep the standard rank-local DCP path. The review fix does change the allowed but unshipped `experts="torch"` + EP-dispatcher combination from an unsafe write-through destination to the rebuilding path, matching the TE implementation that the MoE constructor actually selects at world size greater than one.
- Non-TE grouped backends that support direct loading were already on the standard DCP path.
- Quantized initialization, legacy `.bin` checkpoints, and adapters with full-size transformations keep their existing fallback and remain follow-up work under #3576.
- Pretraining from config is unchanged because it does not load a base checkpoint.

## Scope added during review

The benchmarked behavior change remains the shipped single-device Ling route and the one-process TE/EP fallback described above. Because the implementation changes a shared MoE capability and its in-place-load bookkeeping, review exercised a wider adapter matrix. The resulting changes are classified below rather than presented as one undifferentiated expansion of the optimization.

### Required hardening for the new route

- The storage predicate now mirrors the runtime MoE constructor across maintained EP dispatchers and grouped expert backends. This prevents a configuration from being marked direct-load-safe when its checkpoint destinations are actually temporary tensors.
- In-place expert keys are registered only for `to_hf(for_checkpoint_load=True)`. Qwen3 Omni MoE, HY-V3, and HY-MT2 now forward that option through their public adapters. This prevents save/export conversion from leaving bookkeeping that could incorrectly suppress a later rebuild.
- Composite adapters report view-loaded weights under real model names. Nemotron Omni re-prefixes inner language keys, MiniMax M3 VL delegates the record, and backbone plus MTP merges preserve independent records.
- The newly added MiniMax accounting was completed for its default all-MoE MTP layout: backbone layer 0 and MTP depth 0 now use distinct temporary identities. Without this correction, values still loaded correctly but MTP rebuilt one expert block and missed the intended in-place optimization. Repository MiniMax recipes currently set `num_mtp_modules: 0`, so no shipped recipe was failing.
- Shape-only key discovery has `_extra_state` coverage, and Ling has an end-to-end safetensors reader test for fused QKV plus grouped experts. These tests pin the contracts used by the optimized route.

### Pre-existing or non-headline issues fixed here anyway

These were not needed to produce the reported Ling or Nemotron speedups. They were kept in this draft because review found them in code already being touched and each had a small, focused fix with regression coverage.

- **Grouped expert bias:** the shared adapter did not implement HF per-expert bias split/merge and could otherwise pass incompatible keys through. The low-memory capability is now false for `expert_bias=True`, and genuine split bias input fails loudly. Actual bias conversion remains #3688. Router correction bias was already supported and is unchanged.
- **Dense Nemotron MTP:** the existing MTP path could call MoE conversion even when `moe_config=None`. Dense MTP now bypasses that conversion, and the MTP path consistently forwards conversion options and reports native names. This is orthogonal to the benchmarked Nano V3 MoE recipe.
- **Minor adjacent cleanup:** MiniMax MTP now preserves unknown keys safely and shares its dequantize-and-rename preparation instead of duplicating it; the inaccurate `experts == "te"` constructor comment and stale onboarding guidance for the removed capability flag were corrected.

### Deliberately deferred

- Full grouped expert-bias split/merge support is tracked in #3688.
- Canonical shared MoE runtime capability facts and the remaining composite-adapter audit are tracked in #3689.
- Alternate checkpoint-namespace resolution before DCP destination construction and attempt-scoped cleanup after an aborted load remain tracked on the checkpoint umbrella #3576. These are pre-existing, fail-loud or recovery-path issues; the official benchmarked checkpoints are unaffected.

None of the pre-existing/minor fixes above changes the benchmarked Ling route or the Nemotron one-process fallback.

## Validation

- Review-follow-up checkpoint and affected adapter matrix at `d3680bc5`: **471 passed, 15 skipped**.
- MiniMax M3 MTP, state-dict adapter, and VLM delegation coverage at `ca42a1165`: **15 passed**, including a real `to_hf(for_checkpoint_load=True)` → `from_hf` lifecycle with the default all-MoE backbone layer 0 and MTP depth 0.
- Affected grouped-MoE adapter regression tests: **109 passed, 23 skipped**.
- Model infrastructure and Gemma4 adapter tests: **55 passed, 30 skipped** (CUDA-only cases skipped on the CPU test host).
- Ruff format/check and `git diff --check` pass.
- Nemotron cold-start job `16275130`: exact rebased SHA, 0.0 GiB initial page residency, expected fingerprints, no meta parameters, and 9.3 MiB GPU peak overhead.
- Ling matched A/B job `16275108`: exact rebased SHA, identical parameter/forward probes across parent and treatment, finite logits, and no meta parameters.
- Scoped CI on the behavior-equivalent benchmark head (`1933190d5`): [pipeline `63890795`](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/63890795) passed both [`nemotron_nano_v3_singlegpu_lora`](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/406759440) and the distributed [`nemotron_nano_v3_hellaswag`](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/406759439) checkpoint-robustness job. The restored AutoModel logits matched exactly (max KL `0.0`), and the consolidated checkpoint loaded by vanilla HF remained within tolerance (max KL `0.01525`, threshold `0.07`). Later review commits tighten runtime capability routing, adapter option forwarding, and load bookkeeping without changing the benchmarked Ling route or Nemotron one-process fallback. GitHub CI runs on the current head.

- DGX Spark validation at `d3a171a76`: [pipeline `65125114`](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65125114) / [job `416467831`](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/416467831) passed the checked-in 50-step `ling_mini_2_0_singlegpu_lora` recipe on one GB10 with no runtime overrides. The 30.28 GB checkpoint loaded successfully, all 50 steps had finite loss and gradient norms (about 30.9 GiB reported training memory), validation completed at loss `3.0625`, and checkpoint save succeeded. Load timing was 1098.04 s overall: destinations 0.08 s, storage read 1097.91 s, adaptation 0.03 s, and installation 0.02 s. Because the shared S3-FUSE storage path dominated this run, it is end-to-end compatibility and memory-fit evidence, not a speed comparison.

Closes #3664


yuhezhang-ai:
/ok to test [a6a396e](https://github.com/NVIDIA-NeMo/Automodel/pull/3610/commits/a6a396e28b6118fd2843531f58439a19dac287bb)

yuhezhang-ai:
/ok to test [c768d31](https://github.com/NVIDIA-NeMo/Automodel/pull/3610/commits/c768d3136183f175a37d2ceb6b77ac6dea5d7b53)

yuhezhang-ai:
/ok to test [ca42a11](https://github.com/NVIDIA-NeMo/Automodel/pull/3610/commits/ca42a11653e323afb8114c308d301ceb9ed3de3b)

yuhezhang-ai:
/ok to test [d3a171a](https://github.com/NVIDIA-NeMo/Automodel/pull/3610/commits/d3a171a7689143ec338b2982e902cbb9c7bede3f)

akoumpa (dismissed):

akoumpa (approved):

PR #3616, opened 2026-08-21
perf(checkpoint): load standard HF safetensors with DCP

yuhezhang-ai:
## What does this PR do?

Load standard Hugging Face safetensors directly into model-owned weight storage with DCP, avoiding a complete intermediate CPU checkpoint dictionary. The PR is based on `main` after #3610 merged.

## Changelog

- Use DCP for standard HF-compatible weights, including Falcon H1 and the native dense Llama, Qwen2, and Qwen3 implementations.
- Remove the three dense model adapters and the passthrough adapter class. Their weight names and tensor layouts already match HF. Most custom models still need their existing conversion adapters.
- Restore omitted shared-parameter aliases during HF initialization, preserving actual parameter sharing and checks for missing weights. This also safeguards encoder-decoder checkpoints such as T5/BART.
- Handle module `_extra_state` in shared checkpoint code: omit it from HF safetensors exports, preserve it in native DCP with correctly sized destinations, and retain current metadata when the source omits it.
- Base dequantization routing on the source checkpoint's declared quantization method. Stale dequantization settings on BF16 sources no longer force full CPU loading.
- Register native Qwen TP plans without an indirect distributed-to-checkpoint import.
- Keep onboarding and parity guidance concise: implement a state dict adapter for weight conversion; omit it only when HF names and layouts already match.

## Loading routes

| Base checkpoint | One GPU | Distributed |
| --- | --- | --- |
| HF-compatible safetensors without an adapter or dequantization | DCP | DCP |
| Adapter supporting model-backed destinations and bounded temporaries | DCP | DCP |
| Other custom adapter conversions | Full CPU | DCP with rank-local conversion |
| Standard HF checkpoint requiring dequantization | Full CPU | Full CPU |
| Legacy `.bin` or generic Transformers MoE expert-layout conversion | Full CPU | Full CPU |

Existing quantized-adapter fallback behavior is retained. The adapter removal targets three model implementations; shared checkpoint routing and metadata handling also serve other models.

## Validation

On the latest revision:

- **CPU:** 523 checkpoint, affected-model, and infrastructure tests passed; 12 skipped. Coverage includes real HF/native checkpoint round trips, tied/untied weights, logits, gradients, optimizer updates, LoRA, and variable-size module metadata.
[...]

PR #3619, opened 2026-08-21
perf(checkpoint): bound quantized DCP load memory

yuhezhang-ai:
## What this changes

This PR makes FP8 checkpoint conversion use a small part of the model at a time instead of keeping temporary tensors for the complete checkpoint alive at once.

Main already includes #3616, which chooses DCP according to whether dequantization is actually needed. Two expensive cases remain for Mistral3/Devstral FP8 checkpoints:

- **Distributed 123B model:** DCP reads rank-local FP8 tensors, but the adapter retains all temporary FP8 tensors until the complete read finishes. On TP=8, that peaks at 71.95 GiB of GPU memory per rank.
- **Single-GPU 24B VLM:** the safe fallback first builds and converts the complete checkpoint in CPU memory, then installs it into the model. This uses about 66.5 GiB of process memory in addition to the GPU model and makes loading much slower.

This PR adds an optional adapter hook for checkpoints that require dtype or layout conversion:

1. The adapter describes one self-contained part of the checkpoint.
2. DCP reads only that part.
3. Non-quantized tensors are written directly into model weights. FP8 tensors are converted directly into their final BF16 or FP32 model weights.
4. Temporary tensors are released before the next part is read.

Ordinary adapters do not implement the hook and keep their existing loading path.

The Mistral3 adapter groups at most eight text-decoder layers per part. It supports both the causal-LM layout and the untied full-VLM layout used by `mistralai/Devstral-Small-2-24B-Instruct-2512`. In the VLM layout, the vision encoder and multimodal projector load directly into their final model weights rather than becoming additional temporary copies.

FP32 master-weight configurations use the same bounded path. The model weights already have FP32 storage when loading begins, so each FP8 checkpoint part is converted directly into those FP32 destinations. The later FSDP mixed-precision policy can still use BF16 for compute.

## Safety boundaries

The new path is considered only when all of the following are true:

- this is the initial model load from a safetensors checkpoint;
- dequantization is actually required;
- one complete model state is being loaded, without a requested key subset; and
- the model adapter explicitly provides the load parts.

The Mistral3 implementation additionally requires a complete decoder, BF16 or FP32 final weights, and the supported per-tensor FP8 checkpoint format. Pipeline-parallel ranks that own only part of the decoder and tied VLM checkpoints keep their existing fallback. Unsupported adapters and dtypes are unchanged.

The shared loader rejects missing checkpoint tensors, repeated checkpoint requests, repeated model destinations, and incomplete model coverage instead of allowing a partial load.

## Historical measurements

These measurements predate the September 15 rebase. They compare the original implementation against its then-parent #3616; no new controlled before/after performance comparison on the rebased head is claimed. The fresh GB10 functional CI result is reported separately below.

### Devstral 123B, TP=8 + LoRA

These runs used the same official `mistralai/Devstral-2-123B-Instruct-2512` checkpoint, staged image, TP=8 + LoRA setup, and one-token forward pass.

| Code path | Slurm job | Model load | Peak allocated GPU memory/rank | Peak reserved GPU memory/rank |
|---|---:|---:|---:|---:|
| Parent #3616: retain all rank-local FP8 destinations | `16296055` | 90.08 s | 71.95 GiB | 73.35–73.36 GiB |
| Intermediate: one decoder layer per part | `16295655` | 101.50 s | 30.25 GiB | 30.37 GiB |
| This PR: up to eight decoder layers per part | `16296372` | 91.16 s | 30.44 GiB | 31.48 GiB |

The final path reduces peak allocated GPU memory by **41.51 GiB/rank (57.7%)**. Model-load time changes by **+1.08 seconds (1.2%)**, so the 123B result is a memory/OOM improvement with loading speed preserved, not a speed claim.

The final run used 12 DCP parts; its largest temporary part was 1.29 GiB/rank. It completed with 1,852 sharded parameters, trainable LoRA weights, and finite logits.

### Devstral Small 2 24B VLM, one H100

This is the full `Mistral3ForConditionalGeneration` VLM checkpoint, including its vision encoder and multimodal projector. At the time of this benchmark, the text-only SQuAD recipe used a multi-GPU CI node; the benchmark used one GPU to exercise the path that previously built the complete checkpoint in CPU memory.

Both measurements used the same checkpoint files after download. The comparison uses the fastest observed full-CPU run, making the speed comparison conservative.

| Code path | Slurm job | Model load | Peak CPU/process memory | Peak GPU memory | Peak CPU + GPU memory |
| Parent #3616: full checkpoint converted on CPU first | `16301212` | 56.57 s | 66.48 GiB | 44.73 GiB | 111.21 GiB |
| This PR: DCP load and conversion in six parts | `16301550` | **14.40 s** | **5.19 GiB** | 48.87 GiB | **54.06 GiB** |

For this real single-GPU VLM case, the new path is **3.9x faster**. Peak CPU/process memory falls by **61.29 GiB**, and peak CPU + GPU memory falls by **57.15 GiB (51.4%)**. GPU memory increases by 4.14 GiB because one bounded FP8 decoder group is temporarily present alongside the BF16 model; that allocation is released before the next group.

The loader read the 24.02 GB checkpoint in 12.65 seconds, including 12.55 seconds of storage reads and 0.02 seconds of FP8 conversion. The final one-token forward produced finite logits.

CPU/process and GPU memory were sampled together on a discrete-memory H100 system. Their sum is useful for estimating total loading pressure, but it is not a direct DGX Spark unified-memory measurement.

## Rebase validation — September 15, 2026

Rebased the three bounded-FP8 commits onto main `087f3d34d791ea68cf0b61c774b7037195958bf8`. The rebased implementation head was `7da03d831181c4f0ec0b5336eea50bf0e7b74835`, with seven implementation/test files. The subsequent GB10 CI configuration commit is described below.

- Preserved main's checkpoint and Mistral fixes. Resolved test import conflicts and updated the bounded loader to use main's current metadata helper.
- **437 focused CPU tests passed, no skips**, on this exact head (Slurm `18662592`). Coverage includes checkpoint loading, adapter capabilities, native-HF weight/gradient/export parity, shared parameters, the integrated-memory reader, Mistral causal/VLM models and FP8 adapters, and the existing GPT-OSS adapter.
- Python 3.12; PyTorch 2.10.0+cu130; Transformers 5.12.1.
- Repository Ruff format/check and `git diff --check` passed. All three commits retain DCO signoffs.
- Fresh single-GB10 scoped CI passed on the subsequent recipe-config head below. The historical multi-GPU performance/parity runs were not repeated after rebase; this PR remains draft.

## GB10 release regression coverage

Current head: `4cc19dbcdc76a862f603b2e94380be713a3dcb46`.

The existing Devstral 24B LoRA recipe now selects one GB10 process in release CI. The release override runs five optimizer steps at global/local batch size 1, saves a checkpoint and validates at step 5, and uses four validation samples. The recipe is not added to nightly CI. The normal training settings remain unchanged.

- Local generation confirms the recipe is selected by release CI and absent from nightly CI. Config resolution and Ruff/whitespace checks passed.
- [Scoped GB10 pipeline 68002362](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/68002362) builds only the ARM AutoModel image from this exact pushed commit and selects only `devstral2_small_2512_squad_peft` on `dlcluster_digits`.
- **Passed:** [GB10 job 440920076](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/440920076), with its artifact confirming the exact commit above, release scope, and one GB10. Runtime: PyTorch 2.14.0a0+4fdf77b940.nv26.08, CUDA 13.4, Transformers 5.15.1.
- The real FP8 checkpoint used the bounded path: six parts, with 4.14 GiB of temporary checkpoint destinations in the largest part. All five optimizer steps completed with finite losses and gradient norms; validation loss was 0.3138 and checkpoint saving completed in approximately seven seconds.
- **Loading performance remains unresolved:** the loader reported 1523.35 s total (25m23s), including 36.10 s for metadata and 1487.07 s in DCP load calls (97.6% of the total). The effective rate over those calls was about 16.5 MiB/s. The GB10 launcher uses an S3-FUSE checkpoint cache, which is a likely contributor, but the DCP timer also includes planning, CPU staging, and device copies. This run has no same-host parent baseline or local-cache comparison, so it does not establish the cause or rule out a loading regression. The reported 0.03 s finish-callback time is not a separately synchronized GPU conversion benchmark.
- Logged CUDA allocation peaks were 46.47–48.96 GiB. They exclude total system RAM and swap, which were not recorded; this result does not establish fit on the separate 60 GB machine.
- The GPU job took 35m21s, including 26m56s in the finetune launcher. Its preceding ARM image build took 48m56s, dominated by dependency compilation (37m16s preparing four source-built packages, with MagiAttention finishing last).
- This smoke covers initial loading, LoRA training, validation, and checkpoint saving. It does not assert reload parity or a peak unified-memory limit.
- Timeout caveat: the GB10 Docker launch path does not consume the recipe's `ci.time`; that value is only passed to sbatch on the other launch path. The 35m21s job therefore did not enforce a 30-minute recipe timeout. Enforcing a bounded Docker runtime is a shared CI launcher follow-up; changing this recipe value alone would not enforce it.

## Original GPU and conversion validation

- Real Devstral 123B TP=8 + LoRA load and forward: Slurm `16296372`
- Real Devstral Small 2 24B full-VLM load and forward on one H100: Slurm `16301550`
- Exact tiny TP=2 check: Slurm `16295582`; every reconstructed parameter matched, including 16 real DTensor shards, and the forward was finite
- Focused checkpoint, Mistral3 adapter, and VLM model unit suites: passed
- BF16 and FP32 destination conversion parity and grouped-load unit coverage: passed
- Ruff format/check and `git diff --check`: passed

Addresses the Devstral FP8 loading-memory case discussed in #2114.

### DIFF
--- a/nemo_automodel/components/checkpoint/_backports/hf_storage.py
+++ b/nemo_automodel/components/checkpoint/_backports/hf_storage.py
@@ -276,6 +276,14 @@ class _HuggingFaceStorageReader(FsspecReader):
             super().__init__(path=path)
 
         self.key_mapping = key_mapping
+        self._metadata_cache: Metadata | None = None
+
+    def reset(self, checkpoint_id: str | os.PathLike | None = None) -> None:
+        """Reset reader state and discard metadata only when the checkpoint changes."""
+        previous_path = str(self.path)
+        super().reset(checkpoint_id)
+        if checkpoint_id is not None and str(self.path) != previous_path:
+            self._metadata_cache = None
 
     def read_data(self, plan: LoadPlan, planner: LoadPlanner) -> Future[None]:
         per_file: dict[str, list[ReadItem]] = {}
@@ -372,6 +380,11 @@ class _HuggingFaceStorageReader(FsspecReader):
         return fut
 
     def read_metadata(self) -> Metadata:
+        if self._metadata_cache is not None:
+            if getattr(self._metadata_cache, "storage_meta", None) is not None:
+                self._metadata_cache.storage_meta.load_id = self.load_id
+            return self._metadata_cache
+
         state_dict_metadata: dict[str, TensorStorageMetadata] = {}
         storage_data: dict[MetadataIndex, _HFStorageInfo] = {}
 
@@ -443,6 +456,7 @@ class _HuggingFaceStorageReader(FsspecReader):
             metadata.storage_meta = StorageMeta()
         metadata.storage_meta.load_id = self.load_id  # type: ignore[union-attr]
 
+        self._metadata_cache = metadata
         return metadata
 
 
--- a/nemo_automodel/components/checkpoint/checkpointing.py
+++ b/nemo_automodel/components/checkpoint/checkpointing.py
@@ -1009,13 +1009,14 @@ class Checkpointer:
 
         part_loaded_model_state_dict: dict[str, torch.Tensor] | None = None
         checkpoint_load_parts: Iterator[CheckpointLoadPart] | None = None
+        # Adapter-owned parts name their DCP destinations with exact checkpoint keys, so any generic Transformers
+        # key_mapping is redundant for this path and must not prevent the adapter from describing bounded groups.
         if (
             is_init_step
             and is_safetensors
             and should_dequantize_base_checkpoint
             and isinstance(state_dict_adapter, StateDictAdapter)
             and len(model_state.model) == 1
-            and key_mapping is None
             and not allow_checkpoint_key_subset
         ):
             candidate_state_dict = model_state.state_dict()
--- a/nemo_automodel/components/checkpoint/checkpointing.py
+++ b/nemo_automodel/components/checkpoint/checkpointing.py
@@ -21,6 +21,7 @@ import pickle
 import threading
 import time
 import uuid
+from collections.abc import Iterator
 from dataclasses import dataclass
 from datetime import timedelta
 from pathlib import Path
@@ -70,7 +71,7 @@ from nemo_automodel.components.checkpoint.conversion_mapping import (
     requires_tensor_merging,
 )
 from nemo_automodel.components.checkpoint.lifecycle import CheckpointLifecycle
-from nemo_automodel.components.checkpoint.state_dict_adapter import StateDictAdapter
+from nemo_automodel.components.checkpoint.state_dict_adapter import CheckpointLoadPart, StateDictAdapter
 from nemo_automodel.components.checkpoint.stateful_wrappers import ModelState, OptimizerState
 from nemo_automodel.components.checkpoint.utils import (
     ensure_tied_lm_head,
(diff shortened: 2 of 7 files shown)


yuhezhang-ai:
/ok to test [7da03d8](https://github.com/NVIDIA-NeMo/Automodel/pull/3619/commits/7da03d831181c4f0ec0b5336eea50bf0e7b74835)

yuhezhang-ai:
/ok to test [4cc19db](https://github.com/NVIDIA-NeMo/Automodel/pull/3619/commits/4cc19dbcdc76a862f603b2e94380be713a3dcb46)

akoumpa (approved):

PR #3625, opened 2026-08-22
fix(utils): accumulate param L2 norm in float32 for bf16/fp16 models

ralovets:
# What does this PR do ?

Fixes the `Param L2 norm` logged by `print_trainable_parameters`. `_get_model_param_stats` accumulates `local_sq_norm += p.detach().norm(2) ** 2`, which takes the parameter dtype, so the value drifts for bf16 models and is `inf` for fp16 models — a real 3.62B-parameter bf16 checkpoint logs `3007.3191` against a true value of `3167.8195`, 5.1% low.

The fix passes an accumulation dtype to `norm()`, using `torch.promote_types(p.dtype, torch.float32)` rather than a literal `torch.float32` to avoid narrowing float64 parameters, which `linalg.vector_norm` rejects. Regression from [#1463](https://github.com/NVIDIA-NeMo/Automodel/pull/1463), which dropped the `.float()` upcast along with a per-parameter host sync. Metric only, no effect on training numerics.

## Performance

Benchmarked on GPU against the real 1120-tensor / 3.62B-param shape distribution, bf16, variants interleaved within each round:

| variant | implementation | median | min | p25 |
|---|---|---|---|---|
| before [#1463](https://github.com/NVIDIA-NeMo/Automodel/pull/1463) | `.float()` upcast copy, plus a `.item()` host sync per parameter | 27.59 ms | 27.37 ms | 27.52 ms |
| current main | `norm(2)`, accumulating in the parameter dtype — the bug | 11.65 ms | 11.57 ms | 11.64 ms |
| this PR | `norm(2, dtype=...)`, upcast folded into the reduction | 11.80 ms | 11.71 ms | 11.78 ms |

+1.3% (0.15 ms) against current main, from the per-parameter `promote_types` call. One-time cost, since all call sites run once after sharding.

# Changelog

- `_get_model_param_stats` accumulates the squared parameter norm in at least float32 instead of the parameter dtype, fixing the reported `Param L2 norm` for bf16 (drifts >1%) and fp16 (reports `inf`) models.
- Added unit tests asserting the reported norm is dtype-independent against a float64 reference.

**Pre checks**:


ralovets:
CI is green on `0b49831` and nothing has changed on my side. Anything you need from me? Thanks.

PR #3626, opened 2026-08-22
fix(moe): export fused expert lora in the corrected peft >= 0.19.1 layout

### LINKED ISSUE #3592: fused expert lora adapters can't load on peft >= 0.19 (ParamWrapper layout flipped)
when you train an expert lora on qwen3_moe, minimax_m2 or nemotron_v3, the exported adapter can't be loaded with any recent version of peft. loading fails with a size mismatch error.

what happened: peft 0.19.0 changed how it lays out lora weights for fused moe experts (huggingface/peft#3165, april 2026). our export code in `MoESplitExpertsStateDictMixin._convert_lora_to_paramwrapper` still writes the old 0.18 layout. the repo's own CI pins peft 0.18.1 so tests stay green, but `pyproject.toml` only requires `peft>=0.18.1`, and a plain `pip install peft` today gives 0.20.0. so people train a model here, take the adapter to their own environment, and it won't load.

easy way to see it: create a lora on a tiny `Qwen3MoeExperts` with `LoraConfig(target_parameters=["experts.gate_up_proj", "experts.down_proj"])` and look at the shapes peft creates (E=2, H=8, I=3, r=2):

```
                             peft 0.18.1     peft 0.20.0
base_layer.lora_A.weight     (4, 6)          (4, 8)
base_layer.lora_B.weight     (8, 4)          (6, 4)
lora_A.weight                (4, 8)          (4, 3)
lora_B.weight                (3, 4)          (8, 4)
```

our export produces the left column, so on new peft every tensor mismatches. and if a model happens to have hidden_size equal to 2x intermediate_size, the shapes match by coincidence and the adapter loads with silently wrong math, which is worse than the error.

two more things:

- this went out in the r0.6.0 release too (via the #3458 backport), not just main
- the peft pin can't simply be bumped: the round trip test in `tests/unit_tests/models/test_moe_peft_v5_state_dict_adapters.py` fails on newer peft because of this exact mismatch. so the export bug is also what's blocking the peft upgrade

my suggestion for a fix:

1. export in the new (>= 0.19) layout by default, since that's what people actually install, with a flag for old peft if needed (like `v4_compatible`)
2. on import, detect which layout the file uses instead of assuming, and give a clear error when it's ambiguous
3. raise the peft floor and run the round trip test against both layouts

this is the same version difference that came up in the #3284 review discussion. back then it looked like a question of which layout was correct. turns out both are, just for different peft versions, and that's the real problem.

happy to send a pr for this.

stanley1208:
Fixes NVIDIA-NeMo/Automodel#3592

* export now writes the corrected (peft >= 0.19.1) layout by default. the old layout is still available through a new `legacy_paramwrapper_layout` checkpoint option, separate from `v4_compatible`
* on save we stamp the layout into `automodel_peft_config.json`, and the loader reads that first. for older adapters without the stamp, the layout is detected from tensor shapes. one tensor is enough to decide for the whole file, which handles minimax_m2 where the gate_up shapes look the same in both layouts. if nothing settles it, the load fails with a clear error instead of guessing
* peft minimum bumped to 0.19.1, both lockfiles regenerated

tested loading, merged deltas and forward outputs through real peft with the corrected layout, plus unit tests for both layouts, the stamp, and the ambiguous case.


nemo_automodel/components/checkpoint/config.py
```diff
@@ -122,6 +122,9 @@ class CheckpointingConfig:
     staging_dir: str | None = None
     v4_compatible: bool = False  # If True, save the original pretrained config.json (with quantization_config removed)
     # instead of the in-memory v5 config.  Useful when downstream consumers (e.g. vLLM) expect a v4-format config.
+    legacy_paramwrapper_layout: bool = False  # If True, export fused expert LoRA (PEFT v5 ParamWrapper) weights in
```
yuhezhang-ai, line 125: Thanks for keeping legacy export separate from `v4_compatible`. Could we also thread this option through `Qwen3MoeStateDictAdapter.convert_single_tensor_to_hf`? That override currently calls `_convert_lora_to_paramwrapper(fqn, tensor)` without forwarding `legacy_layout=kwargs.get("legacy_paramwrapper_layout", False)`, so Qwen3 still emits the modern layout when this flag is true. Meanwhile the addon stamps the checkpoint as `peft-0.18`, which makes older PEFT fail and can make AutoModel reload interpret the tensors under the wrong layout. A focused Qwen3 legacy-export/load test would catch this.

nemo_automodel/components/moe/state_dict_mixin.py
```diff
@@ -347,73 +394,151 @@ def _convert_lora_to_paramwrapper(self, fqn: str, tensor: torch.Tensor) -> list[
...
+        if not matches:
+            return state_dict
+
+        # Metadata first: the save path stamps the layout into
+        # automodel_peft_config.json and the checkpoint loader hands it to us
+        # through this attribute; shape detection is the fallback for adapters
+        # saved before the stamp existed.
+        layout_hint = getattr(self, "_paramwrapper_layout_hint", None)
+        if layout_hint is not None:
+            swapped = layout_hint != _PARAMWRAPPER_LAYOUT_LEGACY
```
yuhezhang-ai, line 469: Could we validate the metadata value explicitly before selecting a layout? As written, every non-`None` value other than `peft-0.18`—including a typo, corrupted value, or future incompatible layout—is silently treated as modern, and a stamp that contradicts unambiguous shape votes wins without an error. Since the goal is fail-closed compatibility, please accept only the two known layout identifiers and raise on an unknown value or metadata/shape disagreement, with tests for both cases.

tests/unit_tests/models/test_moe_peft_v5_state_dict_adapters.py
```diff
@@ -124,18 +125,23 @@ def _make_adapter_and_state(family: str, rank: int):
...
         scale: LoRA scaling factor.
 
     Returns:
-        Tensor of shape [experts, input, output].
+        Tensor shaped like the wrapped expert parameter.
     """
     lora_a = lora_a.reshape(num_experts, -1, lora_a.shape[-1])
     lora_b = lora_b.reshape(lora_b.shape[0], -1, num_experts)
-    return torch.einsum("ore,eri->eio", lora_b, lora_a) * scale
+    equation = "ore,eri->eoi" if is_peft_min_version("0.19.1") else "ore,eri->eio"
```
yuhezhang-ai, line 143: Could we make this an independent AutoModel↔PEFT parity check and include Qwen3? This helper mirrors PEFT's implementation, the round trip exercises the same converter in both directions, and `adapted_output != base_output` only proves that the adapter has some effect. A consistent A/B transpose can therefore pass. Please compare the merged PEFT delta/output against the delta computed directly from the native grouped LoRA A/B tensors (or an AutoModel forward), and parameterize the real-PEFT test over Qwen3 as well. This would also exercise the family-specific override.

nemo_automodel/shared/import_utils.py
```diff
@@ -506,6 +506,32 @@ def is_transformers_min_version(version, check_equality=True):
     return get_transformers_version() > PkgVersion(version)
 
 
+def get_peft_version():
```
yuhezhang-ai, line 509: Could we keep this version check out of production code? These two new public helpers are only used by this test, add an untyped API surface, import optional PEFT outside `safe_import`, and `except Exception` hides unrelated import/runtime failures. Since the test dependency floor is now `peft>=0.19.1`, the test can assert the modern equation directly; if multi-version coverage is needed, it would be clearer as a test-local helper or a separate environment matrix.

yuhezhang-ai (changes requested):
Thanks for turning this issue around and for adding modern-by-default export together with legacy import compatibility. The PEFT >= 0.19.1 mapping and the dependency/lock updates look directionally right. Before merging, could we address the compatibility gaps in the inline comments? One additional documentation item: `docs/guides/llm/finetune.mdx` still tells users to install `peft>=0.18.1`; with the new default, that range includes incompatible consumers. Please raise it to `>=0.19.1` and mention `checkpoint.legacy_paramwrapper_layout: true` for users who must export for older PEFT. Once these points are addressed, the overall approach looks good. Thanks again for taking this on.

stanley1208:
@yuhezhang-ai thanks for the review! all five points are addressed in 2528363fe

yuhezhang-ai:
/ok to test [2528363](https://github.com/NVIDIA-NeMo/Automodel/pull/3626/commits/2528363fe3d10559cd9654bc0835467ce04c5c90)

yuhezhang-ai (approved):
Thanks for addressing all of the review feedback. I rechecked the latest head and verified the Qwen3 legacy-option plumbing, fail-closed metadata/shape handling, independent PEFT merge parity across Qwen3-MoE, MiniMax-M2, and Nemotron V3, removal of the test-only production helpers, and the dependency/docs updates. The focused layout and checkpoint tests pass locally, as do the three real-PEFT parity cases under PEFT 0.20.0. LGTM.

kajalj22 (approved):

PR #3605, opened 2026-08-21
feat(dflash): add DFlash 2 draft model, trainer, and recipe

kashif:
# What does this PR do ?

Adds DFlash 2 ([blog](https://inco.ai/blog/dflash2/)) to the speculative-decoding
recipes, next to the existing DFlash, Domino, and JetSpec ones, and makes the
Qwen3.5-family targets (`Qwen/Qwen3.8-27B` among them) trainable with it.

# Changelog

- `Qwen3DFlash2DraftModel` — the DFlash backbone plus the two pieces DFlash 2
  adds: a two-tap dynamic convolution around every sublayer (taps never cross a
  block boundary), and a pairwise path selector that walks one coherent path
  through each position's top-k candidates instead of keeping every top-1 pick.
  Both are initialised to the identity, so a fresh DFlash 2 draft is numerically
  equal to plain DFlash and training moves off from there.
- `DFlash2TrainerModule` — the DFlash block CE plus one candidate-selection CE
  term (`selector_loss_weight`). Positions whose true token missed the candidate
  list carry no selector signal and get reported as `candidate_recall`.
- `spec_generate` with path selection: greedy, and rejection sampling so sampled
  decoding stays lossless.
- `TrainDFlash2Recipe`, plus example configs for Qwen3 and Qwen3.8-27B.
- Qwen3.5-family targets: registry entries, and the decoder config is read from
  the nested `text_config` that a `*ForConditionalGeneration` target carries.
- Optional sliding-window draft attention (`draft_sliding_window`), enforced in
  both mask backends during training and on the decode path, and stamped into
  the saved config as `sliding_attention`. Also keeps the flex `BlockMask` sparse
  on long sequences. Off by default, so plain DFlash is unchanged.
- `draft_num_attention_heads` / `draft_num_key_value_heads` / `draft_head_dim`
  let the draft's attention be sized independently of the target's, which the
  published drafters do.
- Docs: a DFlash drafter guide (the family had none), plus the examples README
  and repository-structure entries.

# Before your PR is "Ready for review"

**Pre checks**:

- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [x] Did you add or update any necessary documentation?

# Additional Information

Building our draft class from the published `z-lab/Qwen3.8-27B-DFlash2` config
reproduces its checkpoint exactly — all 81 tensors, same names and shapes, no key
[...]

PR #3627, opened 2026-08-22
feat(dflash): support top-p and top-k in speculative decoding

kashif:
# What does this PR do ?

Closes the remaining behavioural gaps between our DFlash decode path and the
reference implementation (github.com/z-lab/dflash), found by diffing it symbol
by symbol after #3605 merged, and makes `top_k` expressible in the benchmark
harness.

# Changelog

- `top_p` / `top_k` in `spec_generate`. The truncation applies to the *target's*
  distribution -- the one rejection sampling verifies against -- as the reference
  does; the draft still proposes from plain temperature.
- `top_k` in the benchmark and eval harnesses (`bench_sglang`, `bench_vllm`,
  `bench_sweep`, `decode_eval`). They already had `temperature` and `top_p`, so
  the DFlash 2 evaluation setting (temperature 1.0, top-p 0.95, top-k 20) could
  not be expressed at all. The key is omitted from the request payload unless set:
  the engines spell "no top-k" as `-1`, so sending our `0` default would be read
  as "keep zero tokens".
- `output_multiplier` and `final_logit_softcapping` via a `compute_logits` on the
  draft, applied in **both** training and decoding so the draft is supervised on
  the same distribution it is later verified against. Identity for every
  Qwen3-family target; the published Muse Glimmer drafter ships both.
- `input_embedding_scale` on the embedded draft block.
- Output-head resolution falls back to `get_output_embeddings()` instead of
  hardcoding `target.lm_head`, which a `*ForConditionalGeneration` target lacks.
- Sampling arguments are validated instead of silently misbehaving.
- Stop-token detection is vectorised over the newly produced span rather than
  re-scanning the whole generated prefix each block.

**Pre checks**:

# Additional Information

`sampling_probs` is bit-exact against the reference across every
temperature/top-p/top-k combination tried, including the blog's evaluation
setting. The tests assert exact equality rather than a tolerance, because
rejection sampling compares target and draft probabilities directly and needs
filtered tokens at exactly zero for the residual to be correct.

735 tests pass across the DFlash/DSpark/bench suites. Two deliberate deviations
from the reference are documented in the code: we keep the additive mask
representation (ours dispatches to eager as well as SDPA, and eager *adds* the
mask), and we embed through the embedding module rather than indexing its weight
table, because only the module carries the vocab-parallel DTensor plan under TP.

Scope note: `spec_generate` is the in-repo reference decoder and has no
production caller today -- acceptance length is measured through the bench
harnesses against an SGLang/vLLM server. The `top_k` addition to those harnesses
is the part that affects reproducing published numbers; the `spec_generate`
changes bring the reference decoder back in line with upstream.


nemo_automodel/components/speculative/regenerate.py
```diff
@@ -111,6 +114,7 @@ def _build_manifest(args: argparse.Namespace) -> dict[str, Any]:
         "max_new_tokens": args.max_new_tokens,
         "temperature": args.temperature,
         "top_p": args.top_p,
+        "top_k": getattr(args, "top_k", 0),
```
khazic: This breaks `--resume` for every run started before this PR: `_ensure_manifest_compatible` compares manifests with strict `!=`, and old manifests have no `top_k` key, so resuming an existing output dir now raises even with identical settings. Either default the missing key before comparing (`existing_manifest.setdefault("top_k", 0)`) or leave it out of the manifest until regenerate can actually set it.

nemo_automodel/components/speculative/regenerate.py
```diff
@@ -86,6 +86,9 @@ class GenerationConfig:
     max_new_tokens: int
     temperature: float
     top_p: float
+    # 0 disables the truncation. Not sent to the server unless set: the engines
+    # spell "no top-k" as -1, so a literal 0 would be read as "keep 0 tokens".
+    top_k: int = 0
```
khazic, line 91: The field is needed for the bench harnesses that share this dataclass, but regenerate itself never plumbs it: there is no `--top-k` flag, `gen_cfg` is built without it, and `_regenerate_one`'s payload does not send it (unlike `bench_common._run_workload`). So the manifest records a constant 0. Either plumb it through here too, or drop it from `_build_manifest`.

nemo_automodel/components/speculative/dflash/draft_qwen3.py
```diff
@@ -333,6 +427,14 @@ def __init__(self, config) -> None:
                 "DFlash draft config carries no block_size, neither in dflash_config nor at the top level."
             )
         self.mask_token_id = dflash_config.get("mask_token_id", None)
+        # Output/input transforms some targets apply around the shared embedding
+        # and head; identity for every Qwen3-family target (see compute_logits).
+        self.output_multiplier = float(dflash_config.get("output_multiplier", 1.0))
```
khazic: Softcapping is validated to be > 0 below, but the multiplier is not; a non-positive `output_multiplier` would silently flip (or zero out) the greedy argmax. Worth the same guard.

HuiyingLi (dismissed):

docs/guides/speculative/dflash.mdx
jgerh, line 38:
```suggestion
DFlash 2 keeps drafting in one pass. The convolution and the selector add
parameters inside the stack rather than an autoregressive correction, so it has
```

jgerh, line 47:
```suggestion
`w_k = exp(-(k-1)/loss_decay_gamma)`, so early positions (the ones a verifier
reaches first) dominate the loss.
```

jgerh, line 79:
```suggestion
must fill block slots with the same id.

</Note>
```

jgerh, line 81:
```suggestion
## Run It
```

jgerh, line 91:
```suggestion
Swap the module and config to train a variant. For example,
```

jgerh, line 108:
```suggestion
## Key Config Fields
```

jgerh, line 112:
```suggestion
| `target_model_name_or_path` | frozen target (e.g., `Qwen/Qwen3-8B`) |
```

docs/guides/speculative/dflash.mdx
```diff
@@ -121,9 +121,17 @@ and `val_accept_len`; adding a top-level `wandb:` block uploads all of them.
...
 | `draft_sliding_window` | bounds how far back a block reads the context; unset attends over the whole prefix |
 | `draft_num_attention_heads` / `draft_num_key_value_heads` / `draft_head_dim` | size the draft's attention independently of the target's |
+| `output_multiplier` / `final_logit_softcapping` | the target's logit transform, applied in both training and decoding; identity unless set |
+| `input_embedding_scale` | scales the embedded draft block; identity unless set |
 | `conv_kernel_size` / `conv_group_size` | DFlash 2 convolution taps and channels per dynamic correction |
 | `selector_rank` / `selector_top_k` / `selector_loss_weight` | DFlash 2 path-selector width, candidates scored per position, and objective weight |
 
+`spec_generate` accepts `temperature`, `top_p`, and `top_k`. The truncation
+applies to the *target's* distribution -- the one rejection sampling verifies
+against -- so sampled decoding stays lossless with respect to whatever sampling
```
jgerh, line 131:
```suggestion
applies to the *target's* distribution (the one rejection sampling verifies
against), so sampled decoding stays lossless with respect to whatever sampling
```

examples/speculative/README.md
jgerh, line 9:
```suggestion
NeMo AutoModel trains the draft. You then serve the draft and target together in an
```

jgerh, line 115:
```suggestion
Run DFlash **DLLM SFT** configs under `../dllm_sft/` with the standard NeMo AutoModel
```

jgerh, line 205:
```suggestion
of a trained draft to warm-start the base weights. Adapters over a randomly
```

jgerh, line 207:
```suggestion
`peft:` for full-FT continued training. With a compressed draft vocab the
```

jgerh, line 521:
```suggestion
**DLLM SFT** configs under `../dllm_sft/` use the standard NeMo AutoModel SFT schema
```

jgerh, line 549:
```suggestion
`val_candidate_recall` (how often the true token is in the top-k candidate
list, the ceiling the selector can reach). The reductions sum raw token and block
```

jgerh (commented):
Completed tech pubs review of markdown files and provided a few copyedits.

khazic:
> /ok to test [e9e5709](https://github.com/NVIDIA-NeMo/Automodel/commit/e9e5709be487992c2be33004973f4321b1808959)

Haha bro, /test does absolutely nothing. You just typed it into the void.

khazic:
> /ok 测试[e9e5709](https://github.com/NVIDIA-NeMo/Automodel/commit/e9e5709be487992c2be33004973f4321b1808959)

Alright, I'll take a look at this part of the code.

khazic:
> /ok 测试[e9e5709](https://github.com/NVIDIA-NeMo/Automodel/commit/e9e5709be487992c2be33004973f4321b1808959)

kashif, why you can say chinese? 测试 is a chinese word

kashif:
opps my bad!

nemo_automodel/components/speculative/dflash/core.py
```diff
@@ -548,7 +548,9 @@ def forward(
         )
         # A tensor-parallel target's lm_head is column-parallel and returns
         # vocab-sharded (DTensor) logits; gather to a full tensor for the loss.
-        logits = _to_full_tensor(self.lm_head(output_hidden))
+        # ``compute_logits`` applies the target's output transform (multiplier /
```
khazic, line 564: `input_embedding_scale` is applied on the decode side (`embed_noise_block`) but not on the training side: `_create_noise_embed` and `_create_vp_noise_embed` still embed the noise blocks with a bare `self.embed_tokens(noise_ids)`. A target that sets the scale trains the draft on unscaled noise embeddings and serves it scaled ones, which is the same train/serve mismatch `compute_logits` closes for the logit transform, just on the input side. Should both multiply by `draft_model.input_embedding_scale`?

nemo_automodel/components/speculative/dflash/draft_qwen3_dflash2.py
```diff
@@ -622,12 +636,12 @@ def spec_generate(
                 output_hidden_states=True,
             )
             if temperature > 0:
-                target_probs = torch.softmax(output.logits.float() / temperature, dim=-1)
+                target_probs = sampling_probs(output.logits, temperature, top_p, top_k)
```
khazic, line 645: Tiny edge case, your call whether it is worth handling: `sample()` treats `temperature < 1e-5` as greedy, but this branch gates on `temperature > 0`, so a value like `1e-6` reaches `sampling_probs`, where `logits / temperature` overflows to inf and the softmax can yield NaNs in the rejection compare. The v1 decoder decodes the same value greedily. Gating on `>= 1e-5` would make the two decoders consistent, but the input is exotic and the reference presumably behaves the same, so leaving it as is seems fine too. What do you think?

nemo_automodel/components/speculative/dflash/core.py
kashif: fixed in d759c3a39: both now scale through input_embedding_scale.

nemo_automodel/components/speculative/dflash/draft_qwen3_dflash2.py
kashif: good catch, worth handling: fixed in 6eba5d1be, both the selector walk and the verify branch now gate on the same GREEDY_TEMPERATURE_EPS sample() uses.

nemo_automodel/components/speculative/dflash/core.py
```diff
@@ -548,7 +561,9 @@ def forward(
...
+        # softcapping) so training supervises the distribution serving produces.
+        logits = _to_full_tensor(self.draft_model.compute_logits(output_hidden, self.lm_head))
```
khazic, line 566: The same transform is still missing on the two other trainer modules that `TrainDFlashRecipe` subclasses select: `JetSpecTrainerModule.forward` (jetspec_core.py, `logits = _to_full_tensor(self.lm_head(output_hidden))`) and `DominoTrainerModule.forward` (domino_core.py, `base_logits = _to_full_tensor(self.lm_head(output_hidden))`). Both recipes inherit `_build_dflash_config`, so `output_multiplier` / `final_logit_softcapping` get stamped into the saved draft config and `spec_generate` applies them at decode, while training on those paths supervises raw `lm_head` logits. That is exactly the train/serve gap this change closes for DFlash and DFlash2, so those two sites should go through `compute_logits` as well (for Domino the low-rank correction is then fit on top of the transformed base logits).

HuiyingLi (approved):

HuiyingLi (approved):

PR #3632, opened 2026-08-22
fix(kimi_k25): map lm_head to model.language_model.lm_head

akx:
# What does this PR do ?

Fixes the Kimi-K2.5 state-dict adapter mapping `language_model.lm_head.*` to a top-level `lm_head.*` key that no parameter has, which left the output head at random init on every load through `NeMoAutoModelForCausalLM` / `KimiK25VLForConditionalGeneration`.

# Changelog

- `kimi_k25_vl/state_dict_adapter.py`: `from_hf` maps `language_model.lm_head.*` to `model.language_model.lm_head.*` (the model's actual FQN), using the same `"model." + key rule the adapter already applies to the other `language_model.*` keys.

Background: loading `moonshotai/Kimi-K2.5` logged `Checkpoint key mismatch ... missing=['model.language_model.lm_head.weight'] unexpected=['lm_head.weight']` and continued, because the base-model load uses `set_model_state_dict(strict=False)`. Training then runs against an uninitialised head with a plausible-looking loss curve. With this fix the warning is gone and step-0 SFT loss on chat data is 2.18 (a random head gives ~ln(163840) ≈ 12).

**Pre checks**:

  - updated the test that asserted the buggy key; added a round-trip test
  - none needed — no user-facing option changes

# Additional Information

- This seems to have been a thing ever since Kimi K2.5 support was introduced in #1132. Cc @HuiyingLi as the author of that PR.

### DIFF
--- a/nemo_automodel/components/models/kimi_k25_vl/state_dict_adapter.py
+++ b/nemo_automodel/components/models/kimi_k25_vl/state_dict_adapter.py
@@ -381,8 +381,9 @@ class KimiK25VLStateDictAdapter(MoESplitExpertsStateDictMixin, StateDictAdapter)
                 llm_key = key.replace("language_model.model.", "model.")
                 llm_keys[llm_key] = value
             elif key.startswith("language_model.lm_head."):
-                native_key = key.replace("language_model.lm_head.", "lm_head.")
-                native_state_dict[native_key] = value
+                # The VL model's head lives at model.language_model.lm_head.*; a top-level
+                # "lm_head." key matches no parameter and strict=False leaves the head at init.
+                native_state_dict["model." + key] = value
             elif key.startswith("mm_projector."):
                 # Map HF mm_projector keys to our multi_modal_projector structure
                 # mm_projector.proj.0.* -> multi_modal_projector.linear_1.*
--- a/tests/unit_tests/models/kimi_k25_vl/test_state_dict_adapter.py
+++ b/tests/unit_tests/models/kimi_k25_vl/test_state_dict_adapter.py
@@ -303,7 +303,16 @@ class TestKimiK25VLStateDictAdapterFromHF:
 
         result = adapter.from_hf(hf_state_dict)
 
-        assert "lm_head.weight" in result
+        assert "model.language_model.lm_head.weight" in result
+        assert "lm_head.weight" not in result
+
+    def test_lm_head_round_trip(self, adapter):
+        """to_hf(from_hf(x)) returns the HF lm_head key unchanged."""
+        hf_state_dict = {"language_model.lm_head.weight": torch.randn(2, 3)}
+        native = adapter.from_hf(hf_state_dict)
+        back = adapter.to_hf(native)
+        assert "language_model.lm_head.weight" in back
+        assert torch.equal(back["language_model.lm_head.weight"].float(), native["model.language_model.lm_head.weight"].float())
 
     def test_from_hf_dtype_conversion(self, adapter):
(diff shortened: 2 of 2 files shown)


HuiyingLi:
I think the FQN correction itself is valid, but the current root-cause statement ("the head was left at random init on every load since #1132") does not appear to be accurate.

I tested this with the actual Kimi-K2.5 checkpoint:

- On the PR base with 8xH100 / EP8, every local `lm_head` shard is byte-for-byte equal to the corresponding checkpoint rows, despite the missing/unexpected-key warning.
- The PR base and this branch also produce identical loss with the same model/input.

The reason is that the multi-rank DCP path builds its HF state dict from tensors that alias the final model storage. DCP writes the checkpoint into the real `lm_head` before `from_hf()` returns the incorrect `lm_head.weight` FQN. The subsequent warning is real, but in this path it does not mean the parameter remained at initialization.

I can reproduce an actual regression at `875637f3` / #3574, but only for the `world_size == 1` non-write-through custom-model path:

- Parent `7a6483ba`: loaded head exactly matches the checkpoint.
- `875637f3`: loaded head does not match the checkpoint.
- This PR: loaded head matches again.

That commit routes the single-rank case through the full-state path, where checkpoint tensors no longer alias model storage and the incorrect FQN is silently ignored by `strict=False`.

Could you clarify the topology used for the reported step-0 loss and whether 2.18 was true same-input base/head A/B? If it was a normal multi-rank run, this mapping change should not affect the loaded head or loss.

I suggest reframing this as fixing the Kimi adapter contract and the single-rank full-state regression introduced by #3574. It would also be good to add a regression test through the single-device `Checkpointer` full-state path; the current mapping/round-trip unit tests do not exercise the path that actually fails.

akx:
> Could you clarify the topology used for the reported step-0 loss and whether 2.18 was true same-input base/head A/B? If it was a normal multi-rank run, this mapping change should not affect the loaded head or loss.

Unfortunately I don't currently have access to exact logs anymore, as they were on pre-emptible machines for a PoC training... 😅 The topology was, anyway, 2 x 8xB200, ep_size=16.

I'm also not sure we ever actually hit "~ln(163840) ≈ 12" as truly reported loss -- there were other hurdles to get training to actually step (#3631 having been the critical one), but this patch got rid of the mapping warning, even if it might have been cosmetic.

HuiyingLi:
> > Could you clarify the topology used for the reported step-0 loss and whether 2.18 was true same-input base/head A/B? If it was a normal multi-rank run, this mapping change should not affect the loaded head or loss.
> 
> Unfortunately I don't currently have access to exact logs anymore, as they were on pre-emptible machines for a PoC training... 😅 The topology was, anyway, 2 x 8xB200, ep_size=16.
> 
> I'm also not sure we ever actually hit "~ln(163840) ≈ 12" as truly reported loss -- there were other hurdles to get training to actually step (#3631 having been the critical one), but this patch got rid of the mapping warning, even if it might have been cosmetic.

Thanks @akx ! This patch is good! I just merged #3631 . Were you able to run K25 lora e2e? If not I can help with a ootb recipe on gb200.

akx:
> Were you able to run K25 lora e2e? If not I can help with a ootb recipe on gb200.

Yes, eventually! 😅 
```
-rw------- 1 akx akx 5474589848 Aug 22 14:08 adapter_model.safetensors
```

But I couldn't test it in time, because vLLM doesn't support LORA for K25.

HuiyingLi (approved):

PR #3633, opened 2026-08-22
fix(kimi_k25): honour the quantization flag

akx:
# What does this PR do ?

Removes a hardcoded `quantization = True` in the Kimi-K2.5 state-dict adapter that ignored the checkpointer's flag, and fixes the `dequantize_base_checkpoint` auto-detection it was compensating for (composite models keep `quantization_config` on `text_config`).

# Changelog

- `kimi_k25_vl/state_dict_adapter.py`: drop the override; `convert_single_tensor_to_hf`
  now honours `quantization=` like the other adapters (Kimi-K3 is the in-tree reference).
- `_transformers/infrastructure.py`: `config_has_quantization()` checks the top-level config and its sub-configs (`sub_configs`, `text_config`, `language_config`, `llm_config`)

**Pre checks**:

  - Yep! 
  - none needed

# Additional Information

- Found while loading a BF16-converted K2.5 (`Missing key ... weight_packed`); reproduced on 8e6f854 and d80038384.
- The "try to infer whether the base weights are quantized" fix from #1181 never applied to K2.5 because of the hard-coded `quantization = True` line that has been there ever since K2.5 support was introduced in #1132. cc @HuiyingLi :)


HuiyingLi:
Removing the hardcoded `quantization = True` makes sense. However, Automodel owns `KimiK25VLConfig`, and it is missing the upstream behavior that promotes `text_config.quantization_config` to the top level:

```python
if getattr(self.text_config, "quantization_config", None) is not None:
    self.quantization_config = self.text_config.quantization_config
```

Can we fix the Kimi config instead of changing common infrastructure?

Also, BF16 export must remove the nested `text_config.quantization_config`; otherwise reload will still treat the saved BF16 weights as INT4. Please also test that the exported BF16 checkpoint can be loaded successfully.

akx:
> Can we fix the Kimi config instead of changing common infrastructure?

I'm all ears for a better fix! 

> Please also test that the exported BF16 checkpoint can be loaded successfully.

Unfortunately I don't currently have access to infrastructure to test this with right now (this was found as a side-effect of LoRA training on pre-emptible 2x 8xB200 nodes). I also found out the hard way that vLLM doesn't support LoRAs on Kimi K2.5 yet. 😅

HuiyingLi:
What about fixing it entirely in `KimiK25VLConfig` like this:

```diff
 elif isinstance(text_config, dict):
+    text_config = text_config.copy()
+    quantization_config = text_config.pop("quantization_config", None)
+    if quantization_config is not None:
+        kwargs.setdefault("quantization_config", quantization_config)
     text_config = DeepseekV3Config(**text_config)
```

akx:
@HuiyingLi To be honest, I think you know better if that's the right place to patch this 😄 I don't have the hardware access right now to give that a try, sorry!

HuiyingLi (approved):

PR #3643, opened 2026-08-24
fix(minimax): repair HF parity reference and gate at the measured envelope

yuhezhang-ai:
## What

Closes the MiniMax M2.7 checkpoint-robustness failures (AMINT-286): the catastrophic cross-framework KL (13.1) was an invalid reference, the resume drift matches the shared routed-MoE envelope, and the residual ~0.09 KL is knife-edge-router amplification of bf16 kernel noise — proven not to be an implementation defect — now gated at its measured envelope.

## Changelog

- **Reference repair (root cause of the 13.1):** Transformers 5.12's in-tree `MiniMaxM2Config` drops the legacy `rotary_dim` field, so the vanilla reference silently ran full-head RoPE instead of partial (64/128). The harness now derives `rope_parameters["partial_rotary_factor"] = rotary_dim / head_dim` for legacy configs (13.1 → 0.086). Upstreamed as huggingface/transformers#48241 / #48242.
- **Checkpoint-faithful fp32 router:** the released checkpoint stores `gate.weight` and `e_score_correction_bias` in fp32; AutoModel now scores sigmoid routing in fp32, allocates fp32-contract gate weights in fp32 at construction (keeps FSDP dtype groups uniform), and the harness pins the HF reference's fp32 tensors across vanilla-layout differences. Gate metrics are unchanged by this (the residual was never router precision), but the model now matches the checkpoint contract.
- **Recipe:** shared relaxed resume envelope (restored state and first forward are exact; 0.012–0.020 loss drift at continuation steps 6–7 across three runs), and measured `parity_threshold_overrides` for the two cross-framework gates on the full 2048-token document (mean_kl 0.15 / p95_kl 0.5 / cos 0.95 vs measured 0.086–0.094 / 0.346–0.375 / 0.964–0.965 across scoped CI 64182587, 64344786, 64351027). Overriding all three metrics numerically shadows the profile, so no relaxed-profile selection is used.

## Why the envelope is a closure, not a workaround

Full-weight local investigation on the real 230 GB checkpoint (62-layer streaming sweep, single GPU):

- The in-tree HF reference **against itself** (eager vs sdpa attention) reaches mean KL 0.237 / cos 0.973 at full depth through the same expert-flip cascade (~90 % of tokens re-routed by layer 59) that drives AutoModel's 0.320 / 0.962 in the same protocol — the reference's own kernel-choice variance is the same order as the AutoModel gap.
- Per-layer matched-input deltas are at bf16 ulp scale in every arm; TE and SDPA attention are bit-identical at layer 0; an end-to-end fp32 router leaves the numbers unchanged; HF-rounding-order RMSNorms make the first layernorm bitwise equal and cut 1-layer KL 6×, yet the effect is fully erased by depth 62.
- Router evidence (pipeline 64182587): 68 % of tokens flip an expert within five layers, all flips inside the measured score-noise band, replacements symmetric, no token unflipped across 62 layers.

Real conversion or model-math regressions stay loud: the rope bug measured mean KL 13.1, two orders above the envelope.

## Validation

- **Scoped CI green end to end** (nemo-ci pipeline 64481928, exact job `minimax_m2.7_hellaswag_lora`): all six phases pass — `source_load` 0.0860 / 0.3448 / 0.9644 and `hf_export_reload` 0.0869 / 0.3524 / 0.9646 against the 0.15 / 0.5 / 0.95 envelope; every self-repeat and the AutoModel reload bitwise-exact; resume within the shared relaxed envelope.
- Unit tests: 218 passed, 3 skipped (harness kwargs, MiniMax router precision, MoE layers).

Follow-up (stacked on #3635 once it lands): re-enable `capture_router_diagnostics` / `shape_diagnostic` for this recipe plus the MiniMax router-capture extension — that machinery lives in #3635.

Note: this PR and #3659 add functions in adjacent regions of the same harness file; whichever merges second needs a trivial context-only rebase.


yuhezhang-ai:
Superseded: GitHub closed this PR when its merged base branch was deleted and the stacking metadata prevents retargeting. Replacement (same head branch, rebased onto main): see the new PR linked below.

yuhezhang-ai:
Replacement: https://github.com/NVIDIA-NeMo/Automodel/pull/3674

PR #3661, opened 2026-08-25
feat(recipes): add model-ready hooks

pstjohn:
The finetune recipes currently have no safe extension point for selecting model trainability after PEFT/freezing,but before distributed wrapping and optimizer construction. 

Running such a callback after model construction is too late: DDP may already have captured its reducer parameters, so a newly enabled parameter can be optimized independently on each rank.

For instance, if we're doing LlaVa-style training where we want to train both 1. our input projector and 2. the LoRA parameters, the current sequence is:

```
Build model
→ apply PEFT, which freezes non-LoRA parameters
→ construct DDP/FSDP
→ a monkeypatch to re-enable the projector (which i'd like to remove)
→ construct optimizer
```

This PR adds `model_ready_hooks`, `_target_`-callables accepted by the LLM and VLM finetune setup paths:

```yaml
model_ready_hooks:
  - _target_: your_package.training.configure_model
    trainable_pattern: "extension.*"
```

Each hook receives the unwrapped model and runs once, in list order, after PEFT and configured freezing but before DDP/FSDP/TP/PP wrapping and optimizer construction. In the PEFT path, the hook-selected requires_grad policy is preserved across sharding and checkpoint materialization.

With meta-device initialization or post-shard loading, pretrained values are not available at hook time. Hooks may update trainability and non-tensor metadata on existing modules, but must not register parameters or inspect or restore pretrained tensor data.

The next-token and VLM knowledge-distillation recipes, diffusion-LM SFT recipes, and next-token benchmark recipe inherit the hook through their parent setup. Multi-model composite targets are rejected until target/draft/teacher ownership is defined; unrelated recipe families are not changed. 

The hook path is a no-op when not configured.

For the example above, the flow would now be

```
Build model
→ apply PEFT/freezing
→ model-ready hook selects final trainability
→ construct DDP/FSDP
→ construct optimizer
```

### DIFF
--- a/docs/guides/configuration.mdx
+++ b/docs/guides/configuration.mdx
@@ -59,6 +59,26 @@ Only load configuration files and `_target_` values from trusted sources. A YAML
 
 </Warning>
 
+## Run Hooks Before Distributed Wrapping
+
+The LLM and VLM fine-tuning recipes accept a top-level `model_ready_hooks` list for out-of-tree model setup that must happen before optimizer construction:
+
+```yaml
+model_ready_hooks:
+  - _target_: your_package.training.configure_model
+    trainable_pattern: "extension.*"
+```
+
+Each `_target_` is called once, in list order, with the unwrapped model as its first positional argument and the remaining fields as keyword arguments. For the preceding example, the callable signature is `configure_model(model, trainable_pattern: str)`. The return value is ignored.
+
+Hooks run after configured PEFT and freezing, but before tensor, pipeline, or data-parallel wrapping and before optimizer construction. Use them to configure existing model-owned extensions or select the final trainable parameter set. Running before distributed wrapping ensures DDP and FSDP include the same trainable parameters as the optimizer. An exception aborts recipe setup. Omitting `model_ready_hooks` leaves setup unchanged.
+
+<Note>
+With meta-device initialization or post-shard checkpoint loading, pretrained tensor values are not available at hook time. Hooks may update trainability and non-tensor metadata on existing modules, but must not register new parameters or inspect or restore pretrained tensor data.
+</Note>
+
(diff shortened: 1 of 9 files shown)


docs/guides/configuration.mdx
```diff
@@ -59,6 +59,26 @@ Only load configuration files and `_target_` values from trusted sources. A YAML
...
+    trainable_pattern: "extension.*"
+```
+
+Each `_target_` is called once, in list order, with the unwrapped model as its first positional argument and the remaining fields as keyword arguments. For the preceding example, the callable signature is `configure_model(model, trainable_pattern: str)`. The return value is ignored.
+
+Hooks run after configured PEFT and freezing, but before tensor, pipeline, or data-parallel wrapping and before optimizer construction. Use them to configure existing model-owned extensions or select the final trainable parameter set. Running before distributed wrapping ensures DDP and FSDP include the same trainable parameters as the optimizer. An exception aborts recipe setup. Omitting `model_ready_hooks` leaves setup unchanged.
+
+<Note>
+With meta-device initialization or post-shard checkpoint loading, pretrained tensor values are not available at hook time. Hooks may update trainability and non-tensor metadata on existing modules, but must not register new parameters or inspect or restore pretrained tensor data.
+</Note>
```
jgerh:
```suggestion
With meta-device initialization or post-shard checkpoint loading, pretrained tensor values are not available at hook time. Hooks can update trainability and non-tensor metadata on existing modules, but must not register new parameters or inspect or restore pretrained tensor data.

</Note>
```

jgerh (commented):
Completed tech pubs review of docs/guides/configuration.mdx and provided one copyedit. Otherwise, LGTM.

pstjohn:
closed in favor of #3681

PR #3672, opened 2026-08-25
fix(qwen2_5_omni): put the thinker prefix inside the peft prefix on lora saves

### LINKED ISSUE #3655: qwen2_5_omni peft save puts the thinker. prefix outside base_model.model.
same bug as #3595, just in the qwen2_5_omni adapter.

lora training works, but the saved adapter file has broken key names. the adapter puts `thinker.` in front of every key, even the peft ones that already start with `base_model.model.`. so the file ends up with

```
thinker.base_model.model.model.layers.0.self_attn.q_proj.lora_A.weight
```

but hf peft needs

```
base_model.model.thinker.model.layers.0.self_attn.q_proj.lora_A.weight
```

so the adapter can't be loaded with hf peft. it also breaks the other direction: an external adapter with the correct key names passes through `from_hf` unchanged, so it doesn't load either.

resuming inside automodel still works, only the exported file is affected.

this happens in `components/models/qwen2_5_omni/state_dict_adapter.py`, in `to_hf` and `convert_single_tensor_to_hf`. the same fix as #3630 should work here, and simpler since this model has no moe experts.

stanley1208:
Fixes #3655. same bug #3630 fixes for qwen3_omni_moe, just in this adapter.

when you train a lora and save it, every weight gets a name. this adapter added `thinker.` to the start of every name, but peft names must start with `base_model.model.`. with `thinker.` in front, hf peft doesn't recognize any of the names, so the saved adapter can't be used. loading adapters made outside automodel failed for the same reason.

what changed:

- `thinker.` now goes in the right spot, after `base_model.model.` instead of before it. loading understands both the new and the old naming, so existing saves still work
- weights belonging to the talker (the part of the model we don't train) are now filtered out no matter which naming they use
- the adapter config file now uses the same names as the weights, so peft connects the adapter to the right parts of the model

added cpu tests for all of the above.


yuhezhang-ai (approved):
Reviewed the prefix conversion, metadata target mapping, and legacy loading paths; no blocking findings.

Validation: all 17 focused adapter tests passed. An additional tiny-model CPU check using Transformers 5.15.1 and PEFT 0.19.1 exercised native Checkpointer.save_model and real PeftModel.from_pretrained: all eight adapter tensors loaded exactly, bulk and single-tensor exports agreed, and no adapters were injected into talker/token2wav. Reloaded text logits matched exactly; merged logits differed by at most 5.96e-8. Checkpointer.load_model restored both current and legacy malformed checkpoint names exactly. CI is green on this revision.

This is a suitable incremental fix toward #3867. Please carry the real Qwen2.5 Omni consumer regression into that shared suite as a follow-up.

PR #3659, opened 2026-08-25
fix(test): unblock the Kimi-Linear vanilla-HF parity reference

yuhezhang-ai:
## What

Kimi-Linear's checkpoint remote code targets a pre-v5 Transformers API, so the checkpoint-robustness harness could never load its vanilla-HF reference under the CI stack (Transformers 5.12): phases 0 and 3 crashed before producing a single parity number (AMINT-288). This PR makes the reference loadable through four independent compatibility fixes, none of which changes model math.

Replaces the stacked #3644 (closed): the Kimi work is independent of the MiniMax parity work, so it now sits directly on #3567's branch.

## Changelog

- `apply_cache_compatibility_patches` aliases `OutputRecorder` back into `transformers.utils.generic` for pre-v5 remote code (library-side shim, benefits any stale remote-code model).
- Harness re-resolves vanilla-reference configs from the checkpoint's own `auto_map` when AutoModel's `CONFIG_MAPPING` registrations hijack `AutoConfig` (preserves a load-time FP8 `dequantize` request).
- Harness renames the legacy `input_embeds` kwarg in remote `create_causal_mask` calls (renamed to `inputs_embeds` in 5.x, `cache_position` dropped).
- Harness translates legacy `fla` `fused_kda_gate(g_flat, A, head_k_dim, g_bias=...)` calls to the fla-core 0.4.2 signature (reshape + `dt_bias`; rejects non-default `beta`/`threshold`).

## Validation

- **Final proof on this exact head (61f2845dc, rebased onto post-#3567 main): nemo-ci pipeline [64546415](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/64546415) — GREEN, all six phases** (source 0.00212 / cos 0.99943, export reload 0.00226 / 0.99941, AM reload 0.00232, every self-repeat exact).

- Scoped nemo-ci pipeline 64182587, exact job `kimi_linear_48b_a3b_hellaswag`: **all six phases pass on the standard profile** — first-ever Kimi HF parity numbers: source mean KL 0.00212 / p95 0.00849 / cos 0.99943; export reload 0.00234 / 0.00901 / 0.99941; AM reload 0.00230; resume exact.
- Unit tests: 149 passed (`test_checkpoint_robustness_hf_kwargs.py`, `test_transformers_utils.py`).

Note: this PR and #3674 add functions in adjacent regions of the same harness file; whichever merges second needs a trivial context-only rebase.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3650, opened 2026-08-24
Audit and align MoE router precision with reference implementations


yuhezhang-ai:
## Summary

A public [MoE router precision audit](<https://gist.github.com/akoumpa/87dd2e957b546bfb71a71d7274f6bb7e>) compared AutoModel's effective router behavior with the pinned Hugging Face or checkpoint-owned reference implementations across 188 training, finetune, and benchmark YAMLs.

The audit checks four independent stages:

1. **Param**: stored router projection parameter dtype after checkpoint loading.
2. **Proj**: router linear/matmul arithmetic dtype.
3. **Score**: sigmoid/softmax, correction bias, normalization, and top-k arithmetic dtype.
4. **Out**: selected routing-weight dtype passed to expert computation.

At the audited head, 9 router families matched all four stages, 23 were mismatched or recipe-dependent, and Inkling could not be independently verified against public reference code. These counts describe the original audit baseline, not the current completion status. These are correctness-sensitive differences: projection or score precision can change top-k expert identity, while parameter/output dtype mismatches can change checkpoint fidelity or expert accumulation.

This issue tracks model-owned fixes and validation. It should not be solved by applying one global FP32 policy: several reference routers intentionally use BF16 for projection, scoring, or returned weights.

## Reference and load-convention policy

Checkpoint tensor dtype, reference model code, and observed runtime behavior answer different questions and must be recorded separately. A Hugging Face model loaded with an explicit model-wide BF16 dtype can legitimately downcast an F32 checkpoint outlier; that loaded model is evidence for that named harness convention, not proof that the checkpoint intended the parameter to be BF16.

Production serving implementations such as vLLM, SGLang, and TensorRT-LLM are useful independent references when they have a model-specific implementation, but they are not universal precision authorities either. Record the exact runtime/version, load options, and observed **Param / Proj / Score / Out** chain instead of inferring per-parameter behavior from a global `dtype=auto` setting.

[...]

PR #3674, opened 2026-08-25
fix(minimax): repair HF parity reference and gate at the measured envelope

yuhezhang-ai:
## What

Closes the MiniMax M2.7 checkpoint-robustness failures (AMINT-286): the catastrophic cross-framework KL (13.1) was an invalid reference, the resume drift matches the shared routed-MoE envelope, and the residual ~0.09 KL is knife-edge-router amplification of bf16 kernel noise — proven not to be an implementation defect — now gated at its measured envelope.

## Changelog

- **Reference repair (root cause of the 13.1):** Transformers 5.12's in-tree `MiniMaxM2Config` drops the legacy `rotary_dim` field, so the vanilla reference silently ran full-head RoPE instead of partial (64/128). The harness now derives `rope_parameters["partial_rotary_factor"] = rotary_dim / head_dim` for legacy configs (13.1 → 0.086). Upstreamed as huggingface/transformers#48241 / #48242.
- **Checkpoint-faithful fp32 router:** the released checkpoint stores `gate.weight` and `e_score_correction_bias` in fp32; AutoModel now scores sigmoid routing in fp32, allocates fp32-contract gate weights in fp32 at construction (keeps FSDP dtype groups uniform), and the harness pins the HF reference's fp32 tensors across vanilla-layout differences. Gate metrics are unchanged by this (the residual was never router precision), but the model now matches the checkpoint contract.
- **Recipe:** shared relaxed resume envelope (restored state and first forward are exact; 0.012–0.020 loss drift at continuation steps 6–7 across three runs), and measured `parity_threshold_overrides` for the two cross-framework gates on the full 2048-token document (mean_kl 0.15 / p95_kl 0.5 / cos 0.95 vs measured 0.086–0.094 / 0.346–0.375 / 0.964–0.965 across scoped CI 64182587, 64344786, 64351027). Overriding all three metrics numerically shadows the profile, so no relaxed-profile selection is used.

- **MiniMax-M3-VL:** the released M3 checkpoint stores its router tensors exactly like M2.7 (fp32 gate.weight; fp32 1e-3-quantized bias lattices at magnitudes 4.7–7.9, measured via ranged Hub reads). The same fp32 contract is ported to `minimax_m3_vl` (fp32 gate allocation + weight pin), and the previously uncovered M3-VL LoRA recipe is enrolled in the checkpoint-robustness gates for the AutoModel side: train/save, bitwise reload, and relaxed-envelope resume. **Green end to end in the shipped configuration: nemo-ci pipeline [64755685](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/64755685)** (job 413579005, 31 min: train 401s / reload 243s bitwise-exact / resume 359s, harness status PASS on head 0728023b6). The two vanilla-HF phases are skipped: the 427B (854 GiB bf16) single-node reference OOMs uncapped and, capped, CPU-offloads ~400 GiB while idling all 64 job GPUs for hours per run — M2.7 remains the family'''s blocking cross-framework sentinel with the identical measured router lattice.
- **Harness fixes surfaced by the new coverage:** (1) AutoModel'''s `CONFIG_MAPPING` registrations shadow in-tree config classes, so built-in references paired a nemo-owned config with the in-tree model and crashed (`temporal_patch_size`); `_replace_nemo_owned_reference_config` now falls back to Transformers''' own `CONFIG_MAPPING_NAMES`. GLM-4.7-Flash is verified unaffected (its checkpoint is `glm4_moe_lite`, never hijacked). DeepSeek-V4-Flash is the other affected type, and a local meta-device A/B proves the change strictly recovers a broken reference there too: the in-tree DSV4 model cannot even construct with the nemo-owned config (`'DeepseekV4Config' object has no attribute 'layer_types'`), while the in-tree config constructs cleanly — no working gate changes numbers. (2) `hf_device_map_max_memory_gib`/`_cpu_max_memory_gib` only reached the phase-3 reload; they now also apply to the phase-0 source-load reference (unit-tested).

## Why the envelope is a closure, not a workaround

Full-weight local investigation on the real 230 GB checkpoint (62-layer streaming sweep, single GPU):

- The in-tree HF reference **against itself** (eager vs sdpa attention) reaches mean KL 0.237 / cos 0.973 at full depth through the same expert-flip cascade (~90 % of tokens re-routed by layer 59) that drives AutoModel's 0.320 / 0.962 in the same protocol — the reference's own kernel-choice variance is the same order as the AutoModel gap.
- Per-layer matched-input deltas are at bf16 ulp scale in every arm; TE and SDPA attention are bit-identical at layer 0; an end-to-end fp32 router leaves the numbers unchanged; HF-rounding-order RMSNorms make the first layernorm bitwise equal and cut 1-layer KL 6×, yet the effect is fully erased by depth 62.
- Router evidence (pipeline 64182587): 68 % of tokens flip an expert within five layers, all flips inside the measured score-noise band, replacements symmetric, no token unflipped across 62 layers.

Real conversion or model-math regressions stay loud: the rope bug measured mean KL 13.1, two orders above the envelope.

## Validation

- **MiniMax M2.7 job green end to end at commit 695a9bcca** (nemo-ci pipeline [64546388](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/64546388)); the commits added since (M3-VL contract/coverage, reference-config resolution, phase-0 memory caps, fp32-alias fix) do not touch the M2.7 path — its `minimax_m2` model_type is not registered, so the config-resolution change never fires for it. Full gate readings: (source_load 0.0856 / 0.3565 / 0.9661 and hf_export_reload 0.0890 / 0.3396 / 0.9651 against the 0.15 / 0.5 / 0.95 envelope; all self-repeats and the AutoModel reload bitwise-exact; resume within the shared relaxed envelope).

- **Scoped CI green end to end** (nemo-ci pipeline 64481928, exact job `minimax_m2.7_hellaswag_lora`): all six phases pass — `source_load` 0.0860 / 0.3448 / 0.9644 and `hf_export_reload` 0.0869 / 0.3524 / 0.9646 against the 0.15 / 0.5 / 0.95 envelope; every self-repeat and the AutoModel reload bitwise-exact; resume within the shared relaxed envelope.
- Unit tests: 218 passed, 3 skipped (harness kwargs, MiniMax router precision, MoE layers).

The router/shape diagnostics used during the investigation are deliberately NOT enabled in the recipe: the diagnosis is complete and its evidence is archived (AMINT-286, the 62-layer sweep data, and the router-flip reports from pipeline 64182587), so report-only capture would add per-run cost without decision value. The MiniMax router-capture extension remains recoverable from the investigation branch if ever needed again.

Note: this PR and #3659 add functions in adjacent regions of the same harness file; whichever merges second needs a trivial context-only rebase.

### DIFF
--- a/nemo_automodel/components/models/minimax_m2/model.py
+++ b/nemo_automodel/components/models/minimax_m2/model.py
@@ -144,6 +144,11 @@ class MiniMaxM2Model(nn.Module):
             # The HF reference returns selected weights in the fp32 router
             # logits dtype; keep them fp32 through the expert combine.
             router_weights_fp32=True,
+            # The checkpoint stores the gate weight in fp32; allocate it fp32 so
+            # every construction path (including meta-device init before FSDP
+            # sharding) keeps the gate's dtype group uniform with its fp32
+            # correction-bias buffer.
+            gate_dtype=torch.float32,
             dtype=model_dtype,
         )
         if moe_overrides:
--- a/nemo_automodel/components/models/minimax_m2/model.py
+++ b/nemo_automodel/components/models/minimax_m2/model.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from typing import Any, Union
 
 import torch
@@ -98,6 +98,14 @@ class MiniMaxM2Model(nn.Module):
         moe_overrides: dict | None = None,
     ):
         super().__init__()
+        # Released MiniMax-M2 checkpoints store the router gate weight in fp32,
+        # and the HF reference projects with hidden_states.to(weight.dtype), so
+        # the checkpoint-faithful router runs an fp32 projection, fp32 scoring,
+        # and fp32 selected weights. Keep that default while preserving an
+        # explicit backend override (see AMINT-286; ERNIE follows the same
+        # pattern for its fp32 router).
+        if backend.gate_precision is None:
+            backend = replace(backend, gate_precision=torch.float32)
         self.backend = backend
         self.config = config
         if moe_config is not None and moe_overrides is not None:
@@ -133,6 +141,9 @@ class MiniMaxM2Model(nn.Module):
             expert_activation="swiglu",
             softmax_before_topk=(score_func == "softmax"),
             force_e_score_correction_bias=True,
+            # The HF reference returns selected weights in the fp32 router
+            # logits dtype; keep them fp32 through the expert combine.
+            router_weights_fp32=True,
             dtype=model_dtype,
         )
         if moe_overrides:
@@ -229,7 +240,7 @@ class MiniMaxM2Model(nn.Module):
 
 class MiniMaxM2ForCausalLM(HFCheckpointingMixin, nn.Module, MoEFSDPSyncMixin):
     tie_word_embeddings_support: TieSupport = TieSupport.UNTIED_ONLY
-    _keep_in_fp32_modules_strict = ["mlp.gate.e_score_correction_bias"]
+    _keep_in_fp32_modules_strict = ["mlp.gate.weight", "mlp.gate.e_score_correction_bias"]
 
     @dataclass(frozen=True)
     class ModelCapabilities:
--- a/examples/llm_finetune/minimax_m2/minimax_m2.7_hellaswag_lora.yaml
+++ b/examples/llm_finetune/minimax_m2/minimax_m2.7_hellaswag_lora.yaml
@@ -125,10 +125,28 @@ ci:
     # Use Transformers' built-in MiniMax M2 implementation for the HF+PEFT
     # reload; the checkpoint's remote code targets an older Transformers API.
     trust_remote_code: false
-    # Keep the cross-framework gates blocking: the current built-in-HF mismatch
-    # must be fixed or explicitly proven to be an invalid reference before it is skipped.
     tokenizer_name: MiniMaxAI/MiniMax-M2.7
     hf_device_map_auto: true
+    # Measured envelope for the two cross-framework gates (AMINT-286). The
+    # residual AM-vs-HF divergence is knife-edge routing amplifying bf16
+    # kernel noise, not an implementation defect: the checkpoint's fp32
+    # e_score_correction_bias lattices put top-8 selection inside the
+    # kernel-noise band (68% of tokens flip an expert within five layers,
+    # pipeline 64182587), and the HF reference against itself (eager vs sdpa
+    # attention, 62-layer sweep on real weights) diverges at the same order
+    # (mean KL 0.237 vs AutoModel's 0.320). Gate the full 2048-token document
+    # at the measured envelope (three scoped runs: mean KL 0.086-0.094,
+    # p95 0.346-0.375, cos 0.964-0.965); real conversion or model-math bugs
+    # stay loud (the repaired rope-reference bug measured mean KL 13.1).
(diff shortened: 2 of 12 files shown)


yuhezhang-ai:
/ok to test [695a9bc](https://github.com/NVIDIA-NeMo/Automodel/pull/3674/commits/695a9bccacefa481ec0a6ebfc242cd0f01d649f1)

yuhezhang-ai:
/ok to test [f1e2bed](https://github.com/NVIDIA-NeMo/Automodel/pull/3674/commits/f1e2bededa042bc8c00177b91e7dd05568012fad)

yuhezhang-ai:
/ok to test [ac5194f](https://github.com/NVIDIA-NeMo/Automodel/pull/3674/commits/ac5194ff29b62a137425c62761126dcb1d2f8f47)

yuhezhang-ai:
/ok to test [6fb4b3d](https://github.com/NVIDIA-NeMo/Automodel/pull/3674/commits/6fb4b3d090993d755ba9bae0c79cf47e7e4b4926)

yuhezhang-ai:
/ok to test [0728023](https://github.com/NVIDIA-NeMo/Automodel/pull/3674/commits/0728023b66f35a1e229e222d164ffdd893923c5b)

nemo_automodel/components/models/minimax_m2/model.py
```diff
@@ -98,6 +98,14 @@ def __init__(
         moe_overrides: dict | None = None,
     ):
         super().__init__()
+        # Released MiniMax-M2 checkpoints store the router gate weight in fp32,
+        # and the HF reference projects with hidden_states.to(weight.dtype), so
+        # the checkpoint-faithful router runs an fp32 projection, fp32 scoring,
+        # and fp32 selected weights. Keep that default while preserving an
+        # explicit backend override (see AMINT-286; ERNIE follows the same
+        # pattern for its fp32 router).
+        if backend.gate_precision is None:
```
akoumpa, line 107: 🤔 would it make sense to have this if -> set default in the `__post_init__` of `BackendConfig`? I'm wondering if that would break anything, otherwise looks like it would us some headache.

yuhezhang-ai: I think `None` needs to remain the `BackendConfig` default. It means the caller did not specify a gate-compute dtype, allowing each model to install its reference-compatible policy while preserving an explicit override.

Making FP32 the global default in `BackendConfig.__post_init__` would change projection arithmetic for router families whose references intentionally project in BF16, including Qwen3 MoE, DeepSeek V4, GPT-OSS, Mistral 4, and Laguna.  https://github.com/NVIDIA-NeMo/Automodel/issues/3650 specifically found that a global FP32 router policy would be incorrect.

For MiniMax, the F32 checkpoint gate and FP32 routing chain are model-owned, so I think this conditional belongs here. We could consider a follow-up helper to reduce the repeated immutable `replace(...)` pattern, but it should still receive the model-specific default explicitly and preserve `None` globally.

akoumpa: fair point @yuhezhang-ai , perhaps i should provide more context, what i wanted to ensure is that later the code accesses  `gate_precision` without assuming there's a case its be `None`, right now i see something like

```
            scores = torch.sigmoid(scores.to(dtype=self.gate_precision or torch.float32))
```

If possible, i feel it would be easier to maintain if it used `scores.to(dtype=self.gate_precision)`.

akoumpa (approved):

PR #3675, opened 2026-08-25
fix(peft): support mixed-dtype memory-efficient LoRA backward

### LINKED ISSUE #3652: Memory-efficient LoRA PyTorch backward fails with FSDP2 mixed precision
## Description

LoRA training fails during the first backward pass when using FSDP2 mixed precision with BF16 parameter compute and FP32 outputs.

The initial forward-only validation completes successfully, but the first training step fails in the PyTorch fallback of the memory-efficient LoRA custom autograd function:

```text
File ".../nemo_automodel/components/_peft/lora.py", line 783, in backward
    d_x.addmm_(d_y_lora_B, lora_A, beta=0, alpha=scale)

RuntimeError: self and mat2 must have the same dtype,
but got Float and BFloat16
```

## Downstream scenario

This was encountered downstream in `NVIDIA-NeMo/RL` while running the following SFT recipe:

```text
examples/configs/recipes/llm/sft-llama3.1-8b-1n8g-fsdp2tp1-lora.yaml
```

Configuration summary:

- Model: `meta-llama/Llama-3.1-8B`
- LoRA enabled, rank 128, alpha 128
- FSDP2
- Tensor parallel size: 1
- Precision: BF16
- 1 node, 8 H100 GPUs
- NeMo-RL commit: `98c3f04caf8805bb2e97ed98bb96a9750b8a7c0c`
- Automodel commit: `94387e4318bfd04cb21ff57782061864e0c7bcec`

The downstream FSDP2 policy is:

```python
MixedPrecisionPolicy(
    param_dtype=torch.bfloat16,
    reduce_dtype=torch.float32,
    output_dtype=torch.float32,
)
```

The problem was reproduced in a separate one-node/eight-GPU Slurm run. Initial validation completed with a valid loss, followed by the same exception during the first `loss.backward()` call.

## Analysis

The failure appears to be caused by mixed-dtype handling in the memory-efficient LoRA PyTorch backward path:

```python
d_x = torch.empty_like(x)
d_x.addmm_(d_y_lora_B, lora_A, beta=0, alpha=scale)
```

With the FSDP2 policy above, `x` and therefore `d_x` are FP32, while `lora_A` is BF16 during computation. The in-place `addmm_` operation requires compatible dtypes and fails with FP32 `self` and BF16 `mat2`.

Forward-only validation succeeds because autocast handles the forward matmuls. The mismatch is exposed only in the manually implemented backward.

Although the recipe sets `use_triton: true`, `patch_linear_module()` disables Triton whenever Transformer Engine is available, so this run enters the PyTorch fallback.

Existing LoRA tests appear to cover homogeneous FP32 or BF16 tensors, but not FP32 activations combined with BF16 LoRA parameters.

## Expected behavior

Memory-efficient LoRA backward should support a valid FSDP2 mixed-precision policy where:

- activations/residual outputs are FP32;
- LoRA parameters compute in BF16;
- returned `d_x` matches the original input dtype;
- LoRA parameter gradients match their parameter dtype.

akoumpa:
# What does this PR do ?

Fix memory-efficient LoRA under mixed-precision FSDP2 when activations and adapter weights have different dtypes (the NeMo-RL failure in #3652).

# Changelog

- Recompute fallback LoRA backward operations in the forward compute dtype, then return gradients in the dtype of their corresponding inputs.
- Support mixed activation/adapter dtypes in the fused LoRA MLP forward and backward paths.
- Decline Triton LoRA kernels for mixed dtypes so those calls use the safe PyTorch path.
- Decline fused LoRA MLP dispatch for biased projections, whose bias semantics are not supported by the fused implementation.
- Add CPU autocast, CUDA kernel, fused-MLP, and real two-rank FSDP2 mixed-precision regression coverage.

# Validation

- GitHub CI at exact head `9d4b0ae62`: linting, type checks, CPU/GPU unit tests, and the standard H100 L2 suites passed (remaining GB200 matrix jobs are still running/queued).
- Local CUDA reference parity: zero output and gradient differences.
- Local two-rank FSDP2 mixed-precision backward: passed.
- cw-dfw Slurm job `16590286`: 8/8 H100 ranks passed the focused BF16/FP32 regression.
- cw-dfw Slurm job `16621610`: the exact NeMo-RL issue recipe completed on 8x H100 with exit code 0 using this PR's exact source hashes. [W&B run](https://wandb.ai/Nemo-automodel/nemo-rl/runs/9aukeum8): validation loss `0.8246825933`, train loss `0.8588075314`, grad norm `0.9352310300`, one optimizer step completed.

**Pre checks**:

- [x] Did you add or update any necessary documentation? (No documentation change required.)

# Additional Information

Fixes #3652.

### DIFF
--- a/nemo_automodel/components/_peft/lora.py
+++ b/nemo_automodel/components/_peft/lora.py
@@ -751,11 +751,21 @@ class LoRATritonFunction(torch.autograd.Function):
 
     @staticmethod
     def backward(ctx, d_y):
-        """
-        Backward method for memory-efficient LoRA.
+        """Compute gradients for memory-efficient LoRA.
 
         Reshapes 3D tensors into 2D and then updates d_lora_a, d_lora_b, and dx. The PyTorch matmul
         path recomputes ``x @ lora_A.T`` here instead of saving it from forward.
+
+        Args:
+            ctx: Autograd context containing the saved input and LoRA weights.
+            d_y: Tensor of shape ``[tokens, out_features]`` containing the output gradient in the forward compute
+                dtype.
+
+        Returns:
+            Gradients for the custom autograd inputs. The input gradient has shape ``[tokens, in_features]`` or
+            ``[batch, sequence, in_features]`` and matches the original input dtype. LoRA weight gradients have
+            shapes ``[rank, in_features]`` and ``[out_features, rank]`` and match their weight dtypes. When a
+            residual input is present, its gradient matches the original residual shape.
         """
         x, lora_A, lora_B = ctx.saved_tensors
         scale = ctx.scale
@@ -779,17 +789,23 @@ class LoRATritonFunction(torch.autograd.Function):
         else:
             d_x = d_lora_A = d_lora_B = None
             needs_x, needs_lora_A, needs_lora_B = ctx.needs_input_grad[:3]
+            # Custom backward executes outside autocast. Reuse d_y's forward compute dtype for the recomputed
+            # matmuls, then cast each returned gradient to the dtype of its corresponding input.
+            x_compute = x.to(d_y.dtype) if needs_lora_A or needs_lora_B else x
+            lora_A_compute = lora_A.to(d_y.dtype) if needs_x or needs_lora_B else lora_A
+            lora_B_compute = lora_B.to(d_y.dtype) if needs_x or needs_lora_A else lora_B
             if needs_x or needs_lora_A:
-                d_y_lora_B = torch.matmul(d_y, lora_B)
+                d_y_lora_B = torch.matmul(d_y, lora_B_compute)
+                d_y_lora_B.mul_(scale)
                 if needs_x:
-                    d_x = torch.empty_like(x)
-                    d_x.addmm_(d_y_lora_B, lora_A, beta=0, alpha=scale)
+                    d_x = torch.matmul(d_y_lora_B, lora_A_compute).to(x.dtype)
                 if needs_lora_A:
-                    d_lora_A = torch.matmul(d_y_lora_B.t(), x) * scale
+                    d_lora_A = torch.matmul(d_y_lora_B.t(), x_compute).to(lora_A.dtype)
 
             if needs_lora_B:
-                d_lora_B = torch.empty_like(lora_B)
-                d_lora_B.addmm_(d_y.t(), F.linear(x, lora_A), beta=0, alpha=scale)
+                x_lora_A = F.linear(x_compute, lora_A_compute)
+                x_lora_A.mul_(scale)
+                d_lora_B = torch.matmul(d_y.t(), x_lora_A).to(lora_B.dtype)
 
         if reshape and d_x is not None:
             d_x = d_x.view(bs, seq_len, d)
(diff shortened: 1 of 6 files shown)


HuiyingLi (commented):
[P1] Handle the automatically fused MLP backward in the mixed-dtype fix as well

The per-linear `LoRATritonFunction.backward` change is correct, but it does not cover the other memory-efficient custom-autograd path that the production entry point enables automatically. When all gate/up/down projections are LoRA-patched, `apply_lora_to_linear_modules()` installs the fused MLP wrapper, so execution enters `LoRASwiGLUMLPFunction` instead of the function fixed here.

I reproduced this at head `59abf7495e27f401d510c0c849d21efa36f565b6` with a real `LlamaMLP` and with 2-rank CPU FSDP2 (`param_dtype=bf16`, `reduce_dtype=fp32`, `output_dtype=fp32`) under BF16 autocast. At runtime `_fusible` is true after unshard, the saved activation is FP32, and the LoRA/base compute weights are BF16. The first backward fails at [`Q = F.linear(x, uA)`](https://github.com/NVIDIA-NeMo/Automodel/blob/59abf7495e27f401d510c0c849d21efa36f565b6/nemo_automodel/components/_peft/lora_mlp.py#L184):

```text
RuntimeError: expected m1 and m2 to have the same dtype, but got: float != c10::BFloat16
```

The ReLU² fused backward has the analogous operation [here](https://github.com/NVIDIA-NeMo/Automodel/blob/59abf7495e27f401d510c0c849d21efa36f565b6/nemo_automodel/components/_peft/lora_mlp.py#L327) and fails under the same dtype combination. The new regression test calls `apply_memory_efficient_lora()` directly, so it bypasses this automatic fusion and remains green.

Please either apply the same compute-dtype/returned-gradient-dtype handling to both fused MLP backward functions, or decline fusion for mixed-dtype execution and fall back to the now-fixed per-linear path. A regression test should go through `apply_lora_to_linear_modules()` with a SwiGLU/ReLU² MLP so the production routing is exercised.

akoumpa:
Exact-head downstream reproduction is green.

- PR head: `9d4b0ae62fb6df7a8194e14132de96fd87fb9db7`
- NeMo-RL recipe: `sft-llama3.1-8b-1n8g-fsdp2tp1-lora.yaml` (`use_triton: true`, BF16, LoRA rank 128, FSDP2, 8x H100)
- cw-dfw Slurm: `16621610`, `COMPLETED`, exit `0:0`
- [W&B run](https://wandb.ai/Nemo-automodel/nemo-rl/runs/9aukeum8), state `finished`
- Validation loss: `0.8246825933456421`
- Train loss after step 1/1: `0.8588075314473826`
- Grad norm: `0.9352310299873352`

The launch preflight verified that the remote sources exactly matched the PR implementation:

```text
lora.py     f01629b7faf25f5f7ce27502d5f41904d17bb26c5a557886a758eab51f78b2c1
lora_mlp.py bd36f488b1237dd5c090900089f6fb585e6506f33952956913c3883c1b51b984
TORCH_CUDA_COUNT=8
```

This is the original downstream failure shape, and it now completes validation plus a training/backward/optimizer step without the mixed-dtype matmul error.

HuiyingLi (approved):

PR #3677, opened 2026-08-25
fix(fsdp): restore full-layer activation checkpointing for KV-shared models

athitten:
# What does this PR do ?

Fixes a 16% memory regression on the nightly Gemma4-4B benchmark by putting KV-shared
models (Gemma4 E2B/E4B) back on full-layer activation checkpointing.
Fixs NVBug 6630889 and 6620930

# The bug

Activation checkpointing can be applied two ways:

- **whole block** — wrap the entire decoder layer, so attention is recomputed in backward
- **sub-module** — wrap `mlp` and the norms individually

The sub-module path deliberately **skips `self_attn`** for KV-shared models.

PR #3513 added `and not _has_kv_sharing` to the whole-block branch. Gemma4 E4B has
`num_kv_shared_layers: 18`, so it started taking the sub-module path instead. Attention
activations across all 42 language layers stopped being recomputed and stayed resident:

```
Applied submodule activation checkpointing to 42 layers:
  {'mlp': 42, 'attention': 0, 'pre_norm': 42, 'post_norm': 42, 'mot': 0}
                ^^^^^^^^^^^^^
```

Nothing crashed and the loss did not move, so only the memory metric caught it.

# Why the guard is not needed

The guard exists because recomputing attention could write K/V into the cache a second
time. That does not happen here: Gemma4 passes its shared layers `_Gemma4KVShareHolder`,
whose `update()` just returns what it was given. Replaying a block writes nothing.

# Evidence

2x GB200, `examples/vlm_finetune/gemma4/gemma4_4b_mock.yaml`, activation checkpointing on:

| | peak memory | loss / grad_norm |
|---|---|---|
| `main` today (has the bug) | 87.37 GiB | 6.0085 / 125.5791 … |
| `main` before #3513 landed | 70.06 GiB | identical |
| **`main` + this PR** | **70.06 GiB** | identical |

70.06 GiB matches the nightly's pre-regression value of `7.005e+01` exactly, and losses
and gradient norms agree to four decimals across all 8 steps. This was purely a memory
regression, not a numerics change.

# Changelog

- `parallelizer.py`: drop the `and not _has_kv_sharing` guard so KV-shared models get
  whole-block checkpointing again, keeping #3513's mixed-precision-safe wrapper.
- `test_parallelizer.py`: `test_hf_native_candidate_with_kv_sharing_uses_full_layer_checkpointing`
  now asserts whole-block wrapping. #3513 had rewritten this test to assert the regressed
  behaviour, so it went green on the way in.
- New functional test `L2_Parallelism_VLM_Gemma4_KVShared_AC` (2 GPUs, builds a small
  E4B-shaped model in-process, nothing staged). It asserts every decoder block is
  checkpoint-wrapped, and that gradients match an un-checkpointed run. It lands in the
  `parallelism` folder added by #3529, so no CI wiring changes are needed.

Verified the new test fails on the regressed code:

```
AssertionError: KV-shared Gemma4 decoder blocks [0, 1, 2, 3] are not checkpoint-wrapped;
activation checkpointing fell back to sub-module wrapping, which excludes self_attn
```

and passes with the fix (`OK (63 gradients compared)`).

Peak-memory magnitude stays with the nightly benchmark — a proxy this small can't measure
it without a threshold that says more about the proxy than about the code.

**Pre checks**:

# Additional Information

- Fixes the `memory_usage` regression on `basic/tracked-finetune-gemma4-4b-automodel-ac`
  (nightly-gb200-finetune-automodel), first seen 2026-08-14.
- Caused by #3513.

### DIFF
--- a/nemo_automodel/components/distributed/parallelizer.py
+++ b/nemo_automodel/components/distributed/parallelizer.py
@@ -2113,16 +2113,27 @@ def _kv_sharing_survives_checkpoint_replay(model: nn.Module) -> bool:
 
     ``checkpoint_wrapper`` replays a whole decoder block during backward with the
     arguments the forward saw. Unlike HF's ``GradientCheckpointingLayer.__call__``
-    it cannot drop ``past_key_values`` from that replay, so a model whose shared
-    layers read an accumulating ``Cache`` calls ``Cache.update()`` twice and the
-    recomputed K/V no longer matches the forward. That surfaces as a
-    ``CheckpointError`` about changed tensor metadata (observed on native HF
-    ``Gemma3nForCausalLM`` with ``use_cache=True``), so KV-shared models stay on
-    ``apply_submodule_checkpointing`` by default.
+    it cannot drop ``past_key_values`` from that replay, so every layer that
+    writes to a cache writes to it a second time. A KV-shared model then needs
+    both halves to hold:
 
-    A model whose shared-K/V store is replay-safe -- a pass-through holder rather
-    than an accumulating cache -- opts in by setting the class attribute
-    ``kv_sharing_survives_checkpoint_replay = True``.
+    * the layers that populate the cache -- the *non*-shared ones, which are what
+      call ``Cache.update()`` -- must not accumulate on the replay, or the
+      recomputed K/V stops matching the forward;
+    * the shared layers must still read the K/V their source layer produced.
+
+    Neither holds for a model backed by an accumulating ``Cache``: the second
+    ``Cache.update()`` grows the entry and backward dies with a
+    ``CheckpointError`` about changed tensor metadata (observed on native HF
+    ``Gemma3nForCausalLM`` with ``use_cache=True``). KV-shared models therefore
+    stay on ``apply_submodule_checkpointing``, which leaves attention unwrapped,
+    by default.
+
+    A model that satisfies both halves opts in by setting the class attribute
+    ``kv_sharing_survives_checkpoint_replay = True``. Gemma4 E2B/E4B qualify: a
+    pass-through holder stands in for the cache, and the shared layers read a
+    separate store that the replay does not disturb (see
+    ``gemma4_moe/model.py``).
 
     Args:
         model: The model about to be checkpointed.
--- a/nemo_automodel/components/distributed/parallelizer.py
+++ b/nemo_automodel/components/distributed/parallelizer.py
@@ -406,17 +406,17 @@ class DefaultParallelizationStrategy(ParallelizationStrategy):
                         if m is not None:
                             setattr(layer, attr, checkpoint_wrapper(m, checkpoint_impl=CheckpointImpl.NO_REENTRANT))
             else:
-                if (
-                    _should_use_hf_native_gradient_checkpointing(
-                        model,
-                        layer_groups,
-                        ac_scopes,
-                        enable_compile=enable_compile,
-                    )
-                    and not _has_kv_sharing
+                if _should_use_hf_native_gradient_checkpointing(
+                    model,
+                    layer_groups,
+                    ac_scopes,
+                    enable_compile=enable_compile,
                 ):
                     # Work around a PyTorch FSDP2 bug that skips mixed-precision input casts during
                     # checkpoint recomputation. Remove when the minimum PyTorch version is 2.13.
+                    #
+                    # KV-shared models belong on this path too; they rode the equivalent HF-native
+                    # full-layer path until #3513 rewrote this branch.
                     apply_full_layer_checkpointing_to_layers(model, ac_layers)
                 else:
                     apply_submodule_checkpointing(ac_layers, _has_kv_sharing)
--- a/nemo_automodel/components/distributed/parallelizer.py
+++ b/nemo_automodel/components/distributed/parallelizer.py
@@ -411,12 +411,9 @@ class DefaultParallelizationStrategy(ParallelizationStrategy):
                     layer_groups,
                     ac_scopes,
                     enable_compile=enable_compile,
-                ):
+                ) and (not _has_kv_sharing or _kv_sharing_survives_checkpoint_replay(model)):
                     # Work around a PyTorch FSDP2 bug that skips mixed-precision input casts during
                     # checkpoint recomputation. Remove when the minimum PyTorch version is 2.13.
-                    #
-                    # KV-shared models belong on this path too; they rode the equivalent HF-native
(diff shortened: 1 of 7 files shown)


HuiyingLi (changes requested):
Requesting changes for one correctness blocker and one test-coverage gap found across three independent reviews.

### P1: keep generic KV-sharing models out of whole-layer checkpointing

Removing the generic `_has_kv_sharing` guard in `parallelizer.py` routes every HF-native KV-sharing model through the whole-layer `CheckpointWrapper`. The replay safety described by this PR is specific to the NeMo Gemma4 wrapper, whose `_Gemma4KVShareHolder.update()` is pass-through; it does not hold for native HF models backed by `DynamicCache` while `use_cache=True`.

I reproduced this with the repository-pinned Transformers 5.12.1 in both PyTorch 2.10.0 and the current NGC PyTorch 2.13 container:

- tiny native `Gemma3nForCausalLM`: the plain backward completes with 97 gradients, while this PR's whole-layer route fails during backward because checkpoint recomputation changes saved K/V metadata from `[1, 1, 8, 8]` to `[1, 1, 16, 8]`;
- tiny native `Gemma4UnifiedForCausalLM`: the plain backward completes with 48 gradients, while the whole-layer route fails with the same 8-to-16 cache growth.

The backward replay calls `DynamicCache.update()` a second time. The previous KV-sharing submodule route, which excludes attention from checkpoint wrapping, completes successfully. Please retain the generic guard and make the NeMo custom Gemma4 holder opt in through a model-owned strategy/capability, then add a native-HF `use_cache=True` regression test.

### P2: make the new L2 fail on output or gradient omissions

`run_gemma4_kv_shared_ac.py` computes `checkpointed_logits` and `plain_logits` but never compares them. Its gradient loop also silently skips a parameter when its counterpart is missing or either gradient is `None`, and only requires that at least one gradient was compared.

The current run compares 63 of 79 trainable parameters; the 16 expected omissions are vision-tower parameters not reached by the text-only input, but the test does not assert that omission set. In an adversarial check, adding a detached `1.0` to the checkpointed logits and clearing `layers.0.self_attn.q_proj.weight.grad` still lets the loop pass with 62 gradients compared.

Please assert logits parity, exact parameter-name/gradient-presence parity for the exercised language path, and explicitly freeze or enumerate the expected vision-only omissions.

Validation performed on head `fef1f8e17e083432c0d85470b7c2035227f549ce`:

- the new 2-rank Gemma4 L2 passes on 2x H100 (`63 gradients compared`);
- targeted unit tests: `181 passed, 1 skipped`;
- Ruff check, Ruff format check, and `git diff --check` pass;
- all current CI checks are green, but they do not exercise the native-HF KV-cache replay path above.

athitten:
> Requesting changes for one correctness blocker and one test-coverage gap found across three independent reviews.
> 
> ### P1: keep generic KV-sharing models out of whole-layer checkpointing
> Removing the generic `_has_kv_sharing` guard in `parallelizer.py` routes every HF-native KV-sharing model through the whole-layer `CheckpointWrapper`. The replay safety described by this PR is specific to the NeMo Gemma4 wrapper, whose `_Gemma4KVShareHolder.update()` is pass-through; it does not hold for native HF models backed by `DynamicCache` while `use_cache=True`.
> 
> I reproduced this with the repository-pinned Transformers 5.12.1 in both PyTorch 2.10.0 and the current NGC PyTorch 2.13 container:
> 
> * tiny native `Gemma3nForCausalLM`: the plain backward completes with 97 gradients, while this PR's whole-layer route fails during backward because checkpoint recomputation changes saved K/V metadata from `[1, 1, 8, 8]` to `[1, 1, 16, 8]`;
> * tiny native `Gemma4UnifiedForCausalLM`: the plain backward completes with 48 gradients, while the whole-layer route fails with the same 8-to-16 cache growth.
> 
> The backward replay calls `DynamicCache.update()` a second time. The previous KV-sharing submodule route, which excludes attention from checkpoint wrapping, completes successfully. Please retain the generic guard and make the NeMo custom Gemma4 holder opt in through a model-owned strategy/capability, then add a native-HF `use_cache=True` regression test.
> 
> ### P2: make the new L2 fail on output or gradient omissions
> `run_gemma4_kv_shared_ac.py` computes `checkpointed_logits` and `plain_logits` but never compares them. Its gradient loop also silently skips a parameter when its counterpart is missing or either gradient is `None`, and only requires that at least one gradient was compared.
> 
> The current run compares 63 of 79 trainable parameters; the 16 expected omissions are vision-tower parameters not reached by the text-only input, but the test does not assert that omission set. In an adversarial check, adding a detached `1.0` to the checkpointed logits and clearing `layers.0.self_attn.q_proj.weight.grad` still lets the loop pass with 62 gradients compared.
> 
> Please assert logits parity, exact parameter-name/gradient-presence parity for the exercised language path, and explicitly freeze or enumerate the expected vision-only omissions.
> 
> Validation performed on head `fef1f8e17e083432c0d85470b7c2035227f549ce`:
> 
> * the new 2-rank Gemma4 L2 passes on 2x H100 (`63 gradients compared`);
> * targeted unit tests: `181 passed, 1 skipped`;
> * Ruff check, Ruff format check, and `git diff --check` pass;
> * all current CI checks are green, but they do not exercise the native-HF KV-cache replay path above.

Thanks @HuiyingLi , sorry for missing this.  Addressed in 8f536e841e9f180a16f0d8e60c4db0e62ffba539

P1: Whole-block checkpointing is now gated on a model-owned opt-in, kv_sharing_survives_checkpoint_replay, defaulting to False - so every KV-shared model keeps the sub-module route unless it declares otherwise. Only gemma4_moe.Gemma4ForConditionalGeneration sets it; gemma4_unified and gemma4_drafter ride plain HF and do not.

P2 -  L2 assertions

Now asserts logits parity, exact parameter-name parity, exact gradient-presence parity, and that the ungraded set is exactly the vision path (not a prefix guess - writing that caught model.embed_vision.*, which I had missed).

Both of your adversarial cases now fail:

- +1.0 on the checkpointed logits -> logits diverged under activation checkpointing
- clearing layers.0.self_attn.q_proj.weight.grad -> gradient presence differs: missing-in-checkpointed=[...q_proj.weight]

As shipped: 63 gradients compared, 16 vision-only omissions.

Also added

tests/unit_tests/models/gemma4/test_kv_sharing_ac_optin.py - nothing tied the opt-in attribute to the class that carries it, so deleting it from gemma4_moe/model.py reintroduced the regression with the whole suite green. That deletion now fails 2 of 4 tests.

HuiyingLi (approved):

Issue #3219, opened 2026-07-25
refactor(distributed): unify multimodal module discovery and policy configuration


yuhezhang-ai:
## Motivation

#2763 exposed a recurring problem: several features independently decide which model submodules
are multimodal, and those private taxonomies drift into distributed-correctness bugs. The next
rank-asymmetric mixed-modality safety feature also needs richer shared metadata: modality, role,
and whether a path may be skipped.

This issue proposes:

1. one resolver for multimodal module identity;
2. one `distributed.multimodal` namespace for policies applied to those modules; and
3. a later, separately tested migration of `freeze_config` to the same addressing vocabulary.

## Problem

“Which submodule is the vision/audio tower?” is currently answered independently:

| Owner | Match rule |
|---|---|
| `_get_model_layer_group_specs()` | per-class FQN paths in `language` / `vision` / `audio` groups |
| `shared/multimodal_fsdp.py` | exact leaf-name taxonomy for dense and MoE FSDP |
| `apply_parameter_freezing()` | attributes plus substring matching on lowercased FQNs |

Before #2763, FSDP creation and enumeration also used separate lists. That caused real bugs:

- BAGEL's `vit_model` existed in the MoE tuple but not the shared FSDP list.
- Gemma4's sibling `embed_vision` and `embed_audio` modules—instances of the same class—could
  receive different sharding treatment because only one name was listed.
- `_iter_fsdp_modules()` could miss FSDP units created by `apply_fsdp()`, excluding them from the
  MoE gradient-accumulation synchronization state machine.

#2763 fixes those immediate FSDP inconsistencies by sharing one taxonomy. The architectural
duplication remains: activation checkpointing still uses the layer-group table, freezing uses
substring matching, and future conditional-execution safety needs more than a leaf name.

Three planned PRs need to add policy for the same multimodal paths:

- #2763 initially adds `distributed.frozen_multimodal_sharding` to choose frozen-module FSDP ownership.
- #2990 adds vision frame sharding over CP as `distributed.cp_vision_frame_sharding`.
- A safety follow-up will add a dummy-forward policy for rank-asymmetric mixed-modality batches, as `run_dummy_forward`.

Rather than let each PR introduce an independent distributed config path, all three should extend
`distributed.multimodal`. They govern different behaviors, but each controls how resolved
multimodal modules interact with the existing parallel axes of the text model.

Finally, `freeze_config` is unvalidated. In the current VLM/audio/long-context example scope, 65
[...]

PR #3681, opened 2026-08-26
feat: support generic freeze-config selectors

pstjohn:
Adds support for custom `freeze_modules` / `unfreeze_modules` in `freeze_config`. This version still supports the legacy `freeze_vision_tower` arguments

```yaml
freeze_config:
  freeze_vision_tower: false
  freeze_modules:
  - encoder
  unfreeze_modules:
  - projector
```

this now also supports the same glob-style pattern of peft's `target_modules`:

```yaml
freeze_config:
  freeze_modules: "vision*"
```

porting all these existing [`freeze_vision_tower`](https://github.com/search?q=repo%3ANVIDIA-NeMo%2FAutomodel+freeze_vision_tower&type=code) args over to `freeze_modules` would be a pretty big refactor; for the time i've just emitted a deprecation warning letting people know about the `freeze_modules` option.

Alternative to #3661 to enable freezing / unfreezing modules in between model construction and optimizer creation.


yuhezhang-ai:
Thanks @pstjohn at a high level, I think this is the right feature. We want to support configurations such as LoRA plus a fully trainable multimodal projector.

## Design

Because this introduces a new public configuration, I think we should settle its long-term shape before merging. My suggestion is to keep your `freeze_modules` / `unfreeze_modules` organization, but make each selector explicit and typed:

```yaml
freeze_config:
  freeze_modules:
    - path: vision_tower
    - glob: "*.audio_encoder"

  unfreeze_modules:
    - path: multi_modal_projector
```

The intended semantics would be:

1. Full fine-tuning or PEFT establishes the baseline trainability.
2. `freeze_modules` recursively freezes selected modules.
3. `unfreeze_modules` recursively unfreezes selected modules and wins on overlap.
4. Framework-required freezes for unused/dead parameters remain protected.
5. The final trainable parameter set is validated before optimizer construction.

`path` should mean an exact canonical module path, while `glob` should use case-sensitive shell-style matching against the full path. I would avoid accepting ambiguous bare strings in this new API, since their matching behavior would become a long-term compatibility obligation.

This structure also leaves an additive path for the multimodal resolver planned in #3219:

```yaml
freeze_config:
  freeze_modules:
    - modality: vision
      role: tower
```

The resolver can later translate that semantic selector into module paths without changing the surrounding freeze configuration.

I suggest representing this with a typed `FreezeConfig` and typed selector variants, with strict rejection of unknown fields, invalid selector combinations, and selectors that match no parameters. The existing modality-specific booleans should remain supported without deprecation for now; once #3219 lands, we can verify equivalent frozen-parameter sets before deprecating them.

## Correctness blockers

Separately from the public API, I see several correctness issues that should be addressed before merging:

1. **Trainability does not reliably survive parameter replacement.** The policy is applied before sharding and PEFT overrides are restored using parameter FQNs. Some structural transformations create new parameters or change their FQNs—for example, grouped-expert transformations. A frozen parameter can therefore become trainable under full fine-tuning, or an explicitly unfrozen parameter can fall back to frozen under PEFT. The current replacement test preserves the same names and does not cover this failure.

   The freeze policy should be applied or rebound after model surgery but before DDP/FSDP captures the parameters, followed by final validation before optimizer construction. We should cover name-changing replacement under both full fine-tuning and PEFT.

2. **Unfreezing must not change parameter storage dtype.** The current path casts explicitly selected trainable plain parameters to the compute dtype. Freeze configuration should only control `requires_grad`; it should not turn FP32 resident/master parameters into BF16 parameters. Compute dtype should remain owned by autocast or the distributed mixed-precision policy.

3. **The wildcard implementation is not literal glob matching.** The current implementation builds a regular expression without escaping ordinary characters, so `.` in a module path also acts as a regex wildcard. This can select unintended modules. Please use `fnmatchcase` or escape literal text before expanding `*`.

4. **The proposed legacy migration is not behaviorally equivalent.** For example, the legacy audio option also recognizes names containing `speech` and `sound`, while the documented generic replacement selects only `audio_encoder`. Deprecation should wait until migration is checked using equivalent frozen-parameter FQN sets.

I’m supportive of landing the generic freezing capability in this PR once the new contract is stabilized and these correctness issues are resolved.

pstjohn:
Ok, implemented this. We now have a typed `FreezeConfig` object; and the `ModuleSelector` accepts either path or glob inputs:

```python
@dataclass(frozen=True)
class ModuleSelector:
    path: str | None = None
    glob: str | None = None

@dataclass(frozen=True)
class FreezeConfig:
    freeze_modules: list[ModuleSelector] = field(default_factory=list)
    unfreeze_modules: list[ModuleSelector] = field(default_factory=list)
    freeze_vision_tower: bool | None = None
    freeze_audio_tower: bool = False
    freeze_language_model: bool = False
    freeze_video_embedder: bool = False
 ```
 
Each selector requires exactly one of `path` or `glob`; bare strings, unknown fields, invalid combinations, and unmatched selectors are rejected.

1. **Trainability across parameter replacement:** The FQN snapshot/restore mechanism is gone. We strictly validate selectors against the complete model, then re-resolve the policy after TP/EP/activation-checkpointing surgery immediately before DDP/FSDP construction. We resolve it once more after checkpoint loading and before optimizer construction. The order is baseline → freeze selectors → unfreeze selectors → framework-required freezes. Tests cover name-changing replacement under full fine-tuning and PEFT, wrapper-construction ordering, and real DDP gradient synchronization.
2. **dtype:** Freeze configuration only changes `requires_grad`; explicitly unfrozen parameters retain FP32 storage. Sequence classification now honors `distributed.autocast_dtype`, covering the single-rank case where FSDP mixed precision is skipped, while distributed mixed precision handles sharded execution.
3. **glob literalness:** Freeze selectors use `fnmatch.fnmatchcase`, so matching is case-sensitive and `.` is literal. The PEFT matcher remains unchanged in `module_matcher.py`; this PR no longer shares or modifies its matching implementation.
4. **legacy migration:** The deprecation warning and non-equivalent migration guidance have been removed. The modality-specific booleans remain supported with their established matching behavior. We can revisit deprecation after #3219 with equivalent frozen-parameter-set validation.ptions. Deprecation can be revisited after #3219 with equivalent-frozen-set validation.

File: docs/guides/llm/sequence-classification.mdx
Comment (jgerh) [inline docs/guides/llm/sequence-classification.mdx]:
```suggestion
description: "Train a sequence classification model with NeMo AutoModel using GLUE MRPC, RoBERTa, and optional LoRA."
```

jgerh:
```suggestion
Sequence classification tasks (for example, sentiment analysis, topic classification, and GLUE tasks) map input text to a discrete label. NeMo AutoModel provides a lightweight recipe specialized for this setting that integrates with popular pretrained model formats and dataset sources. Integration with Hugging Face is supported.
```

jgerh:
```suggestion
This guide shows how to train a sequence classification model using the `TrainFinetuneRecipeForSequenceClassification` recipe, including optional Parameter-Efficient Fine-Tuning (PEFT) with LoRA.
```

jgerh:
```suggestion
Use the example config for GLUE MRPC with RoBERTa-large and LoRA:
```

docs/guides/llm/sequence-classification.mdx
```diff
@@ -110,6 +115,8 @@ optimizer:
 
 - `target_modules`: glob to select linear layers (e.g., `"*.proj"`).
 - `dim` (rank), `alpha`, `dropout`: tune per model/compute budget. Values `dim=8, alpha=16, dropout=0.1` are a good starting point for RoBERTa.
+- `freeze_config.unfreeze_modules`: keeps the classification head fully trainable while PEFT freezes other non-LoRA parameters.
+- `distributed.autocast_dtype`: runs the forward pass in the selected compute dtype while trainable parameters retain their configured storage dtype, including when single-rank FSDP is skipped.
```
jgerh:
```suggestion
- `target_modules`: Glob to select linear layers (for example, `"*.proj"`).
- `dim` (rank), `alpha`, `dropout`: Tune per model and compute budget. Values `dim=8, alpha=16, dropout=0.1` are a good starting point for RoBERTa.
- `freeze_config.unfreeze_modules`: Keeps the classification head fully trainable while PEFT freezes other non-LoRA parameters.
- `distributed.autocast_dtype`: Runs the forward pass in the selected compute dtype while trainable parameters retain their configured storage dtype, including when single-rank FSDP is skipped.
```

jgerh (commented):
Completed tech pubs review of docs/guides/llm/sequence-classification.mdx and provided a few copyedits.

yuhezhang-ai:
Thanks @pstjohn

For the ci failure, I think`freeze_embeddings` was never really used (it was ignored by the code). I would say we can remove all occurrences specifically under `freeze_config` from the affected configs and documentation while keeping strict rejection. Please do not remove the separately supported `freeze_embeddings` options in speculative-training recipe arguments.

   If a user genuinely wants to freeze token embeddings, they can use an explicit model-specific selector such as:

   ```yaml
   freeze_modules:
     - path: model.language_model.embed_tokens
   ```

There are a few other items:
1. **Preserve parameters intentionally frozen by the model**

   The current full-fine-tuning baseline [[sets every parameter to trainable](https://github.com/NVIDIA-NeMo/Automodel/blob/cff2a57ed22e38cac6bea3f21bf2b0674ba8246e/nemo_automodel/_transformers/infrastructure.py#L495-L500)](https://github.com/NVIDIA-NeMo/Automodel/blob/cff2a57ed22e38cac6bea3f21bf2b0674ba8246e/nemo_automodel/_transformers/infrastructure.py#L495-L500) whenever generic selectors are used.

   This can unintentionally unfreeze model-owned constants. For example, Bagel has a [[fixed positional-embedding parameter](https://github.com/NVIDIA-NeMo/Automodel/blob/cff2a57ed22e38cac6bea3f21bf2b0674ba8246e/nemo_automodel/components/models/bagel/embeddings.py#L75-L85)](https://github.com/NVIDIA-NeMo/Automodel/blob/cff2a57ed22e38cac6bea3f21bf2b0674ba8246e/nemo_automodel/components/models/bagel/embeddings.py#L75-L85) constructed with `requires_grad=False`.

   Generic freeze/unfreeze selectors should overlay the model’s existing trainability state rather than globally resetting it. Please add a test showing that an unrelated, initially frozen parameter remains frozen.

2. **Distinguish empty selector fields from legacy configuration**

   These currently become indistinguishable:

   ```yaml
   freeze_config:
     freeze_modules: []
   ```

   ```yaml
   freeze_config: {}
   ```

   The first explicitly opts into selector-based configuration but selects nothing; the second retains legacy behavior, including the historical implicit vision freeze.

   Please preserve whether `freeze_modules` or `unfreeze_modules` was declared. Empty lists can mean “no modules selected,” or they can be rejected, but they should not silently activate legacy defaults. Values such as `null`, `false`, and `0` should also be rejected instead of being treated as empty lists.

3. **Add real FSDP2 lifecycle coverage**

   The two-rank DDP test is valuable. However, this PR also changes trainability timing around FSDP2, Megatron-FSDP, TP/EP, and pipeline transformations, while the new FSDP2 tests currently mock the wrapping order.

   Please add the smallest practical two-rank FSDP2 test performing forward, backward, and one optimizer step, verifying that the final selected parameters remain trainable and synchronized after parameter replacement and wrapping.

pstjohn:
Thanks, addressed!

1. **Preserve parameters intentionally frozen by the model**
Generic selectors now overlay the model’s existing requires_grad state during full fine-tuning rather than first making every parameter trainable. Added coverage showing that an unrelated model-owned parameter initialized with requires_grad=False remains frozen through parameter replacement and wrapper construction. The PEFT LoRA-only baseline remains unchanged.
https://github.com/NVIDIA-NeMo/Automodel/blob/dcea089731f8b9f4c238175b9ab4a404fad7045a/tests/unit_tests/_transformers/test_infrastructure.py#L350-L368

2. **Distinguish empty selector fields from legacy configuration**
freeze_modules and unfreeze_modules now preserve whether the field was declared.
  • Omitted fields retain legacy behavior, including the implicit vision freeze.
  • An explicitly declared empty list opts into selector semantics while selecting nothing.
  • null, false, and numeric values are rejected.
  

3. **Add real FSDP2 lifecycle coverage**
Added a real two-rank NCCL/FSDP2 test using FSDP2Manager. It verifies that the selected parameter remains the only trainable parameter after DTensor replacement and wrapping, runs forward/backward and an optimizer step, compares the synchronized gradient and update with a reference, and confirms both ranks have identical final weights.
https://github.com/NVIDIA-NeMo/Automodel/blob/dcea089731f8b9f4c238175b9ab4a404fad7045a/tests/functional_tests/hf_transformer/test_freeze_config_fsdp2.py#L95-L195

yuhezhang-ai:
/ok to test [dcea089](https://github.com/NVIDIA-NeMo/Automodel/pull/3681/commits/dcea089731f8b9f4c238175b9ab4a404fad7045a)

yuhezhang-ai (approved):

PR #3685, opened 2026-08-26
perf(training): fused multi-tensor gradient-norm reduction

akoumpa:
**Draft.** The kernel is correct and fast in isolation, but the end-to-end gain is small — see *Is this worth it?* below. Opening for review of the approach rather than as a merge-ready change.

## What

Gradient clipping reduces every parameter's gradient separately, ~7 kernel launches each. On DiffusionGemma-26B-A4B (8×H100, ep=8) that is **5,434 launches and 74 ms/step** in `_clip_grad_norm_impl` alone — more launches than the entire MoE GEMM path. This replaces it with a Triton multi-tensor reduction for the 2-norm and inf-norm.

The current implementation uses a 32K-element chunk and transfers only a cumulative chunk-count table per tensor. Each Triton program finds its tensor by deterministic binary search, avoiding the Python-built per-chunk tables that dominated the original implementation.

### Transformer Engine comparison

Exact source: `472d6d41606b852b4145e6c89f61174e61a2033f`; TE path proposed in #3878 at `f301f5ee194b1bb79983e3063a01d5d689b54e82`.

| workload | TE | this Triton PR | Triton latency reduction |
| --- | ---: | ---: | ---: |
| 500 mixed-dtype matrices, 225,181,696 elements | 3.9294 ms | **3.7277 ms** | **5.1%** |
| one BF16 `(6,000,261, 256)` tensor, 1,536,066,816 elements | 7.8943 ms | **7.3179 ms** | **7.3%** |

This times the complete local norm calculation plus in-place clipping in the same H100 80GB allocation. Each result is the median of 50 synchronized repetitions after 10 warmups; initial compilation is excluded. Environment: PyTorch `2.12.0a0+0291f960b6.nv26.04.48445190`, Triton `3.6.0`, Transformer Engine `2.14.1+366798ef`.

The Triton result had zero relative error against the fp64 reference on both inputs. A separate BF16 tensor with `2**32 + 1024` elements also returned the exact reference norm, exercising 64-bit indexing beyond the TE metadata limit.

The prefix-table change also reduced measured temporary allocation from 171,008 to 56,320 bytes on the many-tensor case and from 940,544 to 377,344 bytes on the large-tensor case.

### Original benchmark

**In isolation**, 740 tensors:

| | |
| --- | ---: |
| per-tensor loop | 16.77 ms |
| fused multi-tensor | **1.13 ms** (14.8×) |

**End-to-end**, 100 steps, DiffusionGemma-26B-A4B SFT with `dispatcher=deepep`:

| | step time |
| --- | ---: |
| before (per-tensor baseline) | 10.406 s |
| after | 10.370 s |
| gain | **0.35%** |

wandb: [before](https://wandb.ai/Nemo-automodel/dllm-gradnorm/runs/y1p44k5t) · [after](https://wandb.ai/Nemo-automodel/dllm-gradnorm/runs/0mkjgy74)

### Is this worth it?

Honestly: the end-to-end win is marginal. I originally sized this change off a profile that attributed 87 ms/step to gradient clipping, but that profile was captured under tracing, which inflates launch-heavy CPU-side work. The isolated benchmark puts the true cost at ~16.8 ms, so removing it is ~0.15% of a 10.4 s step; the measured 0.35% is consistent with that.

What the change does buy, beyond the 0.35%:

* ~5,400 fewer kernel launches per step, which matters more on a CPU-bound step than this GPU-bound one;
* a **strictly more accurate** norm (fp64 throughout, versus the current storage-dtype scaled formulation);
* it scales with parameter count rather than with model FLOPs.

Reviewers may reasonably decide 300 lines and a Triton kernel is not worth 0.35%. That call is easier to make with the number stated plainly than buried.

## Numerics

Independent before/after training trajectories are not a correctness oracle for the Qwen3-MoE benchmark: the default recipe permits nondeterministic CUDA kernels, and exact repeats of the same implementation produced different full-precision gradient norms. The large separation in the original W&B graph was benchmark drift, not a Triton reduction error.

Two controlled H100 checks isolate the reducer:

* On one backward pass, native, TE, and Triton read the same gradients. TE and Triton agreed within **4.97e-8 relative** over 15 steps.
* With `torch.use_deterministic_algorithms(True)`, math SDPA, and torch model backends, independent five-step native/TE/Triton runs had **exactly identical loss and peak memory at every step**. TE and Triton gradient norms agreed within **4.86e-8 relative**. Native differed from Triton by at most **3.60e-4 relative** because the native path rounds normalized BF16 values before squaring, while Triton widens before the square.

W&B: [native](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260914/runs/gc-20260914-qwen-strict-deterministic-native-reference) · [TE](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260914/runs/gc-20260914-qwen-strict-reducer-te) · [Triton](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260914/runs/gc-20260914-qwen-strict-reducer-triton) · [same-backward shadow](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260914/runs/gc-20260914-qwen-shadow-same-backward)

Against an fp64 reference: **exact** for BF16, 2.5e-16 (fp32), 6.9e-15 (2000-tensor accumulation). The first version of this kernel reduced tiles in fp32 and was quietly ~1e-7; the standalone validation is tight enough to catch that. A CUDA unit regression now checks FP64-reference parity and bit-identical repeatability on the committed kernel.

## Correctness properties

* **Precision** — values widen **in registers** on load, so BF16 is never squared in BF16 (|g| > 256 overflows once squared). The in-tile `tl.sum` runs in fp64 too. Widening *in registers* matters as much as widening at all: upcasting whole gradients would allocate 2–8× gradient memory. Measured peak overhead: **0.12%** of gradient bytes.
* **Determinism** — cumulative chunk counts are computed host-side from sizes only, and each program uses the same binary search. There are no atomics, so the norm is bit-identical run to run.
* **Non-finite** — NaN/Inf propagate through the fp64 sum-of-squares naturally, but `tl.maximum` follows IEEE `maxNum` and **ignores NaN**, so an inf-norm over NaN gradients would have returned finite and silently defeated `error_if_nonfinite`. The kernel counts NaNs and poisons the partial. Both paths were checked for NaN/±Inf.
* **dtype safety** — tensors are bucketed by dtype; a mixed batch cannot reinterpret BF16 storage as FP32.
* **Unsupported inputs** are routed to a slice-bounded reference path rather than converted: non-CUDA, non-contiguous (where `.contiguous()` would allocate a full-size copy), and dtypes outside the supported set — notably **FP8**, whose reduction belongs in a scaled path rather than a raw square.
* **DTensor** — the caller unwraps to local and owns the cross-rank scalar reduction; the kernel never sees a DTensor.

## Validation

* Added a CPU metadata regression plus a CUDA regression against the fp64 oracle with bit-identical repeats (`2 passed` on H100).
* `35 passed, 3 skipped` for the focused training tests on a synthetic merge with current `main`.
* Exact H100 benchmark source matched the committed file hash; the `2**32 + 1024` BF16 indexing check had zero relative error.

### DIFF
--- a/nemo_automodel/components/training/triton/grad_norm.py
+++ b/nemo_automodel/components/training/triton/grad_norm.py
@@ -1,0 +1,301 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Multi-tensor gradient-norm reductions.
+
+Gradient clipping over a sharded MoE model reduces thousands of separate
+gradient tensors. Doing that with per-tensor PyTorch ops costs a handful of
+kernel launches per parameter -- on DiffusionGemma-26B-A4B that measured 5,434
+launches and 74 ms per step, more launches than the entire MoE GEMM path -- so
+the reduction is launch-bound rather than bandwidth-bound.
+
+These kernels reduce every tensor in one launch per dtype, and are explicit
+about the two things that are easy to get wrong in a norm reduction:
+
+**Precision.** Values are widened to fp32 on load, so a BF16 gradient is never
+squared in BF16 (|g| > 256 overflows BF16's range once squared, and the
+mantissa loses most of its bits well before that). Within a tile the reduction
+is a ``tl.sum`` tree reduction carried out **in fp64** (reducing the tile in
+fp32 and widening only the tile total leaves fp32-level error, ~1e-7 relative,
+which is what this kernel exists to avoid); across tiles each program
+accumulates into an fp64 scalar; across programs the fp64 partials are summed
+by PyTorch. Accumulation error therefore stays at fp64 level regardless of how
+many elements a parameter has. ``max|x|`` accumulates nothing and stays in
+fp32, where it is already exact.
+
+**Determinism.** The decomposition into chunks is computed on the host and is a
+pure function of the tensor sizes, and no atomics are used, so the result is
+bit-identical run to run for a given parameter layout. An atomic-accumulate
+version would be faster to write but would make the gradient norm -- and hence
+every subsequent optimizer step -- non-reproducible.
+
+Widening on load also makes the usual "divide by the max first, then square"
+overflow dance unnecessary, so the 2-norm needs a single pass over the
+gradients instead of two.
+"""
+
+from typing import List, Sequence
+
+import torch
+
+from nemo_automodel.shared.import_utils import null_decorator
+
+try:
+    import triton
+    import triton.language as tl
+
+    HAVE_TRITON = True
+except ImportError:  # pragma: no cover - depends on the environment
+    HAVE_TRITON = False
+    triton = None
+
+    class _TritonStub:
+        jit = staticmethod(null_decorator)
+
+        def __getattr__(self, _name):
+            raise RuntimeError("Triton is not available")
+
+    tl = _TritonStub()
+
+
+# Elements handled by one program. Large enough that launch overhead is
+# amortized, small enough that a big parameter is split across programs (which
+# also bounds the fp32-within-tile error before the fp64 accumulate).
+_CHUNK = 8192
+_BLOCK = 1024
+
+_REDUCE_SUMSQ = 0
+_REDUCE_ABSMAX = 1
+
+
+@triton.jit
+def _multi_tensor_reduce_kernel(
+    ptrs_ptr,  # int64[num_tensors]  -- data_ptr() of each tensor
+    numel_ptr,  # int64[num_tensors]
+    chunk_tensor_ptr,  # int32[num_chunks] -- which tensor this program reduces
+    chunk_start_ptr,  # int64[num_chunks] -- element offset within that tensor
+    partial_ptr,  # float64[num_chunks] -- one partial per program
+    REDUCE_OP: tl.constexpr,
+    DTYPE_ID: tl.constexpr,
+    CHUNK: tl.constexpr,
+    BLOCK: tl.constexpr,
+):
+    """Reduce one chunk of one tensor into a single fp64 partial."""
+    pid = tl.program_id(0)
+
+    tensor_idx = tl.load(chunk_tensor_ptr + pid)
+    start = tl.load(chunk_start_ptr + pid)
+    numel = tl.load(numel_ptr + tensor_idx)
+    base_addr = tl.load(ptrs_ptr + tensor_idx)
+
+    # Re-materialize the typed pointer from the raw address. Specializing on
+    # DTYPE_ID keeps this a compile-time choice.
+    if DTYPE_ID == 0:
+        base = base_addr.to(tl.pointer_type(tl.bfloat16))
+    elif DTYPE_ID == 1:
+        base = base_addr.to(tl.pointer_type(tl.float16))
+    elif DTYPE_ID == 2:
+        base = base_addr.to(tl.pointer_type(tl.float32))
+    else:
+        base = base_addr.to(tl.pointer_type(tl.float64))
+
+    end = tl.minimum(start + CHUNK, numel)
+
+    acc = tl.zeros([], dtype=tl.float64)
+    nan_count = tl.zeros([], dtype=tl.float64)
+    for tile in range(0, CHUNK, BLOCK):
+        offs = start + tile + tl.arange(0, BLOCK)
+        mask = offs < end
+        # Widen the BLOCK-sized tile *in registers* -- never a whole tensor.
+        # Squaring in the storage dtype would overflow BF16 above |g| ~ 256 and
+        # lose most of the mantissa well before that, but materializing a
+        # widened copy of the gradient would cost more memory than the
+        # reduction saves in launches.
+        # Literals, not the module constants: a @triton.jit body may only read
+        # globals declared as tl.constexpr. 0 == _REDUCE_SUMSQ, 1 == _REDUCE_ABSMAX.
+        if REDUCE_OP == 0:
+            # fp64 for the square and the in-tile tree reduction, not just for
+            # the tile total: reducing a 1024-element tile in fp32 and widening
+            # afterwards leaves fp32-level error (~1e-7 relative), which defeats
+            # the point of accumulating across tiles in fp64.
+            # NaN/Inf need no special handling here: NaN*NaN and Inf*Inf
+            # propagate through fp64 addition, so a non-finite gradient reaches
+            # the caller's error_if_nonfinite check intact.
+            x = tl.load(base + offs, mask=mask, other=0.0).to(tl.float64)
+            acc += tl.sum(x * x)
+        else:
+            # max|x| accumulates nothing, so fp32 is already exact here.
+            x = tl.load(base + offs, mask=mask, other=0.0).to(tl.float32)
+            acc = tl.maximum(acc, tl.max(tl.abs(x)).to(tl.float64))
+            # tl.maximum follows IEEE maxNum, which *ignores* NaN and returns
+            # the other operand -- so an inf-norm over NaN gradients would come
+            # back finite and silently defeat error_if_nonfinite. Count NaNs
+            # explicitly (x != x) and poison the partial below.
+            nan_count += tl.sum((x != x).to(tl.float64))
+
+    if REDUCE_OP == 1:
+        acc = tl.where(nan_count > 0.0, float("nan"), acc)
+    tl.store(partial_ptr + pid, acc)
+
+
+_DTYPE_IDS = {
+    torch.bfloat16: 0,
+    torch.float16: 1,
+    torch.float32: 2,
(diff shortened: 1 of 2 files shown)


akoumpa:
Resolved the current `main` conflict and removed the fused-grad-norm environment toggle.

- Merge commit: `f4c0f0cc7ec1e8ead339bc887958357d4ff8e2e4` (merged `main` at `ebcd9307e6c7fc73abd081e42d54ba911b312874`). The only manual conflict was `.github/workflows/cicd-main.yml`: both sides already had `timeout: 20`, so I kept `main`'s explanatory comment and value. The PR remains scoped to the same three grad-norm files relative to `main`.
- Toggle removal: `840149fd1779f1ff7d2fc213d8d5cb34eef053e7` deletes `_FUSED_GRAD_NORM_ENV`, the `os` import, and the opt-out branch. Eligible CUDA 2/inf norms now use Triton automatically; unsupported inputs retain the existing fallback.

Validation run on the final head:

```text
uv sync --frozen --group test
# synced the worktree environment to current main (including pytest-timeout)

CUDA_VISIBLE_DEVICES='' .venv/bin/python -m pytest \
  tests/unit_tests/training/test_grad_norm.py \
  tests/unit_tests/training/test_train_utils.py -q
# 35 passed, 4 skipped

PYTHONPATH=$PWD CUDA_VISIBLE_DEVICES=0 pytest \
  --confcutdir=tests/unit_tests/training \
  tests/unit_tests/training/test_grad_norm.py -q
# 2 passed (CUDA 12.8, RTX 5880 Ada; exercises Triton fp64 parity and repeatability)

PYTHONPATH=$PWD CUDA_VISIBLE_DEVICES=0,1 python -m torch.distributed.run \
  --standalone --nnodes=1 --nproc-per-node=2 \
  tests/functional_tests/llm_pretrain_and_kd/run_clip_grad_norm_correctness.py
# 11 passed, 2 skipped, 0 failed; skips require >=4 ranks

ruff format --check \
  nemo_automodel/components/training/triton/grad_norm.py \
  nemo_automodel/components/training/utils.py \
  tests/unit_tests/training/test_grad_norm.py
ruff check \
  nemo_automodel/components/training/triton/grad_norm.py \
  nemo_automodel/components/training/utils.py \
  tests/unit_tests/training/test_grad_norm.py
# passed

git diff --check
# passed
```

I also parsed `.github/workflows/cicd-main.yml` with `yaml.safe_load`; it passed and no conflict markers remain. The first pytest attempt before `uv sync` did not collect because the stale worktree environment lacked the newly required `pytest-timeout` plugin; the post-sync result is reported above.

akoumpa:
### W&B and timing / memory results

#### Controlled Qwen3-MoE 30B training (8×H100)

These five-step deterministic runs use the same recipe and differ only in the gradient-norm reducer.

| reducer | W&B | steady-state median step time¹ | peak allocated / GPU |
| --- | --- | ---: | ---: |
| native reference | [run](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260914/runs/gc-20260914-qwen-strict-deterministic-native-reference) | 3.3151 s | 38.5453 GiB |
| Transformer Engine | [run](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260914/runs/gc-20260914-qwen-strict-reducer-te) | 3.2070 s | 38.5453 GiB |
| Triton (this PR) | [run](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260914/runs/gc-20260914-qwen-strict-reducer-triton) | **3.2054 s** | **38.5453 GiB** |

¹ Median of W&B steps 2–4, calculated as `num_tokens_per_step / tps`; setup/JIT steps 0–1 are excluded. This is a short correctness/memory control, not the reducer microbenchmark. Triton is effectively tied with TE here (0.048% lower median step time) and 3.31% below native.

Loss is exactly `12.4375` at every corresponding step and peak allocation is exactly equal across all three runs. TE and Triton gradient norms agree within `4.85e-8` relative. The [same-backward shadow run](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260914/runs/gc-20260914-qwen-shadow-same-backward) compares all reducers against the exact same gradients.

#### Isolated reducer benchmark (H100 80GB)

Median of 50 synchronized repetitions after 10 warmups; compilation excluded. Exact benchmark source: `472d6d41606b852b4145e6c89f61174e61a2033f`.

| workload | TE time | Triton time | TE temporary allocation | Triton temporary allocation | Triton change |
| --- | ---: | ---: | ---: | ---: | ---: |
| 500 mixed-dtype matrices (225,181,696 elements) | 3.9294 ms | **3.7277 ms** | 171,008 B | **56,320 B** | 5.1% faster; 67.1% less temporary memory |
| one BF16 `(6,000,261, 256)` tensor (1,536,066,816 elements) | 7.8943 ms | **7.3179 ms** | 940,544 B | **377,344 B** | 7.3% faster; 59.9% less temporary memory |

For completeness, the earlier 100-step DiffusionGemma runs are [baseline](https://wandb.ai/Nemo-automodel/dllm-gradnorm/runs/y1p44k5t) and [Triton](https://wandb.ai/Nemo-automodel/dllm-gradnorm/runs/0mkjgy74); both reached the same W&B peak allocation, `59.2373 GiB/GPU`. Those are independent nondeterministic trajectories, so the controlled reducer benchmark above is the TE-vs-Triton performance comparison.

akoumpa:
### Config knob + H100 TE/Triton follow-up

Pushed in `e6a86fb26fd5c6070d8898255a6ffaf22552070e`:

```yaml
clip_grad_norm:
  max_norm: 1.0
  backend: triton  # default; set to te for Transformer Engine
```

The equivalent CLI override is `--clip_grad_norm.backend=te`. An explicit `te` request is L2-only and fails if TE is unavailable, so benchmark arms cannot silently fall back to a different reducer. Focused + wider local coverage: `51 passed, 1 skipped`; the skipped case requires CUDA/TE and was exercised directly on H100 below. Exact-head CI was triggered with `/ok to test e6a86fb26fd5c6070d8898255a6ffaf22552070e`.

#### End-to-end training (EOS H100 80 GB)

The four W&B runs have the derived A/B statistics recorded in their summaries (`steps 10-29`):

| model / topology | TE mean step | Triton mean step | Triton mean change | paired median `Triton - TE` |
| --- | ---: | ---: | ---: | ---: |
| DeepSeek-V4.1-Flash, 16x8 H100 | [12.331664 s](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260915-e2e/runs/p3685-eos-ds41-te-6034591) | [12.259328 s](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260915-e2e/runs/p3685-eos-ds41-triton-6034591) | **0.5866% lower** | +39.245 ms |
| Qwen3.8-Flash-Next, 8x8 H100 | [2.217099 s](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260915-e2e/runs/p3685-eos-qwen38-te-6034590) | [2.184154 s](https://wandb.ai/Nemo-automodel/automodel-gradclip-te-triton-20260915-e2e/runs/p3685-eos-qwen38-triton-6034590) | **1.4860% lower** | -1.369 ms |

This does not reproduce TE being faster. DeepSeek's mean and paired-median directions disagree because late dynamic-shape compilation dominates several steps; Qwen's paired median differs by only 1.37 ms. I would treat both E2E pairs as tied/noise-level rather than claim a model-level speedup.

#### Direct callable benchmark and NCU

I profiled the exact callables with dummy gradients from this head:

- TE: `_local_te_l2_norm(gradients, cuda)`
- Triton: `multi_tensor_sumsq(gradients).sqrt()`

Synchronized H100 medians, compilation excluded:

| dummy input | TE | Triton | Triton latency reduction |
| --- | ---: | ---: | ---: |
| 500 mixed BF16/FP32 tensors, 225,181,696 elements | 0.941099 ms | **0.687845 ms** | **26.9%** |
| 500 BF16 tensors, 225,181,696 elements | 0.811427 ms | **0.583386 ms** | **28.1%** |
| one BF16 tensor, 1,536,066,816 elements | 1.701826 ms | **1.154544 ms** | **32.2%** |

NCU 2026.1.1 on the 500-tensor BF16 case (`job 6040361`, H100 with 132 SMs):

| NCU metric | TE | Triton |
| --- | ---: | ---: |
| main reduction launches | 11 | **1** |
| grid blocks per main launch | 10x320 + 1x300 | **7000** |
| waves / SM | 0.57-0.61 | **3.31** |
| achieved occupancy | 57.6% mean | **89.9%** |
| main reduction duration (NCU) | 233.536 us aggregate | **188.160 us** |
| all GPU kernels in callable | 37 | **5** |
| all-kernel duration (NCU) | 316.416 us | **202.944 us** |

The mechanism is launch fragmentation/underfill, not a slower Triton reduction. TE's fixed argument metadata caps a launch at [320 blocks](https://github.com/NVIDIA/TransformerEngine/blob/366798ef8a0a00d8f2c1650d11e7e623d7c33e26/transformer_engine/common/multi_tensor/multi_tensor_apply.cuh#L17-L25) and relaunches when that table fills ([launch/reset path](https://github.com/NVIDIA/TransformerEngine/blob/366798ef8a0a00d8f2c1650d11e7e623d7c33e26/transformer_engine/common/multi_tensor/multi_tensor_apply.cuh#L77-L111)). On this 132-SM H100, each 512-thread TE grid therefore supplies only about 0.6 theoretical waves and NCU measures ~58% occupancy. The Triton implementation keeps a cumulative chunk table on device and locates a tensor inside one 7,000-block launch, supplying 3.31 waves and reaching ~90% occupancy. TE also adds its cleanup kernel plus the wrapper's scalar validation/combine kernels, yielding 37 kernels versus 5.

The `.ncu-rep` files are retained on EOS under:

`/lustre/fsw/coreai_dlalgo_llm/akoumparouli/codex_remote_work/pr3685_ncu_20260916_e6a86fb26/reports/`

HuiyingLi (approved):

PR #3700, opened 2026-08-27
fix(qwen2_5_omni): place thinker prefix inside base_model.model on PEFT saves

### LINKED ISSUE #3655: qwen2_5_omni peft save puts the thinker. prefix outside base_model.model.

taran-dev4u:
Fixes NVIDIA-NeMo/Automodel#3655.

### Problem
When saving LoRA adapters for \qwen2_5_omni\, parameter keys already start with the PEFT prefix \ase_model.model.\.
\Qwen2_5OmniStateDictAdapter.to_hf\ previously prepended \	hinker.\ directly to the entire key, producing:
\\\	ext
thinker.base_model.model.model.layers.0.self_attn.q_proj.lora_A.weight
\\\
instead of the expected Hugging Face PEFT format:
\\\	ext
base_model.model.thinker.model.layers.0.self_attn.q_proj.lora_A.weight
\\\
This prevented external Hugging Face PEFT from loading saved adapters and broke bidirectional conversion in \rom_hf\.

### Solution
1. In \	o_hf\: Check for \ase_model.model.\ prefix and place \	hinker.\ inside it (\ase_model.model.thinker...\).
2. In \rom_hf\: Strip \	hinker.\ when nested inside \ase_model.model.\, preserving the PEFT prefix.
3. In \convert_single_tensor_to_hf\: Apply identical prefix nesting logic for single tensor conversions.
4. Added comprehensive unit tests for PEFT LoRA key conversion, bidirectional round-trip serialization, and single-tensor export.

### Testing
- Passed 13/13 unit tests in \	ests/unit_tests/models/qwen2_5_omni/test_qwen2_5_omni_state_dict_adapter.py\.

### DIFF
--- a/nemo_automodel/components/models/qwen2_5_omni/state_dict_adapter.py
+++ b/nemo_automodel/components/models/qwen2_5_omni/state_dict_adapter.py
@@ -25,6 +25,7 @@ logger = logging.getLogger(__name__)
 
 
 _THINKER_PREFIX = "thinker."
+_PEFT_PREFIX = "base_model.model."
 _DROP_PREFIXES = ("talker.", "token2wav.")
 
 # Keys the NeMo Thinker class deletes at __init__ time (so the HF base
@@ -71,7 +72,13 @@ class Qwen2_5OmniStateDictAdapter(StateDictAdapter):
         **kwargs,
     ) -> dict[str, Any]:
         if self._uses_thinker_prefix:
-            hf_state_dict = {_THINKER_PREFIX + k: v for k, v in state_dict.items()}
+            hf_state_dict = {}
+            for k, v in state_dict.items():
+                if k.startswith(_PEFT_PREFIX):
+                    new_k = f"{_PEFT_PREFIX}{_THINKER_PREFIX}{k[len(_PEFT_PREFIX) :]}"
+                else:
+                    new_k = _THINKER_PREFIX + k
+                hf_state_dict[new_k] = v
         else:
             hf_state_dict = dict(state_dict)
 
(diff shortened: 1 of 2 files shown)


yuhezhang-ai:
Thank you @taran-dev4u for contributing this fix and adding regression tests! We appreciate the time you put into addressing #3655.

We have now merged #3672, which covers the same PEFT prefix correction and also aligns the exported target-module metadata and filters PEFT-prefixed talker/token2wav weights on import. That resolves #3655, so we’re closing this PR as superseded to avoid duplicating the fix.

Thanks again for helping improve AutoModel’s adapter export support!

taran-dev4u:
Thanks for the update and review!

PR #3703, opened 2026-08-27
fix(fsdp): shard untied embeddings as separate units

yuhezhang-ai:
# What does this PR do ?

Shard large trainable untied input embeddings and LM heads as independent FSDP units instead of placing both gradient tensors in the root unit's reduce-scatter input.

For large-vocabulary dense models with fp32 gradient reduction, grouping both tables at the root requires one contiguous communication buffer containing both gradients. Splitting their FSDP ownership bounds the largest such allocation by one table. Tied parameters remain together to preserve aliasing, and frozen tables remain at the root because they do not allocate gradient communication buffers.

# Changelog

- Give trainable untied input and output embedding modules independent FSDP ownership before sharding the root model.
- Preserve tied-weight aliasing and skip standalone units for frozen tables.
- Follow the existing pipeline-parallel default for `reshard_after_forward`.
- Add unit coverage for untied, tied, and frozen ownership.
- Add real 1-rank and 2-rank Gloo FSDP numerical coverage against an unsharded fp32 reference.

# Validation

- Default parallelization strategy tests — 10 passed.
- Real 1-rank/2-rank FSDP test — 2 passed; per-parameter gradients, global norm, and parameters after one optimizer step match the unsharded reference.
- Linear-CE and Qwen3.5 tied-weight tests — 11 passed, 1 skipped.
- Targeted Ruff formatting/lint and `git diff --check` passed.

GPU validation is intentionally still pending while this PR is a draft:

- Peak-memory and TPS A/B measurements on the reported large-vocabulary multi-node workload.
- Standard and fused-linear-CE loss/gradient/optimizer-step parity with real CUDA FSDP collectives.
- Confirmation that the former root reduce-scatter allocation is split into the expected per-table allocations.

**Pre checks**:

- [x] Added focused unit and real multi-process tests.
- [x] No user-facing documentation change is required for this ownership fix.

# Additional Information

- Related to #2985.
- Draft until real-GPU correctness and memory validation is attached.

### DIFF
--- a/nemo_automodel/components/distributed/parallelizer.py
+++ b/nemo_automodel/components/distributed/parallelizer.py
@@ -17,6 +17,7 @@ import inspect
 import logging
 import warnings
 from abc import ABC, abstractmethod
+from collections.abc import Callable
 from contextlib import contextmanager
 from functools import lru_cache
 from types import FunctionType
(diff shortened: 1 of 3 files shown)


yuhezhang-ai:
Superseded by #3708, which uses a direct NVIDIA-NeMo/Automodel head branch. The replacement also includes the fused-linear-CE lifecycle fix and completed two-H100 CUDA/NCCL correctness, reduction-buffer, memory, and throughput validation.

PR #3708, opened 2026-08-27
fix(fsdp): shard untied embeddings as separate units

yuhezhang-ai:
# What does this PR do?

Shard trainable untied input embeddings and LM heads as independent FSDP units in the generic/default FSDP2 path, instead of placing both gradients in the root unit's reduce-scatter input.

For a vocabulary table with shape `[vocab_size, hidden_size]`, fp32 gradient reduction needs:

```
vocab_size * hidden_size * 4 bytes
```

When input and output tables are untied and owned by one FSDP unit, their gradients can be coalesced into one contiguous allocation twice that size. Splitting ownership does not change the reduction math or total communicated bytes; it bounds the largest contiguous reduction input by one table.

# Original production evidence

This change follows [@zhiqi-li's report from a 64-H100 Qwen3.8-27B training campaign](https://github.com/NVIDIA-NeMo/Automodel/issues/2985#issuecomment-5322924931), reproduced against `main@cb03cf61` on 2026-08-17.

[Qwen3.8-27B](https://huggingface.co/Qwen/Qwen3.8-27B/blob/main/config.json) has:

- `tie_word_embeddings=false`
- `vocab_size=248320`
- `hidden_size=5120`
- two trainable `248320 x 5120` vocabulary tables

Each table contains `1,271,398,400` parameters. Coalescing both gradients for fp32 reduction requires:

```
2 * 248,320 * 5,120 * 4 bytes
  = 10,171,187,200 bytes
  = 9.473 GiB
```

In the reported 64-H100 run, post-backward failed at step 8 while requesting **9.47 GiB**, with approximately **9.42 GiB** driver-visible as free. The request size exactly matched the two-table fp32 reduction input and was not associated with a special or all-padding batch.

With separate embedding/head FSDP ownership, the largest input becomes one table, **4.736 GiB**. Zhiqi's targeted rerun completed **20/20 steps**, and a longer run passed the original failure point.

The issue report does not attach the full training YAML or all topology settings beyond the 64-H100 Qwen3.8-27B campaign, so this PR does not claim an independently reproduced Qwen3.8-27B full-recipe run. The PR-specific real-model validation below uses the tracked Qwen3.5-9B VLM fine-tuning path instead.

# Scope, including EP/MoE

This PR changes the generic/default FSDP2 strategy. It applies automatically when that strategy sees trainable input/output embedding modules whose weights are physically untied. Tied and frozen tables remain in their existing ownership path.

EP/MoE is **not excluded from the memory principle**. Embeddings and the LM head are dense parameters, independent of expert routing. The reason this PR does not modify the EP/MoE implementation is that `ep_size > 1` routes through the dedicated MoE parallelizer, which already:

- restores configured embedding/head ties;
- detects physical tying;
- gives untied `embed_tokens` and `lm_head` separate FSDP ownership before the root wrap.

This PR brings the generic/default dense path to parity with that existing MoE ownership policy. MoE models that do not use the dedicated EP path may still receive this behavior when they route through the default strategy.

# Changelog

- Give trainable untied input and output embedding modules independent FSDP ownership before sharding the root model.
- Preserve tied-weight aliasing, including distinct parameters that share storage.
- Skip standalone units for frozen tables.
- Let the input embedding follow the configured/default reshard policy.
- Keep the output projection gathered through backward, matching its former root-owned lifecycle and allowing `FusedLinearCrossEntropy` to consume the mixed-precision compute weight outside `lm_head.forward()`.
- Add ownership tests and real one-/two-rank Gloo FSDP numerical coverage.

# PR validation

## What was run

No full Qwen3.8-27B AutoModel recipe was run specifically for this PR. The real-GPU PR validation used controlled CUDA/NCCL A/B runs with the PR code:

1. A real Qwen3.5-9B VLM fine-tuning recipe A/B loaded the complete 17.98 GB checkpoint and cached `mmoukouba/MedPix-VQA` dataset. It trained 9,197,093,888 parameters for four AdamW steps with fused linear CE, then completed the full validation set (341,257 label tokens). Topology was DP=2, local batch 1, global batch 2; the vision tower was frozen and the language model, including both untied vocabulary tables, was trainable.
2. A second real Qwen3.5-9B recipe A/B set `model.torch_dtype=float32`, giving FSDP fp32 resident/master parameters, bf16 forward/backward parameters, and fp32 gradient reduction. It used DP=4, local batch 1, global batch 4, trained the same 9,197,093,888 parameters for two AdamW steps with fused linear CE, and completed the same full validation set. Instrumentation recorded both the post-sharding storage dtype and the tensors delivered to fused CE.
3. A small untied causal LM (`vocab_size=257`, `hidden_size=64`, one hidden layer) compared root versus split ownership for standard CE and `FusedLinearCrossEntropy`.
4. An exact Qwen3.8 vocabulary-table-shape benchmark with two trainable `248320 x 5120` tables, one `5120 x 5120` hidden layer, fp32 gradient reduction, and 16 global tokens per step.
5. One warmup plus three measured exact-shape steps for a narrow throughput-overhead guard.

Environment: 2x and 4x NVIDIA H100 80GB, Torch 2.13.0a0+8145d630e8.nv26.06, CUDA 13.3, NCCL 2.30.5, Transformers 5.12.1.

## Correctness

Standard CE and fused linear CE both have exact root-vs-split parity after one SGD step:

| Check | Root-owned | Split units | Maximum absolute difference |
| -- | --: | --: | --: |
| Standard CE loss | 5.5674252510 | 5.5674252510 | 0.0 |
| Standard CE global grad norm | 0.8494461775 | 0.8494461775 | 0.0 |
| Fused linear CE loss | 5.5674200058 | 5.5674200058 | 0.0 |
| Fused linear CE global grad norm | 0.8494154811 | 0.8494154811 | 0.0 |
| Every parameter gradient | — | — | 0.0 |
| Parameters after optimizer step | — | — | 0.0 |

The first GPU run caught a correctness issue in the draft implementation: resharing the standalone LM head exposed its fp32 master shard to fused linear CE beside bf16 hidden states. Keeping the output projection gathered through backward fixes that lifecycle mismatch; the corrected run passes both loss paths exactly.

The real Qwen3.5-9B treatments used identical checkpoint, data order, seed, optimizer, and PR commit. The baseline disabled only the new embedding-split helper. Both completed all optimizer steps and full validation with finite metrics:

| Check | Root-owned | Split units | Absolute difference |
| -- | --: | --: | --: |
| Step-0 loss | 3.2696311 | 3.2696331 | 0.0000019 |
| Step-3 loss | 2.4933345 | 2.4904807 | 0.0028539 |
| Full-validation loss | 1.7855777 | 1.7865659 | 0.0009883 |

Changing FSDP ownership changes bf16 reduction grouping/order, so the multi-step real-model trajectories are not expected to remain bitwise identical. The focused small-model run separately establishes exact loss, gradient, global-norm, and post-update parity; the real recipe establishes stable end-to-end optimization and validation.

The fp32-master recipe directly exercises the LM-head lifecycle raised in review. Immediately after FSDP setup, both treatments exposed an fp32 sharded `lm_head.weight`; only the split treatment reported the head itself as an independent FSDP unit. At the actual fused-CE call in both treatments, the hidden states and materialized LM-head weight were regular bf16 tensors with shapes `[1, 204, 4096]` and `[248320, 4096]`, respectively. Both treatments completed backward, AdamW updates, and validation:

| Check | Root-owned | Split units | Absolute difference |
| -- | --: | --: | --: |
| Step-0 loss | 1.8865188 | 1.8866801 | 0.0001613 |
| Step-1 loss | 2.1660161 | 2.1670527 | 0.0010366 |
| Full-validation loss | 1.8164629 | 1.8129602 | 0.0035027 |

## BF16-storage Qwen3.5-9B recipe memory A/B

Qwen3.5-9B declares `tie_word_embeddings=false`, `vocab_size=248320`, and text `hidden_size=4096`. Its fp32 input-embedding or LM-head gradient is 3.789 GiB.

| Metric | Root-owned | Split units | Delta |
| -- | --: | --: | --: |
| Largest FSDP reduction input | 7.578140 GiB | 3.789062 GiB | -3.789078 GiB (-50.00%) |
| Peak allocated CUDA memory during training | 50.7437 GiB | 46.1408 GiB | -4.6028 GiB (-9.07%) |

The real baseline reduction group was `8,136,966,144` fp32 bytes with shapes `[[248320, 4096], [4096], [248320, 4096]]`; it contained both vocabulary-table gradients plus one 4096-element vector. The split run's largest group was one `248320 x 4096` gradient, `4,068,474,880` bytes. The memory figures are the recipe's per-step peak-allocator metric.

This sequential recipe A/B is not used for a throughput claim: the root treatment ran first and populated shared Triton/kernel caches before the split treatment. The controlled exact-shape harness below remains the narrow overhead guard.

## FP32-master Qwen3.5-9B recipe memory A/B

The four-H100 follow-up kept fp32 resident/model parameters while FSDP used bf16 compute and fp32 reduction. It therefore validates both the intended stable-optimizer configuration and the exact storage/compute dtype boundary discussed in review.

| Metric | Root-owned | Split units | Delta |
| -- | --: | --: | --: |
| Largest FSDP reduction input | 7.578140 GiB | 3.789062 GiB | -3.789078 GiB (-50.00%) |
| Peak allocated CUDA memory during training | 46.9546 GiB | 44.2462 GiB | -2.7084 GiB (-5.77%) |

The reduction groups had the same shapes and fp32 byte counts as in the bf16-storage recipe A/B. The smaller percentage memory change is expected because four-way sharding reduces each rank's resident fp32 parameters and AdamW state, while the unsharded reduction-input saving remains one full vocabulary table. This sequential run is also correctness/memory evidence only, not a throughput comparison.

## Exact-shape memory and overhead A/B

| Metric | Root-owned | Split units | Delta |
| -- | --: | --: | --: |
| Largest FSDP reduction input | 9.473 GiB (both tables) | 4.736 GiB (one table) | -50.0% |
| Peak allocated CUDA memory | 23.945 GiB | 21.630 GiB | -2.316 GiB (-9.67%) |
| Incremental backward peak | 14.298 GiB | 9.565 GiB | -4.733 GiB |
| Synthetic throughput | 265.82 tokens/s | 261.59 tokens/s | -1.59% |

The reduction probe observed one `10,171,187,200`-byte fp32 group before the split and two `5,085,593,600`-byte groups after it. The incremental backward saving matches one table's communication-buffer size.

The throughput result is only an overhead guard from a tiny one-layer, 16-global-token harness; it is not a full-model TPS claim.

## CPU and repository CI

- 7 focused tests passed: untied/tied/frozen ownership and real one-/two-rank Gloo FSDP loss, per-parameter gradient, global-norm, and post-step parameter parity.
- Targeted Ruff formatting/lint, Python bytecode compilation, and `git diff --check` passed.
- The full PR CI matrix passed, including L0 CPU, both L0 GPU shards, L2 Distributed, HF Transformer/finetune/VLM, DCP, PEFT, context-parallel, H100, and GB200 jobs.

**Pre checks**:

- [x] Added focused unit and real multi-process tests.
- [x] Completed real Qwen3.5-9B recipe A/Bs with both bf16 storage and fp32 master parameters, plus CUDA/NCCL correctness and exact-shape memory validation.
- [x] Completed the full repository CI matrix.
- [x] No user-facing documentation change is required for this ownership fix.

# Additional Information

- Related to #2985.
- Supersedes #3703, whose head branch was accidentally submitted from a fork.

### DIFF
--- a/nemo_automodel/components/distributed/parallelizer.py
+++ b/nemo_automodel/components/distributed/parallelizer.py
@@ -276,7 +276,7 @@ def _fully_shard_untied_input_output_embeddings(
     mesh: DeviceMesh,
     mp_policy: MixedPrecisionPolicy,
     offload_policy: OffloadPolicy | None,
-    reshard_after_forward: bool,
+    input_reshard_after_forward: bool,
     fully_shard_fn: Callable[..., nn.Module],
 ) -> None:
     """Give large trainable untied embedding tables independent FSDP buffers.
@@ -295,11 +295,11 @@ def _fully_shard_untied_input_output_embeddings(
         mp_policy: Mixed-precision policy used by the surrounding FSDP units.
         offload_policy: Optional offload policy used by the surrounding FSDP
             units.
-        reshard_after_forward: Whether each embedding unit reshards its
-            parameters after forward.
+        input_reshard_after_forward: Whether the input embedding unit reshards
+            its parameters after forward.
         fully_shard_fn: FSDP sharding callable, injectable for unit tests.
     """
-    ensure_tied_lm_head(model)
+    weights_are_tied = ensure_tied_lm_head(model)
 
     def _resolve(getter_name: str) -> nn.Module | None:
         getter = getattr(model, getter_name, None)
@@ -315,17 +315,21 @@ def _fully_shard_untied_input_output_embeddings(
     output_embeddings = _resolve("get_output_embeddings")
     input_weight = getattr(input_embeddings, "weight", None)
     output_weight = getattr(output_embeddings, "weight", None)
-    weights_are_tied = input_embeddings is not None and (
+    weights_are_physically_tied = input_embeddings is not None and (
         input_embeddings is output_embeddings or (input_weight is not None and input_weight is output_weight)
     )
-    if weights_are_tied:
+    if weights_are_tied or weights_are_physically_tied:
         logger.info("Keeping tied input/output embeddings in the root FSDP unit")
         return
 
     seen: set[int] = set()
-    for role, module in (
-        ("input embedding", input_embeddings),
-        ("output embedding", output_embeddings),
+    for role, module, module_reshard_after_forward in (
+        ("input embedding", input_embeddings, input_reshard_after_forward),
+        # The output projection is the last compute unit. Keep it gathered until
+        # backward, matching the old root-owned behavior and allowing
+        # FusedLinearCrossEntropy to consume its mixed-precision compute weight
+        # outside the module's forward.
+        ("output embedding", output_embeddings, False),
     ):
         if module is None or id(module) in seen:
             continue
@@ -336,7 +340,7 @@ def _fully_shard_untied_input_output_embeddings(
             module,
             mesh=mesh,
             mp_policy=mp_policy,
-            reshard_after_forward=reshard_after_forward,
+            reshard_after_forward=module_reshard_after_forward,
             offload_policy=offload_policy,
         )
         logger.info("Sharded %s as an independent FSDP unit", role)
@@ -522,7 +526,7 @@ class DefaultParallelizationStrategy(ParallelizationStrategy):
             ignored_multimodal_params=ignored_multimodal_params,
         )
 
-        standalone_reshard_after_forward = (
+        input_embedding_reshard_after_forward = (
             reshard_after_forward if reshard_after_forward is not None else not pp_enabled
         )
         _fully_shard_untied_input_output_embeddings(
@@ -530,7 +534,7 @@ class DefaultParallelizationStrategy(ParallelizationStrategy):
             mesh=dp_mesh,
             mp_policy=mp_policy,
             offload_policy=offload_policy,
-            reshard_after_forward=standalone_reshard_after_forward,
+            input_reshard_after_forward=input_embedding_reshard_after_forward,
             fully_shard_fn=fully_shard_fn,
         )
 
--- a/nemo_automodel/components/distributed/parallelizer.py
+++ b/nemo_automodel/components/distributed/parallelizer.py
@@ -17,6 +17,7 @@ import inspect
 import logging
 import warnings
 from abc import ABC, abstractmethod
+from collections.abc import Callable
 from contextlib import contextmanager
 from functools import lru_cache
 from types import FunctionType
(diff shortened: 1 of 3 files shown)


nemo_automodel/components/distributed/parallelizer.py
```diff
@@ -269,6 +270,82 @@ def parallelize(
...
+        # FusedLinearCrossEntropy to consume its mixed-precision compute weight
+        # outside the module's forward.
+        ("output embedding", output_embeddings, False),
+    ):
+        if module is None or id(module) in seen:
+            continue
+        seen.add(id(module))
+        if not any(param.requires_grad for param in module.parameters()):
+            continue
+        fully_shard_fn(
```
akoumpa, line 339: Hi @yuhezhang-ai , any code that accesses `lm_head.weight.dtype` would read the storage dtype not compute dtype, how can we ensure there's no issue introduced by this change?

yuhezhang-ai: I did two validation: 
1 The synthetic two-H100 harness tested the scenario you are concerned about:
- FP32 stored/master weights
- BF16 FSDP compute weights
- FP32 gradient reduction
 
2  The real Qwen3.5-9B MedPix A/B used BF16 model parameters. 

The first GPU test hit a problem with fused cross-entropy. The PR was then changed to keep the output head’s compute version available until backward finishes, and the corrected test passed.

Now I can test a real config with fp32 storage dtype to make sure.

yuhezhang-ai: Follow-up completed: I ran the real Qwen3.5-9B MedPix recipe A/B on four H100s with FP32 stored/master weights, BF16 FSDP compute, and FP32 reduction.

Instrumentation confirmed that `lm_head.weight` was FP32 after FSDP sharding but BF16 when consumed by fused CE alongside BF16 hidden states. Both variants completed two AdamW steps and full validation. The split reduced peak memory from 46.95 to 44.25 GiB and the largest reduction input from 7.58 to 3.79 GiB.

I added the results to the PR body.

akoumpa (approved):

PR #3731, opened 2026-08-28
fix(moe): Kimi-K3 correctness fixes (routing, NCCL warmup, PP, per-GPU tps)

yisongbetter:
Four independent correctness fixes found while benchmarking Kimi-K3 (2.78T
MoE) SFT on 64×GB200 NVL72 (256 GPUs, pp8 × EP32, HybridEP), each reproduced,
root-caused, and validated at that scale.

### 1. `KimiK3MoE` silently ignores `backend.fake_balanced_gate`

**Problem.** The base `MoE` class in `components/moe/layers.py` honors
`backend.fake_balanced_gate`, but `KimiK3MoE.__init__` unconditionally builds
`KimiK3Gate`, so benchmark configs that request forced load balance silently
run the learned gate instead.

**Root cause / failure mode.** With random- or zero-initialized weights the
learned gate produces near-equal scores, so `topk` returns experts
`[0..topk)` for every token and all traffic collapses onto each EP group's
first rank. Per-rank dispatch telemetry on the 64-node run: EP-group rank 0
received a token-slot sum of 524,800 (exactly top-k × microbatch tokens)
while ranks 1–31 received 0; per-device nsys showed one GPU per node running
~26× the expert GEMMs of its peers and 57.8% of GPU time spent in EP
dispatcher spin.

**Fix.** Build `FakeBalancedGate(moe_config, noise=backend.fake_gate_noise)`
when the flag is set, mirroring the base `MoE`.

**Validation.** After the fix, per-rank dispatch counts equalized to
16,384 ± 6% and throughput went from 34.4 to 248.5 tok/s/GPU
(wall-clock-verified, ~6× from this fix alone; end-to-end campaign
5.5 → 362.7 tok/s/GPU). Unit tests assert the gate type for both flag values
and the noise threading (`tests/unit_tests/models/kimi_k3/test_moe_gate.py`).

### 2. Lazy NCCL communicator init collides with the step-0 memory peak

**Problem.** NCCL creates a group's communicator at the first collective on
it. The grad-norm all-reduce first fires at the end of step 0 — the peak
memory point — and NCCL's internal buffer `cudaMalloc` can fail with
`cuda failure 2 (out of memory)` while torch reports free HBM.

**Reproduction.** 2.78T Kimi-K3, pp8, `pp_microbatch_size=2`: all last-stage
ranks (224–255) died identically inside `dist.all_reduce` during
`scale_grads_and_clip_grad_norm` at the end of step 0, on two different node
sets, with >160 GiB free per torch's allocator.

**Fix.** At `run_train_validation_loop` entry, issue one tiny fp32
all-reduce per unique mesh-dimension process group (device mesh + MoE mesh),
one on the default world group, then synchronize. Communicators are then
created while device memory is still empty. No-op when
`torch.distributed` is not initialized.

**Validation.** Unlocked `pp_microbatch_size=2` on the run above:
248.5 → 341.3 tok/s/GPU (+37%). Unit tests with a mocked
`torch.distributed` assert exactly one warmup all-reduce per unique group
plus the world group, dedup of shared groups, and the uninitialized no-op
(`tests/unit_tests/recipes/llm/test_warmup_mesh_collectives.py`).

### 3. MoE PP post-backward crashes for never-forwarded FSDP modules

**Problem.** With `FusedLinearCrossEntropy` under pipeline parallelism the
`lm_head` weight is consumed inside the loss function, so its FSDP module
never runs forward and its `FSDPCommContext` is never lazily initialized.
`_run_post_backward_hooks` in `components/moe/fsdp_mixin.py` then kills the
last stage with `AttributeError: 'FSDPCommContext' object has no attribute
'post_forward_order'`. (K3 declares FusedCE support; the gap was in the MoE
mixin.)

**Fix.** Guard the per-state `post_backward()` calls and the root final
callback against this `AttributeError`. A never-forwarded FSDP group has no
gradients to reduce, so skipping it is safe.

**Validation.** Reproduced and fixed on a 2-stage K3 pipeline probe;
correctness confirmed by comparing per-step losses against a non-fused
control run (no gradient desynchronization). At 64 nodes, FusedCE removed
~40 GiB of materialized logits from the last stage and gave +4.9%
throughput. Unit tests assert the skip and that remaining states still run
(`tests/unit_tests/moe/test_fsdp_mixin.py`).

### 4. `tps_per_gpu` inflated by the pp (and tp) factor

**Problem.** `tps` is global tokens/sec (`num_tokens_in_batch` is summed
over the DP group), but the printed `tps_per_gpu` divided it by
`dp_size * cp_size` only. Under pp8 (dp=32, 256 GPUs) the log line reported
`275.2/gpu` while wall-clock global throughput ÷ 256 GPUs was 34.4 — exactly
8×, the pp factor. Any MFU derived from the printed per-GPU number inflates
identically (a run whose true MFU was ~9.5% appeared as ~75%), which
misleads capacity planning and cross-run comparisons.

**Fix.** Divide by the recipe's world size (`self.dist_env.world_size`),
matching the divisor the built-in MFU computation already uses. Verification
recipe for any parallel shape: per-GPU tps must equal
(step tokens × steps) / wall-clock / total GPUs from raw log timestamps.

**Validation.** Unit test with dp=4, cp=2, pp=2 (world 16) asserts the
divisor is 16, not 8 (`tests/unit_tests/recipes/test_train_ft.py`).
Note: sibling recipes (`llm/kd.py`, `vlm/*`, `dllm/*`, `train_seq_cls.py`)
compute the same metric with the same dp*cp divisor; this PR fixes the LLM
SFT recipe it was validated on and can be extended to the others if
maintainers agree with the semantics.

### Test evidence

- `pytest tests/unit_tests/models/kimi_k3/ tests/unit_tests/moe/test_fsdp_mixin.py tests/unit_tests/recipes/llm/test_warmup_mesh_collectives.py tests/unit_tests/recipes/test_train_ft.py` — all pass (CPU).
- `ruff format --check` / `ruff check` clean on all touched files.

---


nemo_automodel/recipes/llm/train_ft.py
```diff
@@ -110,6 +111,41 @@
 # ---------------------------
 #  Stateless helper functions
 # ---------------------------
+def _warmup_mesh_collectives(*meshes: "DeviceMesh | None") -> None:
```
HuiyingLi: Thank you @yisongbetter ! Does this duplicate prewarm.py?

yisongbetter:
Thanks both — addressed in the latest push:

1. @HuiyingLi you're right, the warmup commit duplicated the new `prewarm.comm_groups` (which wasn't in the base we developed against — and its clip_grad_norm group enumeration is more precise than our version). **Dropped the commit**; we'll flip `prewarm: {comm_groups: true}` in our K3 benchmark configs instead. The PR is now 4 commits (3 fixes + 1 review follow-up).
2. Narrowed the FSDP post-backward guard per the review: it now matches the never-forwarded signature explicitly (missing `post_forward_order`), logs the skip at debug level, and re-raises any other `AttributeError`; added a regression test asserting unrelated errors propagate.

The force-push may need another CI approval when you get a chance. Thanks!

yisongbetter:
Correction to my previous comment: I checked the history — `prewarm.comm_groups` **was already present at the base we developed against** (768f5a3); we simply failed to find it when we root-caused the NCCL failure and wrote a redundant fix. Apologies for the earlier mis-statement, and thanks @HuiyingLi for catching the duplicate — the commit stays dropped, and our benchmark configs now use `prewarm: {comm_groups: true}`.

nemo_automodel/recipes/llm/train_ft.py
yisongbetter: You're right — it does, and on double-checking, `prewarm.comm_groups` was already present at the base we developed against (768f5a3); we failed to find it before writing a redundant fix. Its clip_grad_norm group enumeration is also more precise than ours. **Dropped the warmup commit** in the latest push (the PR is now 3 fixes + 1 review follow-up), and our K3 benchmark configs will use `prewarm: {comm_groups: true}` instead. The force-push may need another CI approval when you get a chance — thanks for catching this!

nemo_automodel/components/moe/fsdp_mixin.py
```diff
@@ -112,11 +112,32 @@ def _configure_fsdp_module(
...
+    """
     fsdp_state = fully_shard.state(fsdp_module)  # type: ignore[attr-defined]
     for state in fsdp_state._state_ctx.all_states:
         if state._fsdp_param_group:
-            state._fsdp_param_group.post_backward()
-    return fsdp_state._root_post_backward_final_callback
+            try:
+                state._fsdp_param_group.post_backward()
+            except AttributeError:
+                continue
```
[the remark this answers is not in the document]
yisongbetter: Narrowed in the latest push (`fix(moe): narrow the never-forwarded FSDP guard to the lazy-init signature`): the guard now matches the missing-`post_forward_order` signature explicitly, logs the skip at debug level, and re-raises any other `AttributeError`. Added a regression test asserting unrelated AttributeErrors propagate.

nemo_automodel/recipes/llm/train_ft.py
yisongbetter: Follow-up: pushed one more commit closing the loop — the K3 recipe itself never enables `prewarm.comm_groups` (it's opt-in and defaults to false), which is why recipe users still hit the exact NCCL failure its docstring describes. The recipe now sets `prewarm: {comm_groups: true}`.

yisongbetter:
Pushed a one-line fix for the `validate-nightly-recipes` failure: the `prewarm` section was appended after `ci`, but the validator requires `ci` to stay last — reordered (`0bd971a`).

The `L2_HF_PEFT` failure (`test_hf_peft_triton_fsdp2_checkpoint`, rank0 SIGABRT in the Triton LoRA path) looks unrelated to this PR: it passed on `a911685`, which already carried all functional changes here, and the only deltas since are the narrowed FSDP guard (strictly stricter re-raise), a recipe yaml line, and the merge from main. Happy to dig further if it reproduces on re-run.

HuiyingLi:
> Pushed a one-line fix for the `validate-nightly-recipes` failure: the `prewarm` section was appended after `ci`, but the validator requires `ci` to stay last — reordered (`0bd971a`).
> 
> The `L2_HF_PEFT` failure (`test_hf_peft_triton_fsdp2_checkpoint`, rank0 SIGABRT in the Triton LoRA path) looks unrelated to this PR: it passed on `a911685`, which already carried all functional changes here, and the only deltas since are the narrowed FSDP guard (strictly stricter re-raise), a recipe yaml line, and the merge from main. Happy to dig further if it reproduces on re-run.

Thank you! Just retriggered

HuiyingLi (approved):

PR #3732, opened 2026-08-28
perf(moe): Kimi-K3 memory and throughput optimizations

yisongbetter:
Three memory/throughput optimizations developed while benchmarking Kimi-K3
(2.78T MoE) SFT on 64×GB200 NVL72 (256 GPUs, pp8 × EP32). The first two are
memory surgery on the two fp32 upcast sites that dominate
activation-checkpointing residency for dispatched `[tokens, *]` tensors; the
third is an opt-in fusion switch. Forward numerics of the default paths are
bitwise-identical to the eager implementations.

### 1. Chunked bf16-saving weighted-SiTU autograd (`kimi_k3/model.py`)

**Problem.** `_weighted_situ` upcasts the whole `[tokens, 2*intermediate]`
expert projection to fp32 and lets autograd save the fp32 intermediates.
Under K3's block-granular activation checkpointing the backward recompute
holds these for every MoE layer of a 12-layer attention-residual block
simultaneously: 7–24 GiB per tensor at 1k–4k dispatched tokens on the 2.78T
model. This drove repeated step-1 backward OOMs at 64 nodes (184 GiB HBM).

**Fix.** `_WeightedSiTUFunction`: computes the identical fp32 math in
32k-row chunks, saves only the low-precision inputs, and recomputes fp32 per
chunk in backward with analytic gradients. Only large 2-D row-aligned
dispatch tensors are routed through it; small/irregular shapes (zero-token
dummy 1-D path, 0-row, broadcast weights) keep the eager path verbatim.

**Validation.** Forward bitwise-identical to eager (fp32 and bf16);
gradients match autograd's fp32 chain within 5e-7 (fp32, CPU); OOM memory
snapshots after deployment confirmed no fp32 SiTU intermediates remained
resident. CPU battery in
`tests/unit_tests/models/kimi_k3/test_weighted_situ.py` covers: bitwise
forward, gradient parity, multi-chunk boundaries, dispatch threshold,
saved-tensor count/dtype (exactly the two inputs), 1-D dummy path, 0-row,
3-D row-aligned, broadcast weights, `linear_beta` on/off, and
`needs_input_grad` combinations.

### 2. Chunked router-weight fp32 multiply (`moe/experts.py`)

**Problem.** With `apply_router_weight_after_down`,
`(output2.float() * permuted_probs.float()).to(dtype)` keeps full-size fp32
`[tokens, hidden]` intermediates alive for backward (upcast input, product,
and a recompute copy under AC). An OOM snapshot on the 64×GB200 run pinned
3 × 7.007 GiB fp32 `[262400, 7168]` blocks per rank to exactly this line;
removing them reclaimed ≥14 GiB per rank at 1k-token rows (scales linearly
with dispatched tokens).

**Fix.** `_RouterWeightMulFunction` + `_apply_router_weight_fp32`, applied
to both fp32 multiply sites (the `GroupedExpertsDeepEP` combine path and the
`GroupedExperts` scatter-add path with fp32 out_dtype). Same chunked design
as above. Deliberate detail: `x` is saved only when `probs.requires_grad` —
`x` is consumed only by `grad_probs`, and with `FakeBalancedGate` (constant
weights, no grad) saving it would pin one full-size `[tokens, hidden]`
tensor per MoE layer across the AC backward window for nothing
(3.5 GiB/layer at 1k rows).

**Validation.** Bitwise-identical forward; CPU gradient parity vs autograd;
saved-tensor count asserted for both `probs.requires_grad` states; chunked
and eager fallback paths, fp32-out mode, broadcast fallback, and 0-row cases
covered in `tests/unit_tests/moe/test_router_weight_mul.py`.

### 3. Optional `torch.compile` of the SiTU chunk cores (`BackendConfig.compile_situ`)

**Problem.** Post-routing-fix profiling showed the fp32 elementwise SiTU
chains as the top kernel class: ~30% of busy GPU time at ~15.9k kernel
launches/s.

**Fix.** New `BackendConfig.compile_situ: bool = False`. When set,
`KimiK3MoE` wraps the two module-level chunk cores
(`_situ_fwd_core`/`_situ_bwd_core`) with `torch.compile(dynamic=True)`, once
per process, lazily compiled at first call.

**Measured.** +13.7% throughput at 2-node scale, +6% at 256 GPUs
(341.3 → 362.7 tok/s/GPU). Notably, broader per-layer compilation gained
+47% at 2 nodes but only +0.6% at 256 GPUs (guard overhead under
launch-bound conditions); the narrow two-function scope is what survives at
scale. Compiled numerics are allclose to eager, not bitwise-identical —
hence opt-in, default False, eager path unchanged.

**Validation.** Config-level tests: field exists and defaults to False;
default backend leaves the cores untouched; enabling wraps both cores
exactly once across repeated construction
(`tests/unit_tests/models/kimi_k3/test_compile_situ.py`).

### Combined campaign context

End-to-end on 64×GB200 (with the routing/warmup fixes of the companion
correctness PR): 5.5 → 362.7 tok/s/GPU (0.2% → 10.1% MFU,
wall-clock-verified, GB200 bf16 dense peak basis). The memory patches are
what make 1k–2k×GBS4096 shapes fit at all on 184 GiB HBM.

### Test evidence

- `pytest tests/unit_tests/models/kimi_k3/ tests/unit_tests/moe/` — all pass
  (CPU; `grouped_gemm`/GPU-only tests skip as usual).
- `ruff format --check` / `ruff check` clean on all touched files.
- Not verifiable locally (CPU-only dev box): actual CUDA memory savings and
  throughput numbers above are from the 64×GB200 campaign logs/snapshots;
  the GPU paths (`gmm`, DeepEP/HybridEP) exercise the same
  `_apply_router_weight_fp32` call sites but were validated at scale, not in
  unit tests.

---


yisongbetter:
Note for reviewers: this PR and #3731 both touch `KimiK3MoE.__init__` (adjacent hunks: gate selection vs. `compile_situ` hook). Whichever merges second will need a trivial rebase — the resolution is to stack both hunks (fake_balanced_gate branch first, then the `if backend.compile_situ:` call). We validated the combined result on hardware.

examples/llm_benchmark/kimi/kimi_k3_gb200.yaml
```diff
@@ -0,0 +1,152 @@
...
+
+distributed:
+  strategy: fsdp2
+  tp_size: 1
+  cp_size: 1
+  pp_size: 8
+  ep_size: 32
+  sequence_parallel: false
+  activation_checkpointing: true
+  enable_fsdp2_prefetch: true
```
HuiyingLi, line 99: Thank you @yisongbetter ! Would you mind also adding
```reshard_after_forward: true``` under distributed section, and then measure again the mem and speed? Thank you!

File: examples/llm_benchmark/kimi/kimi_k3_gb200.yaml [hunk 2]
```diff
@@ -0,0 +1,152 @@
...
+  pipeline:
+    pp_schedule: 1f1b
+    pp_microbatch_size: 2
+    scale_grads_in_schedule: false
+    patch_inner_model: false
+    patch_causal_lm_model: false
+  moe:
+    reshard_after_forward: false
```
HuiyingLi, line 107: This one should be ignored given the top level one.

nemo_automodel/components/models/kimi_k3/model.py
```diff
@@ -973,6 +973,239 @@ def init_weights(self, buffer_device: torch.device, init_std: float) -> None:
...
+# identical fp32 math in row chunks and saves only the low-precision inputs,
+# recomputing fp32 per chunk in backward with analytic gradients. The chain is
+# elementwise per row, so the forward is bitwise-identical to the eager path.
+_SITU_CHUNK_ROWS = 32768
+# Engage the chunked path only for large dispatch tensors, where the memory
+# saving matters; below this the backward recompute is a net compute tax.
+_SITU_CHUNK_THRESHOLD = 12288
+
+
+def _situ_fwd_core(
```
HuiyingLi: would it be possible to have them in a separate file in the same dir? something like optimized_ops or situ.py for example

nemo_automodel/components/moe/experts.py
```diff
@@ -367,6 +367,131 @@ def _apply_bias(value, bias, tokens_per_expert, permuted_probs=None):
     return (flat_value + expanded_bias).view(shape).to(value.dtype)
 
 
+_RW_CHUNK_ROWS = 8192
+# Engage the chunked path only for large dispatch tensors, where the memory
+# saving matters; below this the backward recompute is a net compute tax.
+_RW_CHUNK_THRESHOLD = 12288
+
+
+class _RouterWeightMulFunction(torch.autograd.Function):
```
HuiyingLi: same here

nemo_automodel/components/models/kimi_k3/model.py
```diff
@@ -1187,6 +1434,33 @@ def init_weights(self, buffer_device: torch.device, init_std: float = 0.02) -> N
...
+    """fp32 attention-residual mixing chain.
+
+    The weighted combine is multiply+sum rather than
+    ``torch.matmul(prob[T, 1, B], values[T, B, H])``: cuBLAS dispatches that
+    degenerate batched-GEMM shape to non-tensor-core fp32 kernels
+    (magma_sgemmEx / gemv2 — 4.3% of busy GPU time on a 256×GB200 Kimi-K3
+    profile). multiply+sum runs the same fp32 math on the elementwise/reduce
+    path (identical up to fp32 accumulation order, which is below bf16
+    resolution for typical shapes) and fuses cleanly under ``torch.compile``
+    when ``BackendConfig.compile_situ`` is set.
```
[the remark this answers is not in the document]
yisongbetter: Added — see `4fcf2fb`, which also moves the core into `kimi_k3/situ.py` per the refactor request below; the full `Args`/`Returns` section is there.

examples/llm_benchmark/kimi/kimi_k3_gb200.yaml
yisongbetter: Will do — queued for our next 64-node window; I'll post memory + throughput deltas here. One expectation to set: under pp8 with 64 microbatches/step, upstream itself warns about this combination (`distributed/parallelizer.py` logs "reshard_after_forward=True overrides the pipeline-parallel default of keeping layer weights gathered across microbatches — may increase per-microbatch all-gathers and reduce throughput"), so we expect memory ↓ but throughput ↓ too. Numbers coming either way.

examples/llm_benchmark/kimi/kimi_k3_gb200.yaml
yisongbetter: Small clarification — the two keys drive different wrapping paths rather than shadowing each other: `distributed.reshard_after_forward` feeds the dense/backbone FSDP2 path (`distributed/fsdp2.py:126→189`), while `distributed.moe.reshard_after_forward` is what the MoE parallelizer applies to expert modules (`moe/parallelizer.py:663→812`). For K3 the experts are wrapped via the MoE path, so this line is live rather than ignored. I'll keep both explicit and measure the combination you asked for in the same run.

nemo_automodel/components/models/kimi_k3/model.py
yisongbetter: Done in `4fcf2fb` — SiTU cores + the attn-res chain now live in `nemo_automodel/components/models/kimi_k3/situ.py`. Pure move + imports, no functional change; tests updated and passing.

nemo_automodel/components/moe/experts.py
yisongbetter: Done in `4fcf2fb` — the router-weight Function and its chunk constants now live in `nemo_automodel/components/moe/optimized_ops.py` (your `optimized_ops` suggestion). Tests updated and passing.

examples/llm_benchmark/kimi/kimi_k3_gb200.yaml
yisongbetter: Measured — same 64×GB200 job, back-to-back rungs, no profiler attached, config exactly as in this PR (only delta between rungs is the one line):

| | tok/s/GPU | rank-0 peak mem |
|---|---|---|
| `reshard_after_forward: false` (as in this PR) | 458.2 | 134.9 GiB |
| `reshard_after_forward: true` | 438.1 (**−4.4%**) | 134.3 GiB (−0.6 GiB) |

The result matches the upstream warning in `distributed/parallelizer.py`: under pp8 with 64 microbatches/step, `true` re-gathers layer weights per microbatch, and the memory relief is minimal at this shape because per-rank weight shards (pp8 × dp32) are small relative to activations and pipeline buffers.

Suggestion: keep `false` in this benchmark config (it exists to demonstrate peak throughput); real-training recipes can opt into `true` where memory is the binding constraint. Happy to add a comment line in the yaml documenting the measured trade-off.

examples/llm_benchmark/kimi/kimi_k3_gb200.yaml
HuiyingLi: Thank you so much for testing!

HuiyingLi (approved):
Thank you so much!

PR #3734, opened 2026-08-28
feat(attention): add FA4 backend, enable it by default, fix VLM flash packing

akoumpa:
> **Stacked on #3733** (FA4 pin bump). Base is `akoumparouli/bump-fa4-pin`; retarget to `main` once #3733 lands.

Motivated by a Qwen3.5-VL 4B packed full-finetune on B300 that ran through SDPA's Cutlass mem-efficient kernels instead of FlashAttention.

## Native `backend.attn="fa4"`

- Adds `flash_attn.cute.flash_attn_func` and `flash_attn_varlen_func` dispatch for custom Automodel implementations.
- Preserves native BSHD/THD layouts and translates Automodel's window convention to FA4's `None` convention.
- Uses `safe_import()` and reports incomplete CUDA/CuTe installations as actionable `ImportError`s.
- Maps `model.attn_implementation=flash_attention_4` to native `backend.attn=fa4` before custom-model construction. FA2/FA3 remain unchanged because Automodel has no native FA2/FA3 backend.

## Packed BSHD adapter

- LLM and VLM neat-packed collaters recognize the native `fa4` spelling and compute unpadding indices, `cu_seqlens`, and `max_seqlen` once per batch.
- FA4 preprocessing gathers padded BSHD Q/K/V into the true varlen token stream, calls the varlen kernel, then differentiably scatters outputs back to the original BSHD layout.
- Indexed document boundaries and padding are preserved; no dense `[B, 1, S, S]` mask reaches FA4 and no token can attend across packed documents.
- Qwen3.5 linear-attention and MTP paths retain their existing packed-boundary metadata without receiving FA4-only runtime arguments.

## Default image dependencies

- Keeps `INSTALL_FA4=true` while preserving TileLang and its `apache-tvm-ffi<=0.1.11` runtime.
- Installs only the FA4 dependencies imported at the pinned source ref instead of forcing its broader declared dependency set.
- Aligns `nvidia-cutlass-dsl` on 4.6.2 and `quack-kernels` on 0.6.4 in both CUDA and PyTorch lock variants. The existing `ffpa-attn` 0.2.2 declaration still pins older versions, so the resolver override is documented and should be removed when FFPA publishes a compatible release.

## Testing

- CPU regression spans the production VLM collater → native FA4 preprocessing → varlen kernel adapter → restored BSHD output boundary.
- Compares outputs and Q/K/V gradients against independent per-document PyTorch SDPA.
- Adds a Blackwell functional test using the real `flash_attn.cute` kernel with BF16 forward/backward parity.
- Focused local results: `105 passed, 8 skipped` for attention/packing and `27 passed` for Qwen3.5 backbone/MTP; Ruff and `uv lock --check` pass.

Live validation:

- GitHub copied branch/CI is running on exact head `cd701aaa8a6f05ed727ce373cfa5d408581b8940`.
- BIA/B300 release-style `qwen3_5_4b_neat_packing`: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65016479


akoumpa:
GB200 production-path validation update (4 GPUs, CUDA 13.3, Qwen3.5-VL 4B, MedPix, neat packing 16K, GBS 16):

- Existing AutoModel SDPA: median 10,490 packed tokens/s global (2,623/GPU), peak 138.38 GiB.
- Accelerate FSDP2: median 11,082 packed tokens/s global (2,770/GPU), peak 120.80 GiB.
- Accelerate is 5.6% ahead before the FA4-native changes in this PR.

Matched three-step profiler:

- Both existing paths launch the reported `fmha_cutlass{F,B}_..._sm80` kernels for the native text full-attention layers. They consume 18.9% of self CUDA time in Accelerate and 22.9% in AutoModel.
- Vision uses SM100 cuDNN SDPA in the controlled isolation, under 1.1% of self CUDA time.
- AutoModel issues more FSDP collectives (483 all-gathers / 246 reduce-scatters vs 342 / 171) and uses FP32 gradient reduction by recipe default, while the Accelerate baseline uses BF16 reduction. A BF16-reduction isolation is running now.

Profiler evidence: Accelerate https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/415525887 ; AutoModel https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/415525888

TE-THD VLM packing is separately blocked at step 0: the inherited Qwen3.5 vision attention receives a visual token tensor whose sequence dimension does not match the `image_grid_thw` split lengths (`split_with_sizes`). Evidence: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/415509114

akoumpa:
BF16-reduction follow-up is complete: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/415542777

On the identical data steps 2-7, switching only AutoModel FSDP reduction from FP32 to BF16 changed median throughput from 8,717.7 -> 8,811.4 packed tokens/s (+1.1%) and peak memory from 138.30 -> 128.46 GiB (-9.84 GiB). So the recipe precision mismatch is worth fixing/matching for memory, but it explains only about one percentage point of the 5.6% steady-state gap. The higher FSDP unit/collective launch counts and scheduling/overlap remain the larger structural difference.

nemo_automodel/components/models/qwen3_next/model.py
```diff
@@ -269,6 +269,7 @@ def init_weights(self, buffer_device: torch.device | None = None) -> None:
 
 
 class Qwen3NextForCausalLM(HFCheckpointingMixin, nn.Module, MoEFSDPSyncMixin):
+    _uses_native_fa4 = True
```
yuhezhang-ai, line 272: **Qwen3-Next packed document isolation.** Enabling native FA4 packing sends indexed document masks into linear attention, which multiplies hidden states by document IDs and ignores document boundaries in its recurrent state. Changing only document 1 changed document 2’s output in a CPU reference probe (maximum absolute difference 0.2693). Boolean conversion alone is insufficient; the recurrent path needs boundary handling, or this packing combination should be rejected.

nemo_automodel/components/models/qwen3_5_moe/cp_linear_attn.py
```diff
@@ -189,14 +189,12 @@ def _forward_no_cp(
...
-            if packed_metadata is not None:
-                cu_seqlens = packed_metadata.cu_seqlens
-                cu_seqlens_cpu = packed_metadata.cu_seqlens_cpu
-                indices = packed_metadata.indices
+        if (
+            not use_precomputed_states
+            and is_indexed_packed_mask(attention_mask)
+            and (cu_seqlens is None or indices is None)
+        ):
+            raise ValueError("Packed Qwen3.5 linear attention requires dataset-provided indices and cu_seqlens.")
```
yuhezhang-ai, line 197: **Qwen3.8-Flash-Next regression.** The existing `Qwen3_8_FlashNextGatedDeltaNet.forward` wrapper synthesizes zero-based document IDs and supplies `cu_seqlens` without indices. With three documents, the new parent guard raises: an actual GDN forward with `cu_seqlens=[0,3,6,10]` produces `Packed Qwen3.5 linear attention requires dataset-provided indices and cu_seqlens.` Please preserve this supported caller contract and add a regression test that executes the parent forward.

nemo_automodel/components/attention/utils.py
```diff
@@ -187,6 +316,86 @@ def preprocess_args_and_kwargs_for_attn(
...
+            raise ValueError(
+                "Packed FA4 metadata must be flattened once at model entry after microbatch splitting; "
+                f"got indices={tuple(unpad_indices.shape)} and cu_seqlens={tuple(cu_seqlens.shape)}."
+            )
+
+        if cu_seqlens is not None:
+            if max_seqlen is None:
+                raise ValueError("FA4 cu_seqlens requires max_seqlen")
+            attn_kwargs["cu_seqlens_q"] = cu_seqlens
+            attn_kwargs["cu_seqlens_kv"] = cu_seqlens
```
yuhezhang-ai, line 364: **DeepSeek V3.2 sparse masks are discarded.** FA4 preprocessing accepts `cu_seqlens` before checking for an unsupported dense mask. `DeepseekV32ForCausalLM` inherits native FA4 support from DeepSeek V3, and its MLA passes a per-query sparse mask through this path, so the mask is silently dropped when cumulative lengths are supplied. Reject unsupported masks before selecting varlen execution; sequence boundaries alone do not preserve the sparse attention pattern.

nemo_automodel/components/attention/utils.py
```diff
@@ -187,6 +316,86 @@ def preprocess_args_and_kwargs_for_attn(
...
+            None if left_window is None or left_window < 0 else left_window,
+            None if right_window is None or right_window <= 0 else right_window,
+        )
+        for opt in ("softcap", "learnable_sink"):
+            if kwargs.get(opt) is not None:
+                attn_kwargs[opt] = kwargs[opt]
+
+        cu_seqlens = kwargs.get("cu_seqlens")
+        unpad_indices = kwargs.get("packed_token_indices")
+        max_seqlen = kwargs.get("max_seqlen")
```
yuhezhang-ai, line 337: **THD integration is incomplete.** The FA4 recipe path skips conversion from `seq_lens` to cumulative boundaries: `_prepare_cp_sharder` selects THD conversion only for TE/Magi, so FA4 gets the identity sharder at CP=1. Even with converted metadata, this preprocessing ignores `cu_seqlens_padded`, selecting incorrect physical document spans when padding occurs between documents. For values `[1,2,99,3,4,99]`, cumulative real lengths `[0,2,4]`, and padded offsets `[0,3,6]`, the second FA4 span selects `[99,3]` instead of `[3,4]`. Support these layouts explicitly or reject them before training.

nemo_automodel/components/models/common/packing.py
```diff
@@ -175,38 +353,33 @@ def _model_attn_implementation(model) -> str | None:
...
-        return cfg_model.get("attn_implementation", "sdpa")
-    return "sdpa"
+    if not isinstance(model, torch.nn.Module):
+        raise TypeError(f"Expected a built torch.nn.Module, got {type(model).__name__}")
+    model = getattr(model, "module", model)
+    backend = getattr(model, "backend", None)
+    hf_implementation = _model_attn_implementation(model)
+    uses_native_fa4 = isinstance(model, NativeFA4Consumer) and model._uses_native_fa4
+    if hf_implementation == "flash_attention_4" and not uses_native_fa4:
+        return hf_implementation
```
yuhezhang-ai, line 377: **Packing can disagree with the actual backend.** For unflagged custom models, the HF `flash_attention_4` key overrides the actual backend when selecting packing. Reproduced with Qwen3-MoE running `backend.attn='sdpa'`: this resolver returns `flash_attention_4`, the collater emits indexed document IDs, and SDPA treats those IDs as a binary padding mask, allowing document 2 to attend to document 1. Select the packing contract from the attention path the model actually executes, or reject the inconsistent request.

nemo_automodel/components/models/kimi_linear/model.py
```diff
@@ -1051,12 +1076,18 @@ def forward(
...
+            or attn_kwargs.get("cu_seqlens") is not None
+        )
         packed_context = kimi_packed_context or _packed_context_from_inputs(
             inputs_embeds,
-            attention_mask=attention_mask,
+            attention_mask=packed_seq_ids if packed_seq_ids is not None else attention_mask,
             cu_seqlens=attn_kwargs.get("cu_seqlens"),
         )
-        linear_attn_mask = self._update_linear_attn_mask(attention_mask, cache_position)
+        linear_attn_mask = None if has_packed_inputs else self._update_linear_attn_mask(attention_mask, cache_position)
```
yuhezhang-ai, line 1090: **Kimi padding masks are lost.** Both Kimi Linear and Kimi K3 pass no padding mask to KDA layers when packed metadata exists. Setting `linear_attn_mask=None` bypasses the per-layer derivation of `layer_padding_mask`, and ordinary NEAT batches do not supply a separate `padding_mask`. A CPU argument-capture probe for both text-model forwards produced `attention_mask=None` and `padding_mask=None` on inputs containing padding. The downstream MoE path consequently counts padding as valid tokens. Preserve padding information independently of the attention mask.

nemo_automodel/components/models/qwen3_5/model.py
```diff
@@ -656,6 +660,8 @@ def forward(
 class Qwen3_5ForCausalLM(HFCheckpointingMixin, nn.Module):
     """Qwen3.5 dense causal LM with optional Megatron-style MTP head."""
 
+    _uses_native_fa4 = True
+    requires_packed_sequence_metadata = True
```
yuhezhang-ai, line 664: **Qwen3.5 NEAT metadata survives CP in the wrong format.** This new requirement makes the collater supply batch-major cumulative lengths for TE too. The default load-balanced sharder removes the attention mask but leaves these metadata tensors intact, including sentinel entries. The CP-enabled text backbone skips the non-CP metadata preparation, and TE preprocessing then interprets a tensor such as `[[0,2,4,-1],[0,2,4,6]]` as flat THD boundaries. Translate the metadata for this route or reject the unsupported combination; the current metadata contract is not valid for this CP consumer.

nemo_automodel/_transformers/model_init.py
```diff
@@ -1314,16 +1351,21 @@ def __init_model(
...
                 else:
-                    from nemo_automodel.components.models.common.utils import BackendConfig
-
                     kwargs["backend"] = BackendConfig(**kwargs["backend"])
+            if isinstance(kwargs.get("backend"), BackendConfig):
+                kwargs["backend"] = _resolve_custom_attention_backend(
+                    attn_implementation,
+                    kwargs["backend"],
+                    uses_native_fa4=getattr(model_cls, "_uses_native_fa4", False) is True,
+                )
```
yuhezhang-ai, line 1366: **Omitting the optional backend ignores an explicit FA4 request.** Native-backend resolution only runs when a backend object is supplied. A real `Qwen3_5ForCausalLM` constructed through `_init_model` with `attn_implementation='flash_attention_4'` and no backend retained `backend.attn='sdpa'`. Resolve the default backend for native FA4 consumers before model construction, and cover the omitted-backend case.

yuhezhang-ai (changes requested):
Rechecked at `90a74aad1`. I’m requesting changes for the eight packing and backend-selection issues detailed in the inline comments.

These findings are supported by focused CPU reproductions and source tracing. I have not independently run FA4 GPU parity or multi-rank CP validation.

PR #3739, opened 2026-08-28
feat(dataset): add mask_generation_prompt to keep prompt tokens out of loss

### LINKED ISSUE #3738: mask_reasoning_content does not mask template-inserted empty
# `mask_reasoning_content` does not mask template-inserted empty `<think></think>`

## Problem

When fine-tuning Nemotron 3.5 with:

```yaml
mask_reasoning_content: true
```

the empty `<think></think>` automatically inserted by the chat template is still included in the loss if the dataset has no explicit `reasoning_content`.

Example dataset:

```json
{
  "messages": [
    {
      "role": "user",
      "content": "Solve the following programming problem..."
    },
    {
      "role": "assistant",
      "content": "import sys\n..."
    }
  ]
}
```

After applying the Nemotron chat template, it becomes roughly:

```text
<|im_start|>assistant
<think></think>import sys...
<|im_end|>
```

However, `mask_reasoning_content` only triggers when the original assistant message contains a non-empty `reasoning_content` field.

Therefore, for code-only SFT, the effective loss is roughly:

```text
<think>      -> LOSS
</think>     -> LOSS
code         -> LOSS
```

## Why this may be a problem

With thousands of non-thinking SFT samples, training repeatedly reinforces:

```text
<think> -> </think>
```

But when `enable_thinking=True`, the inference prompt ends with:

```text
<think>
```

and the model is expected to generate reasoning before producing `</think>`.

Repeatedly supervising the immediate `<think> -> </think>` transition during non-thinking SFT may bias the fine-tuned model toward terminating reasoning immediately and potentially degrade its original thinking capability.

## Expected behavior

It would be useful to support an option that keeps the empty `<think></think>` in `input_ids`, but masks those tokens from the loss:

```text
input:
question <think> </think> code

labels:
-100    -100   -100     code
```

For example:

```yaml
mask_empty_reasoning_tags: true
```

This would preserve the Nemotron non-thinking prompt format while avoiding unnecessary supervision of immediate reasoning termination.

khazic:
Fixes #3738.

## Problem

Chat templates for hybrid-reasoning models insert an empty reasoning block into every assistant turn that carries no `reasoning_content`. With Nemotron 3.5, the reporter's example renders as:

```
<|im_start|>assistant
<think></think>import sys...
<|im_end|>
```

Both tag tokens are supervised, and `mask_reasoning_content: true` does not change that. The existing masking isolates a reasoning span by re-rendering the conversation with `reasoning_content` cleared and diffing the two token sequences, so a field that is already empty produces no diff and nothing is masked. I reproduced this with the released Nemotron 3.5 tokenizer: with the flag on and off, the loss mask is identical and `<think>` (id 12) and `</think>` (id 13) are both supervised.

This is not specific to Nemotron. The same pattern appears in Qwen3 (`<think>\n\n</think>\n\n`), GLM-4.5 and GLM-4.6 (`<think></think>`), DeepSeek-V3.1 (`</think>`) and Nemotron Nano v2 (`<think>\n`). In every case the tokens are supplied by the generation prompt at inference, so the model never generates them, and training on them only teaches template boilerplate.

## Fix

This PR adds an opt-in `mask_generation_prompt` to `ChatDataset`, the agent chat dataset, `format_chat_template`, and the spec-decode data path (`build_eagle3_dataloader`, the EAGLE/DFlash/DSpark/ViSpec recipes, the DSpark precompute CLI and cache manifest), mirroring how `mask_reasoning_content` is threaded. It excludes from the loss the tokens of each assistant turn that the chat template's generation prompt supplies: the role header plus any reasoning block the template inserts. `input_ids` are untouched, so the prompt format the model sees is unchanged.

Rather than matching tag strings, the span is detected from the template itself. For each assistant turn, `messages[:idx]` is rendered with `add_generation_prompt=True` in both thinking modes (`enable_thinking` True and False), and the tokens each prompt adds are matched against the start of that turn in the full render. The mode whose prompt reproduces the turn best wins (a complete match first, then the longer one). Rendering both modes rather than deriving the mode from `reasoning_content` matters because the data alone cannot tell the modes apart: a thinking turn may carry its trace inline in `content` (the common distilled-data format), and a template may ignore `enable_thinking` entirely.

The match is anchored at the prompt's own first token when it equals the turn's first token, and otherwise at the last occurrence of that token inside the prompt, since a template appends its generation prompt at the very end of the render and anything earlier belongs to the conversation prefix. This handles templates whose added tokens start before the assistant header (GLM appends `/nothink` to the preceding user turn, SmolLM3 rewrites the system block) and templates whose prompt runs past what the turn contains (SmolLM3's non-thinking prompt ends with an empty reasoning block that its training render omits). A fallback anchor whose remainder contains an end-of-turn token is rejected, because a generation prompt never closes a turn, so rendered history can never be mistaken for the prompt and assistant content is never touched.

Truncation is handled the same way `_build_reasoning_mask` does it: spans are located in the untruncated render and mapped back through the retained window, which must be a contiguous prefix or suffix. Turns whose prefix render cannot be aligned with the full render are left untouched with a once-per-process warning. The prefix renders that `_build_multiturn_assistant_mask` already produces are shared with the new builder through a cache, so the option costs at most two extra `apply_chat_template` renders per assistant turn.

The option is off by default, and it is orthogonal to `mask_reasoning_content`: that one removes rendered reasoning text from the loss, this one removes what the template emits around it. The DSpark cache manifest records the flag as part of the run identity; manifests written before the field existed compare as `False`, so existing caches stay reusable.

## Chat templates verified

Verified against the released chat templates and tokenizers of 17 checkpoints, rendering a single-turn non-thinking conversation, a turn carrying `reasoning_content`, and a multi-turn conversation for each. For every one of them the checks were that `input_ids` are unchanged, that the flag only ever removes supervision, that each removal is a prefix of a supervised run (so content cannot be eaten), and that the assistant content stays supervised.

Templates where the flag removes template-supplied tokens:

| Checkpoint | Tokens removed from the loss (turn without reasoning) |
| --- | --- |
| NVIDIA-Nemotron-3.5-Lightning-30B-A3B | `<\|im_start\|>assistant\n<think></think>` |
| NVIDIA-Nemotron-Nano-9B-v2 | `<SPECIAL_11>Assistant\n<think>\n` |
| Qwen3-4B | `<\|im_start\|>assistant\n<think>\n\n</think>\n\n` |
| GLM-4.5-Air, GLM-4.6 | `<\|assistant\|>\n<think></think>` |
| DeepSeek-V3.1 | `<｜Assistant｜></think>` |
| MiniMax-M2 | assistant header, plus `<think>\n` on reasoning turns |
| gpt-oss-20b | `<\|start\|>assistant` (the `<\|channel\|>final<\|message\|>` the model itself emits stays supervised) |
| SmolLM3-3B | `<\|im_start\|>assistant\n` |
| Seed-OSS-36B-Instruct | `<seed:bos>assistant\n` |
| Qwen2.5-7B-Instruct (non-thinking control) | `<\|im_start\|>assistant\n` |

Templates where the flag is correctly a no-op:

| Checkpoint | Why nothing is removed |
| --- | --- |
| Kimi-K2-Thinking | The generation prompt stops at the role marker, so the model generates `<think></think>` itself and it stays supervised |
| Kimi-K2-Instruct | Same, and the template inserts no reasoning block |
| DeepSeek-R1-0528, ERNIE-4.5-21B-A3B-Thinking, Hunyuan-A13B-Instruct | The template emits the header (and, for ERNIE, the opening `<think>`) as part of the preceding turn, so it was never inside the supervised span |
| Phi-4-reasoning | Its `{% generation %}` block already excludes the header |

Kimi-K2-Thinking is the case that motivates detecting this from the template instead of pattern-matching `<think></think>`: it inserts the same empty block, but its generation prompt does not supply it, so the model does have to produce it and supervising it is correct.

For reasoning turns the flag masks the opening prompt only, so `</think>` and the reasoning text stay supervised unless `mask_reasoning_content` is also set. Combining both leaves exactly the tokens the model must produce.

Also checked on the Nemotron 3.5 tokenizer: a tool-call turn followed by a tool response (header and empty block removed, `<tool_call>` body and the final answer kept), right truncation inside the turn and inside the boilerplate block, `max_length` padding on both sides, and agreement between the shifted `labels` and the unshifted `loss_mask`.

## Testing

- 18 new unit tests in `tests/unit_tests/datasets/llm/test_tokenizer_apply_functions.py` covering the empty block, reasoning turns, inline reasoning in `content` (thinking mode selected, `<think>` opener kept), multi-turn, `train_on_last_turn_only`, `{% generation %}` templates, prompts that extend before and past the turn, plain-text headers, left truncation and a non-contiguous truncation window, unaligned prefixes, a leading assistant turn, prefix-render reuse, and the anchoring rule itself. Two of them pin the safety property: a prompt render that contains an earlier assistant turn verbatim must never anchor on that copy.
- Forwarding tests for `ChatDataset`, `ChatDatasetConfig`, `_format_example`, `AgentChatConfig`, `build_eagle3_dataloader`, the EAGLE-3 recipe, the DSpark distributed precompute (dataloader and manifest), the precompute CLI flag, and legacy manifests without the field.
- The affected unit-test suites (`datasets/llm`, `speculative`, `recipes/llm`) pass with the same results as the base commit (the failures left are pre-existing on `main` and unrelated to formatting).
- `ruff format` and `ruff check` clean on the changed files.

This is a data-formatting change verified at the chat-template and tokenizer level. It does not change any default behavior, so no training run was needed to validate it.

## Docs

`mask_generation_prompt` is documented in the dataset overview and the agent SFT guide, including when to reach for it: fine-tuning a hybrid-reasoning model on non-thinking data.


docs/guides/llm/agent-sft.mdx
jgerh, line 86:
```suggestion
Keep the following two things aligned. `LOCAL_MODELS` ships GGUF repository ids, which vLLM supports only partially and SGLang does not serve, so replace `LOCAL_MODELS[0]` with the unquantized repository you actually serve. The generator sends that entry as the model id, so the served model name has to match it exactly. The placeholder `apiKey: "no-key"` in the provider registration is harmless, because neither server requires a key by default.
```

jgerh, line 168:
```suggestion
- **Dataset size**: Both datasets hold a few thousand traces, so a full pass is fast. SynthTraces traces are long multi-turn coding sessions, so its recipe uses `seq_length: 8192` compared to `seq_length: 4096` for `glaive_toolcall_en`.
```

jgerh, line 315:
```suggestion
Greedy output is deterministic for a given model snapshot and Transformers version. A different snapshot or library version can produce a slightly different trace from the one above.
```

jgerh, line 359:
```suggestion
**Answer-only loss mask.** The agent dataset uses `answer_only_loss_mask` to supervise only the tokens that the model must produce (assistant text and `tool_calls`). User turns and tool responses are masked to `-100`, so the model is never trained to predict the environment. This behavior is enabled by default in `make_agent_chat_dataset`.
```

jgerh, line 364:
```suggestion
`make_agent_chat_dataset` exposes four flags worth knowing for agent traces:
```

docs/guides/llm/agent-sft.mdx
```diff
@@ -366,6 +366,7 @@ tool_call_eval:
...
 - `mask_reasoning_content`: Renders assistant `reasoning_content` (thinking) into the prompt but excludes it from the loss. Defaults to `false`, so a chat template that emits `reasoning_content` inside its `{% generation %}` block trains the model to reproduce that reasoning. The SynthTraces template does exactly this; see the warning in [Train on the Generated Traces](#train-on-the-generated-traces).
 - `drop_history_reasoning_content`: Strips prior-turn thinking from the prompt entirely to match inference. This option works best with `train_on_last_turn_only=true`.
+- `mask_generation_prompt`: Excludes from the loss the tokens of each assistant turn that the chat template's generation prompt supplies at inference: the role header and any template-inserted empty reasoning block (for example the `<think></think>` Nemotron templates emit, or the `<think>\n\n</think>\n\n` Qwen3 emits, for turns without `reasoning_content`). The model never generates those tokens, so supervising them only reinforces template boilerplate. Detected by rendering the generation prompt, so it works for any template. Defaults to `false`.
```
jgerh, line 369:
```suggestion
- `mask_generation_prompt`: Excludes from the loss the tokens of each assistant turn that the chat template's generation prompt supplies at inference: the role header and any template-inserted empty reasoning block (for example, the `<think></think>` Nemotron templates emit, or the `<think>\n\n</think>\n\n` Qwen3 emits, for turns without `reasoning_content`). The model never generates those tokens, so supervising them only reinforces template boilerplate. Detected by rendering the generation prompt, so it works for any template. Defaults to `false`.
```

jgerh (commented):
Completed tech pubs review of docs/guides/llm/agent-sft.mdx and provided a few text revisions for clarity.

HuiyingLi (changes requested):
Thanks for adding this option and plumbing it through the online dataset paths. I reviewed the full diff and exercised it with real tokenizers plus an 8-GPU smoke. I am requesting changes because two paths can silently train with an incorrect loss mask.

## Blocking findings

1. **`_match_generation_prompt` can mask real assistant content when a template uses an EOT distinct from `eos_token_id`.**

   The candidate at offset 0 is accepted without any history/EOT guard, and the fallback candidate only rejects `tokenizer.eos_token_id`. Many chat templates terminate turns with a separate EOT token. If a generation render rewrites or repeats history, a historical assistant header/content can therefore be selected as the current generation prompt.

   I reproduced this through the public `format_chat_template` path with `google/gemma-4-12B-it`, whose EOS is token 1 and turn terminator is token 106. With a tool continuation and repeated assistant text, enabling the new flag changed the final real assistant span from:

   ```text
   mask_generation_prompt=False: [1, 1, 1, 1]
   mask_generation_prompt=True:  [0, 0, 0, 0]
   ```

   This contradicts the invariant in the new docstring that assistant content is never touched. Please make every anchor structurally safe, including offset 0, and fail closed when the prompt tail cannot be proven. A regression test needs an EOT that differs from EOS plus rewritten/repeated assistant history.

2. **The EAGLE-3 offline cache path silently ignores `mask_generation_prompt=True`.**

   `build_eagle3_dataloader` now accepts the flag and the online trainer forwards it, but `components/speculative/precompute_eagle3.py` does not expose the flag in its CLI, pass it to the dataloader, or record it in the manifest. The cached trainer then streams the precomputed `loss_mask` and `position_mask` directly. Consequently, `cached_target_path` together with `mask_generation_prompt: true` starts normally but trains using masks generated with the old behavior.

   If supporting the legacy offline producer is outside this PR scope, the small safe fix is to reject `cached_target_path` with `mask_generation_prompt=True`. Otherwise the producer, manifest, consumer validation, and tests need to be updated together.

## Additional correctness and documentation gaps

- Left truncation is ambiguous when the retained token window matches both the prefix and suffix of the untruncated render. The implementation always selects prefix offset 0 first. A minimal left-truncation case retained `[7, 8]` and produced `[1, 1]` instead of the expected generation-prefix mask `[0, 1]`. Please resolve the side using `tokenizer.truncation_side`.
- The docs say this works for any template and explicitly name Qwen3. Stock Qwen3 multi-turn conversations fail earlier in `_build_multiturn_assistant_mask` because rendering an earlier assistant prefix keeps an empty think block that the full history removes. I reproduced this with both `train_on_last_turn_only` settings. This limitation predates the new option, so I would narrow the claim to prefix-stable or `{% generation %}` templates instead of broadening the implementation here, and add a real-tokenizer coverage case.
- Assistant-first samples are unconditionally skipped at `idx == 0`, even for tokenizers that can render an empty-message generation prefix. That conflicts with the stated behavior of masking every assistant turn.
- `agent-sft.mdx` says there are three flags but now lists four.

## Validation performed

- All seven directly affected test files passed in both host and GPU-container environments: **251 passed**.
- The full affected directory suites collected 2240 tests: **2233 passed, 2 skipped, 5 failed**. All five failures were unchanged object-storage tests caused by the container lacking optional `boto3`; the failing test and implementation files are byte-identical to the merge base.
- Project-pinned Ruff 0.12.12 check and format check passed; `git diff --check` passed; 477 MDX files validated.
- Real-tokenizer single-turn and reasoning smokes passed for Qwen3 and NVIDIA Nemotron 3.5; adversarial coverage also included Gemma, GLM, and tool-call cases.
- On one interactive node with **8x H100 80GB**, an 8-rank NCCL smoke covering real Qwen3 formatting, mask invariants, backward, and optimizer step passed with identical rank signatures. The stock Qwen3 multi-turn case reproduced the prefix-render failure on all ranks.

The first two items need to be fixed, with regression coverage, before this is safe to merge.

khazic:
Thanks for the thorough review, and for the gemma-4 reproduction in particular. Everything below is in 86ad724.

**1. Anchors can reach assistant content when the turn terminator is not `eos_token_id`.** Reproduced on the released gemma-4-12B-it template: the assistant turn that answers a tool response has no header, so its first token is real text, and the `enable_thinking=True` prompt render rewrites the system block, so the prompt's added tokens contain the whole conversation. The last occurrence of that first token sits inside the rewritten user turn, its remainder contains `<turn|>` (106) rather than `<eos>` (1), and the match ran through the answer.

The fix makes every anchor structurally safe instead of tightening the heuristic. For each assistant turn the builder renders the turn once more with its generated text replaced by a sentinel (`content` and any reasoning field swapped for a character that differs from the original's first character, `tool_calls` dropped so the render diverges at the call marker) and takes the common prefix with the real render. That prefix is the part of the turn that does not depend on the message, i.e. the template's own header and inserted block, and the generation-prompt match is capped at it: `min(matched, bound)`. Whichever anchor wins, and whatever token closes a turn, the mask cannot reach text the model produced; if the sentinel render already differs inside the conversation prefix, nothing is marked. The EOS check in `_match_generation_prompt` stays as a scoring preference and its docstring now says so; it is no longer what keeps content out.

Regression coverage: a gemma-like stub (EOT distinct from EOS, header-less continuation after a tool response, thinking-mode render that rewrites the system block, repeated assistant text) asserts the answer stays supervised and only the tool-call turn's header is removed; a second stub whose prompt render echoes the turn verbatim from offset 0 pins the offset-0 case; the perturbation helper has its own unit test. On the real gemma-4 tokenizer the five tool-continuation variants I tried now remove `<|turn>model\n` only, and the 17-template sweep from the description produces byte-identical removed spans to the previous revision.

**2. `precompute_eagle3` ignored the flag.** The producer now takes `--mask-generation-prompt`, forwards it to `build_eagle3_dataloader` and records it in the manifest, and the cached trainer validates it against the recipe setting next to the existing `target_vocab_size` / `aux_hidden_dim` checks (`_validate_cached_eagle3_manifest`), refusing a mismatch with a message that names both sides. `mask_reasoning_content` had the same gap on this producer, so it gets the same treatment. Manifests written before the fields existed are read as `False` (the producer never had either flag, so that is what those caches were built with) through a shared `defaults` argument on `offline_cache.read_manifest`, which also replaces the DSpark-only `_IDENTITY_DEFAULTS`; existing caches stay usable and `--resume` on them still works. The DSpark cached path had the same gap for the two options its manifest already recorded, so `train_dspark` now runs the same check, and a manifest that does not record an option at all is rejected rather than assumed. Tests cover the parser flags, dataloader forwarding, manifest recording, legacy resume, and the consumer checks.

**3. Left-truncation ambiguity.** `_build_reasoning_mask` and the new builder resolve the retained window through one helper that tries `tokenizer.truncation_side` first. Tests use a render whose two-token window equals both its prefix and its suffix, for the generation prompt and for the reasoning mask. While there: `truncation=True` on a sample that fits is now treated as untruncated, so the spec-decode dataloaders (which always pass `truncation=True`) keep sharing prefix renders instead of re-rendering the conversation as a reference.

**4. Qwen3 multi-turn claim.** Narrowed in both docs: the option locates turns the same way answer-only masking does without `{% generation %}` blocks, so multi-turn needs a template that renders earlier turns stably (or a `{% generation %}` template), and stock Qwen3 is single-turn only on this path. Covered by a Qwen3-style stub that drops the reasoning block from history turns: single-turn masks as expected, multi-turn raises before the option runs, with both `train_on_last_turn_only` settings.

**5. Assistant-first samples.** A leading assistant turn is handled like any other: its prefix is the empty conversation and the generation prompt is what the template appends to it. Templates that cannot render an empty message list (the ones that index `messages[0]`) skip that turn with a once-per-process warning; a render error on a non-empty prefix still propagates. Tests cover both.

**6. Docs.** "three flags" is now four, and the tech-pubs suggestions are applied.

nemo_automodel/components/datasets/llm/formatting_utils.py
```diff
@@ -334,6 +411,204 @@ def _build_reasoning_mask(
...
+        if best[1] == 0:
+            continue
+
+        # The tokens of the turn that do not depend on the message, i.e. the template's own
+        # prefix: no anchor may mark anything past them. A render that already differs
+        # inside the conversation prefix proves nothing and marks nothing.
+        perturbed_ids = _tokenize_chat(tokenizer, prefix + [_perturbed_assistant_message(message)], **render_kwargs)
+        bound = 0
+        if perturbed_ids[:start] == reference_full_ids[:start]:
+            bound = _common_prefix_length(perturbed_ids[start:], turn)
```
HuiyingLi: **Blocking: token normalization defeats the structural safety bound.** _sentinel_text guarantees a different character, but not a different token ID: tokenizer normalization or an UNK mapping may collapse the real and perturbed text. I reproduced this through public format_chat_template by extending the new Gemma-like EOT/tool-continuation stub with a valid full-width-zero to ASCII-zero normalization. For the header-less final assistant turn, the result was flag off [1, 1, 1] versus flag on [0, 1, 1]; the first real generated token was removed from loss because the perturbed and real renders shared its token ID, making bound == 1 here. This violates the new invariant that the bound cannot reach assistant content. Please construct the bound from token-level-proven divergent perturbations and fail closed with bound = 0 when divergence at the generated-text boundary cannot be established, with a normalizing/UNK tokenizer regression.

nemo_automodel/components/datasets/llm/chat_dataset.py
```diff
@@ -345,6 +345,9 @@ class ChatDatasetConfig:
     """If set, shuffles Hub/Parquet data before applying a split slice."""
     mask_reasoning_content: bool = False
     """If ``True``, exclude rendered reasoning traces from the loss mask."""
+    mask_generation_prompt: bool = False
```
HuiyingLi: **Blocking compatibility regression:** this public dataclass is not keyword-only, so inserting the new field before existing fields changes the positional constructor ABI. An old call whose next True meant mask_history=True now silently means mask_generation_prompt=True while mask_history falls back to False, changing which tokens receive loss. AgentChatConfig has the same issue at agent_chat.py:872, where old train_on_last_turn_only and later positional values shift. Please append the new field after every pre-existing field in both dataclasses or provide an explicit migration, and add a positional-construction regression.

nemo_automodel/recipes/llm/train_dspark.py
```diff
@@ -910,6 +911,7 @@ def setup(self):
                     distributed=self.dist_env.world_size > 1,
                     shuffle_seed=recipe_cfg.get("shuffle_seed", 42),
                     mask_reasoning_content=recipe_cfg.get("mask_reasoning_content", False),
+                    mask_generation_prompt=recipe_cfg.get("mask_generation_prompt", False),
```
HuiyingLi, line 914: **Documentation gap:** this is a recipe_args.mask_generation_prompt knob, while the new docs only show dataset.mask_generation_prompt. The shipped speculative Qwen/EAGLE/DSpark examples therefore do not show how to enable the feature, and the cached path additionally requires the producer --mask-generation-prompt setting to match or the new manifest check rejects training. Please add at least one runnable speculative recipe/cache example that shows both the trainer key and producer flag.

HuiyingLi (changes requested):
Re-review of 86ad724: requesting changes for two independently reproduced correctness/compatibility blockers.

Validation performed:

- Complete 27-file diff reviewed in three independent passes.
- 413 targeted CPU unit tests passed.
- Real Qwen3 and Nemotron tokenizer checks passed: input IDs stayed identical, while only the generation-supplied header/empty-think tokens left supervision.
- Ruff 0.12.12 check and format check passed for all 25 changed Python files; git diff --check passed.
- No GPU run: the new behavior is deterministic tokenizer/list and cache-manifest processing before device transfer, with no CUDA, dtype, autograd, or collective contract for a GPU to exercise.

Please also add a durable real-tokenizer regression to the existing formatting functional test; the current core cases are all custom render stubs.

khazic:
Thanks for the second pass, both blockers are real. I will fix them the way you describe: the prefix bound will be built from token-level divergent perturbations and fail closed to zero when divergence cannot be shown, with a normalizing tokenizer regression; the new field moves after every pre-existing field in both dataclasses (and the corresponding function signatures) with a positional-construction test. I will also add a speculative recipe/cache example showing the trainer key and the producer flag, and a real-tokenizer regression in the functional formatting test.

khazic:
Both blockers and the two follow-ups are in f5ec11a. The bound is now taken over two sentinel renders that must diverge from each other at the bound (token ids, not characters) and it fails closed to zero when they do not, with a normalizing Gemma-like stub, an UNK stub, and a real Qwen3 tokenizer case in the functional formatting test as regressions. The new field sits after every pre-existing field in ChatDatasetConfig, AgentChatConfig and the agent/eagle3 signatures, with positional-construction tests. The speculative README and eagle3/qwen3_eagle3_perfectblend.yaml show the recipe_args key together with the producer flag.

nemo_automodel/components/datasets/llm/formatting_utils.py
```diff
@@ -231,14 +244,106 @@ def _masked_reasoning_message(message: Dict[str, Any]) -> Dict[str, Any]:
...
+    for sentinel in _SENTINEL_TEXTS:
+        ids = _tokenize_chat(tokenizer, prefix + [_perturbed_assistant_message(message, sentinel)], **render_kwargs)
+        if ids[: len(conversation_ids)] != conversation_ids:
+            return 0
+        tail = ids[len(conversation_ids) :]
+        tails.append(tail)
+        bound = min(bound, _common_prefix_length(tail, turn))
+    if tails[0] == tails[1]:
+        return 0
+    return min(bound, _common_prefix_length(tails[0], tails[1]))
```
HuiyingLi: **Blocking:** two sentinel renders that eventually diverge do not prove that their shared prefix is template-owned. A tokenizer may emit a content-owned prefix token for every value, for example `real = [WORD_START, answer, EOT]`, `sentinel0 = [WORD_START, 0, EOT]`, and `sentinel1 = [WORD_START, 1, EOT]`. This function then returns `bound == 1` even though `WORD_START` is the first model-generated token. I reproduced this through the public `format_chat_template` path with the Gemma-like rewritten-history/tool-continuation case: the final assistant span changed from `[1, 1, 1]` with the flag off to `[0, 1, 1]` with it on. Please derive the bound from an unambiguous content boundary (and fail closed when that boundary cannot be established), with a shared-content-prefix tokenizer regression.

HuiyingLi (changes requested):
I can still reproduce one loss-mask safety issue in the current revision.

nemo_automodel/components/datasets/llm/formatting_utils.py
khazic: Huh? Are there still issues? Let me run a test.

khazic: Agreed, two diverging sentinels only locate where the text starts varying. In 5d5b1ab the bound is also capped by a render with the generated text removed entirely: a masked token must sit at the same position in that render, which has no message text, so a content-owned prefix token such as the word-start marker is never reached. Removing the text can fuse template characters into one token (Qwen3-Thinking turns '<think>\n' + '\n</think>' into a single '\n\n'), so the real token is accepted only when its string is a strict prefix of the fused one via convert_ids_to_tokens, and stays supervised otherwise. Regressions: a shared-prefix tokenizer on the Gemma-like continuation case, the fused-token case with and without token strings, and the render count. Verified on real Qwen3 Instruct and Thinking tokenizers: only the header and the '<think>\n' opener leave the loss.

khazic:
Addressed the remaining blocker in 5d5b1ab. The generation prompt bound is now anchored at the render with the generated text removed entirely, so a token the tokenizer emits for every non-empty value can no longer be masked; the two non-empty sentinels still cap it and it still fails closed to zero when they do not diverge. Fused template tokens in the empty render (Qwen3-Thinking's '\n\n') are accepted only through a strict token-string prefix check. New regressions cover the shared content-prefix tokenizer, the fused-token case with and without token strings, and the render count; real Qwen3 Instruct and Thinking tokenizers remove exactly the header and the '<think>\n' opener.

HuiyingLi:
I found a remaining correctness issue at the current head (`5d5b1ab3`).

`_is_strict_token_prefix()` compares the raw strings returned by `convert_ids_to_tokens()`. A prefix relation between tokenizer-internal strings does not imply a prefix relation between their decoded text, so the exception at `formatting_utils.py:347-350` can advance the empty-render boundary over a real generated token.

There is a real example in the Nemotron 3.5 tokenizer:

- token id 4151: internal token `iá»`, decoded as `i�`
- token id 129573: internal token `iá»ģm`, decoded as `iềm`

The helper returns true although the decoded second token does not start with the decoded first token. I can also reproduce this through the public `format_chat_template()` path using the PR's Gemma-like word-start stub: with the flag off, the final assistant span starts with `<word_start>`; with the flag on, that first real generated token is removed from the loss mask.

Could we remove this token-string prefix exception and fail closed at `empty_bound`? If the fused Qwen newline needs precise handling, it should use a provable rendered-text/offset alignment rather than tokenizer-internal token strings.

Two additional issues:

1. With `mask_generation_prompt=False`, `_render_chat_prefixes()` still retains every full prefix token list. For a 400-message stub, a 1,401-token final render retains 280,800 token references, making the temporary cache approximately O(turns × sequence length). The flag-off path should keep the previous length-only behavior.
2. The new cache example uses stock Qwen3 with the full multi-turn PerfectBlend dataset. That dataset contains 2–72-turn conversations, while stock Qwen3 without `{% generation %}` is not prefix-stable across turns. A real four-turn Qwen3 sample fails while rendering the first two messages, so full precompute will stop on the first such row. The example should filter to single-turn data or provide a prefix-stable/generation-block template.

Validation on this head: 420 changed-area unit tests and 133 adjacent packed/CP tests pass; I also checked real Qwen3/Qwen3-2507/Nemotron 3.5 tokenizers and verified that packed single-turn Qwen samples preserve the loss mask.

khazic:
Thanks, all three are addressed in 73b22bc.

**Token-string prefix exception.** Removed, together with `_is_strict_token_prefix`. The generation prompt bound is now `min(sentinel bound, common prefix of the empty render and the real turn)` on token ids only, so the boundary fails closed at `empty_bound`. For Qwen3-Thinking that means the fused `\n\n` case stops one token earlier: the real Qwen3-4B-Thinking-2507 tokenizer now removes `<|im_start|>assistant\n<think>` and the `\n` after `<think>` keeps its supervision, which I think is the right trade until there is a rendered-text/offset proof for that token. Regression: a Gemma-like word-start stub whose `convert_ids_to_tokens` reports the Nemotron 3.5 pair (`<word_start>` as `iá»`, the empty render's `<eot>` as `iá»ģm`) and asserts `<word_start>` stays supervised; the merged-token test now pins the fail-closed result. The docstring explains why an internal-string prefix relation is not a proof.

**Prefix cache with the flag off.** `format_chat_template` passes a prefix cache to the multiturn builder only when `mask_generation_prompt` is on. Without one the builder keeps prefix lengths alone, as before this PR, so the temporary state is O(turns) rather than O(turns x sequence length). A test spies on the call to check the cache is `None` with the flag off and holds every prefix render with it on, and that both paths place the assistant spans identically. While there, the agent SFT `filter_overlong` probe (which only reads the length) no longer runs the generation-prompt renders.

**Cache example.** You are right that stock Qwen3 with the full PerfectBlend set cannot work. Checking with the `Qwen/Qwen3-8B` template on a four-turn sample, it fails while rendering the first two messages with the flag off as well: the template drops the think block from every assistant turn before the last user query, so this is the existing answer-only prefix-render constraint rather than something the flag adds (Qwen3-Instruct-2507 renders history stably and passes both ways). The README now states the constraint, filters PerfectBlend to single-turn rows before the producer runs, and points at a prefix-stable or `{% generation %}` template as the alternative; the yaml comment carries the same note.

Validation: the formatting, chat dataset, agent chat, eagle3 dataloader and squad unit tests pass (three `TestParquetLoading` cases fail on a hard-coded `/home/TestData` path here, identically on main), and real Qwen3-8B, Qwen3-4B-Instruct-2507 and Qwen3-4B-Thinking-2507 tokenizers keep input ids unchanged and remove exactly the header (plus `<think>` for Thinking) on single-turn samples.

HuiyingLi (changes requested):
I can still reproduce one loss-mask safety issue at the current head (`73b22bc`).

**Blocking: the token-id-only empty-render bound can still cross into model-generated content when the boundary tokens collide.** Starting from the existing `_StubTokenizerChatNoGenGemmaLikeWordStart`, make `_id_for_token("<word_start>")` and `_id_for_token("<eot>")` return the same id (as can happen through normalization or an UNK mapping). Through the public `format_chat_template()` path, the final assistant span changes from:

```text
mask_generation_prompt=False: [1, 1, 1, 1, 1, 1, 1, 1]
mask_generation_prompt=True:  [0, 1, 1, 1, 1, 1, 1, 1]
```

The first zero is the content-owned `<word_start>` token that the model must generate. At `formatting_utils.py:353`, both non-empty sentinel renders share that token and diverge immediately after it, while the empty render has `<eot>` in the same position; because the two strings map to the same token id, every common-prefix calculation includes the real content token. Comparing token ids alone therefore cannot prove ownership at this boundary. Please fail closed at this ambiguous boundary (or establish it from rendered-text offsets) and add the colliding-boundary regression.

The three issues from my previous pass are otherwise addressed: the unsafe internal-token-string exception is gone, the flag-off prefix cache is O(turns), and the PerfectBlend example now filters to a compatible single-turn dataset.

Validation on this head:

- Full affected CPU suites: **2257 passed, 15 skipped**.
- Focused changed-area tests: **421 passed**.
- Real-tokenizer formatting functional test: **5 passed**.
- Additional real-tokenizer checks passed for Qwen3-4B, Qwen3-4B-Instruct-2507, Qwen3-4B-Thinking-2507, and NVIDIA Nemotron 3.5; input ids were unchanged, masking only removed supervision, and answer tokens stayed supervised.
- Ruff 0.12.12 check/format and `git diff --check` passed.

khazic:
Thanks, reproduced and fixed in 28f5ebe.

You are right that ids alone cannot separate a closing token from a content token with the same id when only the prefix is compared. The empty render is now aligned from its end as well: it is `template prefix + closing`, each sentinel render is `template prefix + text + closing`, so their common suffix covers at least the closing, and the bound is capped at `len(empty) - closing` (`max` of the two sentinel suffixes, so any coincidental extra agreement only shortens what may be masked). On your construction the empty render `[X]` is entirely closing, the bound is 0 and the continuation turn keeps every generated token, while the tool-call turn still loses exactly its `<assistant>` header. A template with a two-token closing whose first token collides with the marker is handled the same way, since the suffix alignment consumes both.

Regression added: the Gemma-like word-start stub with `<word_start>` and `<eot>` mapped to one id, checked through `format_chat_template()`, asserting the final span stays `[1] * 8` and that the header is the only position that changes. The changed-area unit tests pass, and the real Qwen3-8B, Qwen3-4B-Instruct-2507 and Qwen3-4B-Thinking-2507 tokenizers produce the same removed spans as at 73b22bc (header only, plus `<think>` for Thinking, plus the empty think block for Qwen3-8B non-thinking turns).

HuiyingLi (changes requested):
I re-reviewed the latest head (28f5ebe). The exact [X] closing/content collision from the previous review is fixed, but the token-only suffix bound still does not establish token ownership in the general case.

A small extension of the same public format_chat_template reproduction still masks generated content. Let the empty render emit an empty-content placeholder before its closing token, and let that placeholder and the non-empty content word-start marker normalize to the same token id:

~~~text
turn:      [X, content..., E]
empty:     [X, E]
sentinel0: [X, 0, E]
sentinel1: [X, 1, E]
~~~

At formatting_utils.py:365-366, closing is 1 and len(empty) - closing is also 1, so the final bound remains 1. Through format_chat_template the final assistant span changes from:

~~~text
mask_generation_prompt=False: [1, 1, 1, 1, 1, 1, 1, 1]
mask_generation_prompt=True:  [0, 1, 1, 1, 1, 1, 1, 1]
~~~

The first zero is still the content-owned word-start token. Suffix alignment proves only which trailing tokens may be closing tokens; it cannot prove that an equal-id token before that suffix has the same source in the empty and non-empty renders.

To keep this PR small, I suggest narrowing the fallback to the directly provable append-only case instead of adding more boundary heuristics:

- Require base_ids to be an exact prefix of the full render.
- Require generation_ids to be base_ids plus an appended prompt tail.
- Mask only the common prefix of that appended tail and the rendered assistant turn.
- If the generation render rewrites the base/history, or the base terminator is replaced, fail closed and leave the turn supervised.

In code, the core condition can be as small as:

~~~python
if base_ids != reference_full_ids[: len(base_ids)]:
    continue
if generation_ids[: len(base_ids)] != base_ids:
    continue
block = generation_ids[len(base_ids) :]
matched = _common_prefix_length(block, turn)
~~~

This removes the need to infer ownership from rewritten-history anchors and sentinel token collisions. Please add the [X, E] empty-placeholder regression above.

Validation on this head: 422 focused changed-area tests and 5 real-tokenizer functional tests pass; Ruff 0.12.12 check/format and git diff --check pass. The broader adjacent suites reached 2252 passed and 16 skipped; the remaining five failures are unchanged object-storage tests caused by boto3 being unavailable in the test environment.

khazic:
Thanks, agreed on the diagnosis: token ids compared across renders of the message cannot establish which render owns a token, and every collision the bound guarded against had a successor. The fallback is now the append-only rule you suggested, in 53a4e54.

**Rule.** For each assistant turn, the plain prefix render must be an exact prefix of the full render (the trailing-terminator tolerance is gone on this path; answer-only masking keeps it), each generation prompt render must be that prefix plus an appended block, and only a block the turn opens with in full is masked (the longer of the two `enable_thinking` modes). Everything else fails closed: rewritten history, a replaced terminator, an empty block, a partly reproduced block. No token of the message is rendered or compared any more.

**One step narrower than the literal suggestion: partial matches are not accepted.** Masking the common prefix of a partly reproduced block reopens the same ownership question in a new place, and a real template shows it. gemma-4-12b-it, reasoning turn: the thinking-mode prompt rewrites the system block and is discarded, so the non-thinking prompt `<|turn>model\n<|channel>thought\n<channel|>` is the only candidate. It matches the turn 6/7 and would mask `<|channel>thought\n`, but the thinking-mode prompt ends at `<|turn>model\n`, so the model generates that opener. With complete matches only, that turn masks nothing.

**Effect on released templates** (assistant span removed from the loss, through `format_chat_template`):

| template | turn without reasoning | turn with `reasoning_content` |
|---|---|---|
| Qwen3-8B | `<\|im_start\|>assistant\n<think>\n\n</think>\n\n` | `<\|im_start\|>assistant\n` |
| Nemotron-3-Nano | `<\|im_start\|>assistant\n<think></think>` | `<\|im_start\|>assistant\n<think>\n` |
| GLM-4.7-Flash | `<\|assistant\|></think>` | `<\|assistant\|><think>` |
| DeepSeek-V3.1 | `<\|Assistant\|></think>` | `<\|Assistant\|></think>` |
| Qwen3-4B-Thinking-2507 | nothing (prompt `<think>\n` vs fused `\n\n` in the render: partial) | `<\|im_start\|>assistant\n<think>\n` |
| gemma-4-12b-it | nothing (non-thinking prompt only partly reproduced) | nothing |

The two "nothing" cells are the coverage given up for provability; both were a role header or a lone `<think>`. The Qwen3-Thinking and Nemotron reasoning turns now lose the whole `<think>\n` prompt, one token more than the old bound allowed.

**Also in this revision.** `enable_thinking` is only probed when the template text reads it, so a template without the variable costs one prompt render and `MistralCommonBackend` (which rejects unknown kwargs) no longer raises. The DSpark manifest backfills `mask_reasoning_content` for caches written before it was recorded, matching the EAGLE-3 defaults, and both distributed precompute examples show the two loss-mask keys.

**Removed.** `_SENTINEL_TEXTS`, `_perturbed_assistant_message`, `_generation_prefix_bound`, `_common_suffix_length`, `_match_generation_prompt` and the second anchor. The rule is `_appended_generation_prompt_length`, four lines.

**Tests.** The `[X, E]` empty-placeholder regression is added as requested, parametrized with the word-start and colliding-id stubs since all three now exercise one path; rewritten history, a partly reproduced prompt, a trailing EOS in the prefix render, and a template that never reads `enable_thinking` each have a fail-closed or single-render test.

nemo_automodel/recipes/llm/train_dspark.py
khazic: Covered in examples/speculative/README.md (the `precompute_eagle3 --mask-generation-prompt` producer run plus the matching `recipe_args.mask_generation_prompt: true` trainer block) and as commented keys in `eagle3/qwen3_eagle3_perfectblend.yaml`; 53a4e54 adds the same commented keys to both distributed DSpark precompute yamls and names `precompute_dspark_dist` in the manifest error.

HuiyingLi (approved):

PR #3740, opened 2026-08-28
fix(fsdp): preserve layer ownership for mixed compute dtypes

akoumpa:
## Stack and scope

This PR is the FSDP2 foundation, plus the optimizer behavior required by the resulting mixed resident-parameter layout.

The Qwen runtime optimizations found while profiling are separate dependent PRs:

- #3750: coalesce packed GatedDeltaNet metadata transfers
- #3751: remove the Qwen3.5-VL dynamic multimodal-splice synchronization

Neither optimization is part of this diff.

## Motivation

Qwen3.5 has ordinary weights that may compute in BF16 alongside precision-sensitive GatedDeltaNet parameters that must remain resident and compute in FP32.

Two resident-storage contracts must remain supported:

- FP32-resident bulk parameters with transient BF16 forward/backward weights; and
- BF16-resident bulk parameters whose separate FP32 master copy is owned by an optimizer such as Transformer Engine FusedAdam.

The previous implementation satisfied FSDP2's unit-wide mixed-precision constraint by wrapping dtype-uniform subtrees separately. That made a dtype choice into an FSDP ownership boundary. In the measured four-GPU, 16K packed profile, this produced 81 FSDP units instead of 57, including 24 `_fp32_params` child units and 576 additional tiny all-gathers across three profiled optimizer steps.

The problem is collective fragmentation, not useful communication volume. Dtype-based wrapping should therefore be a compatibility fallback rather than the normal ownership policy.

## Path selection

Selection uses the actual resident dtypes and layouts of parameters owned by each candidate FSDP unit. It does not infer storage from checkpoint metadata, compute-dtype metadata, or optimizer settings.

| Resident layout | Sensitive module size | Selected path |
|---|---:|---|
| FP32 bulk + FP32 sensitive | At most 8 MiB | Replicate the sensitive module; keep one layer-level FSDP owner for the bulk |
| FP32 bulk + FP32 sensitive | Above 8 MiB | Keep the layer sharded and use per-tensor transient compute materialization |
| BF16 bulk + FP32 sensitive | At most 8 MiB | Replicate the sensitive module; preserve BF16 sharded bulk storage |
| BF16 bulk + FP32 sensitive | Above 8 MiB | Preserve the existing dtype-split sharded fallback |
| CPU offload, compiled autograd, uneven dim-0 sharding, or unsupported layout | Any | Fall back before wrapping |

The 8 MiB limit is per managed module, not model-wide. Repository PEFT recipes use ranks 4-64, predominantly 8 or 16. For an FP32 LoRA A+B pair, `bytes = 4 * rank * (in_features + out_features)`, so 8 MiB covers rank 64 on a square 16K projection while keeping replication bounded.

## Design

- Keep the enclosing transformer layer as the FSDP owner whenever the active configuration supports it.
- Use FSDP2's per-tensor all-gather extension for transient compute materialization without creating dtype-child ownership units.
- Reuse an already-FP32 unsharded tensor directly for FP32 compute instead of recasting it to FP32.
- Exclude selected replicated parameters from every FSDP unit.
- Synchronize their gradients from FSDP's synchronizing post-backward lifecycle using one coalesced FP32 payload per DP mesh dimension. Deferred backward passes accumulate locally; the final synchronizing backward performs communication.
- Carry an error flag and rank-local use bits in that FP32 payload. A rank-local missing gradient contributes zero; a globally unused parameter retains `grad=None`, so weight decay and optimizer state do not advance.
- Split plain replicated tensors from DTensors in optimizer foreach groups.
- With TE FusedAdam, retain optimizer-owned FP32 masters for BF16 parameters while avoiding a redundant `master_param` for resident FP32 parameters. Reapply the same ownership rule after optimizer-state load.

Custom FSDP2 mechanics are self-contained under `nemo_automodel/components/distributed/fsdp2_extensions/`:

- `compute_dtype.py`: path selection and transient per-parameter compute materialization
- `replicated.py`: bounded replication and FSDP-lifecycle FP32 gradient synchronization
- `utils.py`: reusable FSDP2 sharding-policy helpers
- `compat.py`: narrowly scoped PyTorch FSDP2 compatibility hooks

`Qwen3_5ParallelizationStrategy` only declares the precision-sensitive `_fp32_params` set and supplies the FSDP materialization callback. It delegates transformer discovery, root wrapping, prefetch, pipeline resharding, activation checkpointing, frozen multimodal policies, and ignored/frozen parameter handling to `DefaultParallelizationStrategy`.

## Validation

- Core-only targeted unit suite: 408 passed, 7 skipped.
- Local two-GPU functional matrix: ten resident/compute-dtype and ownership cases, independent-reference parity, optimizer parity, accumulation, rank-local/global unused parameters, checkpoint resume, and expected profiler collectives.
- Four-GPU HSDP matrix with Transformer Engine: eleven cases, optimizer parity, and one replicated-gradient all-reduce on each applicable HSDP mesh dimension.
- Real TE GPU probe: typed and factory construction, BF16/FP32 master ownership, optimizer step, and state-dict resume.
- Changed Python files pass Ruff lint and format checks; shell entrypoints pass `bash -n`; the complete diff passes `git diff --check`.

The full three-PR source stack was also profiled successfully on four GB200s with 16K neat packing, MTP disabled, and three active profiler steps: [job 417438959](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/417438959). The history was split after that run; the only subsequent content change is this PR's validation Markdown.

Observed rank-0 communication:

| Metric | Count / type |
|---|---:|
| FSDP ownership units | 57 |
| Parameter all-gathers | 1,332 BF16 |
| Preemption-signal control all-gathers | 3 Int32, one element each |
| Gradient reduce-scatters | 171 FP32 |
| Baseline scalar all-reduces | 15 |
| Replicated-gradient all-reduces | 3 FP32, one per active optimizer step |
| `_fp32_params` child gathers | 0 |

The Int32 all-gathers are the training scheduler's once-per-step distributed preemption-signal poll; they are not parameter materialization or part of the replicated-gradient protocol.

## Performance comparison boundary

The trace confirms the intended FSDP ownership and communication structure. It is not the final matched adoption comparison: the available Accelerate control uses a different resident/master-weight contract and does not preserve the same packed linear-attention document boundaries. A final B1 comparison must match FP32 resident storage and transient compute on both sides; a B2 comparison must match BF16 resident storage and optimizer-owned FP32 masters on both sides.

The complete correctness, topology, and comparison matrix is in `fsdp2_dtype_validation_matrix.md`.


yuhezhang-ai:
@akoumpa my concern with this PR is that we intentionally support BF16 as the resident/storage dtype for the bulk parameters, not only as the compute dtype.

A user may construct the model in BF16 because, for example, a TE optimizer maintains a separate FP32 master copy, or because they intentionally want BF16 resident/master weights. In either case, the configuration should remain supported, while the numerically sensitive parameters stay in FP32.

Could we select the path based on the resident dtype? When the resident parameters are FP32, use the new optimized single-owner path. When the user requests BF16 storage for the bulk parameters, retain the existing dtype-split path.

That would preserve the existing dtype contract while still providing the optimization for FP32-resident configurations.

akoumpa:
Thanks for the feedback @yuhezhang-ai , 

> Could we select the path based on the resident dtype? When the resident parameters are FP32, use the new optimized single-owner path. When the user requests BF16 storage for the bulk parameters, retain the existing dtype-split path.

So if i understand correctly, you have in mind a case like PEFT where base weights are in BF16 and you have PEFT params in fp32 (copy) / bf16 (compute)? If yes, then we agree, i'll update here to make sure this is covered.

yuhezhang-ai:
@akoumpa Thanks. PEFT could be another mixed-storage case, but that is not the main case I had in mind.

I mean ordinary full-parameter training where:

* the model is constructed with `torch_dtype=bfloat16`, so the bulk model parameters remain resident in BF16;
* numerically sensitive parameters such as `A_log` and `dt_bias` remain resident and compute in FP32; and
* TE FusedAdam uses `master_weights: true`, maintaining its own separate FP32 master copy for the BF16 model parameters.

In this case, the FP32 copy belongs to the optimizer—it is not a PEFT parameter. We already have recipes using BF16 model storage with TE FP32 master weights, and I want to make sure this existing configuration remains supported.

For PEFT, I am not sure if BF16 base + FP32 LoRA is fully supported under FSDP2. Most tested PEFT configurations keep LoRA in BF16 to avoid mixed-dtype FSDP issues.

akoumpa:
Exact-image profiler validation is now green at head `97c897a498b970ab84b9b33838b79f18b0f0d0cf`.

The rerun exercises both compatibility fixes:
- compiled-autograd state is read from its owning `torch._dynamo.compiled_autograd` module instead of the removed private FSDP re-export;
- post-all-gather materialization runs under `torch.no_grad()` while retaining the destination parameter storage and version counter.

Validation (4 x GB200, CUDA 13.3, PyTorch 2.13 nightly, neat packing 16,384, MTP disabled, three active profiler steps):
- [Accelerate + FSDP2 job 416279904](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/416279904): success, Slurm `COMPLETED`
- [AutoModel job 416279906](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/416279906): success, Slurm `COMPLETED`
- [leaf pipeline 65103921](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65103921): success

Rank-0 trace comparison:

| Metric | Accelerate + FSDP2 | AutoModel |
|---|---:|---:|
| Mean CPU `ProfilerStep` | 14.516 s | 15.209 s |
| NCCL all-gather kernels | 1,368 | 1,335 |
| NCCL reduce-scatter kernels | 171 | 171 |
| NCCL all-reduce kernels | 15 | 15 |
| Total NCCL kernels | 1,554 | 1,521 |

Note: accelerate uses bf16 (no master copy weight) whereas AM does, thus the degradation is expected.

AutoModel has 1,332 named FSDP gathers plus three scalar control gathers. Both traces retain the same 57 FSDP ownership units, and the AutoModel trace has no `_fp32_params` child-gather paths or 16-element parameter gathers. The mixed FP32/BF16 data is packed into the existing byte collective instead of creating a dtype-specific collective boundary.

Local checks before the rerun: 105 related unit tests passed; Ruff and `git diff --check` passed.

akoumpa:
Exact-head NeMo CI profiler validation passed.

- product head: `51be112fbfdfdaad3475440881b33c6c148415e6`
- profiler-only head: `5816123c446c89de642c00b87ef09a321d36d1b1`
- [parent pipeline](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65139154) -> [child](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65140019) -> [4xGB200 leaf](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65140111)
- [Accelerate + FSDP2](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/416594743): success
- [AutoModel](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/416594744): success

Three-step rank-0 trace result: AutoModel 15.044 s versus Accelerate 15.096 s mean CPU `ProfilerStep` (-0.35%); 1,335 versus 1,368 all-gathers; 57 FSDP units in both; no dtype-child gathers. The replicated precision-sensitive path contributes exactly one 1,585-element FP32 grad all-reduce per step (6,340 bytes). The preceding AutoModel failure exposed meta-to-real `Parameter` replacement; this head resolves through stable owner-module slots and includes the regression test.

akoumpa:
Current-main exact-head profiler rerun is green after incorporating `f40e01619` (`perf(qwen3.5): reuse packed GDN metadata across layers`).

- product head: `dc9315ccbca9a6ce937c4f2e9acd79ca1832541a`
- profiler-only head: `ffb510613b423970f0d844cafc0d8c05607f09e7`
- [parent](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65160361) -> [child](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65160948) -> [4xGB200 leaf](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65160959)
- [Accelerate + FSDP2 job 416771051](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/416771051): success
- [AutoModel job 416771052](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/416771052): success

Trace-confirmed optimization: packed GDN metadata construction falls from 288 layer-local builds to 12 model-forward builds over the three active steps. That removes 276 redundant constructions, including 552 `nonzero`, 828 `item`, 6,762 D2D-copy, and 1,944 `cudaStreamSynchronize` events. NCCL is unchanged at 1,335 AG / 171 RS / 18 AR for AutoModel versus 1,368 / 171 / 15 for Accelerate; both retain 57 FSDP units and no dtype-child gather path.

Fresh CPU `ProfilerStep`: AutoModel 14.986 s mean / 14.623 s median versus Accelerate 14.444 s / 14.532 s (+3.75% mean, +0.63% median). AutoModel improves 0.39% against its prior mean, while Accelerate independently moves 4.32% between runs and the fresh arms ran on different nodes, so the short-run mean delta is within observed allocation variance rather than evidence that the metadata optimization regressed performance.

One comparison boundary remains: the Accelerate harness drops the packed document IDs for its linear-attention path, while AutoModel applies them. This run validates the optimization and FSDP communication structure, but is not yet the final semantic packed-training parity number.

akoumpa:
Final fixes 1/2/3 profiler rerun is complete at product head c530db53d1fd621946b42376ed8beebb34857ba7 (profiler harness ee78b6da630ac80cf4dee81f00017ed1afa76cf1).

Exact-head validation: [parent pipeline](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65189022) -> [child](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65190351) -> [4xGB200 leaf](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65190377) -> [AutoModel trace job](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/417040458). All succeeded; Slurm job 2679168 ran on ptyche0144 with image pipe.65189022. The unchanged Accelerate control is [job 416995337](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/416995337) from the immediately preceding identical profile setup.

| Three profiled optimizer steps | Accelerate + FSDP2 | AutoModel | Delta |
|---|---:|---:|---:|
| Mean CPU step | 14.920 s | 15.006 s | +0.57% |
| Median CPU step | 15.129 s | 14.754 s | -2.48% |
| AG / RS / AR | 1,368 / 171 / 15 | 1,335 / 171 / 18 | -33 / 0 / +3 |
| NCCL rank-0 input payload | 81.707 GB | 115.663 GB | FP32 reduction policy |

The trace confirms the intended local-overhead changes:

- stream synchronizations: 1,272 -> 1,083 (-189)
- direct-copy CUDA kernels: 16,828 -> 14,512 (-2,316)
- fix-3 accumulation conversions: 2,034 per-parameter kernels are replaced by exactly 171 group-level chunk_cat kernels (57 FSDP groups x 3 steps)
- NCCL count, dtype, and bytes are unchanged by the three fixes; parameter gathers remain BF16 and FSDP reduce-scatter remains FP32
- 253 focused tests pass with 7 skipped; the 2-GPU FSDP and 4-GPU HSDP functional matrices retain optimizer parity and expected collective counts

The remaining timing difference is below the resolution of this three-step, different-allocation comparison. The Accelerate probe also omits packed linear-attention document boundaries and reduces in BF16, so this is an FSDP/local-overhead comparison rather than final semantic parity.

PR #3743, opened 2026-08-28
fix(moe): replay HybridEP layouts under activation checkpointing

HuiyingLi:
## Why

Activation checkpointing runs each checkpointed MoE block twice: once in the training forward, then again during backward recomputation. AutoModel already saves the router projection and top-k selection, so both executions choose the same experts. That is necessary, but it is not sufficient for HybridEP.

HybridEP independently rebuilds its distributed receive/permute layout on the second execution. At scale, that rebuild is not reproducible even with bit-identical routing. An 8xH100 reproduction recorded identical top-k checksums on every rank and all 30 MoE layers, while HybridEP returned 2,791 local token rows in forward and 2,701 in recompute. The Kimi K3 32-node Expert LoRA run hit the same failure with per-rank mismatches such as 925 vs 952 and 841 vs 821.

Those row counts must match. The expert GEMM output and dispatched router probabilities are paired one row per routed token. When HybridEP changes the receive layout during recompute, ordinary Expert LoRA exposes it immediately in the weighted SwiGLU multiply; other expert implementations usually surface it later as `torch.utils.checkpoint.CheckpointError` because saved and recomputed tensor metadata differ.

This remains a main-branch bug: the 8-GPU reproduction already included the later HybridEP token-count equalization fix, and `HybridEPDispatch` has not changed between that baseline and current main.

Selective AC does not avoid the problem. HybridEP dispatch is wrapped in a custom `autograd.Function`, so selective checkpoint policies do not observe the inner `torch.ops.hybridep.dispatch` call; the apparent HybridEP save-list entry is therefore inert. Disabling AC avoids recompute but is not a viable memory configuration for K3, and switching away from HybridEP gives up the required large-scale dispatcher path.

## What changes

- record each checkpoint-forward HybridEP dispatch handle and output extent
- reuse that exact layout during activation-checkpoint recompute while redispatching the recomputed activations and router probabilities
- apply the replay contract to both full and selective activation checkpointing
- add a deterministic regression that reproduces the receive-token shape drift and verifies backward succeeds with replay

HybridEP already supports cached-handle dispatch; its existing dispatch backward path uses the same API. The recorder retains layout metadata, not dispatched hidden states, so activation checkpointing keeps its activation-memory savings.

This turns the HybridEP limitation described in #3684 into a correctness fix rather than requiring users to disable AC or switch dispatchers.

## Validation

- `ruff check` on changed files
- `ruff format --check` on changed files
- `pytest -q tests/unit_tests/moe/test_fused_a2a.py tests/unit_tests/moe/test_parallelizer.py`: 99 passed
- regression control without replay deterministically raises `CheckpointError` on a simulated 5-row forward / 3-row recompute
- replay path performs one full dispatch plus one cached dispatch and completes backward

GB200 4-layer K3 Expert LoRA full-AC validation passed 4/4 steps on commit `5b3738554` with HybridEP, `torch_mm`, EP4, ordinary expert LoRA, and reshard-after-forward (train loss 13.2424 -> 12.9688; validation loss 12.8113; peak memory 101.7 GiB/GPU). The first 32-node replay run exposed an extra recompute-only `aten.sum` in the cached-layout path before step 0; commit `d2f3a6382` moves extent caching outside the selective-op context and adds a regression for the exact operator-trace mismatch. The updated 32-node/100-step nightly validation is [NeMo-CI pipeline 65232165](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/65232165).

### DIFF
--- a/nemo_automodel/components/moe/megatron/fused_a2a.py
+++ b/nemo_automodel/components/moe/megatron/fused_a2a.py
@@ -18,6 +18,8 @@
 # Licensed under the MIT License - https://github.com/deepseek-ai/DeepEP/blob/main/LICENSE
 
 import os
+import threading
+from contextlib import contextmanager
 
 try:
     from deep_ep import Buffer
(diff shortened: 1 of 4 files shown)


akoumpa:
Thanks @HuiyingLi ! since the replay recorder code is identical to the one in 3684, would you mind consolidating the fix in 3684 instead? Thanks.

HuiyingLi:
> Thanks @HuiyingLi ! since the replay recorder code is identical to the one in 3684, would you mind consolidating the fix in 3684 instead? Thanks.

yes sure

HuiyingLi:
closing for #3684

PR #3750, opened 2026-08-29
perf(qwen3.5): coalesce packed metadata transfers

akoumpa:
## Stack

- Depends on #3740.
- Contains only the packed GatedDeltaNet metadata-transfer optimization and its tests.
- #3751 is the next PR in the stack and handles the independent multimodal splice synchronization.

## Motivation

Qwen3.5 packed GatedDeltaNet needs device-side indices and cumulative sequence lengths plus a CPU cumulative-length mirror for host-side FLA chunk planning. The prior path made several GPU scalar decisions, built dynamic-size `nonzero` results on the device, and copied cumulative lengths back to the host.

Those boundaries are repeated once per model forward even after the model-level metadata reuse already present on `main`.

## Change

- Prefer the collator-provided `_packed_seq_ids` when it is structurally eligible; fall back to an indexed 2D attention mask.
- Copy the candidate document IDs to CPU once and make the indexed-mask decision there.
- Derive dynamic-size indices and cumulative lengths from that CPU copy.
- Concatenate both metadata tensors and issue one coalesced H2D transfer.
- Retain the CPU cumulative-length tensor directly instead of copying it back from CUDA.

The normal packed path therefore has one D2H and one H2D boundary per model forward, with no dynamic GPU `nonzero` allocation for this metadata.

## Validation

- 20 targeted Qwen3.5 dense-backbone tests passed.
- Tests confirm `_packed_seq_ids` takes precedence and that the fallback to an indexed attention mask preserves reference indices and cumulative lengths.
- The full stacked tree was profiled on four GB200s with the 16K neat-packing recipe. The trace shows twelve model-forward metadata constructions and one boundary in each direction per construction.
- This layer does not change FSDP ownership or NCCL collective count, dtype, or payload.


tests/unit_tests/models/qwen3_5/test_qwen3_5_dense_backbone.py
```diff
@@ -209,6 +209,32 @@ def test_inv_freq_stays_fp32_after_dtype_cast(self):
 
 
 class TestDenseTextBackbone:
+    def test_packed_metadata_prefers_authoritative_ids_and_matches_reference(self):
+        attention_mask = torch.ones((2, 6), dtype=torch.long)
```
yuhezhang-ai: Small, non-blocking test suggestion: could we add a case where both masks contain valid indexed document IDs, but with different boundaries? The all-ones `attention_mask` here is not considered packed, so this test also passes with the previous attention-mask-first implementation. Checking that `document_ids` and `cu_seqlens` follow `_packed_seq_ids` when both masks are eligible would directly cover the new precedence behavior.

akoumpa: Addressed in 82cc98504358199704751578e180ca4036f65edb. The test now supplies conflicting valid indexed document IDs in both masks, so it directly verifies that document_ids and cu_seqlens follow _packed_seq_ids rather than attention_mask.

yuhezhang-ai (approved):
Thanks for adding the conflicting-mask case and retaining the binary-mask coverage. This addresses my test suggestion and directly verifies the intended precedence. Reviewed the latest revision; the implementation is unchanged from the earlier review, and CI is green. Looks good to me.

Issue #3755, opened 2026-08-30
group_by_length silently does nothing for lazily-tokenized datasets


kabirvashisht4-glitch:
**Describe the bug**

`LengthGroupedSampler._compute_lengths` has a "fast path" that unwraps `.dataset`
attributes until it finds a plain `list`, then indexes that list instead of the
dataset itself:

https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/components/datasets/llm/length_grouped_sampler.py#L137-L153

```python
# Fast path: access underlying list directly if available
raw = dataset
while hasattr(raw, "dataset"):
    raw = raw.dataset
if not isinstance(raw, list):
    raw = None
...
sample = raw[i] if raw is not None else dataset[i]
ids = sample.get("input_ids")
if ids is not None:
    lengths[i] = len(ids) if isinstance(ids, list) else ids.numel()
```

The LLM datasets in this repo tokenize lazily in `__getitem__` and keep the
**raw, untokenized** rows in `self.dataset`. `ChatDataset` is the clearest case:
`self.dataset` is the list returned by `_load_openai_messages` (a plain
`List[Dict]` for local JSON/JSONL input), and `input_ids` only exists after
`__getitem__` runs `format_chat_template`.

So the unwrap lands on rows shaped like `{"messages": [...]}`,
`sample.get("input_ids")` returns `None`, and **every length stays at the `0`
initializer**. `sorted()` on all-equal keys is stable, so `sorted_indices` is
just `range(len(dataset))` — the sampler degrades to chunk-shuffled original
order and does no length grouping at all. There is no error and no warning.

The same unwrap is also unsafe for any wrapper that remaps indices
(e.g. `torch.utils.data.Subset`): `raw[i]` is not `dataset[i]`, so lengths get
attributed to the wrong samples, and it can raise `IndexError` when
`len(raw) < len(dataset)`.

Note the fast path buys nothing in the case it is actually correct: when
`dataset` is itself a plain `list`, the loop does not unwrap anything and
`raw[i]` is literally `dataset[i]`. It only changes behaviour in exactly the
cases where it is wrong.

**Steps/Code to reproduce bug**

`group_by_length: true` in the dataloader config with any lazily-tokenizing
dataset, e.g.:

```yaml
dataset:
  _target_: nemo_automodel.components.datasets.llm.chat_dataset.ChatDataset
  path_or_dataset_id: /path/to/train.jsonl

dataloader:
[...]

PR #3756, opened 2026-08-30
fix(datasets): read group_by_length token lengths through the dataset

kabirvashisht4-glitch:
# What does this PR do ?

Fixes `group_by_length: true` silently doing nothing for datasets that tokenize
lazily in `__getitem__` (e.g. `ChatDataset`).

`LengthGroupedSampler._compute_lengths` had a fast path that unwrapped nested
`.dataset` attributes until it found a plain `list`, then indexed that list
instead of the dataset:

```python
raw = dataset
while hasattr(raw, "dataset"):
    raw = raw.dataset
if not isinstance(raw, list):
    raw = None
...
sample = raw[i] if raw is not None else dataset[i]
```

`ChatDataset.dataset` holds the **raw, untokenized** rows returned by
`_load_openai_messages` (a plain `List[Dict]` for local JSON/JSONL), and
`input_ids` only exists after `__getitem__` runs `format_chat_template`. The
unwrap therefore landed on `{"messages": [...]}` rows, `sample.get("input_ids")`
returned `None`, and every length stayed at its `0` initializer. Sorting
all-equal keys is stable, so `sorted_indices` came out as `range(len(dataset))`
and the sampler did no length grouping at all — no error, no warning.

## Why the unwrap is removed rather than guarded

An earlier revision kept the fast path behind a guard: take the inner list only
when it is 1:1 with the dataset (`len(raw) == len(dataset)`) and its first row
already carries `input_ids`. As @HuiyingLi pointed out, equal lengths do not
establish that `raw[i]` is `dataset[i]`. That guard only excludes wrappers that
*shrink* the dataset; it is blind to wrappers that reorder, repeat, or rewrite.
Three length-preserving cases slipped through, all reproduced on that revision:

| wrapper | `dataset[i]` lengths | unwrapped lengths |
|---|---|---|
| `Subset(raw, [3, 1, 0, 2])` — permutation | `[40, 20, 10, 30]` | `[10, 20, 30, 40]` |
| `Subset(raw, [0, 0, 3, 3])` — repeats | `[10, 10, 40, 40]` | `[10, 20, 30, 40]` |
| 1:1 wrapper truncating `input_ids` to 16 | `[10, 16, 16, 16]` | `[10, 20, 30, 40]` |

No guard can be sound here: `.dataset` is an arbitrary attribute on an arbitrary
object, and `__getitem__` may do anything. The impact is mis-grouping and extra
padding, not sample/label mismatch, and it predates this PR.

The fast path also buys almost nothing. Measured over 200k samples, best of 3:

| | time |
|---|---:|
| `raw[i]` (unwrapped list) | 5.9 ms |
| `dataset[i]` | 7.8 ms |
| cost of removing the fast path | **+1.9 ms** |

So `_compute_lengths` now always reads `dataset[i]`, which is the only source
that reflects the samples the sampler will actually yield.

Before / after on an 8-sample dataset with lengths `[8, 128, 16, 64, 4, 256, 32, 512]`
at `batch_size=2`:

| | computed lengths | padding waste |
|---|---|---|
| before | `[0, 0, 0, 0, 0, 0, 0, 0]` | 900 tokens |
| after | `[8, 128, 16, 64, 4, 256, 32, 512]` | 340 tokens |

# Changelog

- Remove the `.dataset` unwrapping from `_compute_lengths`; token lengths are
  always read through `dataset[i]`. The docstring states why an inner list is
  never substituted.
- Warn when no sample yields a token length, so the sampler can no longer become
  a silent no-op.
- Add `TestComputeLengths` to
  `tests/unit_tests/datasets/llm/test_length_grouped_sampler.py` covering the
  lazily-tokenized case, an equal-length permutation `Subset`, an equal-length
  `Subset` with repeated indices, a 1:1 wrapper that shortens `input_ids`, a
  shorter `Subset`, end-to-end grouping over a permuted `Subset`, tensor
  `input_ids`, the new warning, and the empty dataset.

**Pre checks**:

Behaviour matches the existing docstring ("Samples must have an ``input_ids``
key (list or tensor) whose length is used for sorting"), so no doc change was
needed.

Verified locally: `pytest tests/unit_tests/datasets/llm/test_length_grouped_sampler.py`
→ 22 passed. The four wrapper-safety tests fail on the previous revision and on
`main`. `ruff format --check` / `ruff check` clean. CPU only, no GPU needed.

# Additional Information

- Related to #3755


HuiyingLi:
The lazy `ChatDataset` fix works in local validation. One remaining P2 issue in `_unwrap_to_list`: `len(raw) == len(dataset)` does not establish that `raw[i]` is equivalent to `dataset[i]`.

For example, an equally sized `Subset` still bypasses its index mapping:

```python
from torch.utils.data import Subset
from nemo_automodel.components.datasets.llm.length_grouped_sampler import LengthGroupedSampler

raw = [{"input_ids": [0] * n} for n in [10, 20, 30, 40]]
dataset = Subset(raw, [3, 1, 0, 2])

assert [len(dataset[i]["input_ids"]) for i in range(len(dataset))] == [40, 20, 10, 30]
assert LengthGroupedSampler._unwrap_to_list(dataset) is raw
# _compute_lengths therefore reads [10, 20, 30, 40], not [40, 20, 10, 30].
```

This also affects full-length subsets with repeated indices and same-length wrappers that transform/truncate `input_ids` in `__getitem__`. The new Subset test only uses a shorter subset, so its length mismatch avoids the problematic fast path. The impact is incorrect length grouping and extra padding, not sample/label mismatching.

These cases fail on both base and head: this is an incomplete part of the Subset/wrapper safety fix described by the PR, **not a newly introduced regression**. Could we obtain lengths through `dataset[i]` without generically unwrapping arbitrary `.dataset` wrappers, and add equal-length permutation and length-changing-wrapper cases?

Validation at `0445aff94657aa9f3cb645f1c791927e3128f35f`: 19 sampler tests plus 125 related dataset/loader/VLM-sampler tests passed. A real `ChatDataset` with a local HF fast tokenizer reproduces the all-zero lengths on base and correct lengths on head. A two-process Gloo run using the production `ParallelAwareDataloader` passed sample coverage and checkpoint/resume checks with grouping enabled and disabled. The residual wrapper cases above were independently reproduced on exact base/head.

kabirvashisht4-glitch:
Thanks @HuiyingLi — you're right, and the counterexample reproduces exactly as you wrote it. I confirmed two more length-preserving cases that slipped through the same guard:

```
--- Subset(raw, [3, 1, 0, 2]) — full-length permutation ---
  dataset[i]       : [40, 20, 10, 30]
  _unwrap_to_list  : returns raw   <-- guard passed
  _compute_lengths : [10, 20, 30, 40]

--- Subset(raw, [0, 0, 3, 3]) — repeated indices ---
  dataset[i]       : [10, 10, 40, 40]
  _compute_lengths : [10, 20, 30, 40]

--- 1:1 wrapper truncating input_ids to 16 ---
  dataset[i]       : [10, 16, 16, 16]
  _compute_lengths : [10, 20, 30, 40]
```

`len(raw) == len(dataset)` only excludes wrappers that shrink the dataset, so it is blind to reordering, repetition, and rewriting. I don't think the guard can be made sound: `.dataset` is an arbitrary attribute on an arbitrary object and `__getitem__` may do anything, so any predicate is a guess about code we cannot see.

I also measured what the fast path was buying, over 200k samples, best of 3:

| | time |
|---|---:|
| `raw[i]` (unwrapped list) | 5.9 ms |
| `dataset[i]` | 7.8 ms |
| cost of removing it | **+1.9 ms** |

Under 2 ms across 200k samples, so I took the option you suggested rather than tightening the predicate: `_compute_lengths` now always reads `dataset[i]` and the `.dataset` unwrapping is gone. The docstring records why an inner list is never substituted.

Tests added for the cases you asked for:

- `test_equal_length_permutation_subset_follows_index_mapping`
- `test_equal_length_subset_with_repeated_indices`
- `test_length_changing_wrapper_is_measured_through_getitem`
- `test_permuted_subset_is_grouped_on_its_own_order` — end to end, asserting `sorted_indices` follows the samples the sampler will actually yield

All four fail on the previous revision and on `main`; the existing shorter-`Subset` case is kept. The old `test_pretokenized_wrapper_keeps_fast_path` is removed along with the fast path it asserted.

`pytest tests/unit_tests/datasets/llm/test_length_grouped_sampler.py` → 22 passed. `ruff format --check` / `ruff check` clean. PR description updated to describe the current approach.

Agreed on your framing that this was an incomplete part of the wrapper-safety fix rather than a new regression — thanks for catching it before it shipped.

Issue #3767, opened 2026-08-31
max_steps is silently ignored when num_epochs is left unset


kabirvashisht4-glitch:
**Describe the bug**

`StepSchedulerConfig.num_epochs` defaults to `10` instead of `None`, so the
`max_steps`-based derivation in `StepScheduler.__init__` never runs on the
config path the recipes use. A config that sets only `max_steps` silently trains
for `10 * epoch_len` steps instead of `max_steps`.

`StepSchedulerConfig` documents the intended contract
(`nemo_automodel/components/training/step_scheduler.py`):

> `num_epochs`: Number of training epochs. When `None` the builder derives it
> from `max_steps`. Default: 10.
> `max_steps`: **Hard cap** on optimizer steps. `None` means derive from
> `num_epochs * epoch_len`.

and `StepScheduler.__init__` implements that derivation:

```python
if num_epochs is None:
    num_epochs = _calculate_num_epochs(max_steps, self.epoch_len)
```

But the dataclass field is:

```python
num_epochs: int | None = 10
```

`asdict(self)` in `StepSchedulerConfig.build()` therefore always passes
`num_epochs=10` unless the user explicitly writes `num_epochs: null`, and the
`num_epochs is None` branch is dead on this path. Recipes build the scheduler
through this config — `train_ft.py` line 758 calls
`self.cfg.step_scheduler.build(...)`.

The truncation happens because the training loop is bounded by `epochs`, not
only by `max_steps`:

```python
@property
def epochs(self):
    for e in range(epoch, self.num_epochs):
        if self.step >= self.max_steps or self.sigterm_received:
            return
        yield e
```

With `num_epochs` stuck at 10, the generator is exhausted after 10 epochs and
`max_steps` is never reached. There is no warning.

**Steps/Code to reproduce bug**

```yaml
step_scheduler:
  global_batch_size: 1
  local_batch_size: 1
  max_steps: 5000
  # num_epochs intentionally not set
```

Standalone repro (CPU, no GPU or tokenizer needed):

```python
from nemo_automodel.components.training.step_scheduler import StepSchedulerConfig

class DL(list):
    pass

dl = DL(range(100))  # 100 micro-batches; grad_acc = 1 -> epoch_len = 100

cfg = StepSchedulerConfig(global_batch_size=1, max_steps=5000, preemption_signal=None)
ss = cfg.build(dataloader=dl, dp_group_size=1, local_batch_size=1)

total = 0
[...]

PR #3768, opened 2026-08-31
fix(training): honor max_steps when num_epochs is unset

kabirvashisht4-glitch:
# What does this PR do ?

Makes `max_steps` actually cap training when `num_epochs` is left unset.

`StepSchedulerConfig.num_epochs` defaulted to `10` instead of `None`, so the
derivation in `StepScheduler.__init__` never ran on the config path the recipes
use:

```python
if num_epochs is None:
    num_epochs = _calculate_num_epochs(max_steps, self.epoch_len)
```

`asdict(self)` in `StepSchedulerConfig.build()` always passed `num_epochs=10`
unless the user explicitly wrote `num_epochs: null`, leaving that branch dead.
Recipes go through this config — `train_ft.py` line 758 calls
`self.cfg.step_scheduler.build(...)`.

The training loop is bounded by the `epochs` generator as well as by
`max_steps`:

```python
@property
def epochs(self):
    for e in range(epoch, self.num_epochs):
        if self.step >= self.max_steps or self.sigterm_received:
            return
        yield e
```

so with `num_epochs` pinned at 10 the generator ran out before `max_steps` was
reached. A config setting only `max_steps` trained for `10 * epoch_len` steps
and reported success — the symptom is an under-trained model, not an error.

Measured with a 100-step epoch:

| config | `num_epochs` | steps requested | steps executed |
|---|---|---|---|
| `max_steps: 5000` (before) | 10 | 5000 | **1000** |
| `max_steps: 5000` (after) | 50 | 5000 | 5000 |

The field's own docstring already described the intended behaviour ("When
`None` the builder derives it from `max_steps`"), so this makes the code match
the documented contract rather than changing it.

# Changelog

- `StepSchedulerConfig.num_epochs` now defaults to `None` so `max_steps` drives
  the epoch count when `num_epochs` is not pinned.
- Attribute docstring updated to state the resolution order: derived from
  `max_steps`, falling back to 10 when `max_steps` is also unset.
- `tests/unit_tests/training/test_training_config.py::test_defaults` updated for
  the new default.
- Four CPU tests in
  `tests/unit_tests/components/training/test_step_scheduler.py` covering the
  resolution matrix: max-steps-only, num-epochs-only, both set, neither set.

**Pre checks**:

Behaviour now matches the existing docstring, so no separate doc change was
needed.

**Compatibility.** The both-unset case is deliberately unchanged:
`_calculate_num_epochs` returns its `default_num_epochs=10` when `max_steps is
None`, so a config pinning neither field still resolves to 10 epochs and the
same `max_steps`. `test_config_neither_set_keeps_ten_epoch_fallback` locks that
in. Configs that set `num_epochs` explicitly are unaffected.

- `pytest tests/unit_tests/components/training/test_step_scheduler.py tests/unit_tests/training/test_training_config.py` → 47 passed.
- Reverting only the one-line default makes
  `test_config_max_steps_only_derives_num_epochs` and `test_defaults` fail; the
  other three new tests pass either way, which is intended — they guard the
  paths this must not change.
- `pytest tests/unit_tests/recipes/ tests/unit_tests/training/ tests/unit_tests/components/` → 1379 passed, 19 skipped, no regressions.
  (`tests/unit_tests/recipes/dllm` was excluded locally: it fails to import
  `transformers.models.diffusion_gemma` on `main` too, unrelated to this change.)

# Additional Information

- Related to #3767

### DIFF
--- a/nemo_automodel/components/training/step_scheduler.py
+++ b/nemo_automodel/components/training/step_scheduler.py
@@ -355,8 +355,9 @@ class StepSchedulerConfig:
 
     Attributes:
         global_batch_size: Total samples per optimizer step across all GPUs.
-        num_epochs: Number of training epochs.  When ``None`` the builder
-            derives it from ``max_steps``.  Default: 10.
+        num_epochs: Number of training epochs.  When ``None`` (the default)
+            it is derived from ``max_steps``, or falls back to 10 when
+            ``max_steps`` is also unset.
         max_steps: Hard cap on optimizer steps.  ``None`` means derive from
             ``num_epochs * epoch_len``.
         ckpt_every_steps: Save a checkpoint every N optimizer steps.
@@ -377,7 +378,7 @@ class StepSchedulerConfig:
     """
 
     global_batch_size: int = 32
-    num_epochs: int | None = 10
+    num_epochs: int | None = None
     max_steps: int | None = None
     ckpt_every_steps: int | None = 100
     save_checkpoint_every_epoch: bool = True
--- a/tests/unit_tests/components/training/test_step_scheduler.py
+++ b/tests/unit_tests/components/training/test_step_scheduler.py
@@ -661,3 +661,56 @@ def test_step_scheduler_config_exposes_preemption_signal():
         assert scheduler.sig_handler.sigs == [_signal.SIGUSR2]
     finally:
         scheduler.sig_handler.release()
+
+
+# ---------------------------------------------------------------------------
+# num_epochs / max_steps resolution via the typed config
+# ---------------------------------------------------------------------------
+
+
+def _run_all_steps(scheduler):
+    """Drive the scheduler to exhaustion and return the number of steps executed."""
+    total = 0
+    for epoch in scheduler.epochs:
+        scheduler.set_epoch(epoch)
+        for _ in scheduler:
+            total += 1
+    return total
+
+
+def test_config_max_steps_only_derives_num_epochs():
+    """max_steps alone must be honored; num_epochs is derived, not left at a default."""
+    cfg = StepSchedulerConfig(global_batch_size=1, max_steps=5000, preemption_signal=None)
+    scheduler = cfg.build(SizedDataLoader(num_batches=100), dp_group_size=1, local_batch_size=1)
+
+    assert scheduler.epoch_len == 100
+    assert scheduler.num_epochs == 50  # ceil(5000 / 100)
+    assert _run_all_steps(scheduler) == 5000
+
+
+def test_config_num_epochs_only_derives_max_steps():
+    cfg = StepSchedulerConfig(global_batch_size=1, num_epochs=3, preemption_signal=None)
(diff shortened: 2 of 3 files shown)


nemo_automodel/components/training/step_scheduler.py
```diff
@@ -377,7 +378,7 @@ class StepSchedulerConfig:
     """
 
     global_batch_size: int = 32
-    num_epochs: int | None = 10
+    num_epochs: int | None = None
```
akoumpa, line 389: Hi @kabirvashisht4-glitch , thanks for the PR, i'm not sure i follow 100%, i was under the impression that the max_steps' value would be derived from the num_epochs' default value (10).

kabirvashisht4-glitch: Thanks for looking @akoumpa — your understanding is right, and that direction is unchanged by this PR.

`max_steps` unset → derived from `num_epochs` via `_calculate_max_steps`, defaulting to 10 epochs. Still true after this change: `_calculate_num_epochs` returns its own `default_num_epochs=10` when `max_steps is None`, so a config that pins neither field still resolves to `num_epochs=10`, `max_steps=10*epoch_len`. `test_config_neither_set_keeps_ten_epoch_fallback` locks that in.

The case this fixes is the opposite one: **`max_steps` set explicitly, `num_epochs` left unset.**

Because the dataclass default was `10` rather than `None`, `asdict()` in `build()` always passed `num_epochs=10`, so the `num_epochs is None` branch in `__init__` was unreachable on the config path. And since the loop is bounded by the `epochs` generator as well as by `max_steps`, the generator ran out first:

```python
for e in range(epoch, self.num_epochs):   # 10
    if self.step >= self.max_steps: return
```

With a 100-step epoch and `max_steps: 5000`, that is 10 epochs = 1000 steps. The run then reports success, so the symptom is a quietly under-trained model rather than an error:

| config | `num_epochs` | requested | executed |
|---|---|---|---|
| `max_steps: 5000` (before) | 10 | 5000 | **1000** |
| `max_steps: 5000` (after) | 50 | 5000 | 5000 |

This is also what `StepScheduler.__init__` already documents:

> `num_epochs (int | None)`: Total number of epochs. Default: None or calculated from max_steps if num_epochs is None **or 10 if max_steps and num_epochs are both None**.
> `max_steps (int | None)`: Maximum number of steps to run. If None, calculated from num_epochs.

That describes a three-way resolution — explicit `num_epochs` wins; else derive from `max_steps`; else 10 — and the dataclass default of `10` made the middle branch dead. So this is the class's documented contract rather than new behaviour, and the resolution matrix is covered by the four tests added here.

Happy to go a different way if you'd rather keep the `10` default: the alternative is to leave it and instead raise (or warn) in `__init__` when `num_epochs * epoch_len < max_steps`, so the conflict is loud instead of silent. That surfaces the same problem without changing any default — just say which you prefer.

For reference, full CI is green on ed8f52a (L0 CPU, both L0 GPU shards, and the L2 suites).

PR #3777, opened 2026-09-01
fix(loss): prevent TP KD gradient over-reduction

### LINKED ISSUE #3776: kd loss under tensor parallelism trains with gradients scaled by tp_size
training with knowledge distillation and a tensor parallel student gives gradients that are exactly tp_size times too large on the kd side. the loss you see in the logs is perfectly correct, so nothing ever looks wrong. but underneath, at tp=8 with kd_ratio=0.5, the model effectively trains with an 8:1 kd to ce mix instead of 1:1. your kd_ratio setting is silently overridden.

the cause is one line in `components/loss/kd_loss.py` (line 107). the kl sum across tp ranks uses the differentiable `torch.distributed.nn.functional.all_reduce`, whose backward all_reduces the gradients again. since every tp rank computes the same loss and every rank calls backward, that second all_reduce adds up tp identical gradients. each rank's gradient should be counted once, it gets counted tp times.

easy to see on cpu with 2 gloo ranks acting as tp=2, real KDLoss, compared against the plain single process KDLoss on the same inputs:

```
reference (non-TP) loss: 0.368836
rank 0: TP loss=0.368836 | grad/ref ratio = 2.0000
rank 1: TP loss=0.368836 | grad/ref ratio = 2.0000
```

same loss, every gradient exactly doubled. the same thing reproduces through real DTensor plumbing with the tp group auto detected: the ce gradients come out right (1x), the kd gradients come out 2x.

the fix is one line plus a tiny helper: that kl all_reduce should pass gradients through unchanged in backward, since each rank already holds the full gradient of the shared loss. important detail: the softmax denominator all_reduce two lines above must stay as is, that one really needs the gradients summed. verified with the same setup:

```
rank 0 fixed: ratio = 1.0000, max diff from reference = 1.1e-08
rank 1 fixed: ratio = 1.0000, max diff from reference = 7.5e-09
```

why the existing test didn't catch it: the kd tp correctness test checks that gradients exist, not what their values are.

happy to send a pr with the fix and a gradient parity test.

mikemikimike:
## Summary

- Fixes #3776 by preserving the tensor-parallel KD loss forward aggregation while keeping each rank's local vocabulary gradient.
- Replace the final in-place no-grad reduction with a private autograd function that all-reduces a clone in forward and passes the gradient through in backward.
- Add regression coverage for an upstream operation that saves its output for autograd, which would reject an in-place reduction.

## Compatibility

The public API and forward loss values are unchanged. The final KL aggregation now allocates a clone before its reduction; the earlier denominator reductions remain autograd-aware.

## Tests

- `D:\sofware\anaconda3\python.exe -m pytest tests/unit_tests/loss/test_kd_loss.py -q` - 33 passed, 1 skipped on Windows.
- `ruff check .` - passed.
- `ruff format --check nemo_automodel/components/loss/kd_loss.py tests/unit_tests/loss/test_kd_loss.py` - passed.
- `bandit -r nemo_automodel/components/loss/kd_loss.py -t B614` - passed.
- `uvx --from import-linter lint-imports` - passed.
- `ty check nemo_automodel/components/loss/kd_loss.py` reports four pre-existing DTensor fallback diagnostics; this change adds none.

The checked-in two-rank CPU Gloo gradient test remains skipped on Windows. When temporarily unskipped, the prior implementation passes there, while the custom-autograd variant exits in a Windows Gloo child process with `0xc0020043`; this is recorded as a platform limitation. The full locked dependency environment could not be installed on Windows because the locked Triton version has no Windows wheel, so the focused tests above used the available local PyTorch environment.

### DIFF
--- a/nemo_automodel/components/loss/kd_loss.py
+++ b/nemo_automodel/components/loss/kd_loss.py
@@ -104,7 +104,11 @@ def _kl_forward_tp(
 
     # --- Per-token KL: local vocabulary contribution then global reduce ---
     kl_local = _forward_kl_from_log_probs(teacher_log_prob, student_log_prob)
-    kl_local = dist_nn_func.all_reduce(kl_local, op=torch.distributed.ReduceOp.SUM, group=tp_group)
+    # Every rank computes the same global KL value, so its backward pass must
+    # retain the local vocabulary gradient instead of reducing that gradient a
+    # second time through the autograd-aware collective.
+    with torch.no_grad():
+        torch.distributed.all_reduce(kl_local, op=torch.distributed.ReduceOp.SUM, group=tp_group)
 
     return kl_local
 
--- a/nemo_automodel/components/loss/kd_loss.py
+++ b/nemo_automodel/components/loss/kd_loss.py
@@ -66,6 +66,42 @@ def _forward_kl_from_log_probs(
     return (teacher_prob * log_ratio).sum(dim=-1)
 
 
+class _AllReduceForwardPassThroughBackward(torch.autograd.Function):
+    """All-reduce a tensor in forward while preserving its local gradient."""
+
+    @staticmethod
+    def forward(ctx, input_tensor: torch.Tensor, group: torch.distributed.ProcessGroup) -> torch.Tensor:
+        """Reduce a clone of the input tensor without mutating its autograd value.
+
+        Args:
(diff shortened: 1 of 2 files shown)


akoumpa:
Hi @mikemikimike thanks a lot for the fix and for turning this into an autograd function

akoumpa (approved):

PR #3786, opened 2026-09-02
feat(retrieval): context aware reranker training

sahel-sh:
# What does this PR do ?

Adds context-aware cross-encoder reranker training: finetunes a causal LM as a listwise
reranker that scores a query–passage pair using the context the query came from — the
reasoning trace that motivated it and the originating question — not the query alone.

One row per sub-query, 1 positive and 7 hard negatives, one softmax per group. The collator
drops each context field independently, so a single model covers all four prompt shapes.
The model adds no parameters: it scores `logit("yes") - logit("no")` off the pretrained
`lm_head`, and checkpoints serialize as a plain `Qwen3ForCausalLM` that loads in stock HF and
vLLM without `trust_remote_code`.

# Changelog

- `models/qwen3_reranker/model.py`: `Qwen3RerankerForCausalReranking`; raw yes/no log-odds,
  padding-side agnostic, serialized identity rewritten to stock `Qwen3ForCausalLM`.
- `models/qwen3_reranker/collator.py`: per-field context dropout; the `<Instruct>` line is
  chosen from whichever fields survive, and per-field token budgets are applied before
  assembly so a long trace cannot crowd out the query.
- `datasets/llm/retrieval_dataset_inline.py`: context columns and a group-aware validation
  split keyed on a hash of `(seed, group)`, so independently-built train/val sets stay
  complementary regardless of load order.
- `recipes/retrieval/train_cross_encoder.py`: listwise cross-entropy, cast to fp32 before the
  temperature division — these recipes run native bf16, so the loss would otherwise quantize.
- `_transformers/{registry,retrieval}.py`: register the reranker for `score`, keeping the
  generic `AutoModel` fallback for `embedding`.
- `training/step_scheduler.py`, `recipes/retrieval/train_bi_encoder.py`: forward the epoch to
  the training collator only, so validation keeps a fixed prompt mix.
- `examples/retrieval/cross_encoder/qwen3_4b_reranker_agentir.yaml`: example recipe with the
  data schema documented inline.

**Pre checks**:

44 new tests in 4 files, plus 3 in the existing step-scheduler suite: scoring parity with the
published `compute_logits` under both padding sides, checkpoint round-trips reloaded through
`AutoModelForCausalLM`, split independence from load order, and the `embedding` fallback.

# Additional Information

- Related to # (issue)


nemo_automodel/recipes/retrieval/train_cross_encoder.py
```diff
@@ -111,7 +111,17 @@ def _forward_backward_step(self, idx, batch, *, loss_buffer, num_batches, is_tra
...
+            # temperature as the bi-encoder path does. Backbones that scale inside
+            # forward (LlamaBidirectionalForSequenceClassification) apply model.temperature
+            # there; the two compose, and both default to 1.0.
+            #
+            # Up-cast to fp32 BEFORE the division. The backbone emits bf16 logits, and
+            # dividing by a small temperature magnifies them (10x at temperature=0.1),
+            # pushing them onto a coarse part of the bf16 grid -- at magnitude ~32 the
+            # spacing is 0.25, which distorts the softmax and every gradient through it.
+            # Staying in fp32 through cross_entropy costs nothing at this tensor size.
+            outputs.logits = outputs.logits.view(-1, self.train_n_passages).float() / self.temperature
```
gabrielspmoreira, line 124: This is a good addition, upcasting lower precision scores to fp32 is a good practice before softmax cross entropy.
This is helpful for training cross encoders in general.

nemo_automodel/components/models/qwen3_reranker/model.py
```diff
@@ -0,0 +1,320 @@
...
+        )
+        hidden_states = outputs.last_hidden_state
+
+        # Gather the final non-padding hidden state per row, then project only
+        # those positions through the LM head (avoids a full-sequence vocab matmul).
+        batch_size = hidden_states.shape[0]
+        last_idx = _last_token_indices(attention_mask)
+        last_hidden = hidden_states[torch.arange(batch_size, device=hidden_states.device), last_idx]
+
+        last_logits = self.lm_head(last_hidden)  # [B, vocab]
```
gabrielspmoreira, line 317: Here the LM head scores all `151669` tokens. Wouldn't it be more efficient to ajust the MLP head to keep only the "yes" and "no" token outputs, before forwarding `last_hidden` into it, in order to save compute during training and inference?

That efficient implementation would be something like the "Feature Extraction + Pooling + Dense" from sentence-transformers (see [here](https://github.com/huggingface/sentence-transformers/tree/main/examples/cross_encoder/training/multimodal#comparing-the-two-approaches) and [here](https://github.com/huggingface/sentence-transformers/blob/main/docs/cross_encoder/usage/custom_models.rst#module-chains))

sahel-sh: Yes, full head with no slicing is wasteful for sure, both at training and at inference. My main motivation here was to keep everything identical to the base Reranking model, Qwen3-reranker-4b: no custom forward(), no custom HF registry entry, no custom inference code.

nemo_automodel/components/models/qwen3_reranker/model.py
```diff
@@ -0,0 +1,320 @@
...
+    # correctly, where a whole-batch padding-side guess would mislabel the minority.
+    #
+    # A row with an all-zero mask yields -1, which would silently index the last position.
+    # Callers pass masks with at least one attended token per row; this mirrors the prior
+    # behavior rather than paying for a check on every forward.
+    positions = torch.arange(attention_mask.shape[-1], device=attention_mask.device)
+    return positions.masked_fill(~attention_mask.bool(), -1).amax(dim=-1)
+
+
+class Qwen3RerankerForCausalReranking(Qwen3ForCausalLM):
```
oliverholworthy, line 155: There's a test (`test_tie_support_registry`) that checks all models declare `TieSupport` explictly. the qwen3 reranker checkpoint has tie_word_embeddings set to True.  

```python
from nemo_automodel.components.models.common.tie_word_embeddings import TieSupport

class Qwen3RerankerForCausalReranking(Qwen3ForCausalLM):
    tie_word_embeddings_support = TieSupport.BOTH
```

and add a small test like

```python
def test_supports_tied_embeddings():
    config = tiny_config(tie_word_embeddings=True)
    model = Qwen3RerankerForCausalReranking(config)

    assert model.lm_head.weight is model.model.embed_tokens.weight
```

nemo_automodel/_transformers/registry.py
```diff
@@ -203,6 +203,10 @@
                 "Mistral3FP8VLMForConditionalGeneration",
             ),
         ),
+        (
+            "Qwen3RerankerForCausalReranking",
```
oliverholworthy, line 207: Registering Qwen3 as a supported score backbone means the existing `test_build_generic_hf_model_score_task` fixture is no longer testing the generic fallback: its fake model_type = "qwen3" now selects this reranker and attempts to load weights from the empty temporary directory. Could you change that fixture to a new unsupported type, for example:

```python
class FakeConfig:
    model_type = "bert"
```

examples/retrieval/cross_encoder/qwen3_4b_reranker_agentir.yaml
oliverholworthy, line 1: This new recipe/architecture triggers two model-coverage test failures. Could you:

1. Add Qwen3-Reranker to `docs/model-coverage/reranker/index.mdx`, including the literal architecture name `Qwen3RerankerForCausalReranking`.
2. Update `test_embedding_and_reranking_releases_are_discovered_from_recipes` to expect the new checkpoint:
assert models_by_type["Reranking"] == {
    "meta-llama/Llama-3.2-1B",
    "Qwen/Qwen3-Reranker-4B",
}

An entry in the existing reranker overview is sufficient here; I don’t think this needs a separate model page for this MR. A minimal documentation row could be:

```
| Qwen | [Qwen3-Reranker-4B](https://huggingface.co/Qwen/Qwen3-Reranker-4B) | `Qwen3RerankerForCausalReranking` | `NeMoAutoModelCrossEncoder` | Reranking |
```

And the introduction should also stop saying that only Llama has an optimized/custom path.

nemo_automodel/components/models/qwen3_reranker/model.py
oliverholworthy: One additional detail: after adding TieSupport.BOTH, the registry test also checks for a model-local `tie_weights()` using `cls.__dict__`, so inheriting it will still fail. Maybe need to mirror the Qwen3 pattern explictly? with something like:

```python
from nemo_automodel.components.models.common.tie_word_embeddings import (
    TieSupport,
    reject_unsupported_tie_word_embeddings,
)


class Qwen3RerankerForCausalReranking(Qwen3ForCausalLM):
    tie_word_embeddings_support: TieSupport = TieSupport.BOTH
    _tied_weights_keys = {
        "lm_head.weight": "model.embed_tokens.weight",
    }

    def __init__(self, config: Qwen3Config) -> None:
        reject_unsupported_tie_word_embeddings(type(self), config)
        super().__init__(config)

    def tie_weights(self, *_args: object, **_kwargs: object) -> None:
        if getattr(self.config, "tie_word_embeddings", False):
            self.lm_head.weight = self.model.embed_tokens.weight
```

docs/model-coverage/reranker/index.mdx
jgerh:
```suggestion
description: "Reranking model coverage in NeMo AutoModel, including optimized Llama and Qwen3 backbones, Hugging Face Auto fallback, and supported workflows."
```

docs/model-coverage/reranker/index.mdx
```diff
@@ -6,15 +6,18 @@ position: 1
 
 ## Introduction
 
-Reranking models use cross-encoders to score a query-document pair jointly. They are typically used after an embedding model has produced an initial candidate set. NeMo AutoModel supports optimized bidirectional Llama rerankers and falls back to Hugging Face `AutoModelForSequenceClassification` for other architectures.
+Reranking models use cross-encoders to score a query-document pair jointly. They are typically used after an embedding model has produced an initial candidate set. NeMo AutoModel supports optimized bidirectional Llama rerankers and a causal Qwen3 reranker that scores with a yes/no decision at the final token, and falls back to Hugging Face `AutoModelForSequenceClassification` for other architectures.
```
jgerh:
```suggestion
Reranking models use cross-encoders to score a query-document pair jointly. They are typically used after an embedding model has produced an initial candidate set. NeMo AutoModel supports optimized bidirectional Llama rerankers and a causal Qwen3 reranker that scores with a yes or no decision at the final token, and falls back to Hugging Face `AutoModelForSequenceClassification` for other architectures.
```

docs/model-coverage/reranker/index.mdx
jgerh:
```suggestion
- **LoRA and PEFT:** Parameter-efficient fine-tuning for reranking backbones
```

jgerh (commented):
Completed tech pubs review of docs/model-coverage/reranker/index.mdx and provided a few copyedits.

oliverholworthy (dismissed):

gabrielspmoreira (approved):

yuhezhang-ai:
/ok to test [532737d](https://github.com/NVIDIA-NeMo/Automodel/pull/3786/commits/532737d2e2b4dcab46a536f8d3d15c65e5557619)

yuhezhang-ai (commented):
The code looks good to me overall. Thanks for adding the context-aware reranker support and the focused tests! Could you help address these three minor edge cases to make the framework more robust?

These are non-blocking compatibility notes for alternate loading paths; none is triggered by the new example's normal loading configuration.

1. **Preserve the existing Qwen3 classification fallback.** The new `qwen3` → `score` registration also routes checkpoints saved as `Qwen3ForSequenceClassification` into the causal reranker. In a tiny reproduction, this drops the trained `score.weight` and initializes a new `lm_head.weight`, changing the scores. Could we distinguish compatible causal-LM checkpoints from sequence-classification checkpoints and preserve the existing fallback for the latter? This affects loading through `NeMoAutoModelCrossEncoder`; the normal `NeMoAutoModelForSequenceClassification` entry point is unaffected, and I don't know of users relying on this particular combination.

2. **Guard checkpoint weight-tying overrides.** The new model-owned `from_pretrained()` accepts `tie_word_embeddings` overrides that disagree with the checkpoint. Loading a tied checkpoint as untied initializes a new scoring head, while the opposite direction discards the independent head. Could we apply the existing `reject_tie_word_embeddings_flip` guard before loading and cover both directions? Declaring `TieSupport.BOTH` describes the model's capabilities but doesn't protect the saved checkpoint's tying semantics.

3. **Resolve yes/no token IDs when extracting a Qwen3 submodel.** The `extract_submodel` path uses `_load_from_extracted_state`, which calls the base Hugging Face loader directly and bypasses the token-ID resolution in the reranker's `from_pretrained()`. A stock Qwen3 submodel therefore loads but raises on its first forward because `yes_token_id` and `no_token_id` are unset. Could we resolve the IDs for this path too, or preserve the existing extraction fallback until that path is supported?

For validation, the 187 focused CPU tests passed with Transformers 5.15.1, and I reproduced these three cases with tiny models. The normal tied-weight save/reload path also preserved scores. I haven't run full GPU training.

Issue #3798, opened 2026-09-03
VLM LengthGroupedSampler logs a 1.2x pre-filter threshold but applies max_length + 512


kabirvashisht4-glitch:
**Describe the bug**

The VLM `LengthGroupedSampler` pre-filter says it uses 1.2x headroom but adds a
flat 512 tokens, and the log line reports the flat threshold *labelled* as
`1.2 * max_length`. The two diverge as `max_length` grows, so at long context the
sampler drops far more samples than the comment describes and the log misstates
the rule that was applied.

`nemo_automodel/components/datasets/vlm/samplers.py` L117-131:

```python
# Use 1.2x headroom: the estimated length is a heuristic, so only
# drop samples that are clearly overlong.  Borderline samples are
# left to PreTokenizedDatasetWrapper's precise tokenize-and-retry.
filter_threshold = max_length + 512
kept = [i for i in all_indices if self.lengths[i] <= filter_threshold]
n_dropped = len(dataset) - len(kept)
if n_dropped:
    logger.info(
        "LengthGroupedSampler: pre-filtered %d/%d samples with "
        "estimated length > %.0f tokens (1.2 * max_length %d).",
        n_dropped,
        len(dataset),
        filter_threshold,   # <- max_length + 512
        max_length,
    )
```

`%.0f` receives `filter_threshold`, so the message asserts that
`max_length + 512` *is* `1.2 * max_length`. At `max_length=32768` it prints:

```
LengthGroupedSampler: pre-filtered N/M samples with estimated length > 33280 tokens (1.2 * max_length 32768).
```

but `1.2 * 32768` is 39322, not 33280.

**How far the two diverge**

| `max_length` | code (`+512`) | stated (`1.2x`) | effective headroom | delta |
|---:|---:|---:|---:|---:|
| 2048 | 2560 | 2457 | 25.0% | +103 |
| 4096 | 4608 | 4915 | 12.5% | -307 |
| 8192 | 8704 | 9830 | 6.2% | -1126 |
| 16384 | 16896 | 19660 | 3.1% | -2764 |
| 32768 | 33280 | 39321 | 1.6% | -6041 |
| 131072 | 131584 | 157286 | 0.4% | -25702 |

A flat 512 is ~25% headroom at 2k and ~0.4% at 128k. Beyond `max_length=2560`
the filter is stricter than the documented rule, and the gap grows with context
length — 6k tokens' worth of samples at 32k, 25k at 128k.

[...]

PR #3799, opened 2026-09-03
fix(datasets): log the VLM pre-filter threshold that is actually applied

kabirvashisht4-glitch:
# What does this PR do ?

Makes the VLM `LengthGroupedSampler` pre-filter log state the threshold it
actually applies. No behaviour change.

`nemo_automodel/components/datasets/vlm/samplers.py` filtered with
`max_length + 512`, but the comment said "Use 1.2x headroom" and the log line
printed that flat threshold labelled as `1.2 * max_length` — so the message
asserted that `max_length + 512` *is* `1.2 * max_length`. Reproduced on `main`:

```
LengthGroupedSampler: pre-filtered 1/3 samples with estimated length > 2560 tokens (1.2 * max_length 2048).
```

`1.2 * 2048` is 2457, not 2560.

The two readings diverge as context grows, so this is not only cosmetic:

| `max_length` | code (`+512`) | stated (`1.2x`) | effective headroom |
|---:|---:|---:|---:|
| 2048 | 2560 | 2457 | 25.0% |
| 8192 | 8704 | 9830 | 6.2% |
| 32768 | 33280 | 39321 | 1.6% |
| 131072 | 131584 | 157286 | 0.4% |

Beyond `max_length=2560` the filter is stricter than the documented rule, by
~6k tokens' worth of samples at 32k. The comment's stated design is that
borderline samples are *not* dropped here — they go to
`PreTokenizedDatasetWrapper`'s precise tokenize-and-retry, because
`self.lengths[i]` is only an estimate. The log is the only signal a user gets
for why samples left the run, so it has to name the real rule.

After:

```
LengthGroupedSampler: pre-filtered 1/3 samples with estimated length > 2560 tokens (max_length 2048 + 512 headroom).
```

**Scope.** This PR fixes only the reporting, which is true regardless of intent.
Whether the flat `+512` or the `1.2x` it claimed is the right *policy* is a
separate call I have deliberately not made — see #3798. If 1.2x was intended,
say so and I will send `filter_threshold = int(max_length * 1.2)` instead.

# Changelog

- Comment and log message in `LengthGroupedSampler.__init__` now state
  `max_length + 512`; the filter itself is unchanged.
- `%.0f` → `%d` for the threshold, which is an int.
- New `TestPreFilterThreshold` in `tests/unit_tests/datasets/vlm/test_samplers.py`:
  a sample exactly at the threshold is kept, one token over is dropped, and the
  logged threshold matches the applied one.

**Pre checks**:

The stale comment was the documentation; it is corrected here.

**Verification** (CPU, no GPU, dataset or processor needed):

- Reverting the source hunk makes
  `test_log_reports_the_threshold_actually_applied` fail. The two threshold
  tests pass either way by design — they pin the filter behaviour this must not
  change.
- `pytest tests/unit_tests/datasets/vlm/` → 368 passed, no regressions.

# Additional Information

- Related to #3798

### DIFF
--- a/nemo_automodel/components/datasets/vlm/samplers.py
+++ b/nemo_automodel/components/datasets/vlm/samplers.py
@@ -114,8 +114,8 @@ class LengthGroupedSampler(Sampler):
         # rare case where the heuristic underestimates the true tokenized length.
         all_indices = range(len(dataset))
         if max_length is not None:
-            # Use 1.2x headroom: the estimated length is a heuristic, so only
-            # drop samples that are clearly overlong.  Borderline samples are
+            # Fixed 512-token headroom: the estimated length is a heuristic, so
+            # only drop samples that are clearly overlong.  Borderline samples are
             # left to PreTokenizedDatasetWrapper's precise tokenize-and-retry.
             filter_threshold = max_length + 512
             kept = [i for i in all_indices if self.lengths[i] <= filter_threshold]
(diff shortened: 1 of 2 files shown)


yuhezhang-ai (approved):
Thank you, LGTM

PR #3813, opened 2026-09-05
fix(models): honor TE linear backend and native initialization

jquesnelle:
# What does this PR do?

Honor backend.linear="te" for standard projections left by custom-model constructors while preserving native model initialization and parameter identity.

# Changelog

- Extend the existing backend override pass to replace exact nn.Linear modules when the TE linear backend is requested.
- Introduce a TE-backed, nn.Linear-compatible subtype so model-owned initialization continues to recognize the live projection.
- Preserve parameter objects, shared/tied weights, aliases, frozen parameters, training mode, and initialization flags.
- Retain existing TE backend patches and dynamically dispatch to TE’s forward implementation.
- Add CPU-compatible initialization-contract tests and real-TE forward/backward tests.

Pre checks:

- [x] Did you add or update any necessary documentation? The compatibility subtype documents its inheritance and execution contract.

# Additional Information

## Motivation

During Qwen3-style dense pretraining, requesting backend.linear="te" left Hugging Face-derived projections as ordinary nn.Linear modules. The generic backend override pass handled Quack but not TE. Consequently, enabling TE quantization could appear successful while leaving those matrix multiplications unquantized.

Replacing modules alone is insufficient for checkpoint-free training. TE Linear is not normally an nn.Linear subclass, so model initialization based on `isinstance(module, nn.Linear)` can skip the replacement after meta-device materialization.

The fix preserves native initialization dispatch rather than imposing a generic Gaussian initializer or wrapping model owners in closures. This matters for nested configurations, identity-sensitive initialization such as module is `self.score`, and models copied before materialization for pipeline construction.

Specialized Linear subclasses and existing native TE modules remain model-owned; this conversion targets exact `nn.Linear` instances.

## Validation

Executed through SLURM with PyTorch 2.10.0+cu128 and TE 2.15.0:

- Final model-init unit and TE functional suites: 92 passed.
- Coverage includes nested and non-Gaussian initialization policies, zero initialization scales, bias/biasless projections, shared/frozen parameters, initialization flags, and deepcopy before materialization.
- Identity-sensitive BF16 head initialization is tested.
- Real-TE tests cover Qwen3 BF16/MXFP8 forward/backward and parity against native TE.
- Scoped Ruff lint/format and whitespace checks passed.

## Test command:

    python -m pytest tests/unit_tests/_transformers/test_model_init.py tests/functional_tests/training/test_te_linear_backend_init.py -q

## Broader test disclosure

The broader Transformers/Qwen3/TE test selection produced 749 passed, 1 failed, 2 skipped. The failure was:

    TestApplyPreloadOverridesPacked::test_packed_non_flash_falls_back_to_available_flash

It expected flash_attention_2 but received sdpa. The same failure reproduced on the pre-fix base, so it is not introduced by this change.

## Limitations

Validation does not establish numerical parity for every custom architecture or a complete multi-rank training matrix. No performance claims or cluster-specific dependency overrides are included.

### DIFF
--- a/nemo_automodel/components/models/common/te_linear.py
+++ b/nemo_automodel/components/models/common/te_linear.py
@@ -1,0 +1,64 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Linear type compatibility for projections replaced in HuggingFace models."""
+
+import torch
+
+from nemo_automodel.components.models.common.utils import _patch_te_modules
+from nemo_automodel.shared.import_utils import safe_import_from
+
+HAVE_TE, TransformerEngineLinear = safe_import_from("transformer_engine.pytorch", "Linear")
+if not HAVE_TE:
+    raise ImportError("linear='te' requires Transformer Engine. Install nemo-automodel[cuda].")
+
+
+class TELinear(torch.nn.Linear, TransformerEngineLinear):
+    """Execute TE kernels while retaining native ``nn.Linear`` initialization hooks.
+
+    ``nn.Linear`` must precede TE in the MRO: TE's cooperative base constructor
+    must reach ``nn.Module.__init__``, not ``nn.Linear.__init__`` (which requires
+    dimensions). Explicitly select TE's constructor, forward, and reset so the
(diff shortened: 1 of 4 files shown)


nemo_automodel/components/models/common/te_linear.py
```diff
@@ -0,0 +1,64 @@
...
+        Args:
+            inp: Tensor of shape [..., in_features], with arbitrary leading dimensions.
+            is_first_microbatch: Whether to refresh the cached FP8 weights.
+
+        Returns:
+            Tensor of shape [..., out_features], preserving the leading dimensions.
+        """
+        return TransformerEngineLinear.forward(self, inp, is_first_microbatch=is_first_microbatch)
+
+    reset_parameters = TransformerEngineLinear.reset_parameters
```
HuiyingLi, line 64: Since this class is also an `nn.Linear`, model-owned initialization may call `module.reset_parameters()`. This currently uses TE’s default initializer rather than `nn.Linear`’s Kaiming initializer. Full-checkpoint SFT is unaffected because those values are overwritten, but checkpoint-free or missing-weight paths may not preserve the original initialization contract. Could we preserve the no-argument `nn.Linear.reset_parameters()` behavior, or document this limitation?

PR #3814, opened 2026-09-05
fix(checkpoint): restore serialized module extra state safely

jquesnelle:
# What does this PR do?

Restore serialized module extra state correctly when its saved size differs from a freshly constructed module’s placeholder, without silently discarding checkpoint state.

# Changelog

- Read typed checkpoint storage metadata when checking module extra-state compatibility.
- Allocate checkpoint-sized destination buffers for one-dimensional uint8 serialized extra state, leaving interpretation and restoration to the owning module.
- Recognize dictionary/list extra state represented by flattened DCP descendant keys.
- Reject missing required extra state during resume, while retaining compatibility for initialization and empty serialized placeholders.
- Preserve ordinary tensor shape validation.
- Add real DCP/safetensors round-trip tests and Transformer Engine restore tests.

Pre checks:

- [x] Did you add or update any necessary documentation? Implementation comments and test documentation describe the supported compatibility behavior.

# Additional Information

## Motivation

We encountered this while resuming an NVFP4-trained model with BF16 execution. Transformer Engine serializes quantizer metadata into `_extra_state`; a saved payload can have a different length from the empty placeholder returned by a fresh module. DCP rejected the load with a size mismatch before the module could interpret its state.

Simply dropping mismatched entries is unsafe: the same size difference can occur when resuming the same recipe, and discarding that payload loses valid scaling/amax history. Dropping entries can also fail later during strict module loading.

This implementation restores the saved serialized payload rather than treating its length as a model-parameter shape. It also addresses nested extra state: DCP can flatten dictionaries into keys such as `layer._extra_state.counter`, which must not be mistaken for absent module state.

## Validation

Executed on B200 through SLURM with PyTorch 2.10.0+cu128 and Transformer Engine 2.15.0:

- Complete checkpoint unit directory plus real-TE restore tests: 387 passed, 18 skipped.
- Real DCP and safetensors tests cover growing, shrinking, equal-length, and empty byte payloads.
- Tests cover dictionary-valued state, missing required state, and rejection of ordinary tensor shape mismatches.
- Real TE tests cover delayed/current/NVFP4 checkpoint metadata, preservation of delayed-scaling history, and subsequent BF16 or current-scaling execution.
- Scoped Ruff lint/format and whitespace checks passed.

## Test command:

    python -m pytest tests/unit_tests/checkpoint tests/functional_tests/training/test_te_checkpoint_extra_state.py -q

## Limitations

These tests use locally generated, trusted checkpoints. TE’s upstream deserializer interprets serialized byte payloads; this change does not make untrusted checkpoints safe. Validation was single-GPU, not a complete multi-rank restore matrix.

### DIFF
--- a/nemo_automodel/components/checkpoint/checkpointing.py
+++ b/nemo_automodel/components/checkpoint/checkpointing.py
@@ -48,6 +48,7 @@ from safetensors.torch import load as safetensors_load
 from safetensors.torch import load_file, save_file
 from safetensors.torch import save as safetensors_save
 from torch import nn
+from torch.distributed.checkpoint.metadata import BytesStorageMetadata, TensorStorageMetadata
 from torch.distributed.checkpoint.storage import StorageReader, StorageWriter
 from torch.distributed.device_mesh import DeviceMesh
 from torch.nn.parallel import DistributedDataParallel
@@ -314,14 +315,14 @@ def _summarize_state_dict_key_diff(
     }
 
 
-def _get_checkpoint_metadata_keys(
+def _get_checkpoint_state_dict_metadata(
     path: str,
     storage_reader: StorageReader | None = None,
-) -> set[str]:
-    """Return checkpoint FQNs present in metadata."""
+) -> dict[str, TensorStorageMetadata | BytesStorageMetadata]:
+    """Return checkpoint FQN -> storage metadata mapping."""
     reader = storage_reader if storage_reader is not None else FileSystemReader(path)
     metadata = reader.read_metadata()
-    return set(metadata.state_dict_metadata.keys())
+    return dict(metadata.state_dict_metadata)
 
 
 if _is_geq_torch_2_9():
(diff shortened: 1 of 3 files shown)


HuiyingLi:
I independently ran a full-model save/resume check on an H100 at `37210f4a`:

- Stock `Qwen/Qwen3-0.6B` configuration: 28 layers and 596,049,920 parameters, freshly initialized.
- Converted all 197 linear projections to real TE Linear modules.
- Ran a DelayedScaling FP8 forward/backward, saved with `Checkpointer.save_model(..., model_save_format="safetensors")`, rebuilt a fresh model, and resumed with `load_model(..., is_init_step=False)`.
- The checkpoint contained 197 `_extra_state` entries; the fresh destination had zero initialized entries before load.
- After resume, 310/310 parameters and 197/197 TE FP8 states (forward/backward scale and amax history) matched exactly.
- The next FP8 training step also matched exactly: loss `11.995684624` on both models, 310/310 gradients equal, maximum gradient difference `0`.

This validates the full-model, no-state-dict-adapter DCP save/resume path. NVFP4 was not exercised because the node was H100.

HuiyingLi:
There is one remaining path for custom models with a `state_dict_adapter`: checkpoint save and DCP load-destination construction currently call `to_hf(..., exclude_key_regex=r".*_extra_state.*")`, so the module state never reaches the resize/restore logic in this PR.

I opened #3823 as a separate stacked draft. It keeps ordinary weight conversion in the adapter, carries native `_extra_state` FQNs through resumable DCP checkpoints, reattaches them after `from_hf`, and leaves them out of consolidated HF exports. I validated torch-save DCP, sharded safetensors, consolidated export, and a real TE FP8 restore on H100. The draft is currently stacked on this PR and will be rebased after this lands.

PR #3815, opened 2026-09-05
perf(training): accelerate grad clipping with fused Triton kernel

piyushumate:
# What does this PR do ?

Accelerates gradient norm calculation in `_clip_grad_norm_impl` by up to **6.49x** on NVIDIA H100 using a fused, zero-allocation 2D multi-tensor Triton reduction kernel, resolving CPU driver bottlenecks in distributed and MoE LLM training without memory overhead (#3719).

# Changelog

- Implemented `nemo_automodel/components/training/_fused_grad_clipping.py` with 2D chunk-parallel Triton kernels (`_multi_tensor_max_2d_kernel`, `_multi_tensor_scaled_l2_2d_kernel`) supporting `bfloat16`, `float16`, `float32`, and `float64`.
- Updated `_clip_grad_norm_impl` in `nemo_automodel/components/training/utils.py` to dispatch to the fused Triton kernel when running on CUDA, with clean eager fallback for CPU/non-CUDA and preservation of all DTensor sharding placements and device mesh communications.
- Added comprehensive unit and functional tests in `tests/functional_tests/llm_pretrain_and_kd/test_fused_grad_clipping.py`.
- Verified strict numerical parity against PyTorch eager baseline on NVIDIA H100 (relative difference: 0.0002%, well within BF16 1.6% precision).

### Benchmark Results (NVIDIA H100 SXM 80GB)

Workload: 500 gradient matrices totaling **10.26 Billion elements (21.00 GB)** across realistic 70B/MoE model layers:

| Implementation | Latency (ms) | Speedup vs Baseline | Speedup vs Foreach | Memory Overhead |
| :--- | :---: | :---: | :---: | :---: |
| Sequential Baseline (`main`) | 176.211 ms | 1.00x | — | 0 MB (Sequential) |
| PyTorch Foreach | 33.833 ms | 5.21x | 1.00x | ~21.00 GB (Temporary Buffer) |
| **Fused Triton 2D Kernel** | **27.150 ms** | **6.49x** | **1.25x** | **0 MB (Zero-Alloc)** |

* **Latency reduction**: From **176.2 ms down to 27.1 ms** (saving ~149 ms per step, or ~4.14 hours of cluster GPU time per 100k steps).
* **Zero extra memory**: In-place fused accumulation avoids the ~21 GB intermediate buffer required by `torch._foreach_div`.

# Before your PR is "Ready for review"

**Pre checks**:

[...]

PR #3817, opened 2026-09-05
fix(training): reduce grad-clip norms across the expert-parallel axis

### LINKED ISSUE #3816: grad clipping computes a different "global" norm on every EP group with TE experts
**what happens**

with `model.backend.experts: te` and `ep_size > 1`, grad clipping does not compute one global norm for the whole model. each EP group computes its own norm and clips with it.

the reason is that TE expert params are plain tensors (or DTensors sharded only on ep_shard), so the clip code never adds up their part of the norm across the EP axis. it only reduces over the dims it can see on each param's own mesh.

**why it matters**

dense params (attention, embeddings, etc) are shared across EP groups through FSDP. each group clips its shard of the same dense param by a different factor, so the shared weights slowly drift apart between groups. clipping is on by default (max_norm 1.0), so this is active in normal runs. nothing in the logs shows it, every rank prints a grad norm that looks fine.

**repro**

cpu, 2 gloo ranks acting as 2 EP groups, real `clip_grad_norm`. each rank has different expert grads plus one dense param sharded across both ranks:

```
correct global grad norm (same on every rank): 40.201493
rank 0: total_norm=40.199502  dense_grad_after_clip=2.487593e-02
rank 1: total_norm=4.019950   dense_grad_after_clip=2.487592e-01
total_norm ratio rank0/rank1: 10.000x
```

same dense param, clipped 10x differently depending on the rank. with ep_shard > 1 (experts DTensor-wrapped on the ep_shard mesh only) the norms agree inside an EP group but differ 7.1x across groups.

**who is affected**

every shipped yaml with `experts: te` and `ep_size > 1` that uses the shared MoE experts, 20 configs on main today: qwen3 moe 30b/235b, deepseek_v3, gpt_oss 20b/120b, and moonlight. models that leave `experts` at its default (`torch_mm`) are not affected.

**related**

while fixing this we found the ep grad scaling regex only matches `mlp.experts.*`. models that attach the moe block as `moe` (gemma4_moe, diffusion_gemma, step3p5) miss the existing ep grad scaling too.

**fix**

pr with the fix and multi-rank tests: #3817

stanley1208:
Fixes #3816.

grad clipping never reduced the TE expert part of the norm across the EP axis, so each EP group computed its own "global" norm and clipped the same dense params by different factors (10x apart in the repro). repro and details are in the issue.

**what changed**

- expert params are found by structure, not by name. `GroupedExpertsTE` sets a marker when it is built, and one scan of the model feeds both the existing ep grad scaling and the new ep reduction inside the clip. this replaces the `mlp.experts` regex, which missed gemma4_moe, diffusion_gemma and step3p5 (they attach the moe block as `moe`), so their ep grad scaling is fixed too
- the sharding groups come from the model structure instead of whatever grads exist this step. every rank runs the same collectives in the same order, so a rank with no expert grads joins with zero instead of hanging the others
- the ep mesh is resolved with `get_flat_mesh` (flattened axes are not in `mesh_dim_names`), the torch fast path is skipped when ep expert params exist, the comm-group prewarm also warms the ep group, and the seq-cls recipe and the pipelining guide now pass the moe mesh through

**tests**

they run the real `scale_grads_and_clip_grad_norm` over gloo ranks: both ep layouts (plain tensors and ep_shard DTensors), the production shape with ep grad scaling on, a rank without expert grads, the torch fast path, inf norm, ep_size 1, and no moe mesh. all fail on main except the two fallback cases.

**note**

the ep reduction adds two small scalar collectives per expert group per step. #3685 and #3815 are both reworking these loops for perf and could fold the ep_shard+ep chain into one flattened group there.

### DIFF
--- a/docs/guides/pipelining.mdx
+++ b/docs/guides/pipelining.mdx
@@ -610,7 +610,9 @@ from nemo_automodel.components.training.utils import clip_grad_norm
 # Call these after backward on every pipeline rank.
 scale_grads_by_divisor(ap.info.stages, divisor=8)
 
-# Use the PP-aware clipping helper across the local model parts.
+# Use the PP-aware clipping helper across the local model parts. For MoE
+# models with expert parallelism, also pass moe_mesh/ep_axis_name so expert
+# gradient norms are reduced across the EP axis.
 grad_norm = clip_grad_norm(
     1.0,
     ap.parts,
@@ -618,6 +620,8 @@ grad_norm = clip_grad_norm(
     pp_enabled=True,
     device_mesh=world_mesh,
     pp_axis_name="pp",
+    moe_mesh=moe_mesh,
+    ep_axis_name="ep" if moe_mesh is not None and "ep" in (moe_mesh.mesh_dim_names or ()) else None,
 )
 ```
 
--- a/docs/guides/pipelining.mdx
+++ b/docs/guides/pipelining.mdx
@@ -1,21 +1,21 @@
 ---
 title: "Pipeline Parallelism with AutoPipeline"
-description: ""
+description: "Configure AutoPipeline for Hugging Face models, mix pipeline parallelism with other strategies, and debug pipeline stages."
 position: 3
 ---
 ## Introduction
 
 As large language models continue to grow in size, training and fine-tuning them efficiently across multiple GPUs has become increasingly challenging. While data parallelism works well for smaller models, models with billions of parameters require more sophisticated parallelization strategies to overcome memory constraints and communication overhead.
 
-Pipeline parallelism addresses these challenges by splitting a model's layers across different devices and processing them in a pipelined fashion. Each device processes a different stage of the model, enabling training of models that wouldn't fit on a single device while maintaining high GPU utilization through overlapped computation.
+Pipeline parallelism addresses these challenges by splitting a model's layers across different devices and processing them in a pipelined fashion. Each device processes a different stage of the model, enabling training of models that would not fit on a single device while maintaining high GPU utilization through overlapped computation.
 
-AutoPipeline is NeMo AutoModel's high-level pipeline parallelism interface for Hugging Face-compatible models. Built on PyTorch's native `torch.distributed.pipelining`, it expects an HF-style model with a `config`, a supported decoder-layer layout, and pipeline-compatible forwards. Recipe-driven pipeline parallelism also requires the selected model implementation to advertise a `_pp_plan` (or use the supported MoE path) and pass the repository's pipeline validation; encoder-decoder models and models with actually tied input/output embedding weights are rejected.
+AutoPipeline is NeMo AutoModel's high-level pipeline parallelism interface for Hugging Face-compatible models. Built on PyTorch's native `torch.distributed.pipelining`, it expects a Hugging Face-style model with a `config`, a supported decoder-layer layout, and pipeline-compatible forwards. Recipe-driven pipeline parallelism also requires the selected model implementation to advertise a `_pp_plan` (or use the supported MoE path) and pass the repository's pipeline validation; encoder-decoder models and models with actually tied input/output embedding weights are rejected.
 
-The functional module at `nemo_automodel.components.distributed.pipelining.functional` exposes the same HF-specific `pipeline_model()` splitter plus lower-level stage and schedule helpers. Custom PyTorch architectures can reuse helpers such as `stage_ids_this_rank()`, `calculate_virtual_stages()`, and `build_pipeline_schedule()` after their own code constructs valid `PipelineStage` objects; `pipeline_model()` is not a generic `nn.Module` splitter.
+The functional module at `nemo_automodel.components.distributed.pipelining.functional` exposes the same Hugging Face-specific `pipeline_model()` splitter plus lower-level stage and schedule helpers. Custom PyTorch architectures can reuse helpers such as `stage_ids_this_rank()`, `calculate_virtual_stages()`, and `build_pipeline_schedule()` after their own code constructs valid `PipelineStage` objects; `pipeline_model()` is not a generic `nn.Module` splitter.
 
-This guide walks you through AutoPipeline for compatible Hugging Face models and the lower-level helpers for manually constructed stages. You'll learn how to configure pipeline stages, integrate with existing training workflows, optimize performance, and combine pipeline parallelism with other parallelization strategies.
+This guide walks you through AutoPipeline for compatible Hugging Face models and the lower-level helpers for manually constructed stages. You will learn how to configure pipeline stages, integrate with existing training workflows, optimize performance, and combine pipeline parallelism with other parallelization strategies.
 
-**Prerequisites:**
+## Prerequisites
 
 ```bash
 # Install uv from https://docs.astral.sh/uv/getting-started/installation/
@@ -29,8 +29,8 @@ uv pip install nemo-automodel
 uv pip install git+https://github.com/NVIDIA-NeMo/Automodel.git
 ```
 <Info>
-Before proceeding with this guide, please ensure that you have NeMo AutoModel installed on your machine.
-For a complete guide and additional options please consult the AutoModel [Installation Guide](/get-started/installation).
+Before proceeding with this guide, ensure that you have NeMo AutoModel installed on your machine.
+For a complete guide and additional options, consult the NeMo AutoModel [Installation Guide](/get-started/installation).
 
 </Info>
 
@@ -38,14 +38,14 @@ For a complete guide and additional options please consult the AutoModel [Instal
 
 AutoPipeline provides the following capabilities:
 
-- **Hugging Face-Compatible Model Support**: Works with decoder-only causal language models and explicitly supported VLMs that satisfy the pipeline model contract
-- **PyTorch Native Integration**: Built on PyTorch's `torch.distributed.pipelining` for optimal performance
-- **Flexible Configuration**: Multiple scheduling strategies, configurable microbatch sizes, and automatic or manual layer splitting
-- **Mixed Parallelism Support**: Combine pipeline parallelism with data parallelism, tensor parallelism, and FSDP
-- **Modular Functional API**: Lower-level stage and schedule helpers can be reused with manually constructed `PipelineStage` objects
-- **Minimal Opinions**: Easy to extend and integrate with existing training workflows
+- **Hugging Face-Compatible Model Support**: Works with decoder-only causal language models and explicitly supported VLMs that satisfy the pipeline model contract.
+- **PyTorch Native Integration**: Built on PyTorch's `torch.distributed.pipelining` for optimal performance.
+- **Flexible Configuration**: Multiple scheduling strategies, configurable microbatch sizes, and automatic or manual layer splitting.
+- **Mixed Parallelism Support**: Combine pipeline parallelism with data parallelism, tensor parallelism, and FSDP.
+- **Modular Functional API**: Lower-level stage and schedule helpers can be reused with manually constructed `PipelineStage` objects.
+- **Minimal Opinions**: Easy to extend and integrate with existing training workflows.
 
-## Quick Start with AutoPipeline (Hugging Face Models)
+## Quickstart with AutoPipeline (Hugging Face Models)
 
 Here's a minimal example to get started with AutoPipeline using 2 pipeline stages with a Hugging Face model:
 
@@ -91,7 +91,7 @@ if __name__ == "__main__":
     print(ap.pretty_print_stages())
 ```
 
-### Run the Quick Start Example
+### Run the Quickstart Example
 
 Save the above code as `pipeline_example.py` and run with:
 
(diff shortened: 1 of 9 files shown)


docs/guides/pipelining.mdx
jgerh:
```suggestion
description: "Configure AutoPipeline for Hugging Face models, mix pipeline parallelism with other strategies, and debug pipeline stages."
```

jgerh:
```suggestion
Pipeline parallelism addresses these challenges by splitting a model's layers across different devices and processing them in a pipelined fashion. Each device processes a different stage of the model, enabling training of models that would not fit on a single device while maintaining high GPU utilization through overlapped computation.
```

jgerh:
```suggestion
AutoPipeline is NeMo AutoModel's high-level pipeline parallelism interface for Hugging Face-compatible models. Built on PyTorch's native `torch.distributed.pipelining`, it expects a Hugging Face-style model with a `config`, a supported decoder-layer layout, and pipeline-compatible forwards. Recipe-driven pipeline parallelism also requires the selected model implementation to advertise a `_pp_plan` (or use the supported MoE path) and pass the repository's pipeline validation; encoder-decoder models and models with actually tied input/output embedding weights are rejected.
```

jgerh:
```suggestion
The functional module at `nemo_automodel.components.distributed.pipelining.functional` exposes the same Hugging Face-specific `pipeline_model()` splitter plus lower-level stage and schedule helpers. Custom PyTorch architectures can reuse helpers such as `stage_ids_this_rank()`, `calculate_virtual_stages()`, and `build_pipeline_schedule()` after their own code constructs valid `PipelineStage` objects; `pipeline_model()` is not a generic `nn.Module` splitter.
```

jgerh:
```suggestion
This guide walks you through AutoPipeline for compatible Hugging Face models and the lower-level helpers for manually constructed stages. You will learn how to configure pipeline stages, integrate with existing training workflows, optimize performance, and combine pipeline parallelism with other parallelization strategies.
```

jgerh:
```suggestion
Before proceeding with this guide, ensure that you have NeMo AutoModel installed on your machine.
For a complete guide and additional options, consult the NeMo AutoModel [Installation Guide](/get-started/installation).
```

jgerh:
```suggestion
- **Hugging Face-Compatible Model Support**: Works with decoder-only causal language models and explicitly supported VLMs that satisfy the pipeline model contract.
- **PyTorch Native Integration**: Built on PyTorch's `torch.distributed.pipelining` for optimal performance.
- **Flexible Configuration**: Multiple scheduling strategies, configurable microbatch sizes, and automatic or manual layer splitting.
- **Mixed Parallelism Support**: Combine pipeline parallelism with data parallelism, tensor parallelism, and FSDP.
- **Modular Functional API**: Lower-level stage and schedule helpers can be reused with manually constructed `PipelineStage` objects.
- **Minimal Opinions**: Easy to extend and integrate with existing training workflows.
```

jgerh:
```suggestion
## Prerequisites
```

jgerh:
```suggestion
## Quickstart with AutoPipeline (Hugging Face Models)
```

jgerh:
```suggestion
### Run the Quickstart Example
```

jgerh:
```suggestion
### Model Patching Flags
```

jgerh:
```suggestion
AutoPipeline splits a model by deep-copying it per stage and pruning away modules that do not belong to that stage. Many Hugging Face models assume the full module tree is present and return `ModelOutput` objects; after pruning, their original `forward()` often breaks (or returns objects that are awkward to pipeline).
```

jgerh:
```suggestion
  For compilation and performance, this patched forward prefers a precomputed `causal_mask_mapping` dict (it will fall back to computing masks and warn if you do not provide it).
```

jgerh:
```suggestion
#### When to Change These Flags
```

jgerh:
```suggestion
  - Handles sliced layer containers (for example, `layers` becoming dict-like after stage pruning) and returns a **tensor** of hidden states so stages can be chained.
```

jgerh:
```suggestion
- **Leave both `True` (default)** for standard Hugging Face `AutoModelForCausalLM` / `...ForCausalLM` models. This is the common case and gives the expected behavior: token IDs to hidden states to logits across stages.
- **Set both `False`** when your model already has a pipeline-friendly forward. That forward returns tensors and can accept hidden states when embeddings are absent. Set both to `False` when the model needs custom kwargs or paths that the Hugging Face patch does not preserve. That case is common for NeMo AutoModel-native implementations, packed-sequence/`thd` paths, and extra arguments such as `padding_mask`. Many benchmark configs for NeMo-native models do this (for example, `examples/llm_benchmark/qwen/qwen3_moe_30b_torch.yaml`).
```

jgerh:
```suggestion
### Automatic Compared to Manual Layer Distribution
```

jgerh:
```suggestion
When AutoPipeline splits your model, it distributes components across pipeline stages. A typical split looks like this:
```

jgerh:
```suggestion
- **Embeddings** only exist on the first stage.
- **Language modeling head** only exists on the last stage.
- **Rotary embeddings** are shared across all stages (for position encoding).
- **Transformer layers** are evenly distributed.
```

jgerh:
```suggestion
### Add a Parallelization Callback to Hugging Face-Compatible Splitting
```

jgerh:
```suggestion
When using the Hugging Face-compatible `pipeline_model()` path, a callback can apply additional parallelism to each model part. Its signature must match `ParallelizeFnProtocol`:
```

jgerh:
```suggestion
1. **Module Naming**: Ensure your model has consistent module naming that can be mapped to stages.
2. **State Management**: Handle model state (embeddings, buffers) carefully across stages.
3. **Communication**: First and last stages need special handling for inputs and outputs.
4. **Splitting Ownership**: Keep custom model splitting in the model implementation; do not pass an arbitrary `nn.Module` to `pipeline_model()`.
5. **Testing**: Start with a small model and verify correct splitting before scaling up.
```

jgerh:
```suggestion
- `--distributed.pipeline.pp_schedule`: Pipeline schedule (`1f1b`, `interleaved1f1b`, `LoopedBFS`, and so on)
```

jgerh:
```suggestion
#### Pipeline and Data Parallelism (4 GPUs Total)
```

jgerh:
```suggestion
#### Pipeline and Tensor Parallelism (4 GPUs Total)
```

jgerh:
```suggestion
#### Full Hybrid with PP, DP, and TP (8 GPUs Total)
```

jgerh:
```suggestion
AutoPipeline integrates with NeMo AutoModel's recipe system. The following YAML configuration is a complete example:
```

jgerh:
```suggestion
**The model does not fit in memory:**
- Increase the number of pipeline stages.
- Reduce microbatch size.
- Enable gradient checkpointing.
```

jgerh:
```suggestion
- Increase `step_scheduler.local_batch_size` to have more microbatches.
- Try different schedules (for example, `interleaved1f1b`).
- Adjust virtual stages configuration.
```

jgerh:
```suggestion
- Use manual module assignment for fine control.
- Adjust `layers_per_stage` parameter.
- Check parameter counts with `get_stage_param_counts()`.
```

jgerh:
```suggestion
AutoPipeline and `pipeline_model()` provide the repository's Hugging Face-compatible splitting path, while lower-level functional helpers can be reused after custom code constructs its own pipeline stages.
```

jgerh:
```suggestion
- Pipeline parallelism enables training of models too large for a single GPU.
- AutoPipeline provides a simple API for Hugging Face-compatible models with customization options.
- Lower-level functional helpers can schedule manually constructed stages for custom models.
- Both can be combined with other parallelization strategies for optimal performance.
- Use built-in monitoring tools to understand and optimize your pipeline.
```

jgerh (commented):
Completed tech pubs review of docs/guides/pipelining.mdx, providing copyedits and readability enhancements to ensure adherence to style guides.

stanley1208:
thanks @jgerh, just applied all 32 suggestions

stanley1208:
merged main in again, would need someone start CI, thanks!

stanley1208:
@akoumpa merged main in after #3685 landed
the two new fast paths were skipping the expert-parallel reduction, so i routed them through it and added tests for both.

nemo_automodel/components/training/utils.py
```diff
@@ -323,12 +370,10 @@ def _clip_grad_norm_impl(
...
-            if is_dtensor and not has_partial:
-                mesh = first.device_mesh
-                for dim_idx, pl in enumerate(first.placements):
-                    if isinstance(pl, Replicate):
-                        continue
-                    group_val = _all_reduce_scalar(group_val, reduce_op, mesh, dim_idx)
+            # multi_tensor_* hands back a CPU zero when the group has no local
+            # gradients, so place it before any collective. Same helper as the other
+            # paths: an EP-local group is reduced over the EP axis too.
+            group_val = _reduce_group_scalar(group_val.to(target_device), reduce_op)
```
yuhezhang-ai, line 388: Thanks for addressing the EP norm issue. Could we make the fused/fallback selection consistent across ranks when a group has no local gradients before merging?

With Triton importable, `_use_fused_grad_norm([], 2.0)` returns `True` because `all([])` is true, while a nonempty group of CPU gradients returns `False`. Since this PR now retains groups with no local gradients, a CPU/Gloo run (or CPU-offloaded gradients) can take different paths for the same EP group:

- Rank with no expert gradients: fused path, one `SUM` reduction.
- Rank with CPU expert gradients: fallback path, `MAX` followed by `SUM`.

The collective operations and counts therefore differ, which can fail or hang. I verified the selector behavior against this head, but have not run a distributed reproduction.

Could we select a consistent reduction path even for empty groups and extend the no-expert-gradients regression case to cover CPU gradients with Triton installed? The CUDA case also needs to remain consistent; simply rejecting empty lists would reverse the mismatch when peers use the CUDA fused path.

stanley1208: thanks for the catch.

before:
    return all(p.grad is not None and p.grad.is_cuda for p in params)
after:
    return representative.is_cuda

the old one asked about this rank's grads, so an empty group said True and a
cpu group said False. the representative is the same param on every rank, so
everyone picks the same path: empty and cuda groups both.

yuhezhang-ai:
/ok to test [b7eeb14](https://github.com/NVIDIA-NeMo/Automodel/pull/3817/commits/b7eeb145725e31cee09517da3dd89853d96d3a87)

tests/unit_tests/training/test_clip_grad_norm_ep.py
```diff
@@ -0,0 +1,384 @@
...
+
+def _scenario_fused_backend(rank: int) -> None:
+    """The fused multi-tensor reducer must not bypass the EP reduction."""
+    moe_mesh = init_device_mesh("cpu", (1, _WORLD), mesh_dim_names=("ep_shard", "ep"))
+
+    expert = nn.Parameter(torch.zeros(2, 8))
+    expert.grad = torch.full((2, 8), 10.0 if rank == 0 else 0.1)
+    model = _expert_model(expert)
+
+    with _fused_backend_forced():
```
yuhezhang-ai, line 290: Thanks for updating the selector—the representative-based decision addresses the original empty-gradient mismatch. One small test-coverage follow-up: `_fused_backend_forced()` now sets `HAVE_FUSED_GRAD_NORM = True`, but `_use_fused_grad_norm()` still returns `False` for the CPU parameters in this scenario. As a result, `_scenario_fused_backend` exercises the fallback path rather than the fused EP reduction.

Could we keep the new CPU empty-gradient regression using the real selector, and separately restore the selector override just for `_scenario_fused_backend` (as in the previous version), retaining the reference reducers and multi-rank numerical assertions? An assertion that the fused reducer was actually called would also help prevent this coverage from silently disappearing again.

This is a coverage gap, not a new observed production failure: the current assertions can pass on the fallback path, so green CI would not verify the fused EP reduction.

PR #3821, opened 2026-09-05
fix(deepseek-v4): advance PP media cursor for text-only microbatches

### LINKED ISSUE #3820: DeepSeek-V4-Vision PP media cursor stalls on text-only microbatches
### Bug

DeepSeek-V4-Flash-Vision pipeline training fails when a text-only microbatch precedes an image microbatch in the same batch.

Reproduced on `main@db379c1c` with the native Vision model, the real pipeline splitter, and `prepare_vlm_media_for_pp`, using CPU/Gloo. The media preparation function produces one slot per microbatch, including empty slots for text-only samples. The model advances `_vlm_chunk_idx` inside its `has_visual_tokens` branch, so a text-only microbatch leaves the cursor pointing at its empty slot. The next image microbatch receives that slot.

A two-microbatch batch with `n_images_per_sample=[0, 1]` produces:

```text
media_chunk_sizes [0, 9]
cursor_after_text_microbatch 0
ValueError: Found 1 image spans but got 0 grids
```

The failure occurs in the first pipeline stage's image-token merge. The integration from #3772 added this native media-consumption path. The existing collator emits zero image counts for text-only samples.

The cursor needs to consume one prepared media slot per first-stage microbatch, including text-only microbatches. I can prepare a scoped fix with mixed-media pipeline regression coverage.

ZenAlexa:
## Summary

Fixes #3820.

Advance the DeepSeek-V4-Vision staged-media cursor once per first-stage microbatch, including text-only microbatches. The media chunker reserves one slot per microbatch, including empty slots for text-only microbatches. The previous cursor skipped that slot, and the next image microbatch failed with `Found 1 image spans but got 0 grids`.

Media tensors are assigned to image microbatches, preserving the text-only input contract. Existing bounds checks, later-stage isolation, metadata-probe reset, and schedule cleanup remain in place.

This fixes the native Vision pipeline integration added in #3772 for [DeepSeek-V4-Flash-Vision-Exp](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-Vision-Exp).

## Validation

Python 3.12.14, PyTorch 2.14.0, Transformers 5.15.1, CPU:

- Expanded the existing staged-media regression to cover image-only and text/image/text/different-image sequences. Each result exactly matches a direct call with that microbatch's media. The original implementation produces **1 failed, 1 passed**.
- `pytest -q --cpu tests/unit_tests/models/deepseek_v4/test_dsv4_vision.py`: **60 passed, 1 skipped**. The skipped case requires CUDA.
- Real `prepare_vlm_media_for_pp` and `split_model_into_stages` reproduce the original failure. After the fix, `[0, 9]` patch slots advance the cursor to 1 after the text microbatch and complete the image microbatch successfully.
- A six-layer tiny Vision model split into two real model parts with `LoopedBFS`, hosted on one CPU/Gloo rank, matches the unsplit model exactly for both text and image logits. All 138 text-path and 157 image-path parameter gradients are finite and match exactly. This is a CPU model-part parity experiment; distributed GPU execution is outside this validation.
- Ruff lint, format checks, and `git diff --check` pass for both changed files.

### DIFF
--- a/nemo_automodel/components/models/deepseek_v4/model.py
+++ b/nemo_automodel/components/models/deepseek_v4/model.py
@@ -1379,24 +1379,25 @@ class DeepseekV4ForCausalLM(HFCheckpointingMixin, nn.Module, MoEFSDPSyncMixin):
         # PP VLM batches keep variable-length patch tensors off the pipeline
         # schedule and stage them per microbatch on the first model part. Pull
         # the matching chunk back here, where the DSV4-owned embedding bridge
-        # consumes it. Later PP stages have no embed_tokens and must neither
+        # consumes it. Text-only microbatches consume their empty media slots.
+        # Later PP stages have no embed_tokens and must neither
         # read nor advance the shared media cursor.
         on_first_stage = getattr(self.model, "embed_tokens", None) is not None
         if pixel_values is None and on_first_stage and getattr(self, "_vlm_pixel_values_chunks", None) is not None:
+            chunk_idx = int(getattr(self, "_vlm_chunk_idx", 0) or 0)
+            if chunk_idx >= len(self._vlm_pixel_values_chunks):
+                raise RuntimeError(
+                    f"DeepSeek-V4 PP media cursor {chunk_idx} exceeds "
+                    f"{len(self._vlm_pixel_values_chunks)} staged chunks"
+                )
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):
Thank you so much!

Issue #3825, opened 2026-09-07
Checkpoint size estimation runs on every rank before the is_rank_0 guard


kabirvashisht4-glitch:
**Describe the bug**

`_warn_if_large_inline_consolidation` performs its size estimation on **every
rank** and only then checks `is_rank_0()`. The work on all non-zero ranks is
discarded — but it includes a filesystem read of the Hugging Face safetensors
index and a walk of the full state dict, on every checkpoint save.

`nemo_automodel/components/checkpoint/checkpointing.py` L435-452:

```python
def _warn_if_large_inline_consolidation(config, state_dict, fqn_to_index_mapping, is_final_checkpoint=False):
    if not _should_write_consolidated_safetensors(config, is_final_checkpoint):
        return
    if config.save_consolidated != SaveConsolidatedMode.EVERY:
        return
    estimated_bytes = _get_original_hf_index_total_size(config)   # opens + json.load()s the HF index
    is_hf_index_estimate = estimated_bytes is not None
    if estimated_bytes is None:
        estimated_bytes = estimate_state_dict_bytes(state_dict)   # walks every tensor
    if not is_rank_0():                                           # <-- guard is here
        return
```

Both calls are purely local:

- `_get_original_hf_index_total_size` -> `get_safetensors_index_total_size`
  (`checkpoint/utils.py` L82-96) resolves a path and does `open()` +
  `json.load()`.
- `estimate_state_dict_bytes` (`utils.py` L71-79) iterates the whole state dict.

The function immediately above it, `_warn_if_inline_consolidation_enabled`
(L421-427), gets the ordering right — it returns on non-rank-0 *before* doing
anything — so the two adjacent warn helpers disagree.

**Consequence.** The output is a rank-0 log line, so this is not a correctness
bug. It is redundant work that scales with world size: on an N-rank job every
rank opens and parses the *same* HF index file from shared storage on every
checkpoint, which is the metadata-storm pattern that shared filesystems handle
worst. When the index is absent, all N ranks instead walk their full state dict
so that N-1 of them can discard the number.

It only triggers under `checkpoint.save_consolidated=every`, which is exactly
the configuration the warning exists to discourage — so the ranks paying this
cost are the ones already on the slow path.

**Steps/Code to reproduce bug**

CPU only, no GPU or distributed setup:

```python
from unittest import mock
import nemo_automodel.components.checkpoint.checkpointing as ckpt
[...]

PR #3826, opened 2026-09-07
fix(checkpoint): skip size estimation on non-rank-0 processes

kabirvashisht4-glitch:
# What does this PR do ?

Moves the `is_rank_0()` guard in `_warn_if_large_inline_consolidation` above the
size estimation, so only the rank that actually logs does the work.

`components/checkpoint/checkpointing.py` estimated first and checked rank second:

```python
estimated_bytes = _get_original_hf_index_total_size(config)   # open() + json.load()
is_hf_index_estimate = estimated_bytes is not None
if estimated_bytes is None:
    estimated_bytes = estimate_state_dict_bytes(state_dict)   # walks every tensor
if not is_rank_0():                                           # <- guard was here
    return
```

The output is one rank-0 log line, so every other rank paid for a result it
discarded. The estimate is not free:

- `_get_original_hf_index_total_size` → `get_safetensors_index_total_size`
  (`checkpoint/utils.py` L82-96) resolves a path, `open()`s it and `json.load()`s it.
- `estimate_state_dict_bytes` (`utils.py` L71-79) iterates the whole state dict.

On an N-rank job that is N processes opening the *same* HF index on shared
storage on every checkpoint save — the metadata-storm pattern shared
filesystems handle worst — or, when the index is absent, N full state-dict
walks so N-1 ranks can throw the number away. It only fires under
`save_consolidated=every`, which is exactly the config the warning exists to
discourage, so the ranks paying the cost are already on the slow path.

The helper immediately above it, `_warn_if_inline_consolidation_enabled`
(L421-427), already returns on non-rank-0 before doing anything, so the two
adjacent warn helpers disagreed. This makes them consistent.

**Why moving the guard is safe.** Neither helper is collective — one is a local
file read, the other is documented as estimating "without materializing
tensors" — and nothing between the two guards communicates. Had either
participated in a collective, having only rank 0 execute it would deadlock and
the original ordering would have been deliberate. I checked that first; it is
not.

This is redundant I/O, not a correctness bug: the emitted warning is unchanged.

# Changelog

- `is_rank_0()` check moved above the estimation in
  `_warn_if_large_inline_consolidation`, with a comment recording why skipping
  the helpers on other ranks cannot deadlock.
- Two CPU tests in `TestOfflineConsolidationScriptAndWarnings`:
  a non-rank-0 process calls neither helper; rank 0 still does.

**Pre checks**:

No user-facing behaviour changed, so no doc update was needed.

- Reverting the source hunk fails `test_non_rank_0_skips_size_estimation_entirely`.
  `test_rank_0_still_estimates` passes either way by design — it pins the
  behaviour this must not break, so the guard cannot silence the warning.
- The tests assert on the externally observable effect (whether the helpers are
  invoked) rather than mirroring the implementation.
- `pytest tests/unit_tests/checkpoint/` → 361 passed, 19 skipped. The 3 failures
  in that directory (`test_optimizer_state.py` x2,
  `test_rank_local_rng_checkpoint.py` x1) reproduce identically on unmodified
  `main` — they need multiple ranks — and are unrelated to this change.

# Additional Information

- Related to #3825

### DIFF
--- a/nemo_automodel/components/checkpoint/checkpointing.py
+++ b/nemo_automodel/components/checkpoint/checkpointing.py
@@ -443,12 +443,16 @@ def _warn_if_large_inline_consolidation(
         return
     if config.save_consolidated != SaveConsolidatedMode.EVERY:
         return
+    # Only rank 0 emits this warning, so bail out before estimating. Neither
+    # helper below is collective -- one is a local file read, the other walks
+    # the state dict without materializing tensors -- so skipping them on the
+    # other ranks cannot deadlock.
+    if not is_rank_0():
+        return
     estimated_bytes = _get_original_hf_index_total_size(config)
     is_hf_index_estimate = estimated_bytes is not None
     if estimated_bytes is None:
         estimated_bytes = estimate_state_dict_bytes(state_dict)
-    if not is_rank_0():
-        return
     if estimated_bytes is None or estimated_bytes < _CONSOLIDATED_SIZE_WARNING_THRESHOLD_BYTES:
         return
     world_size = get_world_size_safe()
--- a/tests/unit_tests/checkpoint/test_checkpointing.py
+++ b/tests/unit_tests/checkpoint/test_checkpointing.py
@@ -2591,6 +2591,56 @@ class TestOfflineConsolidationScriptAndWarnings:
         assert "1 output file, world_size=64" in caplog.text
         assert "~64.0 GiB" in caplog.text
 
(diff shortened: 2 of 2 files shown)


yuhezhang-ai:
/ok to test [4093c77](https://github.com/NVIDIA-NeMo/Automodel/pull/3826/commits/4093c771033ca0cb84523614b45b2b97809e968b)

yuhezhang-ai (approved):

PR #3830, opened 2026-09-08
fix(recipes): normalize seq-cls loss and gradients by a global label count

### LINKED ISSUE #3829: Sequence-classification recipe scales gradients by grad_accumulation_steps
**Describe the bug**

`TrainFinetuneRecipeForSequenceClassification` accumulates gradients over
micro-batches without dividing by the micro-batch count, so the gradient
magnitude scales linearly with `grad_accumulation_steps`. Training with
gradient accumulation applies an effectively multiplied learning rate.

`nemo_automodel/recipes/llm/train_seq_cls.py` L90 and L227-242:

```python
self.loss_fn = torch.nn.CrossEntropyLoss()          # reduction="mean"
...
for batch in batches:                                # the grad-accumulation loop
    ...
    loss = self.loss_fn(logits, labels.view(-1))     # mean over THIS micro-batch
    ...
    (loss * self._get_dp_group_size(include_cp=True)).backward()
```

Each `.backward()` adds to the existing `.grad`, so accumulating `N`
micro-batch *means* gives `N x` the mean over the full accumulated batch.
Nothing divides by `len(batches)`.

The `* dp_group_size` factor is unrelated — it converts the DP-averaged
gradient back to a sum and is correct; the missing term is the accumulation
divisor.

**Contrast with the sibling recipes.** `recipes/llm/train_ft.py` normalizes by
`num_label_tokens`, the DP-reduced count over the whole accumulated batch
(L1211-1213, threaded into the loss at L1161), so its gradients are invariant to
how a global batch is split into micro-batches. `train_seq_cls.py` is the only
training recipe whose gradient depends on that split.

CPU only, no GPU or checkpoint needed. This reproduces the exact accumulation
the recipe performs:

```python
import torch, torch.nn as nn

torch.manual_seed(0)
V, C, N, B = 8, 3, 4, 2          # 4 micro-batches of 2 samples
lin = nn.Linear(V, C, bias=False)
loss_fn = nn.CrossEntropyLoss()
X = torch.randn(N * B, V)
Y = torch.randint(0, C, (N * B,))

# What the recipe does: backward each micro-batch mean, never divide by N.
lin.weight.grad = None
for i in range(N):
    loss_fn(lin(X[i * B:(i + 1) * B]), Y[i * B:(i + 1) * B]).backward()
g_recipe = lin.weight.grad.clone()

# Reference: one mean over the whole accumulated batch.
lin.weight.grad = None
loss_fn(lin(X), Y).backward()
g_ref = lin.weight.grad.clone()

print(g_recipe.norm() / g_ref.norm())
```

```
tensor(4.0)
```

The ratio is exactly the number of micro-batches.

**Consequence.** With `global_batch_size: 32` and `local_batch_size: 8`,
`grad_accumulation_steps` is 4 and every gradient is 4x too large — the same as
quadrupling the learning rate, silently. Raising `global_batch_size` to reduce
gradient noise instead *increases* the effective step size, which is the
opposite of the intent. Nothing errors; the run reports a normal loss, because
only the gradient is affected, not the logged loss value.

The bug is invisible at `grad_accumulation_steps == 1` (one micro-batch, ratio
1.0), so a smoke test with `global_batch_size == local_batch_size * dp_size`
passes.

**Expected behavior**

The accumulated gradient equals the gradient of the mean loss over the full
global batch, independent of how that batch is split into micro-batches —
matching `train_ft.py`.

**Environment overview**

- `main` at 3ddef9b1, CPU only.

**Additional context**

Dividing the per-micro-batch loss by `len(batches)` cancels the factor exactly
when micro-batches are equally sized, which is the normal case here (the
sampler drops the tail). Weighting each micro-batch by its own sample count
would also be exact for a ragged final batch — happy to do it either way; say
which you prefer.

While confirming this I also noticed `train_seq_cls.py` is the only recipe that
never calls `_set_moe_aux_loss_backward_scale` (`train_ft.py`, `llm/kd.py`,
`vlm/finetune.py` and `vlm/kd.py` all do). That one is **latent, not a live
bug**: the only sequence-classification architecture in `MODEL_ARCH_MAPPING` is
`LlamaBidirectionalForSequenceClassification`, which is dense, so no MoE
auxiliary loss currently flows through this recipe. Flagging it only because the
recipe does wire up `moe_mesh` and `moe_parallel_config`, so it would bite the
day a MoE sequence-classification model is registered.

### LINKED ISSUE #3846: Sequence-classification recipe over-scales gradients by dp_size
**Describe the bug**

Identified by @yuhezhang-ai while reviewing #3830 — filing it so it is tracked
outside that PR thread.

`TrainFinetuneRecipeForSequenceClassification` multiplies the loss by `dp_size`
before `backward()`, but the loss is a **local** per-micro-batch mean. That
factor is only valid when the loss is already normalized by a **global**
denominator, so on `dp_size > 1` the gradient is over-scaled by `dp_size` — on
top of the separate `grad_accumulation_steps` factor in #3829.

`nemo_automodel/recipes/llm/train_seq_cls.py` L90 and L242:

```python
self.loss_fn = torch.nn.CrossEntropyLoss()     # local mean
...
(loss * self._get_dp_group_size(include_cp=True)).backward()
```

DDP/FSDP **average** gradients across DP ranks, so:

```
(1/dp) * sum_r [ dp * local_mean_r ] = sum_r local_mean_r = dp x global_mean    (wrong)
```

`recipes/llm/train_ft.py` uses the same `* dp_size` at L1193 and is correct,
because its loss is divided by `num_label_tokens`, which is DP-allreduced
(L1211-1213):

```
(1/dp) * sum_r [ dp * local_sum_r / N_global ] = global_sum / N_global          (right)
```

So `* dp_size` is not wrong in itself — it cancels the DP average. It is only
correct alongside a global denominator, and `train_seq_cls.py` does not have one.

CPU only. Simulates the DP gradient average against the true global-mean
gradient (4 DP ranks x 2 accumulation steps x 2 samples):

```python
import torch, torch.nn as nn, torch.nn.functional as F

torch.manual_seed(0)
V, C, DP, ACC, B = 8, 3, 4, 2, 2
lin = nn.Linear(V, C, bias=False)
W = lin.weight.detach().clone()
N = DP * ACC * B
X, Y = torch.randn(N, V), torch.randint(0, C, (N,))
shards = [[(X[(r*ACC+a)*B:(r*ACC+a+1)*B], Y[(r*ACC+a)*B:(r*ACC+a+1)*B]) for a in range(ACC)] for r in range(DP)]

def grad(fn):
    lin.weight.data.copy_(W); lin.weight.grad = None
    fn(); return lin.weight.grad.clone()

truth = grad(lambda: F.cross_entropy(lin(X), Y, reduction="mean").backward())

def ddp(per_rank):                       # DDP averages the per-rank grads
    gs = []
    for r in range(DP):
        lin.weight.data.copy_(W); lin.weight.grad = None
        per_rank(r); gs.append(lin.weight.grad.clone())
    return torch.stack(gs).mean(0)

def today(r):                            # what the recipe does
    for xb, yb in shards[r]:
        (F.cross_entropy(lin(xb), yb, reduction="mean") * DP).backward()

def train_ft_convention(r):              # local SUM over a GLOBAL denominator
    for xb, yb in shards[r]:
        (F.cross_entropy(lin(xb), yb, reduction="sum") / N * DP).backward()

print((ddp(today).norm() / truth.norm()).item())                 # 8.0
print((ddp(train_ft_convention).norm() / truth.norm()).item())   # 1.0
```

Driving the recipe's real `_run_train_optim_step` gives the same factors:

| `dp_size` | `grad_acc` | gradient vs. global mean |
|---:|---:|---:|
| 1 | 1 | 1.00x |
| 1 | 4 | 4.00x |
| 2 | 1 | **2.00x** |
| 2 | 2 | **4.00x** |
| 4 | 2 | **8.00x** |

The factor is exactly `dp_size * grad_accumulation_steps`. On 8 GPUs with
`grad_accumulation_steps=4` that is 32x — an effectively multiplied learning
rate, with no error and an unaffected logged loss, since only the gradient is
wrong.

Both factors are 1 at `dp_size=1, grad_acc=1`, so a single-GPU smoke test passes.

**Expected behavior**

The DP-averaged gradient equals the gradient of the mean loss over the global
batch, independent of `dp_size` and of how the batch is split into
micro-batches — matching `train_ft.py`.

**Environment overview**

- `main` at 3ddef9b1, CPU only; the DP average is simulated, no multi-GPU needed.

**Additional context**

Fixing only the accumulation half (#3829) leaves the `dp_size` half in place —
the middle row of the table above. Both come from the same root cause, so the
single fix is to adopt the `train_ft.py` convention: normalize by a DP-allreduced
global label count and keep `* dp_size`. I have pushed that to #3830, which now
supersedes its original narrower change and closes both this and #3829.

As @yuhezhang-ai noted on #3830, learning rates in existing sequence-classification
recipes may need restating afterwards. Worth flagging that the correction is not
uniform: each run absorbed a `dp_size * grad_acc` factor specific to its own
parallelism and batch config, so a learning rate tuned under one configuration
will not transfer to another. I do not have multi-GPU hardware to run those
experiments.

kabirvashisht4-glitch:
# What does this PR do ?

Puts the sequence-classification **gradient** and **reported loss** on one
normalization: the global summed loss over a DP-allreduced global label count.
Both were previously derived from per-micro-batch *means*, which cannot be added
back together.

`torch.nn.CrossEntropyLoss` reduces with `mean`, so each micro-batch loss is a
local average. Summing those averages over-counts by `grad_accumulation_steps`,
summing them across ranks over-counts by `dp_size`, and neither weights an
unevenly sized micro-batch correctly.

`* dp_size` is **not** the bug and is kept. It cancels the DDP/FSDP gradient
average — but only reconstructs the global mean when the denominator is already
global, which is exactly what `train_ft.py` does with `num_label_tokens`
(L1211-1213):

```
(1/dp) * sum_r [ dp * local_sum_r / N_global ] = global_sum / N_global   correct
(1/dp) * sum_r [ dp * local_mean_r ]           = dp x global_mean        wrong
```

## Measured against the global-batch reference

Driving the real `_run_train_optim_step` and averaging per-rank gradients as the
framework does:

| splits | dp | gradient on `main` | reported loss on `main` | after |
|---|---:|---:|---:|---:|
| `[2]` | 1 | 1.00x | 1.00x | exact |
| `[2, 2]` | 1 | 2.00x | 1.00x | exact |
| `[1,1,1,1]` | 1 | 4.00x | 1.00x | exact |
| **`[3, 1]`** | 1 | **2.69x** | **0.98x** | exact |
| `[2]` | 2 | 2.00x | **2.00x** | exact |
| `[2, 2]` | 2 | 4.00x | **2.00x** | exact |
| **`[3, 1]`** | 2 | **4.90x** | **1.96x** | exact |
| `[2, 2]` | 4 | 8.00x | **4.00x** | exact |
| **`[3, 1]`** | 4 | **11.94x** | **3.92x** | exact |

Even splits give clean integer factors (`dp x acc`). **Uneven splits give
non-integer ones** — 2.69x, 4.90x, 11.94x — because a mean of means also
misweights the shorter micro-batch. Equal splits cannot expose that, which is
why `3 + 1` is covered at every `dp_size`.

Only `splits=[2], dp=1` is correct on `main`, and that is the configuration a
single-GPU smoke test runs.

# Changelog

- **Gradient**: normalized by a DP-allreduced global label count instead of a
  local per-micro-batch mean. `loss * labels.numel()` recovers the summed loss
  from the mean `self.loss_fn` already returns, so there is no second
  cross-entropy pass and `self.loss_fn` stays authoritative.
- **Reported loss**: accumulates summed losses and divides by that same global
  count, rather than summing means and dividing by `len(batches)`.
- **Tests** (`tests/unit_tests/recipes/test_train_seq_cls_grad_accum.py`):
  gradients *and* reported loss asserted against
  `F.cross_entropy(..., reduction="mean")` over the global batch, across a
  `splits x dp_size` matrix including uneven `3 + 1` at dp 1, 2 and 4. A further
  test pins that `3 + 1` genuinely separates the two conventions, so the uneven
  cases cannot pass for the wrong reason.
- The DP average is simulated in **two phases** rather than as `value * dp_size`:
  ranks hold different data, so that shortcut is valid for counts but not for
  losses. Phase one records each rank's reduction inputs; phase two replays the
  true cross-rank sum in the same call order.

**Pre checks**:

- On unmodified `main`, 8 of 9 gradient cases and 5 of 9 reported-loss cases
  fail, with the factors in the table. Restoring the fix passes all 19.
- `pytest tests/unit_tests/recipes/` → 1180 passed, 12 skipped, no regressions.
  (`recipes/dllm` excluded locally: it fails to import
  `transformers.models.diffusion_gemma` on unmodified `main` too — my
  `transformers` is older than the pinned 5.12.1.)

**Learning rates.** The correction is not uniform: each existing run absorbed a
`dp_size x grad_acc` factor specific to its own parallelism and batch config, so
a rate tuned under one configuration will not transfer to another. I do not have
multi-GPU hardware for those experiments and can only offer the CPU invariant
tests.

# Additional Information

- Closes #3829, closes #3846

### DIFF
--- a/nemo_automodel/recipes/llm/train_seq_cls.py
+++ b/nemo_automodel/recipes/llm/train_seq_cls.py
@@ -229,10 +229,18 @@ class TrainFinetuneRecipeForSequenceClassification(BaseRecipe):
         )
         num_tokens_in_batch = self._dp_allreduce(num_tokens_in_batch).item()
 
-        # Each backward() adds to .grad, so accumulating N micro-batch means
-        # gives N x the mean over the full accumulated batch. Divide it back out
-        # so gradients do not depend on how a global batch is split.
-        num_batches = len(batches)
+        # Global number of labels over the whole accumulated batch and every DP
+        # rank. Normalizing the backward loss by this -- rather than by a local
+        # per-microbatch mean -- is what makes the gradient invariant to both the
+        # micro-batch split and dp_size. It is the convention train_ft.py uses
+        # with num_label_tokens, and it is what makes the `* dp_size` below
+        # correct: that factor cancels the DP gradient average, which only
+        # reconstructs the global mean when the denominator is already global.
+        num_label_samples = torch.tensor(
+            sum(batch["labels"].numel() for batch in batches),
+            dtype=torch.long,
+        )
+        num_label_samples = max(int(self._dp_allreduce(num_label_samples).item()), 1)
 
         for batch in batches:
             batch = {
@@ -250,7 +258,13 @@ class TrainFinetuneRecipeForSequenceClassification(BaseRecipe):
             preds = torch.argmax(logits, dim=-1)
             all_preds.append(preds.detach())
             all_labels.append(labels.view(-1).detach())
-            (loss * self._get_dp_group_size(include_cp=True) / num_batches).backward()
+            # `loss` is the mean over this micro-batch, so `loss * n_local` is its
+            # summed loss; dividing by the global count gives this micro-batch's
+            # share of the global mean. Every sequence-classification sample
+            # carries exactly one label, so `numel()` is the supervised count and
+            # matches how num_label_samples above was accumulated.
+            local_loss = loss * labels.numel() / num_label_samples
+            (local_loss * self._get_dp_group_size(include_cp=True)).backward()
 
         # Calculate gradient norm (distributed-aware)
         grad_norm = clip_grad_norm(
--- a/nemo_automodel/recipes/llm/train_seq_cls.py
+++ b/nemo_automodel/recipes/llm/train_seq_cls.py
@@ -229,6 +229,11 @@ class TrainFinetuneRecipeForSequenceClassification(BaseRecipe):
         )
         num_tokens_in_batch = self._dp_allreduce(num_tokens_in_batch).item()
 
+        # Each backward() adds to .grad, so accumulating N micro-batch means
+        # gives N x the mean over the full accumulated batch. Divide it back out
+        # so gradients do not depend on how a global batch is split.
+        num_batches = len(batches)
+
         for batch in batches:
             batch = {
                 k: (v.to(self.dist_env.device, non_blocking=True) if v is not None else None) for k, v in batch.items()
@@ -245,7 +250,7 @@ class TrainFinetuneRecipeForSequenceClassification(BaseRecipe):
             preds = torch.argmax(logits, dim=-1)
             all_preds.append(preds.detach())
             all_labels.append(labels.view(-1).detach())
-            (loss * self._get_dp_group_size(include_cp=True)).backward()
+            (loss * self._get_dp_group_size(include_cp=True) / num_batches).backward()
 
         # Calculate gradient norm (distributed-aware)
         grad_norm = clip_grad_norm(
--- a/nemo_automodel/recipes/llm/train_seq_cls.py
+++ b/nemo_automodel/recipes/llm/train_seq_cls.py
@@ -252,7 +252,10 @@ class TrainFinetuneRecipeForSequenceClassification(BaseRecipe):
                 out = model(**batch)
                 logits = getattr(out, "logits", out)
                 loss = self.loss_fn(logits, labels.view(-1))
-            losses.append(loss.detach().clone())
+            # Keep the summed loss, not the per-microbatch mean: means cannot be
+            # added back together, and the reported metric is normalized by the
+            # same global label count the gradient uses.
+            losses.append((loss * labels.numel()).detach().clone())
 
             # Collect predictions for accuracy calculation
             preds = torch.argmax(logits, dim=-1)
@@ -330,9 +333,13 @@ class TrainFinetuneRecipeForSequenceClassification(BaseRecipe):
                     reference_mfu=mfu_calculator.reference_mfu,
                 )
 
+        # Global summed loss over every micro-batch and DP rank, divided by the
+        # same global label count as the gradient. Dividing by len(batches)
+        # instead would report a mean of means: inflated by dp_size, and wrong
+        # whenever micro-batches are unevenly sized.
         total_loss = torch.sum(torch.stack(losses))
         total_loss = self._dp_allreduce(total_loss, include_cp=True).detach()
-        loss = total_loss / len(batches)
+        loss = total_loss / num_label_samples
 
         return MetricsSample(
             step=self.step_scheduler.step,
--- a/tests/unit_tests/recipes/test_train_seq_cls_grad_accum.py
+++ b/tests/unit_tests/recipes/test_train_seq_cls_grad_accum.py
@@ -1,0 +1,120 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Gradients must not depend on how a global batch is split into micro-batches.
+
+``torch.nn.CrossEntropyLoss`` reduces with ``mean``, and every ``backward()``
+adds to ``.grad``. Accumulating N micro-batch means therefore yields N x the
+mean over the full accumulated batch unless the count is divided back out, which
+silently multiplies the effective learning rate by ``grad_accumulation_steps``.
+"""
+
+from types import SimpleNamespace
+from unittest import mock
+
+import pytest
+import torch
+import torch.nn as nn
+
+from nemo_automodel.recipes.llm import train_seq_cls as seq_cls_mod
+from nemo_automodel.recipes.llm.train_seq_cls import TrainFinetuneRecipeForSequenceClassification
+
+VOCAB, N_CLASSES, HIDDEN = 16, 3, 8
+
+
+class _TinyClassifier(nn.Module):
+    """Smallest model with the sequence-classification forward contract."""
+
+    def __init__(self):
+        super().__init__()
+        self.embed = nn.Embedding(VOCAB, HIDDEN)
+        self.head = nn.Linear(HIDDEN, N_CLASSES, bias=False)
+
+    def forward(self, input_ids, attention_mask=None):
+        pooled = self.embed(input_ids).mean(dim=1)
+        return SimpleNamespace(logits=self.head(pooled))
+
+
+def _make_recipe(model):
+    """A recipe instance with only what ``_run_train_optim_step`` touches."""
+    recipe = object.__new__(TrainFinetuneRecipeForSequenceClassification)
+    recipe.model_parts = [model]
+    recipe.dist_env = SimpleNamespace(device=torch.device("cpu"), world_size=1)
+    recipe.loss_fn = nn.CrossEntropyLoss()
+    recipe.optimizer = [torch.optim.SGD(model.parameters(), lr=0.0)]
+    recipe.lr_scheduler = None
+    recipe.max_grad_norm = 1.0
+    recipe.device_mesh = None
+    recipe.timestamp = 0.0
+    recipe.mfu_calculator = None
+    recipe.step_scheduler = SimpleNamespace(step=0, epoch=0)
+    recipe._get_dp_group_size = lambda include_cp=False: 1
+    recipe._get_cp_group_size = lambda: 1
+    recipe._get_pp_rank = lambda: 0
+    recipe._dp_allreduce = lambda t, *a, **k: t
+    return recipe
+
+
+def _grads_for(split, seed=0):
+    """Run one optimizer step over ``split`` micro-batches; return the accumulated grads.
+
+    Gradients are captured at ``clip_grad_norm``, which runs after the
+    accumulation loop and before ``optimizer.step()`` zeroes them.
+    """
+    torch.manual_seed(seed)
+    model = _TinyClassifier()
+    recipe = _make_recipe(model)
+
+    torch.manual_seed(123)
+    input_ids = torch.randint(0, VOCAB, (8, 4))
+    labels = torch.randint(0, N_CLASSES, (8,))
+
+    batches = [
+        {
+            "input_ids": input_ids[i : i + 8 // split],
+            "attention_mask": torch.ones_like(input_ids[i : i + 8 // split]),
+            "labels": labels[i : i + 8 // split],
+        }
+        for i in range(0, 8, 8 // split)
+    ]
+    assert len(batches) == split
+
+    captured = {}
+
+    def _capture(**kwargs):
+        captured["grads"] = [p.grad.detach().clone() for p in model.parameters() if p.grad is not None]
+        return torch.tensor(0.0)
+
+    with (
+        mock.patch.object(seq_cls_mod, "clip_grad_norm", _capture),
+        mock.patch.object(torch.cuda, "max_memory_allocated", lambda: 0),
+    ):
+        recipe._run_train_optim_step(batches)
+
+    assert captured["grads"], "no gradients were captured"
+    return captured["grads"]
+
+
+@pytest.mark.parametrize("split", [2, 4, 8])
+def test_gradients_are_invariant_to_microbatch_split(split):
+    """Splitting one global batch into more micro-batches must not change grads."""
+    single = _grads_for(1)
+    accumulated = _grads_for(split)
+
+    assert len(single) == len(accumulated)
+    for g_one, g_many in zip(single, accumulated):
+        ratio = (g_many.norm() / g_one.norm()).item()
+        assert torch.allclose(g_many, g_one, rtol=1e-5, atol=1e-6), (
+            f"{split} micro-batches scaled the gradient by ~{ratio:.2f}x"
+        )
--- a/tests/unit_tests/recipes/test_train_seq_cls_grad_accum.py
+++ b/tests/unit_tests/recipes/test_train_seq_cls_grad_accum.py
@@ -118,3 +118,86 @@ def test_gradients_are_invariant_to_microbatch_split(split):
         assert torch.allclose(g_many, g_one, rtol=1e-5, atol=1e-6), (
             f"{split} micro-batches scaled the gradient by ~{ratio:.2f}x"
         )
+
+
+# ---------------------------------------------------------------------------
+# Data-parallel scaling
+# ---------------------------------------------------------------------------
+
+
+def _ddp_averaged_grads(dp_size, accum, per_rank_bs, seed=0):
+    """Run one step on every simulated DP rank and average the grads, as DDP does.
+
+    ``_dp_allreduce`` is a SUM across ranks, so with equally sized shards it
+    multiplies a per-rank count by ``dp_size``.
+    """
+    torch.manual_seed(seed)
+    reference = _TinyClassifier()
+    init_state = {k: v.clone() for k, v in reference.state_dict().items()}
+
+    total = dp_size * accum * per_rank_bs
+    torch.manual_seed(123)
+    input_ids = torch.randint(0, VOCAB, (total, 4))
+    labels = torch.randint(0, N_CLASSES, (total,))
+
+    per_rank_grads = []
+    for rank in range(dp_size):
+        model = _TinyClassifier()
+        model.load_state_dict(init_state)
+        recipe = _make_recipe(model)
+        recipe._get_dp_group_size = lambda include_cp=False: dp_size
+        recipe._dp_allreduce = lambda t, *a, **k: t * dp_size
+
+        batches = []
+        for a in range(accum):
+            lo = (rank * accum + a) * per_rank_bs
+            hi = lo + per_rank_bs
+            batches.append(
+                {
+                    "input_ids": input_ids[lo:hi],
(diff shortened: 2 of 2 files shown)


yuhezhang-ai:
/ok to test [0d96d0d](https://github.com/NVIDIA-NeMo/Automodel/pull/3830/commits/0d96d0d6af72b24e8564160b6b8ee7e860d6aa3f)

yuhezhang-ai:
Thanks @kabirvashisht4-glitch  for catching and fixing the gradient-accumulation scaling, the proposed change looks correct for that issue.

@NVIDIA-NeMo/nemotron_retriever_maintainers , while looking at this, I noticed the sequence-classification recipe may already be over-scaling gradients even without accumulation when `dp_size > 1`: `CrossEntropyLoss` returns a local mean, but the recipe multiplies it by `dp_size` before DDP/FSDP averages the gradients. This does not appear to match the intended global-mean objective.

What do you think? Ideally, we should correct the full scaling and run a few short experiments to determine whether the learning rates in the existing recipes need adjustment?

kabirvashisht4-glitch:
You are right, and thanks for catching it — my PR fixes only one of two factors.

I simulated the DP average on CPU (4 DP ranks x 2 accumulation steps x 2 samples, gradients averaged across ranks as DDP/FSDP does) against the true mean cross-entropy over the whole global batch:

| variant | ratio vs. global-mean gradient |
|---|---:|
| `main` today | **8.00x** (= `dp_size` x `grad_acc`) |
| with only this PR's `/ num_batches` | **4.00x** (= `dp_size`) |
| `train_ft.py`'s convention | **1.00x** (exact match) |

So the two factors are independent and multiply. On 8 GPUs with `grad_accumulation_steps=4` that is 32x.

The root cause is that `* dp_size` is only valid when the loss is already normalized by a **global** denominator. `train_ft.py` does exactly that — the loss is divided by `num_label_tokens`, which is DP-allreduced (L1211-1213) — so `* dp_size` there cancels DDP's averaging and reconstructs the global mean:

```
(1/dp) * sum_r [ dp * local_sum_r / N_global ] = global_sum / N_global   ✓
```

`train_seq_cls.py` normalizes locally instead (`CrossEntropyLoss` mean over the local micro-batch), so `* dp_size` has nothing to cancel against and multiplies through:

```
(1/dp) * sum_r [ dp * local_mean_r ] = sum_r local_mean_r = dp x global_mean   ✗
```

The complete fix is to adopt the `train_ft.py` convention rather than patch the two factors separately — reduce with `sum`, divide by the DP-allreduced global label count, keep `* dp_size`:

```python
num_label_samples = torch.tensor(sum(b["labels"].numel() for b in batches), dtype=torch.long)
num_label_samples = self._dp_allreduce(num_label_samples).item()
...
loss = F.cross_entropy(logits, labels.view(-1), reduction="sum") / num_label_samples
(loss * self._get_dp_group_size(include_cp=True)).backward()
```

That is row (c) above, and it subsumes the `/ num_batches` change in this PR: one global denominator makes the gradient invariant to both the micro-batch split and `dp_size`, and it also removes the ragged-final-batch caveat I noted, since it weights by actual sample count rather than assuming equal micro-batches.

Happy to extend this PR to that, with the DP simulation added as a unit test so the invariant is pinned in CI rather than argued about. Or if you would rather keep this PR as the narrow accumulation fix and track the `dp_size` half separately, that works too — just say which you prefer.

On the learning rates: agreed, and worth flagging that the correction is not uniform across existing runs. Any recipe that was tuned on this path absorbed a `dp_size x grad_acc` factor specific to its own parallelism and batch config, so the same LR will not transfer between two configs that previously "worked". Short A/B runs at a couple of `(dp_size, grad_acc)` combinations would tell you whether the published seq-cls recipes need their LR restated.

I do not have multi-GPU hardware to run those experiments, so I can only offer the CPU-level invariant test — happy to hand the empirical side over.

kabirvashisht4-glitch:
Pushed the full fix in 69be991 — this PR now supersedes its original `/ num_batches` change and covers both factors. Filed #3846 so the `dp_size` half is tracked outside this thread.

Driving the recipe's real `_run_train_optim_step` and averaging the per-rank gradients as DDP does, against the gradient of the mean loss over the global batch:

| `dp_size` | `grad_acc` | on `main` | with only `/num_batches` | after 69be991 |
|---:|---:|---:|---:|---:|
| 1 | 1 | 1.00x | 1.00x | 1.00x |
| 1 | 4 | 4.00x | 1.00x | 1.00x |
| 2 | 1 | **2.00x** | **2.00x** | 1.00x |
| 2 | 2 | **4.00x** | **2.00x** | 1.00x |
| 4 | 2 | **8.00x** | **4.00x** | 1.00x |

The middle column is exactly your point: the narrow fix removes the accumulation factor and leaves `dp_size` untouched. On unmodified `main` 7 of the 8 parametrized cases fail; the only passer is `dp=1, acc=1`, which is why a single-GPU smoke test never caught either.

The change follows `train_ft.py` rather than patching the two factors separately:

```python
num_label_samples = torch.tensor(sum(b["labels"].numel() for b in batches), dtype=torch.long)
num_label_samples = max(int(self._dp_allreduce(num_label_samples).item()), 1)
...
local_loss = loss * labels.numel() / num_label_samples
(local_loss * self._get_dp_group_size(include_cp=True)).backward()
```

Three notes on the shape of it:

- `loss * labels.numel()` recovers the summed loss from the mean `self.loss_fn` already returns, so there is no second cross-entropy pass and `self.loss_fn` stays authoritative for the logged metric — the reported per-micro-batch loss values are unchanged.
- `* dp_size` is kept deliberately. It is not the bug; it cancels the DP gradient average and is correct once the denominator is global, exactly as in `train_ft.py`.
- Because the denominator is a real sample count rather than a micro-batch count, this also removes the ragged-final-batch caveat I flagged on the first version — no assumption that micro-batches are equally sized.

Tests are parametrized over `dp_size x grad_acc` and simulate the DP average on CPU, so the invariant is pinned in CI rather than argued about. `pytest tests/unit_tests/recipes/` → 1169 passed, 12 skipped.

On the learning-rate question: agreed the recipes should be re-checked, and worth flagging that the correction is not uniform — each existing run absorbed a `dp_size * grad_acc` factor specific to its own parallelism and batch config, so an LR tuned under one configuration will not transfer to another. I do not have multi-GPU hardware, so I can offer the CPU invariant test but not the empirical runs; happy to hand that part over.

If you would rather land the accumulation fix alone first and take `dp_size` separately, say the word and I will split it back out.

yuhezhang-ai:
@kabirvashisht4-glitch Thanks for updating this, the global label-count normalization now looks correct for both DP scaling and gradient accumulation.

Could you also update the reported training loss in this PR so it uses the same global sum/global label-count normalization? Please add coverage showing that both the gradients and reported loss match the global-batch reference, including an uneven microbatch split such as `3 + 1`.

The PR description and test module docstring also still describe the earlier `/ len(batches)` approach, so could you update those to reflect the new DP-global normalization?

kabirvashisht4-glitch:
Thanks — all three done in 0d986e4.

**Reported loss.** You were right that it was still on the old convention. It accumulated per-micro-batch means, summed those across ranks, and divided by `len(batches)` — the same "averages cannot be added" mistake as the gradient, just in the metric. It now accumulates the summed loss and divides by the same DP-allreduced global label count:

```python
losses.append((loss * labels.numel()).detach().clone())   # summed, not mean
...
total_loss = self._dp_allreduce(torch.sum(torch.stack(losses)), include_cp=True)
loss = total_loss / num_label_samples
```

**Coverage, including `3 + 1`.** Both gradients and reported loss are now asserted against `F.cross_entropy(..., reduction="mean")` over the global batch, across a `splits x dp_size` matrix:

| splits | dp | gradient on `main` | reported loss on `main` |
|---|---:|---:|---:|
| `[2]` | 1 | 1.00x | 1.00x |
| `[2, 2]` | 1 | 2.00x | 1.00x |
| **`[3, 1]`** | 1 | **2.69x** | **0.98x** |
| `[2]` | 2 | 2.00x | **2.00x** |
| **`[3, 1]`** | 2 | **4.90x** | **1.96x** |
| `[2, 2]` | 4 | 8.00x | **4.00x** |
| **`[3, 1]`** | 4 | **11.94x** | **3.92x** |

The uneven rows were worth insisting on: they produce **non-integer** factors, because a mean of means also misweights the shorter micro-batch — something an even split can never expose. I added a further test asserting that `3 + 1` genuinely separates the two conventions, so those cases cannot pass for the wrong reason.

One correction to my own harness while doing this: I had been faking the DP reduction as `value * dp_size`, which is only valid when every rank holds the same value. True for label counts, false for losses. It now runs two phases — record each rank's reduction inputs, then replay the real cross-rank sum in the same call order. The reported-loss tests failed until I fixed that, and they were right to.

**Prose.** PR description and the test module docstring both rewritten; both still described `/ len(batches)`.

`pytest tests/unit_tests/recipes/` → 1180 passed, 12 skipped. On unmodified `main`, 8 of 9 gradient cases and 5 of 9 reported-loss cases fail with the factors above.

rnyak:
> Thanks @kabirvashisht4-glitch for catching and fixing the gradient-accumulation scaling, the proposed change looks correct for that issue.
> 
> @NVIDIA-NeMo/nemotron_retriever_maintainers , while looking at this, I noticed the sequence-classification recipe may already be over-scaling gradients even without accumulation when `dp_size > 1`: `CrossEntropyLoss` returns a local mean, but the recipe multiplies it by `dp_size` before DDP/FSDP averages the gradients. This does not appear to match the intended global-mean objective.
> 
> What do you think? Ideally, we should correct the full scaling and run a few short experiments to determine whether the learning rates in the existing recipes need adjustment?

@yuhezhang-ai not sure this MR affects the cross encoder training which uses `class TrainCrossEncoderRecipe(TrainBiEncoderRecipe):`

For cross-encoder training, the loss scaling is separate. It does:
```
loss = F.cross_entropy(outputs.logits, labels)
scaled_loss = loss / num_batches
scaled_loss.backward()
```
The cross-encoder divides each microbatch’s mean loss by the number of accumulated microbatches. On the final backward pass, DDP/FSDP synchronizes and averages the accumulated gradients across data-parallel ranks, so no explicit dp_size multiplier is needed (if I am not mistaken).

do you mind elaborating if we still need the learning rates  adjustments in the existing [recipe](https://github.com/NVIDIA-NeMo/Automodel/blob/main/examples/retrieval/cross_encoder/llama3_2_1b.yaml)? thanks.

yuhezhang-ai:
> > Thanks @kabirvashisht4-glitch for catching and fixing the gradient-accumulation scaling, the proposed change looks correct for that issue.
> > @NVIDIA-NeMo/nemotron_retriever_maintainers , while looking at this, I noticed the sequence-classification recipe may already be over-scaling gradients even without accumulation when `dp_size > 1`: `CrossEntropyLoss` returns a local mean, but the recipe multiplies it by `dp_size` before DDP/FSDP averages the gradients. This does not appear to match the intended global-mean objective.
> > What do you think? Ideally, we should correct the full scaling and run a few short experiments to determine whether the learning rates in the existing recipes need adjustment?
> 
> @yuhezhang-ai not sure this MR affects the cross encoder training which uses `class TrainCrossEncoderRecipe(TrainBiEncoderRecipe):`
> 
> For cross-encoder training, the loss scaling is separate. It does:
> 
> ```
> loss = F.cross_entropy(outputs.logits, labels)
> scaled_loss = loss / num_batches
> scaled_loss.backward()
> ```
> 
> The cross-encoder divides each microbatch’s mean loss by the number of accumulated microbatches. On the final backward pass, DDP/FSDP synchronizes and averages the accumulated gradients across data-parallel ranks, so no explicit dp_size multiplier is needed (if I am not mistaken).
> 
> do you mind elaborating if we still need the learning rates adjustments in the existing [recipe](https://github.com/NVIDIA-NeMo/Automodel/blob/main/examples/retrieval/cross_encoder/llama3_2_1b.yaml)? thanks.

Thanks @rnyak you’re right. sorry I initially thought the generic sequence-classification recipe belongs to the retrieval reranker. This PR does not affect `TrainCrossEncoderRecipe`, so no LR adjustment is needed for that recipe.

The cross-encoder scaling is correct when its microbatches have equal sizes. A shorter final accumulation window is also fine because it divides by the number of batches actually present.

The edge case is a window containing differently sized microbatches. For example, if it contains batches of 4 and 1 samples, the current calculation is:

`(mean_of_4 + mean_of_1) / 2`

This gives the single-sample batch 50% of the gradient weight. The correct global mean is:

`(4 × mean_of_4 + 1 × mean_of_1) / 5`

where that sample contributes 20%. Since `drop_last=False`, this can occur in the final optimizer step of an epoch. It is a real correctness edge case, but likely affects only one step per epoch. We can address it separately and no need to retune the existing LR.

yuhezhang-ai:
/ok to test [5614a87](https://github.com/NVIDIA-NeMo/Automodel/pull/3830/commits/5614a87b3da0555a536b794c67569cf27b4fd1ee)

yuhezhang-ai:
Following my offline discussion with Ronay, I ran an extra real-GPU check to confirm the fix.

I tested this fix on **two H100 80GB GPUs**, using both DDP and FSDP2, at commit `5614a87b3da0555a536b794c67569cf27b4fd1ee`.

The check was straightforward: start with identical model weights and the same training examples, then compare:

- One update using the entire batch on a single GPU.
- One update using the PR's actual training step, with those examples split across two GPUs and accumulated over smaller batches.

I tested equal batch sizes, a smaller final batch, different numbers of examples on each GPU, and gradient clipping.

**All 10 GPU cases passed.** The gradients, updated weights, reported loss, and gradient norm matched the single-GPU reference within tight float32 tolerances (`rtol=1e-5`, `atol=1e-6`).

I also ran the old code through the same check. It failed as expected, reproducing **4× gradients and 2× reported loss** with two GPUs and two accumulation steps.

The PR's **19 focused unit tests also passed**.

This was a focused check of the training step using a tiny deterministic classifier; it does not establish full training convergence or mixed-precision behavior.

yuhezhang-ai (approved):

PR #3831, opened 2026-09-08
fix(vlm): skip dense masks for compact packing backends

Butterfingrz:
# What does this PR do ?

Skip redundant CPU dense-mask allocation and mask transfers for NEAT VLM packing at
`cp_size=1` when the attention path can consume compact document boundaries.

`VlmDataloaderConfig.build` selects compact `[B, S]` document maps for the existing
`flash_attention_2/3/4` packing integration, or when the model declares
`consumes_packed_seq_ids=True`, regardless of the resolved backend label.
Gemma4-MoE opts in for its vision-aware HF eager/SDPA text path, which rebuilds
full and sliding attention masks from `_packed_seq_ids`.

**Collator benchmark (dense → compact)**

Original measurements retained from this PR: PyTorch 2.10.0+cu130, one CPU thread,
and two synthetic text documents per pack. Each case ran in a separate process.
Collate times are medians of five calls after warm-up; H2D times are medians of
five synchronized copies of the pinned mask.

| Batch × tokens | Collate (ms) | Peak host RSS increase | Mask H2D (ms) |
|---|---:|---:|---:|
| 1 × 3072 | 30.65 → 0.02 | 53.0 → 0.02 MiB | 0.30 → 0.02 |
| 4 × 16384 | 3766 → 0.45 | 3.25 GiB → 0.5 MiB | 19.34 → 0.03 |

The benchmark directly toggled the collator's `materialize_4d_mask` with its
`te` setting. It measured collation and mask copies; it did not run a TE attention
kernel or the model. The consumer-declaration change adds no new performance
measurements. Gemma4 still builds quadratic full/sliding masks on the GPU;
compact inputs do not imply varlen execution or linear GPU memory use.

# Changelog

- Select compact masks through the existing FA packing integration or a model's
  explicit consumer declaration. `_COMPACT_MASK_BACKENDS` contains only FA2/3/4;
  `te` alone does not select compact inputs.
- Add Gemma4's declaration for the active MoE text implementation with
  `enable_moe_block=True`, `use_bidirectional_attention="vision"`, and actual
  HF eager/SDPA dispatch. The shipped Gemma4 26B packing recipe uses HF attention
  despite its `backend.attn: te` label, and receives compact maps through this declaration.
- Read the live model's declaration once in training setup, unwrap one `.module`
  wrapper, and pass a strict boolean to the dataloader's runtime build argument.
- Select the collator mode that also emits `_packed_seq_ids` for single-document
  packs, preserving Gemma4-MoE's sliding-window mask reconstruction.
- Build additive packed masks for Gemma4-MoE eager attention, which adds the mask
  to attention logits; a boolean mask would add `+1/+0` without blocking attention.

- [x] Followed the contributor guidelines.
- [x] Added regression tests.
- [x] Updated the collator documentation.

## TEST

- `pytest -q tests/unit_tests/datasets/vlm/test_vlm_loader.py tests/unit_tests/models/gemma4/test_gemma4_packed_mask_eager.py tests/unit_tests/recipes/test_finetune_vlm_cp_wiring.py`: 78 passed (CPU).
- `ruff format --check` / `ruff check` on `nemo_automodel/components/datasets/vlm/loader.py`, `nemo_automodel/components/models/gemma4_moe/model.py`, `nemo_automodel/recipes/vlm/finetune.py`: passed.

# Additional Information

For CP=1 NEAT packing, non-FA models without a consumer declaration retain their
pre-PR dense input path, including those configured with `te`. This preserves
the old path's limitations: it adds no packing support for those models, no
setup-time rejection, and no guarantee that every TE integration fails on its
first forward.

The declaration is wired into training setup only. Validation dataloader wiring,
CP>1 handling, THD packing, non-packed inputs, and example YAML configurations
remain unchanged.

### DIFF
--- a/nemo_automodel/components/datasets/vlm/collate_fns.py
+++ b/nemo_automodel/components/datasets/vlm/collate_fns.py
@@ -1513,12 +1513,17 @@ def neat_packed_vlm_collater(
     Packs arrive with **variable lengths** (no pre-padding).  This collater:
 
     1. Pads all text tensors to a common length.
-    2. Converts the indexed ``attention_mask`` to the appropriate format:
-       - ``flash_attention_2``: keeps the indexed ``[B, S]`` mask (values
-         1, 2, … for documents, 0 for padding).  The monkey-patched
-         ``_get_unpad_data`` converts this to ``cu_seqlens`` for
-         ``flash_attn_varlen_func``.
-       - ``sdpa`` / ``eager``: converts to a 4D block-causal bool mask.
+    2. Converts the indexed ``attention_mask`` to the representation
+       ``materialize_4d_mask`` selects:
+       - True: a 4D block-causal bool mask. SDPA reads it directly;
+         HF eager consumers must convert it to an additive mask.
+       - False: the indexed ``[B, S]`` map itself (values 1, 2, … for
+         documents, 0 for padding), for consumers that recover document
+         bounds from ``_packed_seq_ids`` rather than from a mask.
+       ``attn_implementation="flash_attention_2"`` also keeps the compact map,
+       whatever ``materialize_4d_mask`` says: the monkey-patched
+       ``_get_unpad_data`` converts it to ``cu_seqlens`` for
+       ``flash_attn_varlen_func``.
     3. Concatenates media tensors across the batch dimension.
 
     **No autoregressive shift** — it was already applied during packing.
@@ -1530,13 +1535,16 @@ def neat_packed_vlm_collater(
             If ``None`` (default), pad to the longest pack in the batch.
             A fixed length avoids recompilation with ``torch.compile``
             and ensures uniform tensor shapes across steps.
-        attn_implementation: Attention backend (``"flash_attention_2"``,
-            ``"sdpa"``, or ``"eager"``).
-        materialize_4d_mask: Whether SDPA/eager packing should expand the
-            indexed ``[B, S]`` document map into a dense
-            ``[B, 1, S, S]`` block-causal mask. Context-parallel VLM paths
-            rebuild their local mask from ``_packed_seq_ids`` and set this to
-            False to avoid the quadratic allocation.
+        attn_implementation: Attention backend the batch is collated for. Only
+            ``"flash_attention_2"`` changes behaviour here, by keeping the
+            compact map; every other value defers to ``materialize_4d_mask``.
+        materialize_4d_mask: Whether packing should expand the indexed
+            ``[B, S]`` document map into a dense ``[B, 1, S, S]`` block-causal
+            mask. Callers set this to False when the consumer recovers document
+            boundaries from ``_packed_seq_ids`` instead of from the mask:
+            context-parallel paths rebuild their local mask from it, and flash
+            attention and Transformer Engine cannot use a dense mask at all.
+            False therefore also forces ``_packed_seq_ids`` to be emitted.
 
     Returns:
         Dict with batched tensors ready for model forward.
--- a/nemo_automodel/components/datasets/vlm/loader.py
+++ b/nemo_automodel/components/datasets/vlm/loader.py
@@ -150,6 +150,19 @@ class VlmDataloaderBuild:
     processor: ProcessorMixin | None
 
 
+# Attention backends the packed collater must hand the compact ``[batch, sequence]`` document map
+# instead of a dense block-causal mask, because a dense mask is not a usable input for them: flash
+# attention rebuilds ``cu_seqlens`` from the compact map (``models/common/packing.get_unpad_data``),
+# and Transformer Engine reduces any non-None mask to a padding mask and drops ``cu_seqlens``
+# (``attention/utils.preprocess_args_and_kwargs_for_attn``), so document isolation has to reach it
+# through the model instead. A ``te`` model that does not rebuild its mask from ``_packed_seq_ids``
+# cannot honour the map either way; this set is about which representation a backend can use, not
+# about whether one arrives.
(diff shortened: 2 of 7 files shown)


Butterfingrz:
Hi, could you please take a look at this PR when you have a chance. @akoumpa @HuiyingLi @athitten  It avoids unnecessary dense mask allocation in VLM packing and fixes Gemma4-MoE eager attention masking. Thanks!

HuiyingLi:
@Butterfingrz Thank you! taking a look

HuiyingLi:
Thanks for the fix. I validated the PR head `ea2ad87994fc4428fb2028ebb9097e13e60b786e` on H100. The intended Gemma4 path looks correct, but I think the loader-side TE selection is currently too broad and should be addressed before merge.

`_COMPACT_MASK_BACKENDS` makes every neat-packed VLM with `backend.attn=te` receive the compact `[B, S]` document map. However, the generic TE preprocessing in `components/attention/utils.py` does not understand indexed document IDs. For any non-None mask it calls `logical_not()`, selects `padding_causal`, and does not use the `cu_seqlens` branch. Thus a map such as `[1, 1, 1, 2, 2, 2]` becomes an all-False padding mask and loses every document boundary.

A real TE H100 probe confirmed the semantic failure: after changing document 1 K/V, document 2 output changed by `5.3125`. This can silently enable cross-document attention for any TE VLM that does not explicitly consume `_packed_seq_ids` and reconstruct the isolation mask.

The current shipped Gemma4 recipe is safe: Gemma4 explicitly rebuilds its full/sliding vision-aware masks from `_packed_seq_ids`. I validated the exact critical chain:

`loader(te, cp=1) -> compact two-document batch -> Gemma4 HF eager attention -> forward/backward/optimizer step`

Results:

- packed vs separately evaluated documents: max abs diff `7.63e-06`
- changing document 1 produced document 2 delta `0`
- finite loss and gradients, optimizer parameters updated
- VLM/packing GPU suite: `388 passed`
- Gemma4 CUDA unit tests: `127 passed`
- Ruff and `git diff --check`: passed

The shipped-recipe impact is narrow: the Gemma4 26B packing recipe is the only current CP=1 NEAT VLM recipe I found that newly changes from dense to compact due to `backend.attn=te`. Qwen3-VL FA2 recipes were already compact, Qwen3.5 packed recipes use SDPA, MuseGlimmer TE packing uses THD, and CP>1 was already compact.

Suggested focused fix: gate compact TE masks on an explicit model capability/opt-in that guarantees the model consumes `_packed_seq_ids`. For TE consumers without that capability, fail fast or require THD/`cu_seqlens`. I would not add full packing support to unrelated models as part of this PR.

One related distinction: retaining the old dense `[B,1,S,S]` mask is not a generic TE fallback. The current TE preprocessor treats it as a padding mask and adds two dimensions, so it is not consumed as an arbitrary block-causal mask. The important requirement is an explicit supported consumer, not merely choosing dense versus compact.

Butterfingrz:
Thanks for pointing that out! @HuiyingLi  The updated PR ([d05704f0](https://github.com/Butterfingrz/Automodel/commit/d05704f063ef49824e05506308ac6f50cb04769e)) covers:

- **Reduce CPU overhead:** At `cp_size=1`, NEAT packing uses compact document maps only for FA2/3/4 or models declaring `consumes_packed_seq_ids=True`, skipping CPU dense mask construction and transfer. Gemma4 26B uses this declaration for its HF eager path and rebuilds the required masks on the GPU.
- **Fix eager masking:** Gemma4-MoE eager now uses additive masks to enforce document isolation and sliding-window constraints. The previous boolean masks were added as `+1/+0` and blocked nothing.
- **Preserve document IDs:** Compact mode always includes `_packed_seq_ids`, so Gemma4 rebuilds its masks and applies the sliding window even for single-document packs.

With FA3/4 included in `loader.py`'s `_COMPACT_MASK_BACKENDS`, future NEAT packing recipes using those backends can also benefit from the same CPU and transfer savings at `cp_size=1`.

At CP=1 NEAT packing, undeclared non-FA models retain pre-PR dense inputs and their limitations. This compatibility choice forgoes setup-time rejection, adds no packing support, and does not guarantee a first-forward error. Declaration wiring is training-only.

CPU tests cover representation selection and declaration wiring; CPU fp32 tests compare packed and separate-document eager/SDPA logits.

**Design at a glance**

```mermaid
flowchart LR
    G["Gemma4-MoE<br/>consumes_packed_seq_ids = True<br/>(MoE text backend + vision mask rule + eager/sdpa)"]
    O["Other models<br/>no attribute → False"]
    R["VLM recipe setup()<br/>reads the declaration once, passes a bool to build()"]
    L{"VlmDataloaderConfig.build · NEAT · cp_size=1<br/>label in FA2/3/4 or declaration True?"}
    C["compact [B, S] map + _packed_seq_ids<br/>model rebuilds its own masks"]
    D["dense [B, 1, S, S] mask<br/>= pre-PR input"]
    G --> R
    O --> R
    R --> L
    L -- yes --> C
    L -- no --> D
```

**Other recipes are unaffected**

- The declaration exists only on Gemma4-MoE; every other model reads as `False`, so the loader rule reduces to `cp_size <= 1 and label not in {FA2, FA3, FA4}`: the pre-PR rule with `te` removed. FA2 recipes stay compact, SDPA and other labels stay dense, `te` returns to dense.
- THD packing, `cp_size > 1`, non-packed inputs, and the validation dataloader never read the declaration.
- Among shipped VLM recipes only Gemma4 26B packing changes representation relative to the PR parent, and it already did at `ea2ad879`.

PR #3834, opened 2026-09-08
fix(peft): export merged grouped-expert adapters

pstjohn:
# What does this PR do ?

Fix a gap where AutoModel can successfully train and save a LoRA adapter that cannot subsequently be used by inference or RL workflows requiring canonical Hugging Face weights.

This occurs for MoE models using `GroupedExpertsLoRA` with `moe_rank_scaling`. AutoModel correctly trains and checkpoints a smaller native rank per expert, while `adapter_config.json` records the requested global rank. The existing Hugging Face PEFT merge path constructs each fused expert tensor using that global rank, so loading fails with a shape mismatch before the adapter can be merged.

For example, a rank-192 adapter over 128 experts may contain the correct AutoModel-native folded rank dimension of `128 * 32 = 4,096`, while PEFT constructs `128 * 192 = 24,576`.

The result is a valid, completed training checkpoint that cannot be merged for standard inference deployment or consumed by RL workflows expecting canonical merged Hugging Face weights.

This PR adds a public AutoModel-native export operation that:

1. Reconstructs the original AutoModel PEFT topology from checkpoint metadata.
2. Restores the adapter through the model-family state-dict adapter.
3. Folds both `LinearLoRA` and `GroupedExpertsLoRA` into the base weights.
4. Converts native grouped-expert weights into canonical Hugging Face expert tensors.
5. Writes a standard sharded Hugging Face checkpoint.

The existing merge behavior remains the default. AutoModel PEFT-v5 grouped-expert checkpoints opt into the native route explicitly:

```bash
python tools/merge_lora.py \
  --base-model <base-model> \
  --adapter-path <checkpoint>/model \
  --output-dir <output> \
  --automodel-native-export
```

# Changelog

- Add the public `export_merged_peft_checkpoint` operation.
- Add native weight materialization for `GroupedExpertsLoRA`.
- Add the opt-in `--automodel-native-export` route to `tools/merge_lora.py`.
- Keep the existing Hugging Face PEFT route as the default, including for AutoModel metadata, QLoRA, CPU, and non-causal model classes.
- Force the torch linear backend for offline native export so dense and mixed adapters can be materialized.
- Switch the model to eval mode before materialization so checkpointed LoRA dropout does not block export.
- Support standalone, non-distributed loading of AutoModel adapter directories.
- Document standalone PEFT checkpoint export and provide a runnable usage example.

**Pre checks**:

- [x] Make sure you read and followed the Contributor guidelines

Validation performed:

- Focused export and CLI tests: 44 passed
- Related LoRA expert and checkpoint unit tests: 211 passed, 44 skipped
- Ruff formatting and lint checks passed
- Python compilation and `git diff --check` passed

# Additional Information

The regression test models the observed failure where AutoModel trains with a requested LoRA rank but scales it to a smaller native rank for each grouped expert. It verifies that restoration uses the AutoModel state-dict adapter instead of asking Hugging Face PEFT to independently reconstruct the fused expert layout.


tools/merge_lora.py
```diff
@@ -212,52 +283,69 @@ def merge_lora(
...
             the class is inferred from the adapter's ``task_type``.
     """
-    from peft import PeftModel
     from transformers import AutoTokenizer
 
-    auto_cls = _resolve_auto_cls(adapter_path, model_class)
     torch_dtype = getattr(torch, dtype)
+    is_automodel_checkpoint = os.path.isfile(os.path.join(adapter_path, "automodel_peft_config.json"))
+
+    if is_automodel_checkpoint:
```
yuhezhang-ai: Could we narrow this branch to only adapters that the native exporter actually supports, or fall back to the existing HF PEFT path for the rest? automodel_peft_config.json is emitted for every AutoModel PEFT checkpoint, not only rank-scaled grouped-expert ones. As written, previously supported AutoModel QLoRA (--qlora), CPU, non-causal task types, and explicit non-causal --model-class invocations now enter this branch and raise. It would be good to keep the old path for those cases and add regression tests showing that metadata alone does not change their behavior.

tools/merge_lora.py
```diff
@@ -83,6 +87,73 @@
...
+            f"--model-class={model_class!r} is not supported"
+        )
+    if device.startswith("cuda:"):
+        torch.cuda.set_device(int(device.split(":", 1)[1]))
+
+    from nemo_automodel import NeMoAutoModelForCausalLM, export_merged_peft_checkpoint
+
+    peft_config = _load_automodel_peft_config(adapter_path)
+    logger.info("Loading base model with AutoModel PEFT topology: %s", base_model)
+    model = NeMoAutoModelForCausalLM.from_pretrained(
```
yuhezhang-ai, line 152: Could we construct the offline export model with an explicitly merge-compatible backend? On GPU, BackendConfig defaults to linear='te' when TE is installed. Those linears are patched with delegated super_fwd, while LinearLoRA.materialize_effective_weight() explicitly raises for delegated implementations. This means dense or mixed dense+expert AutoModel adapters can fail in the normal GPU container even though the new CPU test passes because it explicitly uses BackendConfig(linear='torch', ...). Forcing a Torch linear backend here (or supporting TE materialization), plus exercising the actual CLI construction path, would close the gap.

nemo_automodel/_transformers/peft_export.py
```diff
@@ -0,0 +1,170 @@
...
+
+    for module_name, module in model.named_modules():
+        if isinstance(module, GroupedExpertsLoRA):
+            gate_and_up, down = module.materialize_effective_weights()
+            module.gate_and_up_projs.copy_(gate_and_up)
+            del gate_and_up
+            module.down_projs.copy_(down)
+            del down
+        elif isinstance(module, LinearLoRA):
+            module.weight.copy_(module.materialize_effective_weight())
```
yuhezhang-ai, line 84: Could we switch the model to eval mode before merging? Custom AutoModel construction does not call eval(), and materialize_effective_weight() rejects active dropout when self.training and dropout_p > 0. AutoModel checkpoints record dropout, and several recipes set it to 0.1, so those adapters would fail to export. A small regression test using a training-mode model with nonzero dropout would cover this case.

docs/guides/llm/finetune.mdx
jgerh:
```suggestion
description: "Fine-tune LLMs with NeMo AutoModel using SFT or LoRA PEFT, from config and training through evaluation, publishing, and vLLM deployment."
```

jgerh:
```suggestion
**The `recipe` key.** Every config file includes a top-level `recipe` key that tells the CLI *which recipe class* to run. You can write it as a **short name** or as a **fully qualified Python path**, and both resolve to the same class:
```

tools/merge_lora.py
pstjohn: good call out, i'm making this an explicit opt-in `--automodel-native-export` cli option in merge_lora.py. That way we can just use it for certain checkpoints where we need it, and we can possibly remove it if we align some of this with PEFT down the road

docs/guides/llm/finetune.mdx
jgerh:
```suggestion
| `target_modules` | A glob pattern matched against fully qualified layer names (for example, `"*_proj"` matches Llama projection layers such as `q_proj`, `k_proj`, and `down_proj`). |
```

jgerh:
```suggestion
QLoRA requires `bitsandbytes`, which is provided by NeMo AutoModel's `cuda_source`
```

jgerh:
```suggestion
You can run the recipe using the `automodel` CLI or directly with `torchrun` (advanced).
```

docs/guides/llm/finetune.mdx
```diff
@@ -395,7 +395,7 @@ Each log line reports the current loss, gradient norm, peak GPU memory, and toke
...
 
-**PEFT checkpoints** contain only the adapter weights (megabytes instead of gigabytes) and are saved directly under `model/`. They do not use `model/consolidate.sh`. At inference time, you must load the original base model and apply the adapter on top. This distinction affects every downstream step, including inference, publishing, and deployment.
+**PEFT checkpoints** contain only the adapter weights (megabytes instead of gigabytes) and are saved directly under `model/`. They do not use `model/consolidate.sh`. At inference time, you can load the original base model and apply the adapter on top. To produce a standalone checkpoint instead, run `python tools/merge_lora.py --base-model <base-model> --adapter-path <checkpoint>/model --output-dir <output>`. For AutoModel PEFT-v5 checkpoints with grouped-expert adapters, add `--automodel-native-export` to restore the adapter through the model family's state-dict adapter and write canonical sharded Hugging Face weights. This distinction affects every downstream step, including inference, publishing, and deployment.
```
jgerh, line 398:
```suggestion
**PEFT checkpoints** contain only the adapter weights (megabytes instead of gigabytes) and are saved directly under `model/`. They do not use `model/consolidate.sh`. At inference time, you can load the original base model and apply the adapter on top. To produce a standalone checkpoint instead, run `python tools/merge_lora.py --base-model <base-model> --adapter-path <checkpoint>/model --output-dir <output>`. The tool detects NeMo AutoModel PEFT metadata, restores fused dense and grouped-expert adapters through the model family's state-dict adapter, and writes canonical sharded Hugging Face weights. This distinction affects every downstream step, including inference, publishing, and deployment.
```

docs/guides/llm/finetune.mdx
jgerh:
```suggestion
This section documents all available config fields for the fine-tuning recipe. For the Quickstart config, see [Configure Your Training Recipe](#configure-your-training-recipe).
```

jgerh:
```suggestion
# Use a short name (auto-discovered) or a fully qualified Python path:
```

jgerh:
```suggestion
  target_modules: "*_proj" # glob pattern matched against fully qualified layer names;
```

jgerh:
```suggestion
  dataset_name: rajpurkar/squad  # Hugging Face Hub dataset ID
```

jgerh:
```suggestion
| `recipe` | CLI only | Required by `automodel`; select `TrainFinetuneRecipeForNextTokenPrediction` by short name or fully qualified path. |
```

jgerh:
```suggestion
### Data Parallelism as the Default
```

jgerh:
```suggestion
### Tensor Parallelism
```

jgerh:
```suggestion
### Pipeline Parallelism
```

jgerh:
```suggestion
### Context Parallelism
```

jgerh:
```suggestion
### Expert Parallelism for MoE Models
```

jgerh:
```suggestion
### Combine Multiple Dimensions
```

tools/merge_lora.py
pstjohn: thanks for the pointer, yup, I added `backend=BackendConfig(linear="torch")` here to force the torch linear backend

nemo_automodel/_transformers/peft_export.py
pstjohn: done, added on [nemo_automodel/_transformers/peft_export.py#L139](https://github.com/pstjohn/Automodel/blob/9db374131815ad584e578ead55c1999b683043ba/nemo_automodel/_transformers/peft_export.py#L139)

jgerh (commented):
Completed tech pubs review of docs/guides/llm/finetune.mdx. Provided copyedits and text revisions to ensure alignment with style guides, and fixed some mismatched heading levels.

nemo_automodel/_transformers/peft_export.py
```diff
@@ -0,0 +1,156 @@
...
+
+    state_dict = {
```
yuhezhang-ai, line 86: Thanks for addressing the earlier points. Could you please restore the fail-closed check for unsupported LoRA tensors before filtering all `lora_*` keys here, and add a small mixed-module regression test? As written, a model containing a supported dense `LinearLoRA` plus an unsupported adapter such as `GroupedExpertsDeepEPLoRA` passes the non-empty `lora_modules` check, merges only the dense adapter, and then silently drops the unmerged expert LoRA tensors. The previous `unsupported_lora_keys` guard would make this fail clearly instead of producing an incomplete checkpoint.

pstjohn: restored the `unsupported_lora_keys` guard before any weights are merged or adapter keys are filtered and added a regression test combining `LinearLoRA` with `GroupedExpertsDeepEPLoRA` that verifies export raises clearly and leaves all model tensors unchanged.

pstjohn:
/ok to test [96c51da](https://github.com/NVIDIA-NeMo/Automodel/pull/3834/commits/96c51da6ff14422a98ee9ebf6f0865d1fd883799)

pstjohn:
/ok to test [93b351b](https://github.com/NVIDIA-NeMo/Automodel/pull/3834/commits/93b351b5695ded9e32d1b570ee020d896d3bffb9)

nemo_automodel/_transformers/peft_export.py
```diff
@@ -0,0 +1,172 @@
...
+        checkpointer.load_model(model, str(adapter_dir))
+    finally:
+        checkpointer.close()
+
+    model.eval()
+    state_dict = _merge_and_get_hf_state_dict(model)
+    destination = Path(output_dir)
+    destination.mkdir(parents=True, exist_ok=True)
+
+    config.save_pretrained(destination)
```
yuhezhang-ai: Could we strip stale `quantization_config` when exporting dequantized base weights, using the existing checkpoint metadata export handling, and add a reload regression? Custom models loaded from FP8 checkpoints can retain that field, but this exporter calls `to_hf(..., quantization=False)` and then saves the original config unchanged. The result contains BF16 weights while advertising an FP8 checkpoint.

I reproduced this with a tiny LoRA model: export retained the FP8 config without any scale tensors. Exercising Transformers' FP8 reload path on CPU, with only hardware validation mocked, created `FP8Linear` layers and reported seven missing `weight_scale_inv` tensors. Removing the stale config made reload preserve every saved weight exactly. The normal checkpoint metadata path already handles this in `components/checkpoint/addons.py` via `_maybe_strip_quantization_config`; this export path needs the same behavior.

pstjohn: Fixed in 8b9663ee6. The exporter now uses `ConsolidatedHFAddon` to reuse the existing quantization metadata cleanup. Added a CPU regression that exports a BF16 LoRA model carrying stale FP8 metadata, reloads it through Transformers, and verifies every weight matches exactly with no missing or unexpected keys. All 76 focused tests pass.

pstjohn:
/ok to test [2903f72](https://github.com/NVIDIA-NeMo/Automodel/pull/3834/commits/2903f72a8901b15344aed107978e775bff395906)

yuhezhang-ai (approved):
Reviewed the latest revision. The outstanding FP8 export metadata issue is fixed, with a regression verifying exact weights after export and reload. Focused local tests passed (191 passed, 5 skipped), and CI passed on this revision. No remaining blocking findings.

PR #3848, opened 2026-09-09
fix(loss): avoid overflow in fully ignored KD batches

MrCapricornLiu:
# What does this PR do ?

After #3897, a fully ignored KD microbatch returns `student_logits.sum() * 0.0`. This keeps the loss connected to the graph, but the sum can overflow in FP16 even when every logit is finite. For example, 48 logits of value 2000 produce a NaN loss instead of zero.

Sum an empty student-logit view to return a differentiable zero without reading the ignored values. For DTensor logits, use the local tensor so backward preserves the original shard placement and local gradient shape. Supervised batches keep their existing loss and normalization.

# Changelog

- Avoid overflow when an entire KD microbatch is ignored.
- Cover FP32/FP16/BF16, chunked loss, accumulated optimizer updates and two-process vocabulary-sharded DTensor gradients.

**Pre checks**:

- [x] Added regression tests for the affected behavior.
- [x] Checked documentation impact; no public API or configuration changes.

# Additional Information

`pytest --cpu tests/unit_tests/loss/test_kd_loss.py -q`: 42 passed on the updated main branch, including the CPU/Gloo distributed case. Changed-file pre-commit checks and `git diff --check` pass. No new GPU run was needed for this update.

### DIFF
--- a/nemo_automodel/components/loss/kd_loss.py
+++ b/nemo_automodel/components/loss/kd_loss.py
@@ -245,8 +245,10 @@ class KDLoss(nn.Module):
         # Exclude padding / ignored tokens from the loss.
         valid_mask = (labels != self.ignore_index).view(-1)
         if valid_mask.sum() == 0:
-            # Entire batch contains only padding - return zero to keep gradients finite.
-            return student_logits.new_tensor(0.0)
+            # Keep a zero student gradient without reducing the ignored logits.
+            if _HAVE_DTENSOR and isinstance(student_logits, DTensor):
+                student_logits = student_logits.to_local()
+            return student_logits[..., :0].sum()
 
         if student_logits.ndim > 2:
             student_logits = student_logits.view(-1, student_logits.shape[-1])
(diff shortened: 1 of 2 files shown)


HuiyingLi (dismissed):

MrCapricornLiu:
Merged upstream main at `4e00f6be078f`, including the upstream CPU test budget and timeout-handling changes. The previous CPU job hit its 900-second limit on all three attempts. The focused KD tests pass: 41 tests, including the CPU Gloo/DTensor cases.

Could you authorize CI for `376b4b90f945839390cfe9d606b61d880fafe7c1`?

MrCapricornLiu:
Merged current main and updated the scope after #3897. Its graph-connected zero still overflows for finite FP16 logits (for example, 48 values of 2000), while the empty-view reduction returns zero without reading them. Kept the new upstream tests and the DTensor coverage; all 42 KD tests pass on CPU, including the two-process case.

PR #3441, opened 2026-08-06
fix(retrieval): repair Ministral3 training recipe

yuhezhang-ai:
# What does this PR do?

Fix the shipped Ministral3 bi-encoder retrieval recipe end to end: use the BF16 training checkpoint, repair tokenizer construction, make native-FP8 loading safe when requested explicitly, and support the registered tokenizer backend in the collator.

# Root cause

The recipe forced raw Hugging Face tokenizer dispatch and passed `add_eos_token`. With the current Transformers Tekken dispatch, that resolves to `MistralCommonBackend`, which rejects both `add_eos_token` and Hugging Face's injected `_from_auto` kwarg. AutoModel already has a registered `mistral3` backend with the intended BOS/no-EOS behavior, so the forced raw-HF route was unnecessary.

After tokenizer construction succeeded, scoped CI exposed the original checkpoint's native-FP8 scalar `weight_scale_inv` parameters. FSDP2 cannot shard scalar parameters. The shipped training recipe now uses Mistral's BF16 checkpoint. As a safeguard for users who explicitly select a native-FP8 parent checkpoint, the retrieval loader follows the documented Transformers API and passes `FineGrainedFP8Config(dequantize=True)` before extracting the trainable backbone.

The next scoped run then reached collation and showed that `BiEncoderCollator` always passed `return_token_type_ids=False`, although `MistralCommonBackend` explicitly does not support that kwarg. The collator now passes it only to tokenizers that advertise token-type IDs.

# Changelog

- Use AutoModel's registered `mistral3` tokenizer backend in `ministral3_3b_instruct.yaml`.
- Use `mistralai/Ministral-3-3B-Instruct-2512-BF16` for the shipped training recipe.
- Dequantize explicitly selected native-FP8 parent checkpoints through Transformers' `FineGrainedFP8Config` API.
- Avoid unsupported token-type kwargs for strict tokenizer backends.
- Add focused tokenizer, FP8 forwarding, and collator regression tests.
- Add a 2-GPU, 10-step nightly smoke for the shipped Ministral3 retrieval recipe.

# Validation

- Current-head tokenizer, retrieval, collator, and CI generator/resolver suite: 79 passed.
- Broader tokenizer, retrieval, and CI coverage before the final collator follow-up: 123 passed.
- `ruff format .`
- `ruff check --fix .`
- Instantiated the edited YAML's real cached tokenizer and verified `MistralCommonBackend`, left padding, BOS insertion, and no EOS insertion.
[...]

PR #3102, opened 2026-07-16
feat(tokenizer): add explicit backend selection

oliverholworthy:
# What does this PR do?

Add an explicit tokenizer backend selector and use it to make tokenizer selection deterministic for the maintained Ministral retrieval recipe.

For Ministral checkpoints, Transformers `AutoTokenizer` can select `MistralCommonBackend` when `tekken.json` and the optional `mistral-common` package are available. This makes the tokenizer implementation depend on the loading environment. The new selector names all four supported loading routes while retaining `force_default` and `force_hf` as compatibility aliases.

# Changelog

- Add `tokenizer_backend` to `NeMoAutoTokenizer.from_pretrained`:
  - `nemo_auto`: preserve the existing NeMo registry dispatch and wrapped `AutoTokenizer` fallback.
  - `nemo_wrapped_auto`: bypass registered tokenizers and use Transformers `AutoTokenizer` with NeMo's tokenizer compatibility wrapper.
  - `transformers_auto`: delegate directly to Transformers `AutoTokenizer` without the NeMo wrapper.
  - `tokenizers`: load `tokenizer.json` directly through Transformers `TokenizersBackend`, bypassing automatic backend dispatch.
- Keep `force_default=True` as an alias for `tokenizer_backend="nemo_wrapped_auto"` and `force_hf=True` as an alias for `tokenizer_backend="transformers_auto"`.
- Repair a missing pad token ID after direct `TokenizersBackend` loading so padded batches remain usable.
- Reject unknown selectors and contradictory legacy flag or selector combinations with clear errors.
- Configure the Ministral-3B retrieval recipe to use direct `TokenizersBackend` loading with an explicit retrieval policy:
  - corrected Mistral regex;
  - split literal special-token spellings;
  - BOS enabled; and
  - EOS disabled.
- Replace `force_hf: true` in the Nemotron-1B distillation recipe with the behaviorally equivalent `tokenizer_backend: transformers_auto`.
- Document the four public loading routes and their legacy flag compatibility.
- Add selector, exhaustive legacy compatibility, pad-token, policy-forwarding, recipe-policy, deterministic token-ID, and portable save/reload coverage.

# Behavioral compatibility

Before this PR, the legacy flags selected distinct, deterministic routes:

- with neither flag, NeMo used a registered tokenizer when available and otherwise used its wrapped Transformers fallback;
- `force_default=True` bypassed the NeMo tokenizer registry but still used the NeMo-wrapped Transformers fallback;
[...]

PR #3849, opened 2026-09-09
fix(data): support Mistral tokenizers in the distillation collator

MrCapricornLiu:
# What does this PR do ?

`BiEncoderDistillCollator` passes `return_token_type_ids=False` to every tokenizer call. `MistralCommonBackend` does not accept this argument, so retrieval distillation fails while encoding the first query. The regular `BiEncoderCollator` already handles this backend in #3441.

Apply the same convention to queries, positive documents and hard negatives: disable token type IDs explicitly only when the tokenizer advertises them in `model_input_names`.

# Changelog

- Support tokenizer backends without the `return_token_type_ids` argument throughout retrieval distillation collation.
- Extend cached-teacher coverage with a strict tokenizer and verify token contents with the real Mistral SentencePiece and Tekken backends using their bundled vocabularies.

**Pre checks**:

- [x] Added necessary regression tests.
- [x] Checked documentation impact; no new API, configuration or dependency requirement.

# Additional Information

`pytest tests/unit_tests/datasets/llm/test_retrieval_distill_collator.py tests/unit_tests/datasets/llm/test_bi_encoder_collator.py -q`: 17 passed with both Transformers 5.3.0 and the current 5.15.1 pin, using mistral-common 1.11.7. The updated distillation tests give 3 failures and 1 pass against the original collator. Ruff 0.16.4, Bandit B614 and `git diff --check` pass on the changed files.

The Mistral tests use local package assets and skip when the optional backend is absent. No GPU or complete retrieval training run was needed or performed. Related #3102 changes backend selection; its distillation configuration still dispatches through AutoTokenizer and does not modify this collator.

### DIFF
--- a/nemo_automodel/components/datasets/llm/retrieval_distill_collator.py
+++ b/nemo_automodel/components/datasets/llm/retrieval_distill_collator.py
@@ -213,19 +213,23 @@ class BiEncoderDistillCollator:
         neg_lists_raw = [docs[1:] for docs in doc_lists]
         neg_lists = [[self._apply_passage_prefix(n, pi) for n in negs] for negs, pi in zip(neg_lists_raw, passage_inst)]
 
+        token_type_kwargs = {}
+        if "token_type_ids" in getattr(self.tokenizer, "model_input_names", []):
+            token_type_kwargs["return_token_type_ids"] = False
+
         q_enc = self.tokenizer(
             query_texts,
             max_length=self.q_max_len,
             padding=PaddingStrategy.DO_NOT_PAD,
             truncation=True,
-            return_token_type_ids=False,
+            **token_type_kwargs,
         )
         d_enc = self.tokenizer(
             pos_docs,
             max_length=self.p_max_len,
             padding=PaddingStrategy.DO_NOT_PAD,
             truncation=True,
-            return_token_type_ids=False,
+            **token_type_kwargs,
         )
 
         q_features = [{k: q_enc[k][i] for k in q_enc.keys()} for i in range(len(query_texts))]
@@ -300,7 +304,7 @@ class BiEncoderDistillCollator:
             max_length=self.p_max_len,
             padding=PaddingStrategy.DO_NOT_PAD,
             truncation=True,
-            return_token_type_ids=False,
+            **token_type_kwargs,
         )
         n_features = [{k: n_enc[k][i] for k in n_enc.keys()} for i in range(len(flat_negs))]
         n_batch = self.tokenizer.pad(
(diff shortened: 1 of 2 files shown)


MrCapricornLiu:
Merged upstream main at `4e00f6be078f` and corrected the optional Mistral test guard. Transformers 5.15.1 requires mistral-common >= 1.11.5, so the previous import-only guard incorrectly ran the backend tests with the CI image's 1.11.0.

The two collator modules pass all 17 tests with mistral-common 1.11.7 on this revision. The earlier 1.11.0 check passed 15 tests and skipped the two backend cases; the guard and collator code are unchanged by this merge. This also picks up the upstream CPU CI timeout changes.

Could you authorize CI for `0a6637a921e258d8cda5db7bf21e2a6370a69092`?

nemo_automodel/components/datasets/llm/retrieval_distill_collator.py
```diff
@@ -213,19 +213,23 @@ def __call__(self, batch: List[dict[str, Any]]) -> dict[str, torch.Tensor]:
         neg_lists_raw = [docs[1:] for docs in doc_lists]
         neg_lists = [[self._apply_passage_prefix(n, pi) for n in negs] for negs, pi in zip(neg_lists_raw, passage_inst)]
 
+        token_type_kwargs = {}
```
vinay-raman, line 218: Add a comment here:
"Some tokenizer backends, including MistralCommonBackend, do not support
he return_token_type_ids kwarg and do not advertise token type IDs."

vinay-raman (approved):

nemo_automodel/components/datasets/llm/retrieval_distill_collator.py
MrCapricornLiu: Added the comment next to `token_type_kwargs` explaining that MistralCommonBackend neither accepts the kwarg nor advertises token type IDs. Both collator test modules pass (17 tests).

yuhezhang-ai:
/ok to test [35e6082](https://github.com/NVIDIA-NeMo/Automodel/pull/3849/commits/35e6082c9a83d6bd51c79917f2b6e2c48e07689c)

yuhezhang-ai (approved):
Reviewed the tokenizer compatibility change and its regression coverage. No actionable findings. All 17 focused collator tests passed, including the real Mistral SentencePiece and Tekken cases; Ruff lint and formatting checks also passed. Local validation used Transformers 5.12.1 and mistral-common 1.11.0.

PR #3850, opened 2026-09-10
fix(loss): scale InfoNCE logits before masking padded negatives

MrCapricornLiu:
# What does this PR do ?

Padded hard negatives can produce a finite InfoNCE loss but a NaN gradient for a learned temperature. Their logits are first set to `-inf` and then divided by the temperature, so division backward encounters an infinite operand even though the padded candidate contributes no loss.

Scale positive/in-batch and hard-negative logits before applying the padding mask in both `infonce_loss` and `infonce_distill_loss`. Teacher logits remain detached. The probabilities and embedding gradients are preserved, while infinite mask values stay out of the temperature derivative.

# Changelog

- Keep learned temperature gradients finite for padded negatives in ordinary and distillation InfoNCE.
- Compare losses and student/temperature gradients against per-query references that physically omit padded candidates, including KL, CE and MSE distillation.
- Cover valid q2d/symmetric configurations and fully valid, ragged and fully padded negative sets; preserve the configuration checks added in #3870.

**Pre checks**:

- [x] Added regression tests.
- [x] Checked documentation impact; no new API or configuration.

# Additional Information

`pytest --cpu tests/unit_tests/loss/test_infonce.py -q`: 98 passed. Before the distillation fix, 18 of its 27 new cases failed at the temperature-gradient check; the nine fully valid cases already passed. The references independently compare losses and all affected gradients after removing padded candidates.

Changed-file pre-commit checks and `git diff --check` pass. This update was validated on CPU; it does not claim a new distributed or GPU benchmark.

### DIFF
--- a/nemo_automodel/components/loss/infonce.py
+++ b/nemo_automodel/components/loss/infonce.py
@@ -70,14 +70,14 @@ def infonce_loss(
             logits = pos
             target = torch.zeros(batch, device=device, dtype=torch.long)
 
+        logits = logits / temperature
         if neg is not None:
-            sim_qn = torch.einsum("bd,bkd->bk", query, neg)
+            sim_qn = torch.einsum("bd,bkd->bk", query, neg) / temperature
             if hard_negatives_mask is not None:
                 pad = hard_negatives_mask.to(sim_qn.dtype) == 0
                 sim_qn = sim_qn.masked_fill(pad, float("-inf"))
             logits = torch.cat([logits, sim_qn], dim=-1)
 
-        logits = logits / temperature
         return F.cross_entropy(logits, target)
 
     if direction == "q2d":
--- a/tests/unit_tests/loss/test_infonce.py
+++ b/tests/unit_tests/loss/test_infonce.py
@@ -431,6 +431,52 @@ def test_infonce_module_learnable_temperature():
     assert torch.allclose(loss_fn.current_temperature(), torch.tensor(0.05), atol=ATOL)
 
 
+@pytest.mark.parametrize("direction", ["q2d", "symmetric"])
+@pytest.mark.parametrize("in_batch", [False, True])
+@pytest.mark.parametrize("mask_kind", ["all_valid", "ragged", "all_padding"])
+def test_infonce_learnable_temperature_with_masked_negatives(direction, in_batch, mask_kind):
+    torch.manual_seed(42)
+    inputs = [torch.randn(*shape, requires_grad=True) for shape in [(3, 8), (3, 8), (3, 2, 8)]]
+    mask = torch.tensor([[1, 0], [0, 0], [1, 1]], dtype=torch.bool)
+    if mask_kind == "all_valid":
+        mask.fill_(True)
+    elif mask_kind == "all_padding":
+        mask.fill_(False)
+    loss_fn = InfoNCELoss(
+        temperature=0.2,
+        learnable_temperature=True,
+        direction=direction,
+        use_in_batch_negatives=in_batch,
+        cross_device_negatives=False,
+    )
+    reference_inputs = [x.detach().clone().requires_grad_() for x in inputs]
+    log_inv_tau = loss_fn.log_inv_tau.detach().clone().requires_grad_()
+    q, d, negatives = [torch.nn.functional.normalize(x, dim=-1) for x in reference_inputs]
+
+    def reference_side(query, docs, include_negatives):
+        losses = []
+        for i in range(len(query)):
+            candidates = docs if in_batch else docs[i : i + 1]
+            if include_negatives:
+                candidates = torch.cat([candidates, negatives[i, mask[i]]])
+            scores = (candidates @ query[i]) * log_inv_tau.exp()
+            positive = i if in_batch else 0
+            losses.append(torch.logsumexp(scores, dim=0) - scores[positive])
+        return torch.stack(losses).mean()
+
+    expected = reference_side(q, d, True)
+    if direction == "symmetric":
+        expected = (expected + reference_side(d, q, False)) / 2
+    expected.backward()
+    actual = loss_fn(*inputs, hard_negatives_mask=mask)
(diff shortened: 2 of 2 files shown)


MrCapricornLiu:
The branch now includes the CPU CI timeout changes from upstream main at `4e00f6be078f`. The previous run exhausted the 900-second limit; the focused InfoNCE suite passes all 61 tests on the updated branch.

Could you authorize CI for `3346123fb376db5de8422821a3b9aefe01e85d04`?

nemo_automodel/components/loss/infonce.py
```diff
@@ -70,14 +70,14 @@ def _one_direction(query: torch.Tensor, doc: torch.Tensor, neg: torch.Tensor | N
             logits = pos
             target = torch.zeros(batch, device=device, dtype=torch.long)
 
+        logits = logits / temperature
```
vinay-raman, line 120: explicity call out that temperature is learnable, add relevant comments

nemo_automodel/components/loss/infonce.py
vinay-raman: do you also want to add same scale & then mask logic here as well? right now temperature = float for this, change it to temperature = float | torch.Tensor. 
add a corresponding test, thanks!

tests/unit_tests/loss/test_infonce.py
```diff
@@ -431,6 +431,52 @@ def test_infonce_module_learnable_temperature():
...
+        return torch.stack(losses).mean()
+
+    expected = reference_side(q, d, True)
+    if direction == "symmetric":
+        expected = (expected + reference_side(d, q, False)) / 2
+    expected.backward()
+    actual = loss_fn(*inputs, hard_negatives_mask=mask)
+    actual.backward()
+
+    torch.testing.assert_close(actual, expected)
```
vinay-raman, line 565: add explicit assert torch.isfinite(loss_fn.log_inv_tau.grad)

vinay-raman (changes requested):

nemo_automodel/components/loss/infonce.py
MrCapricornLiu: The docstrings now explicitly describe a learned scalar temperature, and both loss paths explain why scaling must precede the padding mask.

nemo_automodel/components/loss/infonce.py
MrCapricornLiu: Applied scale-before-mask to `infonce_distill_loss` as well. The temperature annotation accepts `float | torch.Tensor`. Added 27 KL/CE/MSE cases comparing loss and student/temperature gradients against references that remove padded candidates. The 18 padded cases failed before the fix; the updated module passes all 98 tests. Also aligned the test configurations with #3870.

tests/unit_tests/loss/test_infonce.py
MrCapricornLiu: Added the explicit finite-gradient assertion for `loss_fn.log_inv_tau.grad`, alongside the reference-gradient comparison. The new distillation tests explicitly check their temperature gradient too.

rnyak:
@MrCapricornLiu thanks for the MR. Is the goal to use `learnable_temperature: true` with this distillation recipe? This creates log_inv_tau as an `nn.Parameter`, so it can receive gradients when the InfoNCE loss is active. However, the optimizer is built only from the student model’s parameters, before self.infonce_loss is instantiated. Therefore, log_inv_tau does not appear to be included in the optimizer, so optimizer.step() will not change the temperature.  You can verify that at your end, and in case you want the  `log_inv_tau` param be added to the optimizer, you can make changes accordingly.

nemo_automodel/components/loss/infonce.py
rnyak, line 370: @MrCapricornLiu, Is the goal to use learnable_temperature: true with this distillation recipe? This creates log_inv_tau as an nn.Parameter, so it can receive gradients when the InfoNCE loss is active. However, the optimizer is built only from the student model’s parameters, before self.infonce_loss is instantiated. Therefore, log_inv_tau does not appear to be included in the optimizer, so optimizer.step() will not change the temperature. You can verify that at your end, and in case you want the log_inv_tau param be added to the optimizer, you can make changes accordingly.

yuhezhang-ai:
Thanks for the fix. Scaling before masking is correct, it gives the same result for a fixed float temperature, and the reference-based tests cover it well.

Building on @rnyak's note above (thanks for catching that!): `learnable_temperature` isn't a working feature today, so this change only makes the gradient finite for a parameter that's never trained. Two options:

1. **Merge as is.** The fix is correct and harmless, and we can merge it now.
2. **If you actually need a learned temperature,** please extend this PR to make it a real feature. Beyond adding `log_inv_tau` to the optimizer as @rnyak pointed out, a few more pieces are needed:
   - **Checkpointing:** `BaseRecipe` skips any attribute with `loss` in its name (`base_recipe.py:212`), so `self.infonce_loss` is never saved, and a resumed run would reset τ to its initial value.
   - **Gradient sync:** `log_inv_tau` sits outside FSDP/DDP, so its gradient isn't all-reduced. Without cross-device negatives, each rank computes a different loss and τ would drift apart across ranks.
   - **Gradient clipping and device:** clipping only covers `model_parts`, and the loss module is never moved to the training device.
   - **Tests:** show that τ changes after an optimizer step, survives a checkpoint save and resume, and stays identical across ranks in a 2-GPU run.

   A learned temperature makes sense for the contrastive loss only. `InfoNCEDistillLoss` should keep a fixed τ: the teacher logits are computed under `no_grad`, so a shared τ would only get the student-side gradient, and learning it is degenerate anyway because KL(p_t(τ) ‖ p_s(τ)) goes to 0 as τ → ∞.

Let us know which way you'd like to go.

Nit, non-blocking: in `infonce_distill_loss`, the "a 0-dim Tensor when it is learned" docstring and the "student temperature can be learned" comment describe a mode that doesn't exist. Consider dropping them.

PR #3853, opened 2026-09-10
fix(loss): apply per-token gradients in parallel cross entropy

MrCapricornLiu:
# What does this PR do ?

`TEParallelCrossEntropy(reduction="none")` currently scales every token gradient by the first element of `grad_output`. A weighted token loss therefore produces incorrect parameter gradients; if its first weight is zero, every gradient becomes zero even when other tokens contribute to the loss.

Read the upstream gradient for each row when the loss is unreduced, materializing noncontiguous upstream gradients before the kernel call. Reduced losses and the soft-CE caller retain scalar scaling.

The existing two-GPU DTensor functional test now compares weighted and reduced losses, vocabulary-sharded projection gradients, their global norm, and an equivalent SGD update with a full-vocabulary PyTorch reference. FP32/BF16 cover transposed, contiguous and expanded weights, plus mean/sum reductions. The same ten cases pass with TP1, and all 21 existing soft-CE tests pass. The unmodified kernel fails the first weighted-gradient comparison. Complete trainer runs, multi-node and DP/CP/PP combinations were not exercised.

- [x] Added regression coverage.
- [x] Checked documentation impact; no new public API or configuration.

### DIFF
--- a/nemo_automodel/components/loss/triton/te_cross_entropy.py
+++ b/nemo_automodel/components/loss/triton/te_cross_entropy.py
@@ -260,6 +260,7 @@ def element_mul_kernel(
     grad_output_ptr,
     n_cols,
     BLOCK_SIZE: tl.constexpr,
+    PER_ROW: tl.constexpr = False,
 ):
     """
     This function multiplies each element of the tensor pointed by X_ptr with the value pointed by grad_output_ptr.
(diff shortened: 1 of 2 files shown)


HuiyingLi (approved):

MrCapricornLiu:
Merged upstream main at `4e00f6be078f` to pick up the upstream CPU CI budget and timeout handling after all three previous attempts hit the 900-second limit. The loss kernel and its GPU regression file are unchanged; the existing TP1/TP2 numerical results still apply to those files. No new GPU run was performed for this merge.

Could you authorize CI for `5a1dd794f805e3f9e127a0600d26ec405ccd9738`?

PR #3699, opened 2026-08-27
feat(glm5-next): add GLM-5.3-Flash training support

HuiyingLi:
# What does this PR do ?

Adds native AutoModel training support for [zai-org/GLM-5.3-Flash](https://huggingface.co/zai-org/GLM-5.3-Flash), including its hybrid Kimi Delta Attention (KDA) / DeepSeek Sparse Attention (DSA) decoder, MoE routing, VLM processing, Hugging Face checkpoint conversion, and packed-sequence context parallelism.

The included MedPix recipe uses the validated EP72/CP2 topology: TP1, PP1, 9 nodes / 72 GPUs, packed sequence length 2048, and 100 optimizer steps. It now selects the shared FlashMLA-forward/cuDNN-backward kernel for sparse MLA layers; KDA layers remain on FLA.

> [!IMPORTANT]
> Full-model single-GPU checkpoint loading and training are not supported. The supported base-checkpoint initialization path is distributed DCP loading.
>
> User training does not require Transformers 5.16 or a nightly build. This implementation supports AutoModel's pinned `transformers==5.12.1`, provides its own GLM-5.3 configuration and processor registration, and does not require `trust_remote_code`.

# Changelog

- Register the GLM-5.3-Flash Transformers architecture and native AutoModel configuration.
- Implement the multimodal processor, vision tower, mHC decoder blocks, dense/MoE layers, and HF state-dict adapter with FP8 checkpoint dequantization.
- Run KDA with FLA production kernels and preserve its left-to-right recurrence with contiguous packed CP shards.
- Implement packed DSA CP with document isolation and differentiable full-sequence gather, including empty-local-shard backward participation.
- Reuse a model-independent FlashMLA-forward/cuDNN-backward latent sparse-attention adapter for GLM-5.2 and GLM-5.3; preserve GLM-5.2's model-specific `D=576`, top-k <= 2048 contract.
- Absorb GLM-5.3's K projection into the query, attend over its shared 512-wide latent K/V, then apply the model-owned V projection.
- Align MoE routing with the HF reference: correction bias selects experts while routing weights use unbiased sigmoid scores.
- Add a MedPix packed-2K EP72/CP2 recipe plus unit, numerical, distributed, and functional parity coverage.

# Validation

## Hugging Face numerical parity

Reference setup:

- Validation-only truncated model with the first 4 decoder layers: KDA, KDA, KDA, DSA
- Sequence length: 4096
- Full vocabulary: 154,880 logits for all 4096 token positions
[...]

PR #3855, opened 2026-09-10
feat: add DeepSeek-V4.1-Flash model training support

khazic:
## What does this PR do?

Add DeepSeek-V4.1-Flash full-sequence training and fine-tuning support to AutoModel, including the text backbone, optional image inputs, checkpoint adaptation, sharded Engram memory, and a HellaSwag fine-tuning recipe.

## Changes from DeepSeek-V4

image

- Implement the V4.1 causal encoder/decoder layer schedule and CSA2 cross-layer sharing. Full layers produce compressed KV and index keys, Reindex layers refresh the selection against shared keys, and Reuse layers consume the shared KV and selection.
- Apply hierarchical candidate restrictions to later decoder indexers. The current implementation computes full index scores before masking; candidate-only scoring acceleration is not implemented.
- Implement single-pass mHC coefficient handoff, with each sublayer consuming the preceding sublayer's input-mixing coefficients.
- Integrate tokenizer-dependent Engram hashing, gated residual updates, and row-owner-sharded embedding tables.
- Preserve FP8 SWA and FP4 compressed-KV/indexer quantize/dequantize boundaries for floating-point training. This does not provide packed FP4 inference-cache storage.
- Support optional vision encoding, image projection, learned image delimiters, and modality-aware routing. The default model configuration enables vision; the text-only HellaSwag recipe disables it.
- Add typed text/vision configuration, processor support, released-checkpoint mappings, and FSDP2/EP integration. Reuse shared V4 attention, vision, and MoE components where applicable.

CSA2 state is passed explicitly between layers and retains the autograd history needed by shared-KV consumers and activation recomputation.

Additions and deletions by @HuiyingLi :
  - Engram parallelism: reused Qwen3.8 Next’s row-sharded embedding, including distributed lookup, gradient ownership, and checkpoint shard handling.
  - Reference numerical alignment:
    - Aligned mHC/Sinkhorn, complex RoPE, and MX quantization rounding.
    - Corrected indexer/compressor precision and sparse-attention forward rounding.
  - Model and loading alignment:
    - Migrated ds41’s configuration, initialization, parameter naming, and FP32 parameter/logit handling.
    - Fixed automatic DCP dequantization and distributed loading of FP8 dense/Engram weights and FP4 experts.
    - Restored tokenizer-derived Engram buffers after model materialization.
  - Integration and validation:
    - Added optional TE RMSNorm.
    - Added the 16-node recipe.
    - Full model training validation [W&B run](https://wandb.ai/Nemo-automodel/deepseek-v41/runs/pr3855-te-7074698).
  - Removed or replaced:
    - Extra Engram enable/freeze and KV quantization switches.
    - Legacy checkpoint decoders and V4.1-specific DSV4 FSDP extensions.
  
## Scope and limitations
- Indexers remain frozen; indexer distillation is outside this implementation. The recipe fine-tunes the remaining backbone parameters and enabled Engram tables.
- The recipe uses FusedAdam/AdamW with fixed router correction biases. It is a fine-tuning configuration, not a reproduction of the paper's Muon and Sinkhorn-balanced pretraining optimizer.
- Encoder-only prefill, incremental generation caches, Decoder SWA bounded replay, DSpark draft training, and million-token training are outside the current scope. No prefill-halving or packed-cache memory benefit is claimed.
- Native HybridEP combine introduces BF16 rounding before adding routed and shared expert outputs, so it does not preserve the reference’s FP32 accumulation behavior and remains a source of numerical differences.

## Validation

The [versioned validation record](https://github.com/NVIDIA-NeMo/Automodel/blob/e24155cdad777fdc8a7e9152b6edd00e1940d4e0/docs/model-coverage/llm/deepseek-ai/deepseek-v41-flash.mdx#L123) reports a real-weight comparison covering all 40 layers, 4,096 positions, and all 129,280 vocabulary entries against the unchanged official inference implementation at checkpoint revision `df42c109f1defefcbfcedbe7d905718a12266e40`.

The first four released layers and the full layer-1 Engram table were loaded through the standard NeMo Checkpointer/DCP path. The forward used four resident layers with EP4, four Engram owners, torch_mm experts, native HybridEP dispatch/combine, and TileLang attention.

The comparison covered 4,096 tokens and all 129,280 vocabulary entries against the unchanged official inference implementation at checkpoint revision df42c109f1defefcbfcedbe7d905718a12266e40. Reference weights were decoded for BF16 compute while retaining the original cache/index quantization.

| Metric | Recorded result |
| :--- | ---: |
| Mean KL | 0.0062182555 |
| P95 KL | 0.0357103683 |
| Maximum KL | 0.2955792844 |
| Logit cosine | 0.9993935077 |
| Top-1 agreement | 94.50683594% |

The full model completed 100 HellaSwag training updates using the released weights for all 40 backbone layers and both Engram tables on 16 nodes × 4 GB200 GPUs, with EP64 and 64 Engram owners. The run used `torch_mm` experts and HybridEP. Training loss decreased from 2.9959 to 1.8177; validation losses were 2.0687 and 2.0173 at updates 50 and 100. All recorded losses and gradient norms were finite. [W&B run](https://wandb.ai/Nemo-automodel/deepseek-v41/runs/pr3855-te-7074698).
  
image

### Other historical evidence

- The earlier [six-layer no-Engram results](https://github.com/NVIDIA-NeMo/Automodel/pull/3855#issuecomment-5621814232) belong to an earlier implementation and included uncommitted fixes. They are historical diagnostics, not validation of the current head. Earlier packed-sequence, unit-test-count, and restoration claims likewise must not be interpreted as current capabilities or a current-head pass.

## Before this PR is ready to merge

- [x] Add model documentation, a fine-tuning recipe, and unit/functional validation harnesses.
- [ ] Complete numerical parity and distributed validation for the proposed revision.
- [ ] Validate checkpoint restoration for the supported training configuration.
- [ ] Pass the required CI checks.


khazic:
## Paper alignment and implementation scope

Scope checked at `9b4c21153120b3724110794689e54729011b265c`. This PR provides full-sequence training and fine-tuning support, including optional image inputs. It does not reproduce the paper's complete pretraining or serving system.

### Implemented with differences from the paper

| Area | Current implementation | Difference and consequence |
| :--- | :--- | :--- |
| Hierarchical Reindex scoring (Section 2.3.2) | Computes scores against all index keys, then applies the shared candidate mask. | Candidate restriction is implemented, but later indexers do not score only the candidates. The paper's bounded candidate-scoring cost is not achieved. |
| FP4 KV representation (Section 2.4.4) | Applies FP4 quantize/dequantize boundaries and returns floating-point tensors. | This models quantization effects during training; it does not implement packed FP4 inference-cache storage or its memory/bandwidth savings. |
| Indexer training | Indexer parameters are frozen and their forward runs without autograd. | Fine-tuning updates the other trainable modules. Indexer distillation/training is not implemented; this does not reproduce optimization of the indexers under the paper's training procedure. |
| Optimizer and load balancing (Sections 2.5 and 4.2.2) | The recipe uses FusedAdam/AdamW; default router bias updates and auxiliary balancing loss are disabled. | This is a fine-tuning recipe, not the paper's Muon/head-wise Muon, Sinkhorn-balanced embedding/head updates, and pretraining load-balancing configuration. |
| Engram storage and scheduling (Sections 2.4.2 and 3.1.3) | Uses floating-point row-owner-sharded tables with differentiable lookup. | The paper's FP8 storage/fetch path, dedicated prefetch/communication overlap, and specialized embedding optimizer are not fully reproduced. |

### Not implemented in this PR

- Encoder-only prefill, incremental generation KV caches, and Decoder SWA bounded replay. Full-sequence CED training does not imply the paper's prefill compute savings.
- Candidate-only Reindex scoring acceleration and packed FP4 serving-cache management.
- DSpark draft training and indexer distillation.
- Mega-mHC deployment fusion. The shifted mHC coefficient handoff is implemented; the serving kernel optimization is separate.
- THD packed inputs, TP, PP, CP, and million-token training. Both text and image inputs currently require unpacked sequences.

These scope exclusions are distinct from demonstrated mathematical bugs. The current PR still needs correct shared-state semantics, gradients, and checkpoint behavior within its supported training scope.

### Validation still required

- Complete loaded-weight auditing and numerical localization for the revision proposed for merge.
- Establish acceptable forward numerical parity and distributed/multi-step training behavior; successful execution or finite gradients alone is insufficient.
- Validate fresh checkpoint reload and training resume for the supported recipe. Checkpoint saving is disabled in the supplied recipe, and bitwise TileLang training continuation is not established.
- Run the relevant current-revision tests and required CI checks. Historical results must remain attributed to their tested commits.

The PR description contains the newer, commit-specific 40-layer streamed validation record. The earlier six-layer measurements below are retained for history and do not establish validation of this revision.


<details>
<summary>Historical six-layer experiment and metrics</summary>

## Six-layer real-weight validation update

The six-layer, no-Engram subset completes forward execution on both attention backends, but numerical parity is **not yet passing**. This describes the earlier tested implementation only.

### Setup and revision

- Original zero-based layers `[1, 2, 3, 20, 24, 25]`, renumbered to `0..5`, plus embeddings, final norm, and LM head. This covers SWA, encoder Full/Reuse, and decoder Full/Reindex/Reuse.
- Real checkpoint weights and BF16 compute with FP32-sensitive operations. The reference uses independently decoded FP4/FP8 weights and the official inference implementation with MP8; AutoModel uses FSDP2/EP8.
- TileLang results use **TileLang sparse attention with torch indexer scoring**. The original indexer kernel exceeded the available dynamic shared-memory limit in this validation configuration.
- The tested revision is based on `bd4ce5bea72de3da089d419216b953d50e48026d` with additional uncommitted configuration/router-loading fixes. These measurements must not be attributed to the committed PR revision alone. The new 4096-token run records the working-tree patch and its SHA256.
- The numerical reporting follows the truncated-model portion of #3699. This is not a reproduction of its full-model, multi-node CP training or performance tests.

### Completed forward measurements

Code, English, and Chinese inputs at 128, 512, and 1025 tokens: reference **9/9**, SDPA **9/9**, TileLang **9/9**, with finite logits. Every token position and all **129,280 vocabulary entries** are included.

Each paired entry below is **SDPA / TileLang**. Cosine is computed over flattened logits; KL is `KL(reference || AutoModel)` with natural logarithms.

| Case | Logits cosine | Top-1 agreement | Mean KL |
| :--- | ---: | ---: | ---: |
| code_128 | 0.985250 / 0.986639 | 64.06% / 73.44% | 0.145623 / 0.133700 |
| code_512 | 0.993015 / 0.991695 | 86.52% / 83.01% | 0.075493 / 0.094328 |
| code_1025 | 0.994374 / 0.994413 | 89.07% / 89.07% | 0.062605 / 0.061263 |
| english_128 | 0.962678 / 0.964212 | 57.03% / 59.38% | 0.412085 / 0.390124 |
| english_512 | 0.985466 / 0.987391 | 82.03% / 81.84% | 0.164041 / 0.136067 |
| english_1025 | 0.988963 / 0.989122 | 80.39% / 81.17% | 0.118051 / 0.115625 |
| chinese_128 | 0.973177 / 0.976250 | 55.47% / 66.41% | 0.280769 / 0.251475 |
| chinese_512 | 0.990759 / 0.989080 | 79.69% / 77.54% | 0.099711 / 0.120134 |
| chinese_1025 | 0.991563 / 0.991783 | 81.07% / 82.93% | 0.090142 / 0.090453 |

Additional full-distribution metrics for `code_1025`:

| Metric | SDPA | TileLang |
| :--- | ---: | ---: |
| P95 per-token KL | 0.228187 | 0.242855 |
| Maximum per-token KL | 3.961746 | 3.704509 |
| Mean JSD | 0.013858 | 0.013696 |
| Mean absolute logit difference | 0.186872 | 0.187517 |

The existing strict forward gate (`max_abs < 0.01` and minimum per-token cosine `> 0.9999`) passes **0/9 cases for either backend**. Similar mean NLL is not sufficient evidence of parity. Error sources remain under investigation; these differences are not being classified as expected BF16 noise.

### Additional completed checks

- The unit-test run recorded with this earlier experiment passed **140 tests**. This run includes uncommitted fixes and is separate from the earlier 135-test baseline in the PR description.

- One real-weight backward pass and SGD update completed for each backend on all eight ranks. Gradient statistics are finite and no trainable gradients are missing. This is a finite-backward/update smoke, **not reference gradient parity**.
- Different ranks reported different losses for the same nominal input. Cross-rank input/output consistency and repeated-forward diagnostics remain necessary before claiming distributed training parity.
- **72/72 identical-score candidate-selection comparisons match the unchanged reference function exactly.** Cases include empty visibility, partial blocks, ties, explicit pooled positions, and widths exceeding the released 2048-block candidate budget. This is component-level coverage, not end-to-end real-weight pruning validation.
- At 128 tokens, `index_topk=512` can retain every visible compressed position and the candidate budget is not saturated. Indexer top-k cutoff flips alone therefore do not explain the short-sequence discrepancy.

### Historical follow-up status

The earlier 4096-token runs produced finite outputs for all three input types on the reference, SDPA, and TileLang. The subsequent audit/training launcher failed before those follow-up tests started. This historical record does not claim that a job is currently queued or running, or that W&B upload completed. Current merge requirements are listed above.

Engram and full-model validation remain outside this six-layer no-Engram experiment.

</details>

docs/model-coverage/llm/deepseek-ai/deepseek-v41-flash.mdx
```diff
@@ -0,0 +1,150 @@
...
+| Model ID | `deepseek-ai/DeepSeek-V4.1-Flash` |
+| Reference revision | `df42c109f1defefcbfcedbe7d905718a12266e40` |
+| Configuration | `DeepseekV41Config` with typed `text_config` and `vision_config` |
+
+<Note>
+The numerical results below belong to the original implementation at commit
+`0d5919a5a506dd35dc4f55769410feb98ae854b8`. Final validation of the current PR
+implementation remains pending. The 16-node and 32-node validation runs are
+also pending; neither establishes successful resident 100-update training.
+</Note>
```
jgerh:
```suggestion
also pending; neither establishes successful resident 100-update training.

</Note>
```

docs/model-coverage/llm/deepseek-ai/deepseek-v41-flash.mdx
```diff
@@ -0,0 +1,150 @@
...
+
+## Fine-tuning
```
jgerh:
```suggestion
## Fine-Tuning
```

docs/model-coverage/llm/deepseek-ai/deepseek-v41-flash.mdx
```diff
@@ -0,0 +1,150 @@
...
+
+Both recipes train all 40 text-backbone layers and both Engram tables. Dense
+parameters use generic FSDP2; routed experts use the shared AutoModel MoE
```
jgerh:
```suggestion
parameters use generic FSDP2; routed experts use the shared NeMo AutoModel MoE
```

docs/model-coverage/llm/deepseek-ai/deepseek-v41-flash.mdx
```diff
@@ -0,0 +1,150 @@
...
+implementation with HybridEP dispatch. Engram reuses the Qwen3.8 Flash Next
+row-owner embedding and distributes each table across the same world ranks
+without a full-table all-gather. Its gradients return to the row owners and
+receive the owner divisor once before clipping.
+
+| Recipe | Topology and batch | Optimizer moments and checkpoints |
```
jgerh:
```suggestion
| Recipe | Topology and Batch | Optimizer Moments and Checkpoints |
```

docs/model-coverage/llm/deepseek-ai/deepseek-v41-flash.mdx
```diff
@@ -0,0 +1,150 @@
...
+```bash
+export NUM_OF_HYBRID_EP_RANKS_PER_NVLINK_DOMAIN=4
+```
+
+This gives HybridEP equal four-rank, node-local NVLink groups when the
+allocation spans partial or unequal NVL72 domains. Every training rank must
+inherit it. The EP64 single-domain allocation uses its verified 64-rank domain
+instead of this four-rank override.
+
+Set `DS41_CHECKPOINT` to the absolute path of the shared local HF snapshot for
```
jgerh:
```suggestion
Set `DS41_CHECKPOINT` to the absolute path of the shared local Hugging Face snapshot for
```

docs/model-coverage/llm/deepseek-ai/deepseek-v41-flash.mdx
```diff
@@ -0,0 +1,150 @@
...
+released image encoder, projector, image delimiters, and visual router bias,
+using the existing DeepSeek V4 vision modules.
+
+See the [Slurm launcher guide](/job-launchers/slurm-cluster),
+[installation guide](/get-started/installation), and
+[model README](https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/components/models/deepseek_v41/README.md)
+for the surrounding workflow. Compilation caches and communication prewarming
+must match the chosen backend, topology, and local input shapes.
+
+## Historical validation and current scope
```
jgerh:
```suggestion
## Historical Validation and Current Scope
```

docs/model-coverage/llm/deepseek-ai/deepseek-v41-flash.mdx
```diff
@@ -0,0 +1,150 @@
...
+official `inference/` code. Official TP1 is compared with native EP4 and four
+Engram row owners using HybridEP and TileLang. Rank 0 supplies one valid global
+sample; the other ranks use masked inputs and participate in communication.
+
+The native run streams one production block at a time while carrying its own
+residual, mHC, and CSA2 state. Embedding, final norm, and output head weights are
+replicated. This is complete-backbone streamed-forward evidence; it does not
+establish resident multi-node training or a new parity result for this PR.
+
+| Metric | Original streamed-40 `torch_mm` result |
```
jgerh:
```suggestion
| Metric | Original Streamed-40 `torch_mm` Result |
```

jgerh (commented):
Completed tech pubs review of docs/model-coverage/llm/deepseek-ai/deepseek-v41-flash.mdx and added a few edits for alignment with our style guidelines.

khazic:
## Why router correction biases stay frozen (`gate_bias_update_factor: 0`)

This is a deliberate default for the fine-tuning recipe, and it differs from the DeepSeek-V4 model in this repository, which defaults to `1e-3`. The reasoning:

- V4.1 routes with two selection biases (report Section 2.1.1): text tokens use `e_score_correction_bias` and image-span tokens use `bias_vl`. The report updates the two biases independently from their own modality's expert load after every step.
- The shared `Gate.update_bias` path is not modality-aware. It accumulates one combined expert load over text and image tokens and applies the sign update to `e_score_correction_bias` only; `bias_vl` is never updated. Enabling the update as-is would let image-token load steer the text bias while the image bias stays fixed, which is further from the report than freezing both.
- For fine-tuning, keeping the pretrained routing fixed is the standard choice for the V3 family and has no measurable effect on the short recipe shipped here. Routing weights still come from the unbiased scores, so the model function is unchanged.

Modality-split load accumulation and a `bias_vl` update would be a separate change to the shared gate, and the same gap applies to the existing V4 vision gate, so it is left for a follow-up rather than this PR. The value is a `MoEConfig` field, so a text-only run can still opt in by supplying a custom `moe_config` with `gate_bias_update_factor > 0`.

khazic:
## Follow-up: adopt the V4.1 mHC kernels

The Single-Pass mHC in this PR is numerically aligned with the release, but it is not running on the V4.1 kernels yet:

- `DeepseekV41HyperConnection` reuses the V4 TileKernels Sinkhorn kernel (`dsv4_sinkhorn_normalize`) only for the doubly stochastic `comb` normalization. The projection, RMS scaling, affine `base + mixes * scale`, sigmoid split, stream collapse, and stream expand are separate torch ops. The official `inference/kernel.py` fuses the affine, sigmoid, and Sinkhorn stages into one `hc_split_sinkhorn` kernel, and the PR keeps the fused arithmetic order so the FP32 coefficients match, but it does not call that kernel.
- The report's Mega-mHC (Section 2.4.1) goes further and fuses residual update, input mixing, and coefficient prediction into a single pass over the streams, with the input pre-norm and FP8 conversion included. That kernel is a deployment kernel in DeepGEMM and, per the report, pretraining keeps the multi-kernel implementation, so there is no released forward-plus-backward version to plug into training today.

Proposed follow-up once a differentiable kernel is available:

1. Replace the torch affine/sigmoid/Sinkhorn sequence with `hc_split_sinkhorn` (or a backward-capable port of it) behind the existing `sinkhorn_backend` switch, keeping `torch` as the reference path for parity tests.
2. If a training-capable Mega-mHC lands, fuse `collapse` + `attn_norm`/`ffn_norm` + coefficient prediction the same way, so the streams are read once per sublayer instead of three times.
3. Gate both on the existing HC reference-parity test so the FP32 coefficient rounding and the BF16 stream collapse stay bit-aligned with the release.

Tracking this here so the kernel swap is not mistaken for a correctness gap in the current implementation.

khazic:
Support for the paper's Sinkhorn-balanced optimizer updates will be added in a follow-up PR, covering Engram embedding tables, token embeddings, and the prediction head. The current fine-tuning recipe uses AdamW; the Sinkhorn normalization already used by mHC is a separate operation.

HuiyingLi (approved):

chtruong814 (approved):

khazic:
Embrace it and enjoy it, that is the last time I'm doing support for DeepSeek's models.

PR #3856, opened 2026-09-10
fix(qwen3_5): keep vision tower rotary inv_freq in fp32 across the dt…

haowen-han:
### What does this PR do?                                                                                                                        
                                                                                                                                                  
 Fixes a silent numerical corruption in the Qwen3.5 vision tower: initialize_weights() calls cast_model_to_dtype(self, bf16), whose internal      
 model.to(dtype) rounds the vision rotary's non-persistent fp32 buffer inv_freq down to bf16 (e.g. 0.5994842 -> 0.59765625). Since RoPE phases    
 are computed as pos * inv_freq, the rounding error scales linearly with the position id — vision position encodings are systematically corrupted 
 for large-coordinate patches (measured: vision activations diverge by ~1.5e3 max vs upstream transformers on Qwen3.8-27B under the same bf16     
 checkpoint and inputs).                                                                                                                          
                                                                                                                                                  
 The text backbone already guards against exactly this via Fp32SafeQwen3_5TextRotaryEmbedding; this PR guards the vision tower the same way:      
 after the cast, re-derive inv_freq at full precision from theta/dim stored on the module (pure function of config, no snapshot needed, no-op     
 when the target dtype is fp32).                                                                                                                  
                                                                                                                                                  
 ### Changelog                                                                                                                                    
                                                                                                                                                  
 - nemo_automodel/components/models/qwen3_5/model.py (Qwen3_5ForConditionalGeneration.initialize_weights): after cast_model_to_dtype(...),        
   re-derive model.visual.rotary_pos_emb.inv_freq in fp32 when the cast target is not fp32 (+17 lines, single site)                               
                                                                                                                                                  
 ### Reproduction                                                                                                                                 
                                                                                                                                                  
 <details>                                                                                                                                        
 <summary>Script: before vs after the fix (single GPU, ~2 min)</summary>                                                                          
                                                                                                                                                  
 ```python                                                                                                                                        
 # repro_vision_inv_freq.py -- run in the Automodel venv (transformers 5.15.1)                                                                    
 import torch                                                                                                                                     
                                                                                                                                                  
 from nemo_automodel import NeMoAutoModelForImageTextToText                                                                                       
 from transformers import Qwen3_5ForConditionalGeneration as UpstreamHF, AutoProcessor                                                            
                                                                                                                                                  
 CKPT = "/path/to/Qwen3.8-27B"   # any qwen3_5 VLM checkpoint works                                                                               
 IMG = "/path/to/some_image.jpg"                                                                                                                  
                                                                                                                                                  
 processor = AutoProcessor.from_pretrained(CKPT)                                                                                                  
 msgs = [{"role": "user", "content": [{"type": "image", "image": IMG},                                                                            
                                      {"type": "text", "text": "describe this image"}]}]                                                          
 inputs = processor.apply_chat_template(                                                                                                          
     msgs, tokenize=True, add_generation_prompt=True, return_dict=True, return_tensors="pt")                                                      
                                                                                                                                                  
 pixel, grid = inputs["pixel_values"].bfloat16().cuda(), inputs["image_grid_thw"].cuda()                                                          
                                                                                                                                                  
 backend = BackendConfig(attn="sdpa", linear="torch", rms_norm="torch", rope_fusion=False)                                                        
                                                                                                                                                  
 # Reference: upstream HF (its from_pretrained does NOT round inv_freq)                                                                           
 hf = UpstreamHF.from_pretrained(CKPT, dtype=torch.bfloat16).cuda().eval()                                                                        
 with torch.no_grad():                                                                                                                            
     out = hf.model.visual(pixel, grid)                                                                                                           
     ref = (out[0] if isinstance(out, tuple) else out.last_hidden_state).float().cpu()                                                            
                                                                                                                                                  
 # Automodel wrapper                                                                                                                              
 am = NeMoAutoModelForImageTextToText.from_pretrained(                                                                                            
     CKPT, dtype=torch.bfloat16, attn_implementation="sdpa", backend=backend).cuda().eval()                                                       
                                                                                                                                                  
 print("inv_freq dtype:", am.model.visual.rotary_pos_emb.inv_freq.dtype)                                                                          
 print("inv_freq[:3] :", am.model.visual.rotary_pos_emb.inv_freq[:3].tolist())                                                                    
                                                                                                                                                  
 with torch.no_grad():                                                                                                                            
     out = am.model.visual(pixel, grid)                                                                                                           
     got = (out[0] if isinstance(out, tuple) else out.last_hidden_state).float().cpu()                                                            
                                                                                                                                                  
 diff = (got - ref).abs()                                                                                                                         
 print(f"vision output max diff vs upstream: {diff.max().item():.4e}")                                                                            
 ```                                                                                                                                              
                                                                                                                                                  
 **Before this fix**:                                                                                                                             
 ```                                                                                                                                              
 inv_freq dtype: torch.bfloat16                                                                                                                   
 inv_freq[:3] : [1.0, 0.59765625, 0.359375]      # rounded: should be 0.5994842, 0.3593814                                                        
 vision output max diff vs upstream: ~1.5e+03     # position encodings corrupted                                                                  
 ```                                                                                                                                              
                                                                                                                                                  
 **After this fix**:                                                                                                                              
 ```                                                                                                                                              
 inv_freq dtype: torch.float32                                                                                                                    
 inv_freq[:3] : [1.0, 0.5994842052459717, 0.35938137769699097]                                                                                    
 vision output max diff vs upstream: 0.0          # bit-identical                                                                                 
 ```                                                                                                                                              
                                                                                                                                                  
 </details>                                                                                                                                       
                                                                                                                                                  
 ### Before your PR is "Ready for review"                                                                                                         
                                                                                                                                                  
 Pre checks:                                                                                                                                      
                                                                                                                                                  
 - Make sure you read and followed Contributor guidelines (https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md)                    
 - Did you write any new necessary tests?  (see note below)                                                                                       
 - Did you add or update any necessary documentation?  (comment in code covers the rationale; no user-facing behavior change)                     
                                                                                                                                                  
 ### Additional Information                                                                                                                       
                                                                                                                                                  
 - Related to # (issue) — none filed yet; happy to open one if maintainers prefer issue-first                                                     
 - Verified end-to-end: with the fix, all 333 visual tensors load equal to upstream and the vision tower forward is bit-identical to transformers 
   Qwen3_5ForConditionalGeneration (same checkpoint, bf16, sdpa, same inputs)                                                                     
 - The fix is a no-op when the cast target dtype is fp32                                                                                          
 - Same class of issue as the text backbone's Fp32SafeQwen3_5TextRotaryEmbedding guard — this PR extends that protection to the vision tower

### DIFF
--- a/nemo_automodel/components/models/qwen3_5/model.py
+++ b/nemo_automodel/components/models/qwen3_5/model.py
@@ -870,14 +870,15 @@ class Qwen3_5ForConditionalGeneration(HFCheckpointingMixin, HFQwen3_5ForConditio
     cp_mesh = None
 
     tie_word_embeddings_support: TieSupport = TieSupport.BOTH
-    # ``cast_model_to_dtype`` snapshots/restore matched fp32 buffers around its
+    _tied_weights_keys = {"lm_head.weight": "model.language_model.embed_tokens.weight"}
+
+    # ``cast_model_to_dtype`` snapshots/restores matched fp32 buffers around its
     # bulk cast, and ``cast_frozen_modules_to_compute_dtype`` exempts matched
     # names, so this keeps the vision tower's rotary ``inv_freq`` buffer at its
-    # exact fp32 values under any bf16/fp16 cast (including the frozen-vision
+    # exact fp32 values under shared bf16/fp16 casts (including the frozen-vision
     # recipes' cast). Class-level: ``super().__init__()`` runs ``post_init`` ->
     # ``initialize_weights`` -> the first cast before the constructor body.
     _keep_in_fp32_modules: list[str] = ["rotary_pos_emb"]
-    _tied_weights_keys = {"lm_head.weight": "model.language_model.embed_tokens.weight"}
 
     @dataclass(frozen=True)
     class ModelCapabilities:
--- a/nemo_automodel/components/models/qwen3_5/model.py
+++ b/nemo_automodel/components/models/qwen3_5/model.py
@@ -1509,6 +1509,23 @@ class Qwen3_5ForConditionalGeneration(HFCheckpointingMixin, HFQwen3_5ForConditio
         # Keep the fp32 SSM-gating params fp32 (skip them in the dtype cast); each
         # ``_fp32_params`` holder is sharded as its own fp32 FSDP group.
         cast_model_to_dtype(self, dtype, skip_modules=("_fp32_params",))
+        # The cast above also rounds the vision tower's rotary ``inv_freq`` — a
+        # non-persistent fp32 buffer that feeds every patch's RoPE — down to bf16
+        # (0.5995 -> 0.5977), corrupting all vision position encodings (observed:
+        # vision activations diverge by ~1.5e3 vs upstream HF). The text backbone
+        # guards against this via Fp32SafeQwen3_5TextRotaryEmbedding; re-derive the
+        # vision table at full precision after the cast (theta/dim live on the module,
+        # so this is a pure function of config — no snapshot needed).
+        vision_rotary = getattr(getattr(self.model, "visual", None), "rotary_pos_emb", None)
+        if vision_rotary is not None and hasattr(vision_rotary, "inv_freq") and dtype != torch.float32:
+            inv_freq_fp32 = 1.0 / (
+                vision_rotary.theta
+                ** (
+                    torch.arange(0, vision_rotary.dim, 2, dtype=torch.float32, device=vision_rotary.inv_freq.device)
+                    / vision_rotary.dim
+                )
+            )
+            vision_rotary.inv_freq = inv_freq_fp32
 
 
 ModelClass = Qwen3_5ForCausalLM
--- a/nemo_automodel/components/models/qwen3_5/model.py
+++ b/nemo_automodel/components/models/qwen3_5/model.py
@@ -870,6 +870,13 @@ class Qwen3_5ForConditionalGeneration(HFCheckpointingMixin, HFQwen3_5ForConditio
     cp_mesh = None
 
     tie_word_embeddings_support: TieSupport = TieSupport.BOTH
+    # ``cast_model_to_dtype`` snapshots/restore matched fp32 buffers around its
+    # bulk cast, and ``cast_frozen_modules_to_compute_dtype`` exempts matched
+    # names, so this keeps the vision tower's rotary ``inv_freq`` buffer at its
+    # exact fp32 values under any bf16/fp16 cast (including the frozen-vision
+    # recipes' cast). Class-level: ``super().__init__()`` runs ``post_init`` ->
+    # ``initialize_weights`` -> the first cast before the constructor body.
+    _keep_in_fp32_modules: list[str] = ["rotary_pos_emb"]
     _tied_weights_keys = {"lm_head.weight": "model.language_model.embed_tokens.weight"}
 
     @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


yuhezhang-ai:
Thanks for identifying this precision issue and providing the reproduction!

Could we use the existing `_keep_in_fp32_modules` declaration for `rotary_pos_emb` instead of reconstructing `inv_freq` after the cast? `cast_model_to_dtype()` already snapshots matching buffers before casting and restores their original FP32 values, so this fits our shared precision-handling mechanism.

There is also a correctness reason to use the declaration: after initialization, `apply_model_infrastructure()` calls `cast_frozen_modules_to_compute_dtype()`. With a frozen vision tower and BF16 compute, that helper casts the repaired `inv_freq` back to BF16 unless it matches the FP32 declarations. Several shipped Qwen3.5/3.6 recipes freeze the vision tower, so the one-time repair leaves those configurations exposed to the same issue.

I suggest declaring this on `Qwen3_5ForConditionalGeneration` and removing the added reconstruction block:

```python
_keep_in_fp32_modules: list[str] = ["rotary_pos_emb"]
```

The existing constructor preserves this entry when it appends `_fp32_params`. Please put the declaration at class level: `super().__init__()` already invokes `initialize_weights()`, so adding it afterward would miss the first cast. This is a buffer-preservation requirement and does not need a strict FP32 FSDP group.

Please also add a small CPU regression test that checks exact frequency values and rotary outputs after initialization, repeated BF16/FP16 casts, and the frozen-module cast. Checking only the dtype would miss rounded values subsequently promoted to FP32.

This feedback is based on source inspection of f44ebe62ee2ab70c4a1c2d1c6c9e1b2949a8805a; I have not independently verified the reported full-model GPU parity.

haowen-han:
Thanks for the detailed review — the _keep_in_fp32_modules approach is clearly better, and the frozen-vision gap you pointed out was a real hole in my 
 original fix.                        

 Addressed in 10e6e51bb:              

 - Removed the post-cast reconstruction block and added the class-level declaration instead: _keep_in_fp32_modules: list[str] = ["rotary_pos_emb"].     
   Verified the constructor's _fp32_params append preserves the entry.      
 - Added a CPU regression test (tests/unit_tests/models/qwen3_5/test_vision_rotary_fp32.py) covering exact inv_freq values and rotary outputs after     
   initialization, repeated BF16/FP16 casts, and the frozen-module cast — all with rtol=atol=0 against values recomputed from theta/dim, so a           
   rounded-then-promoted value would fail, not just a dtype check.          

 Also re-verified on the full Qwen3.8-27B checkpoint (bf16, official NeMoAutoModelForImageTextToText.from_pretrained entry point): with frozen vision   
 parameters plus an explicit cast_frozen_modules_to_compute_dtype(model, torch.bfloat16) call, inv_freq stays at its exact fp32 values and the vision   
 tower output is bit-identical to upstream transformers (max diff 0.0). On unpatched main the same script shows inv_freq rounded to bf16 with a max     
 output divergence of ~2.9e2.

yuhezhang-ai (dismissed):
Thanks for addressing the feedback! The class-level FP32 declaration fits the shared precision-handling mechanism and protects the vision rotary buffer through both initialization and the later frozen-module cast. The new regression tests check exact values and rotary outputs.

Validated the five new tests together with the common dtype-cast tests: 50 passed on CPU (PyTorch 2.10.0, Transformers 5.12.1). Formatting and lint checks also passed. Full-checkpoint GPU parity was not independently rerun. No blocking code-review findings remain.

yuhezhang-ai (dismissed):
Thanks for identifying the precision issue and updating the fix and regression tests! This now fits our shared FP32 handling.

My latest commit, 84ba56881, is just a small code-order cleanup: it moves `_keep_in_fp32_modules` and its comment below the tied-weight keys, with minor comment wording fixes. There are no functional changes.

The 50 focused CPU tests and formatting/lint checks passed. Looks good to me!

haowen-han:
DCO has been resolved. I added the required Signed-off-by lines and force-pushed the updated commits. Thanks!

yuhezhang-ai (approved):

yuhezhang-ai:
/ok to test [e082050](https://github.com/NVIDIA-NeMo/Automodel/pull/3856/commits/e0820509fd7564271e29841a51fc2a50d1e0cd78)

haowen-han:
@yuhezhang-ai  @akoumpa Thanks for the review and CI rerun! On the latest commit `e082050`, `cicd-container-build` is cancelled, the downstream tests are skipped, and `Nemo_CICD_Test` is failing.

Could you please check what caused the cancellation and rerun CI when possible? Please let me know if anything is needed from my side. Thanks!

yuhezhang-ai:
Thanks for flagging this! CI timed out waiting for a runner, so the tests never started. I’ve opened #3912 as a CI proxy to rerun the same commit using the maintainer CI runners. No action needed from your side for this timeout.

PR #3859, opened 2026-09-10
test: enforce a 5s per-test timeout

akoumpa:
## What

Adds `pytest-timeout` and applies a 5s default to tests collected under `tests/unit_tests`. A unit test that needs longer than 5s is either doing too much work or is stuck, and today neither shows up as anything more specific than the whole job hitting its 15-minute wall.

Functional and integration tests deliberately keep their workflow-level 20–60 minute budgets; the 5s default does not leak into `tests/functional_tests`.

54 unit-test modules exceed the budget and carry an explicit module-level timeout with a comment naming the reason: 49 use 60s, four cold-compile modules use 120s, and the MAGI parity module uses 300s. The debt is enumerated and greppable rather than implicit, while every **new** unit test is held to 5s.

The exemption list isn't guesswork. The original green run's `--durations 32` bottoms out at **4.87s**, so it enumerates every CPU phase at or above the budget; all 32 land in an exempted module (verified programmatically). Exact-head GPU CI then identified 14 additional cold-compilation modules, and the same-base green `main` run measured those paths at roughly 9–205s.

## On "use a smaller model / fewer steps"

I went looking for that and it isn't where the time is. The models in the slow tests are **already toy-sized** — `hidden_size=2`, `num_hidden_layers=1`, `vocab_size=32`, and similar. There are no training loops with reducible step counts. Breaking down the 335s in the original 32 slowest tests by actual cause:

| cause | time | share | |
|---|---:|---:|---|
| `mp.spawn` | 237.6 s | 71% | 27 files, 53 call sites, 24 gloo-only |
| `sys.executable` subprocess | 36.0 s | 11% | 8 files |
| git subprocesses | 24.2 s | 7% | throwaway repositories |
| everything else | 37.3 s | 11% | `torch.compile`, repo-wide file scans |

**82% is a fresh interpreter re-importing torch in a child process** — measured at ~9s for the parent's own startup, and leaving a ~4.9s floor under even a trivial two-rank test (`test_comm_groups_prewarm_warms_groups_on_two_ranks`, which does almost nothing, costs 4.87s). Shrinking a model that is already `hidden_size=2` cannot move that.

So this PR deliberately does **not** include cosmetic model-shrinking. The change that actually cuts the ~270s is reusing worker processes instead of spawning fresh ones (or a fork context — the CPU job never initializes CUDA), and that deserves its own PR where CI can validate it in isolation. The timeout added here is what makes that attempt safe: a fork-related hang becomes a bounded test failure naming the test, instead of a 15-minute job timeout followed by two full retries.

## Notes

- **Scope.** The default marker is added during unit-test collection only when a test has no explicit timeout marker. Functional tests receive no per-test default.
- **Margin.** Everything in the original CPU measurement that was not exempted ran under 4.87s. Exact-head GPU failures were compared with the same-base green `main` run before assigning 60–300s cold-compilation budgets.
- **`uv.lock` is edited surgically**, not regenerated: CI pins uv 0.11.26 and regenerating under a newer uv rewrites unrelated marker lines.
- **Lint.** Verified with ruff 0.12.12 (the version `uv run --active ruff` resolves in CI, per `uv.lock`), repo-wide: `check` passes, `format --check` reports 804 files already formatted. Note `.pre-commit-config.yaml` pins ruff 0.16.4 while the `linting` group pins `~=0.12.0`; the hook reformatted three files it touched, and 0.12.12 accepts the result. That version split is worth reconciling separately.

## Related

Stacks conceptually on #3858, which removes the redundant `-rA` report (~128s/attempt). Neither fixes the growth trend on its own; the spawn cost is the durable fix.


tests/unit_tests/checkpoint/test_rank_local_rng_checkpoint.py
```diff
@@ -18,13 +18,18 @@
...
 import torch
 import torch.distributed as dist
 import torch.multiprocessing as mp
 
 from nemo_automodel.components.checkpoint.checkpointing import CheckpointingConfig
 from nemo_automodel.components.training.rng import StatefulRNG, init_all_rng
 
+# Over the default 5s budget on purpose: this module spawns worker processes; every child re-imports torch from scratch.
+# Shrink the work or the process count before raising this further.
+pytestmark = pytest.mark.timeout(60)
```
yuhezhang-ai, line 31: [P2] Ensure a timeout also stops spawned workers

On Linux, this marker uses pytest-timeout's default `signal` method. If the `mp.spawn(..., join=True)` below hangs, the timeout raises in the parent, but does not terminate its non-daemon workers. Python then waits for those workers during interpreter shutdown, so CI can still hit the workflow timeout and retry—the failure mode this PR is intended to bound.

I reproduced the lifecycle issue with a minimal stdlib `multiprocessing` spawn-and-join test: a 0.5s marker reported the test failure, but pytest remained alive and required an external kill at 3s. Please ensure timeout handling terminates the workers, or use a process-terminating watchdog for these tests, and add a hanging-worker regression test that checks the pytest process actually exits.

tests/unit_tests/conftest.py
```diff
@@ -48,6 +48,14 @@
     sys.path.insert(0, str(_PROJECT_ROOT))
 
 
+def pytest_collection_modifyitems(items):
+    """Give unit tests a 5s default while preserving explicit overrides."""
+    unit_tests_root = Path(__file__).parent
+    for item in items:
+        if unit_tests_root in item.path.parents and item.get_closest_marker("timeout") is None:
+            item.add_marker(pytest.mark.timeout(5))
```
yuhezhang-ai, line 66: [P2] Preserve explicit global timeout overrides

Adding a per-test marker gives this fallback higher priority than `--timeout` and `PYTEST_TIMEOUT`. Consequently, even otherwise unmarked unit tests ignore an explicit `--timeout=30` or `--timeout=0`. Using the extracted collection hook and pytest-timeout 2.4.0, I confirmed that a 5.5s test still fails at 5s when invoked with `--timeout=0`.

This prevents developers from adjusting the default for slower machines or temporarily disabling it during investigation through the plugin's normal options. Please apply the 5s fallback only when no global timeout was explicitly selected, while preserving genuine per-test/module markers, and cover the CLI/environment override behavior.

tests/unit_tests/checkpoint/test_rank_local_rng_checkpoint.py
akoumpa: Thanks @yuhezhang-ai resolved both; ptal

yuhezhang-ai (dismissed):

yuhezhang-ai (dismissed):

yuhezhang-ai (approved):

PR #3746, opened 2026-08-29
fix: preserve selected routing weights in FP32 in deepseek

edjson:
# What does this PR do ?

Add a one line overview of what this PR aims to accomplish.

# Changelog

- `nemo_automodel/components/models/deepseek_v3/model.py`: `router_weights_fp32=True` in `DeepseekV3Model.moe_defaults`, above the `moe_overrides` update so a caller override wins. `DeepseekV3ForCausalLM.__init__` installs the FP32 `gate_precision` default on a `replace` copy instead of mutating the caller's `BackendConfig`.
- `nemo_automodel/components/models/deepseek_v32/model.py`: same two defaults on V3.2's own path (`DeepseekV32Model.moe_defaults`, `DeepseekV32ForCausalLM.__init__`); both call `nn.Module.__init__` directly and skip the V3 code.
- `tests/unit_tests/models/test_deepseek_gate_precision_policy.py` (new; mirrors `test_glm_gate_precision_policy.py`): parametrized over V3, V3.2, Kimi K2 config on the V3 path, and a Moonlight-16B-A3B-shaped V3 config. Contents mapped to the requested evidence below.
- `tests/unit_tests/models/deepseek_v3/test_dsv3_model.py`: V3 gate-precision cases folded into the shared file; one test kept for "default lands on a copy". String-form `gate_precision` normalization is covered in `tests/unit_tests/moe/test_backend_config.py`.
- `tests/unit_tests/models/deepseek_v32/test_dsv32_model.py`: `test_from_config_with_explicit_backend` now expects `replace(backend, gate_precision=torch.float32)` and asserts the caller's object is untouched.

# Before your PR is "Ready for review"

**Pre checks**:

- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?

# Additional Information
 scope:
- [x] V3 - Out stage 
- [x] V3.2 -  Cover Proj/Score as well as Out for V3.2, while preserving an explicit caller override.
- [x] Kimi K2 - verify that Kimi K2 inherits the corrected DeepSeek V3 policy; it should not need a separate implementation change.

validation:
- [x] a deterministic tiny router/component comparison with grouped routing (n_group > 1);
- [x] reduced-layer V3 and V3.2 construction tests covering Param / Proj / Score / Out;
- [x] default and explicit-override tests;
[...]

PR #3861, opened 2026-09-11
fix(ling): fp32 default router projection and selected weights

edjson:
# What does this PR do ?

Defaults the Ling / Bailing MoE V2 router projection to FP32 and preserves the
selected routing weights in FP32, matching the checkpoint-owned reference.

# Changelog

- `ling_v2/model.py`: `moe_defaults` sets `router_weights_fp32=True`, above the
  `moe_overrides` update so a caller override still wins.
- `ling_v2/model.py`: `BailingMoeV2ForCausalLM.__init__` defaults
  `backend.gate_precision` to `torch.float32` when unset, via
  `dataclasses.replace` so a shared `BackendConfig` is not mutated.

Proj and Out moved; Param and Score already matched on main.

**Pre checks**:

- [ ] Did you add or update any necessary documentation? — n/a

## Audit checklist (#3650)

- [x] **Identify an immutable reference implementation and, when relevant, checkpoint
  tensor dtype evidence.**
  `BailingMoeV2Gate` transcribed from the checkpoint-owned `modeling_bailing_moe_v2.py`
  at revision `920c3fd9916e3d5e543fc4f609e827cad8a32983`. Stored dtypes read from the
  safetensors headers — BF16 gate weight, F32 `expert_bias`, across all 19 MoE layers —
  and re-derived on every smoke run.

- [x] **State every reference load convention explicitly; do not treat a framework
  loaded with a model-wide dtype override as a checkpoint-storage oracle.**
  The reference's casts are documented in the test module. Storage comes from the file
  itself: header bytes, no `from_pretrained`, no model-wide dtype override in the path.

- [x] **Where a model-specific production implementation exists, cross-check a pinned
  serving runtime and record its actual loaded parameter and routing arithmetic dtypes.**
  vLLM `v0.29.0`, which registers `BailingMoeV2ForCausalLM` to `bailing_moe.py`: gate
  weight fp32, expert bias fp32, Proj fp32, Out fp32 — derived from the same
  `router_dtype` field. Score is bf16 on the Python fallback and unverified on the fused
  path, since `torch.ops._moe_C.grouped_topk` does not ship in that repo at that tag.

- [x] **Match Param / Proj / Score / Out independently; do not infer full correctness
  from FP32 logits alone.**
  Asserted separately at the Gate, at the expert boundary, and on the constructed model.
  The checkpoint run's logits check is a liveness check; the stage checks are separate
  and gate its `PASS`.

- [x] **Keep model-specific defaults in the owning model package. Preserve explicit
  caller overrides where supported.**
  Both defaults in `ling_v2/model.py`. `gate_precision` via `dataclasses.replace` so a
  shared `BackendConfig` is untouched; `router_weights_fp32` above the `moe_overrides`
  update. Both override paths tested.

- [x] **Add focused tests for stored gate dtype and exact precision-critical values,
  projection/scoring arithmetic dtype, selected-weight dtype, and default/explicit-override
  behavior.**
  All five clauses covered. The correction bias is pinned by **value**, not only dtype,
  across a cast and a load; demoting it to bf16 reroutes 27 of 512 tokens.

- [ ] **Run a tiny numerical router/component comparison against the reference,
  including input/gate gradients, followed by the relevant end-to-end checkpoint/parity
  test. Finite training loss or first-few-layer agreement alone does not establish
  full-depth logits parity.**
  Router comparison bitwise at three shapes; gate-weight and router-input gradients
  bit-identical, with negative controls on both. **Full-depth HF-vs-AutoModel logits
  parity is outstanding** — the checkpoint run establishes a clean load and finite
  logits, which per this criterion is not parity.

- [x] **Update this checklist and link the fixing PR plus exact-head validation
  evidence.**
  This section; validation output above, produced at head `d2d8668e`.
# Additional Information

- Related to #3650
- Follows #3746 (DeepSeek V3/V3.2, same two-stage change)


yuhezhang-ai:
/ok to test [f471c0a](https://github.com/NVIDIA-NeMo/Automodel/pull/3861/commits/f471c0a91d7e8020fb24aabf34e2fc083b00914f)

yuhezhang-ai (approved):
The router precision changes and regression tests look good.

I added Ling-mini checkpoint-robustness coverage and fixed the legacy HF compatibility/loading issues it exposed. Follow-up diagnostics support relaxing only the HF source-load and export/reload comparisons; full 2K checks and native reload/resume policies remain unchanged.

[Scoped CI](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/438997837) passed all six phases at `f471c0a91`, with exact native reload and resume in this run.

Issue #3869, opened 2026-09-11
InfoNCE d2q/symmetric with use_in_batch_negatives=false trains on a loss of exactly 0


kabirvashisht4-glitch:
## Describe the bug

`infonce_loss` and `infonce_distill_loss` accept `direction="d2q"` (and
`"symmetric"`) together with `use_in_batch_negatives=False`. In that
configuration the document-to-query direction has **no candidates to score
against**, so its cross-entropy is exactly `0.0` with a zero gradient. Nothing
raises, nothing is logged — the run just trains on nothing.

Hard negatives are documents mined *per query*, so they are candidates for the
query-to-document direction only. `_one_direction` correctly passes `None` for
the reverse direction:

https://github.com/NVIDIA-NeMo/Automodel/blob/main/nemo_automodel/components/loss/infonce.py

```python
if direction == "q2d":
    return _one_direction(q, d, n)
if direction == "d2q":
    return _one_direction(d, q, None)          # <- negatives intentionally dropped
return 0.5 * (_one_direction(q, d, n) + _one_direction(d, q, None))
```

With `use_in_batch_negatives=False` that branch builds a single-column logit
tensor and a target of all zeros:

```python
pos = (query * doc).sum(dim=-1, keepdim=True)   # [B, 1]
logits = pos
target = torch.zeros(batch, device=device, dtype=torch.long)
...
return F.cross_entropy(logits, target)          # one class -> log(1) == 0
```

The existing guard only catches the case where *no* negatives were passed at
all, so it never fires here:

```python
if not use_in_batch_negatives and not has_hard_negs:
    raise ValueError("infonce_loss: no negatives provided")
```

`infonce_distill_loss` has the same shape of problem: the single-candidate
softmax makes both the student and the (detached) teacher distributions
degenerate, so KL/CE/MSE over them is identically zero.

## Steps/Code to reproduce bug

```python
import torch
from nemo_automodel.components.loss.infonce import infonce_loss

torch.manual_seed(0)
q = torch.randn(4, 8, requires_grad=True)
d = torch.randn(4, 8, requires_grad=True)
n = torch.randn(4, 3, 8, requires_grad=True)

for direction in ("q2d", "d2q", "symmetric"):
    for in_batch in (True, False):
        for t in (q, d, n):
            t.grad = None
        loss = infonce_loss(
            q, d, hard_negatives=n, direction=direction, use_in_batch_negatives=in_batch
        )
        loss.backward()
[...]

PR #3870, opened 2026-09-11
fix(loss): reject InfoNCE directions without candidates to score

kabirvashisht4-glitch:
# What does this PR do ?

Rejects the InfoNCE configurations whose loss is mathematically forced to `0.0`,
instead of training on them silently.

`infonce_loss` / `infonce_distill_loss` accept `direction="d2q"` or
`"symmetric"` together with `use_in_batch_negatives=False`. Hard negatives are
documents mined *per query*, so they are candidates for the query-to-document
direction only, and `_one_direction` correctly passes `None` for the reverse
direction. With in-batch negatives also switched off, that direction is left
with a single candidate column:

```python
pos = (query * doc).sum(dim=-1, keepdim=True)   # [B, 1]
logits = pos
target = torch.zeros(batch, device=device, dtype=torch.long)
...
return F.cross_entropy(logits, target)          # one class -> log(1) == 0
```

The existing guard only caught the case where *no* negatives were passed at all,
so it never fired:

```python
if not use_in_batch_negatives and not has_hard_negs:
    raise ValueError("infonce_loss: no negatives provided")
```

Measured on `main` @ `b1f4031d`, `B=4`, `K=3`, CPU:

| direction | `use_in_batch_negatives` | loss | sum abs grad (queries) | sum abs grad (hard negs) |
|---|---|---:|---:|---:|
| `q2d` | true | 8.986291 | 25.4025 | 4.2906 |
| `q2d` | false | 5.643632 | 17.9486 | 20.5227 |
| `d2q` | true | 5.800508 | 20.0058 | — |
| `d2q` | false | **0.000000** | **0.0000** | — |
| `symmetric` | true | 7.393400 | 20.2241 | 2.1453 |
| `symmetric` | false | 2.821816 | 8.9743 | 10.2613 |

`2.821816 == 5.643632 / 2` exactly — the `d2q` half of `"symmetric"` contributes
literally nothing, so a config asking for a symmetric objective silently gets a
halved one-directional one. `infonce_distill_loss` degenerates identically: the
single-candidate softmax makes the student and the detached teacher
distributions both one-hot, so KL/CE/MSE over them is `0.0` with zero student
gradients.

This is reachable from YAML: `RetrievalDistillBiEncoderRecipe` builds both
modules from `self.cfg.get("infonce_loss")` / `self.cfg.get("infonce_distill_loss")`,
and `direction` and `use_in_batch_negatives` are plain keys on that block (see
`examples/retrieval/distillation/nemotron3_embed_1b_distill.yaml`). The shipped
example is safe; flipping those two keys is not.

## Why reject rather than feed the negatives into `d2q`

Reusing `hard_negatives` for the reverse direction would be numerically wrong,
not just unhelpful: they are *document* embeddings, and the document-to-query
direction needs query-side candidates. The `None` in `_one_direction(d, q, None)`
is correct; the missing piece is that the caller was never told the resulting
candidate set is empty. So the guard is extended instead, in the same place and
the same style as the existing one, and the failure is raised before any work is
done. This matches the precedent already in the recipe:

```python
# Guard against a configured intermediate loss whose layer_pairs are empty
# while the weight is active: hooks would fire but the loss would silently
# short-circuit to zero (contributing no gradient).
```

No behavior changes for any configuration that was producing a real loss.

# Changelog

- `infonce_loss` and `infonce_distill_loss` now raise `ValueError` when
  `use_in_batch_negatives=False` is combined with `direction="d2q"` or
  `"symmetric"`, explaining that the document-to-query direction can only draw
  candidates from the batch and naming both ways out.
- Both functions gained full Google-style docstrings recording the tensor
  layouts (`[batch, hidden]`, `[batch, negatives, hidden]`, `[batch, negatives]`
  masks), the per-query/document-side semantics of `hard_negatives` that make
  the direction restriction necessary, and the `Raises` contract.
- `tests/unit_tests/loss/test_infonce.py`: added the rejection tests for both
  functions and for the `InfoNCELoss` module (the config-driven entry point),
  plus parametrized guards asserting every *accepted* `(direction,
  use_in_batch_negatives)` combination yields a strictly positive loss and
  non-zero input gradients — the invariant the bug violated.

**Pre checks**:

The only documentation of this contract was the functions' one-line docstrings,
which are expanded here. No user-facing docs or example configs reference the
rejected combination, so nothing else needed updating.

Verified locally: `pytest tests/unit_tests/loss/test_infonce.py` -> 62 passed.
The 5 new rejection tests fail on `main` (`DID NOT RAISE ValueError`) and pass
here. `ruff format --check` / `ruff check` clean. CPU only, no GPU needed.

# Additional Information

- Related to #3869

### DIFF
--- a/nemo_automodel/components/loss/infonce.py
+++ b/nemo_automodel/components/loss/infonce.py
@@ -22,7 +22,34 @@ def infonce_loss(
     use_in_batch_negatives: bool = True,
     normalize: bool = True,
 ) -> torch.Tensor:
-    """InfoNCE contrastive loss with optional hard negatives."""
+    """InfoNCE contrastive loss with optional hard negatives.
+
+    Args:
+        queries: Tensor of shape [batch, hidden] -- pooled query embeddings.
+        documents: Tensor of shape [batch, hidden] -- pooled embeddings of the
+            positive document for each query, row-aligned with ``queries``.
+        hard_negatives: Optional Tensor of shape [batch, negatives, hidden] --
+            mined negative *documents* for each query. Rows are per-query, so
+            these are candidates for the ``"q2d"`` direction only.
+        hard_negatives_mask: Optional Tensor of shape [batch, negatives] --
+            1 for a real negative, 0 for padding in a ragged batch. Padded
+            columns are scored as ``-inf`` and contribute nothing.
+        temperature: Logit scale divisor; a 0-dim Tensor when it is learned.
+        direction: ``"q2d"``, ``"d2q"``, or ``"symmetric"``.
+        use_in_batch_negatives: Score every query against every document in the
+            batch instead of only its own positive.
+        normalize: L2-normalize the embeddings before scoring.
+
+    Returns:
+        Scalar Tensor: the mean cross-entropy over the batch.
+
+    Raises:
+        ValueError: If the embedding shapes disagree with the layouts above, if
+            ``direction`` is unknown, or if the requested directions have no
+            candidates to score against -- ``use_in_batch_negatives=False``
+            needs ``hard_negatives`` *and* ``direction="q2d"``, because the
+            document-to-query direction can only draw candidates from the batch.
+    """
     if queries.dim() != 2 or documents.dim() != 2:
         raise ValueError(
             f"infonce_loss: queries and documents must be 2-D [B, D]; got queries={tuple(queries.shape)}, "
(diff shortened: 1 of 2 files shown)


yuhezhang-ai (approved):
This is useful housekeeping for an uncommon configuration. The supplied example is unaffected, and no affected training run is reported here. Rejecting d2q/symmetric with in-batch negatives disabled makes the configuration contract explicit and prevents silently dropping the reverse-direction loss.

Validated the focused loss suite locally on CPU: 62 tests passed.

yuhezhang-ai:
/ok to test [0583390](https://github.com/NVIDIA-NeMo/Automodel/pull/3870/commits/0583390d86ef7624f5e4d469f37631a791a02428)

PR #3878, opened 2026-09-13
perf(training): use Transformer Engine for local L2 gradient norms

HuiyingLi:
## Enable TE gradient norms

The knob is **`clip_grad_norm.use_te`**, default **`false`** in the validated opt-in implementation. Enable it in the native `llm_finetune` recipe YAML:

```yaml
clip_grad_norm:
  max_norm: 1.0
  use_te: true
```

Set `use_te: false` or omit it to use native PyTorch norm computation. `use_te: true` enables TE local L2 reduction for eligible gradients when TE is installed; unsupported cases retain the documented fallback. Distributed norm combination and in-place clipping remain in AutoModel. The three local DS41 CP recipes explicitly enable this flag.

**Published implementation:** the PR head is `f301f5ee194b1bb79983e3063a01d5d689b54e82`, which exposes the explicit `clip_grad_norm.use_te` opt-in (default `false`). The isolated microbenchmarks below were measured at `e0b5657c`; DS41 results use the frozen integration commit `30f41b61`, which includes the published opt-in change.

DS41 CP integration is published in #3879, with report-cleanup commit `15d21d99` on its branch. Production code, recipes and formal regression tests match experiment commit `30f41b61`; subsequent changes added DCO trailers and removed local validation-report files. Original experiment artifacts remain archived on the cluster.

**TE vs Triton:** on the identical contiguous DS41-shaped BF16 gradient, local norm + clipping took **6.37 ms with TE**, **68.10 ms with #3815**, and **29.59 ms with #3685**. For many smaller tensors, TE and #3815 were close (**4.31 vs 4.14 ms**). The full input definitions, pinned revisions, precision checks and measurement boundaries are in the comparison below; these are local clipping results, not end-to-end TE-vs-Triton throughput.

# What does this PR do ?

Use Transformer Engine's existing multi-tensor L2 norm for eligible local gradients to reduce gradient-clipping latency and temporary memory. This PR compares the TE approach with main and the implementations in #3815 and #3685.

For a contiguous BF16 gradient with shape `(6000261, 256)`, the measured local norm + clipping time drops from **20.29 ms to 6.37 ms**, and peak extra allocated memory drops from **17.17 GiB to 4 KiB** on one GB200. These are isolated clipping measurements. The DS41 integration results below additionally cover full-model training; OFF OOM prevents a steady-state OFF/ON throughput comparison.

# Changelog

- `nemo_automodel/components/training/utils.py`: optionally import TE optimizers and group contiguous CUDA FP16/BF16/FP32 local gradients by device/dtype for `multi_tensor_l2norm`. Pass gradient storage directly to TE without concatenation or a full-gradient copy.
- Retain scalar MAX/SUM reduction across DTensor shards, replica counting, Partial materialization, pipeline norm combination, and PyTorch in-place clipping. TE computes only the local L2 norm.
- Use native PyTorch fallback when TE is unavailable or the input layout/dtype is unsupported. Retry tiny/non-finite TE results in FP64 to handle FP32 squared-sum range limits; retain scaled reduction for FP64 extremes.
- `tests/functional_tests/training/test_te_grad_norm.py`: add numerical and real two-GPU tests for local gradients, DTensor placement behavior, and FSDP2 backward/clipping/SGD updates.

## Comparison on identical inputs

Hardware/software: one NVIDIA GB200, PyTorch `2.13.0a0+8145d630e8.nv26.06`, Transformer Engine `2.14.1+366798ef`.

Each implementation receives the same gradients, seed `20260913`, and `max_norm=1.0`. After one initial call and two warmups, measure 20 repetitions. Times below are median synchronized wall time for the **complete local norm + in-place clipping function**, including the TE range guard. Gradient resets, FP64 reference construction, distributed communication, and optimizer updates are outside timing.

Extra memory is the peak PyTorch-allocated GPU memory above already-live tensors, excluding parameter/gradient storage and allocations outside the PyTorch allocator.

- **Many tensors:** 500 matrices, 225,181,696 elements total. Matrix `i` has shape `(512 + (i % 4) * 128, 512 + (i % 3) * 128)`; every fifth tensor is FP32 and the rest are BF16.
- **Large tensor:** one contiguous BF16 matrix of shape `(6000261, 256)`, 1,536,066,816 elements. Inputs are generated with `torch.empty(...).normal_()`.

| Implementation | Many tensors | Large tensor | Large tensor extra memory | Large tensor norm relative error vs FP64 |
| --- | ---: | ---: | ---: | ---: |
| main | 36.94 ms | 20.29 ms | 17.17 GiB | 5.62e-6 |
| TE (this PR) | 4.31 ms | 6.37 ms | 4 KiB | 4.29e-8 |
| #3815 | 4.14 ms | 68.10 ms | 5 KiB | 1.86e-16 |
| #3685 | 7.85 ms | 29.59 ms | 3.58 MiB | 0 |

TE and #3815 have similar latency in the many-tensor case. The workload dependence is substantial; these measurements do not establish a general winner across models or hardware.

TE norm relative error is `6.05e-9` for the many-tensor case and `4.29e-8` for the large-tensor case. For these measured TE inputs, clipped gradients match the FP64-coefficient reference exactly in the destination dtype, and all 20 repeated norms are identical. Reported norms are not bitwise identical to main's BF16 scaled-square calculation. The two Triton approaches use FP64 reduction and are closer to the FP64 reference in these cases.

### Non-contiguous input

The same large matrix, transposed without changing its values:

| Implementation | Norm + clipping | Extra memory |
| --- | ---: | ---: |
| main | 47.78 ms | 17.17 GiB |
| TE integration (PyTorch fallback) | 11.73 ms | 11.45 GiB |
| #3815 | 123.21 ms | 2.86 GiB |
| #3685 | 96.64 ms | 2.88 GiB |

**The TE integration's fallback can allocate a full widened gradient. The KiB-scale memory result applies to the eligible contiguous path.** The range guard also synchronizes a CUDA scalar per TE device/dtype group; this is documented and included in the timing.

### Pinned source revisions

| Implementation | Revision |
| --- | --- |
| main | `f7ccd6f7902634af34c2f31b3294ac250dc97670` |
| TE | `e0b5657c1e6efe5226e63471590586e1e7493b08` |
| #3815 | `13b7b4c3cc7044f709c43e206482f9249bd8ccea` |
| #3685 | `7bbf0cd532fab33d1ff3c241709fbf5ac929f725` |

Comparison functions and their helpers were exported from these revisions without modification. The other PRs were compared on local clipping inputs; their distributed paths were not validated in this experiment.

## Validation

**Import compatibility follow-up (`f301f5ee`): 167 tests passed** across training utilities, LLM fine-tuning recipe tests and real two-GPU TE/DTensor clipping tests. Ruff and `git diff --check` passed. The new regression reproduces the pre-fix import failure from a broken optional TE binary and verifies module import, CPU clipping and an SGD update after the fix. TE is now loaded only for eligible CUDA gradients on the opt-in path, through the existing `safe_import_te()` compatibility guard; a missing or incompatible TE installation retains the PyTorch fallback. The `[cuda]`/`[all]` installation checks exposed TE 2.15.0 loading against cuBLAS 13.1.0.3 with a missing `cublasLtGroupedMatrixLayoutInit_internal` symbol. Fresh CI on `f301f5ee` passed both previously failing installation jobs: [`[cuda]`](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/34809421338/job/103868445502) and [`[all]`](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/34809421338/job/103868445564), including their import checks. The remaining installation and main CPU/GPU checks are still running.

**55 tests passed:** 37 existing unit tests plus 18 new CUDA cases. Ruff check/format and `git diff --check` passed.

- Compare norm, clipped gradients, and an SGD update against an independent FP64 reference across FP16/BF16/FP32/FP64/complex gradients, contiguous/transposed/strided/offset layouts, and forced no-TE fallback.
- Cover empty/no gradients, zero, NaN/Inf, huge and tiny finite values, and mixed subnormal squares.
- Exercise real two-GPU NCCL with Shard(0), Shard(1), Replicate, Partial, an empty shard, unequal local tensor counts and TE eligibility, pipeline scalar reduction, and a small FSDP2 model's actual backward, clipping, and SGD update.
- A separate real dense `F.embedding` backward with the large BF16 weight shape produced a contiguous gradient with stride `(256, 1)`. The norm was exactly 1024, clipping matched the expected result, gradient storage was preserved, and extra allocation was 4 KiB. This does not exercise full-model training or distributed owner routing.

```bash
python -m pytest tests/unit_tests/training/test_train_utils.py tests/functional_tests/training/test_te_grad_norm.py -q
```

The tests above describe the published microbenchmark revision. Full-model DS41 integration results and their separate source revisions are recorded below; a steady-state OFF/ON throughput comparison and full 100-step numerical parity remain unavailable.

## Full-model DSV4.1-Flash SFT: TE OFF / ON (2026-09-13)

**TE ON completed all 100 updates without OOM; TE OFF completed 2 updates, then OOMed during gradient-norm calculation on the third attempted update. Full 100-step OFF/ON parity was not established.**

Tested a frozen local DS41 integration checkout, `30f41b61d8e93ff1a60a76e48d8bd7fae6527d7b`, containing the TE opt-in change `8a0b15bf60e92ff90ddaffd7f6e692cf93a6b221`. The opt-in commit `8a0b15bf` is included in the published PR head. The frozen DS41 integration commit `30f41b61` is separate from the PR; the isolated microbenchmarks above remain pinned to `e0b5657c`.

Both runs used the full pretrained 40-layer DSV4.1-Flash model, including both Engram layers; packed Tulu3 with global sequence length 4096; CP2 / EP64 / world64 on 16 nodes x 4 GPUs; GBS64 / LBS1; seed1234; per-block activation checkpointing and FSDP synchronization every microbatch. Both used the same frozen source; only `clip_grad_norm.use_te=false/true` differed in the training configuration. Each had a separate 30-minute allocation, on different physical node sets, with no extension, retry or continuation.

| Variant | Job | Completed updates | Result / allocation elapsed | Native W&B |
| --- | --- | ---: | --- | --- |
| TE OFF | 7128311 | 2 / 100 | Grad-norm OOM; 6m38s | [OFF: 2 exact rows, failed](https://wandb.ai/Nemo-automodel/automodel-ds41-gradnorm-packed4k-20260913/runs/ds41te4k07128311) |
| TE ON | 7128312 | 100 / 100 | COMPLETED, exit0; 19m17s | [ON: 100 exact rows, finished](https://wandb.ai/Nemo-automodel/automodel-ds41-gradnorm-packed4k-20260913/runs/ds41te4k17128312) |

ON's final training loss was **0.571274**, final grad norm **0.162518**, and validation loss **0.546551** (optimizer index99). All 100 training losses/norms were finite; all 16 nodes recorded completion. The unchanged native-history verifier checked the actual row counts (OFF2, ON100), including ON validation at indices49/99: every original logger field matched exactly, with no missing rows or backfilled values.

### Memory and throughput

OFF failed at `g.square().sum(dtype=torch.float64)` in the native PyTorch norm reduction. The first reporting GPU (rank47) had **184.31 GiB** capacity and **10.81 GiB** free when an additional **11.45 GiB** allocation failed; total process GPU usage, including non-PyTorch allocations, was **173.48 GiB**. This identifies the allocation failure; it does not identify the particular parameter being reduced.

The native `mem` metric uses `torch.cuda.max_memory_allocated()`, read after clipping and the optimizer step and reset after main-rank training logging. It measures the **main rank's PyTorch allocator peak over that logging interval**, including successful temporary allocations; it is not instantaneous live memory or the maximum total device usage across ranks. The maximum reported value was **132.07866 GiB OFF** versus **132.08661 GiB ON**. OFF's failed third update emitted no completed training row. These values do not isolate the gradient-norm workspace, and the incomplete OFF run does not establish a full-training peak-memory delta.

ON averaged **29,947.58 logged tokens/s** across all 64 GPUs (467.93/GPU) after excluding the first 10 updates; mean timestamp interval was **7.4466 s/update**. OFF's two updates are insufficient for a steady-state throughput comparison, so no end-to-end speedup is claimed.

### Why the DS41 workload benefits

The native PyTorch norm path processes gradients one tensor at a time, so the largest **local gradient tensor** strongly affects temporary memory. DS41's Engram embedding shard is much larger than any single gradient in the measured Qwen3-MoE EP4 workload:

| Workload | Largest relevant local gradient | Elements | Same-shape FP64 tensor size |
| --- | --- | ---: | ---: |
| Qwen3-30B-A3B, EP4 | `[32, 2048, 1536]`, FP32 | 100,663,296 | 0.75 GiB |
| DS41 Engram, tested CP2/EP64 layout | `[6000261, 256]`, BF16 | 1,536,066,816 | 11.45 GiB |

The FP64 column is a tensor-size calculation, not a measured per-module peak. For the actual DS41-shaped embedding-backward gradient, the isolated measurement above found **17.17 GiB** extra allocation with native PyTorch versus **4 KiB** with TE. TE reduces the eligible gradient directly in the kernel, avoiding full-gradient temporary copies for these intermediate operations.

Qwen3-MoE's main/TE runs both reported about **125.55 GiB**: reducing norm workspace need not lower the maximum allocation reached elsewhere in the training interval. DS41's much larger single shard creates a larger allocation request near the device limit; OFF's failed request was 11.45 GiB with only 10.81 GiB free, while ON completed all 100 updates. The traceback identifies the norm operation, but not which parameter was being reduced. This is evidence for this tested layout, not a universal memory reduction for every model.

### Numerical comparison: common first two updates only

Optimizer indices are zero-based.

| Index | Loss OFF | Loss ON | Absolute loss difference | Grad norm OFF | Grad norm ON |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 0 | 0.843463540 | 0.843463540 | 0 | 2.953675749 | 2.954571863 |
| 1 | 0.740743339 | 0.740498304 | 0.000245035 | 2.861239832 | 2.861029354 |

Initial parameter samples matched on all 64 ranks. The common three input audits matched; OFF's third audited input did not complete an update. LR and supervised-token counts matched for both completed updates.

For the **two-update prefix**, maximum / mean loss differences were **0.000245035 / 0.000122517**, within the predeclared 0.01 / 0.002 thresholds. Maximum relative grad-norm difference was **0.0303389%**, passing the separate 5% diagnostic for this prefix.

The comparison was still run with **100 expected updates** and returned failure: completion, step-count, full-input-audit and validation gates remain failed because OFF ended early and has no validation rows. These results do not establish 100-step loss, gradient-norm or validation parity.

**Pre checks**:

- [x] Read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md).
- [x] Added necessary numerical and distributed tests.
- [x] Updated API docstrings for fallback behavior and scalar synchronization.

# Additional Information

Related to #3719; alternative approach for comparison with #3815 and #3685.


akoumpa:
@HuiyingLi can you do a/b comparison with triton kernel to see if TE has any perf gain over triton? if not i'm inclined to keep the triton kernel; we got 2-3 PRs already adding the triton k.

PR #3880, opened 2026-09-14
chore: stop eagerly importing the transformers model zoo

akoumpa:
## What

`nemo_automodel.components.distributed.parallelizer` took **~7.2 s** to import. Only ~1.2 s of that was torch. The rest came from a dozen concrete `transformers.models.*.modeling_*` imports at module scope — used only for `isinstance` checks, dict keys and type annotations.

Importing any single one of them drags the whole model zoo along:

```
transformers.models.gemma3
  -> transformers.generation.candidate_generator -> sklearn -> pandas, scipy
  -> transformers.processing_utils -> image_utils   -> torchvision
  -> transformers.generation.continuous_batching    -> opentelemetry
```

Measured incrementally:

```
1.74 s   import torch
2.92 s   + transformers                                  (+1.18)
5.16 s   + transformers.models.gemma3.modeling_gemma3    (+2.24)  <- more than torch itself
```

## The fix

Three changes, no behaviour change:

- **`parallelizer.py`** — the 12 modeling imports move into the three functions that use them (`get_hf_tp_shard_plan`, `validate_tp_mesh`, `_get_model_layer_group_specs`). Nothing referenced them at module scope.
- **`optimized_tp_plans.py`** — `from __future__ import annotations` makes the annotation-only uses free, the imports move behind `TYPE_CHECKING`, and the `PARALLELIZE_FUNCTIONS` keys become literal qualnames. That's the same trick this file **already** used for qwen3_5 and falcon_h1, with comments saying exactly why.
- **Gemma4** keeps its optional-import fallback, now behind an `lru_cache`'d accessor so the placeholder remains one stable class object (callers use it for identity comparisons and as a dict key).

## Result

```
import parallelizer                     7.31 7.22 7.22 7.47  ->  4.19 3.77 4.10 4.11
import test_fsdp_embedding_split_gloo   7.31 7.19 7.23       ->  3.94 4.23 4.11
```

**−43%**, about 3.1 s off every import of that module. Local numbers (py3.10, torch 2.10); CI runs py3.12 and has `flash-linear-attention` installed, which isn't here, so the CI figure will differ.

This cost is paid by every process that touches the module: each of the ~50 `mp.spawn` children in the CPU unit-test suite (which re-import from scratch — measured at **83% of a spawn test's runtime**, e.g. 10.90 s of imports for 2.27 s of work in `test_split_embedding_fsdp_matches_unsharded_reference[1]`), plus pytest's own startup and the 76–81 s collection phase.

## Verification

- `PARALLELIZE_FUNCTIONS` (24 entries) and `_get_model_layer_group_specs()` (26 entries) serialize **byte-identically** before and after.
- The qualname literals were **generated by running `_get_class_qualname` over the real classes**, not written by hand — note the aliased ones resolve to their true names (`CustomLlamaForCausalLM` → `...models.llama.model.LlamaForCausalLM`).
- Nothing outside these modules imported the moved names (checked across `nemo_automodel/` and `tests/`).
- `ruff check` + `format --check` clean under **both** 0.12.12 (what `uv run --active ruff` resolves in CI) and 0.16.4 (what `.pre-commit-config.yaml` pins).
- `nemo_automodel/components/distributed/**` is not in `ty`'s scope (that override block is commented out in `pyproject.toml`), so no type-check impact.

## Not addressed here

48 modules across the package carry 98 of these eager modeling imports. This PR fixes only the two in the parallelizer chain, which is the one the unit tests actually import. The `components/models/*` ones are loaded per-model and want their own pass.

## Related

- #3858 — drops the redundant `-rA` report (~128 s/attempt)
- #3859 — adds the 5 s per-test timeout


akoumpa:
Added `nemo_automodel/components/distributed/AGENTS.md` (+ `CLAUDE.md` symlink, matching the root's arrangement) so this doesn't quietly regress — one new `from transformers.models.<arch>.modeling_<arch> import ...` in either file pulls the whole graph back in.

It records the rule, the measured numbers, the three replacements (literal qualname registry keys / `TYPE_CHECKING` for annotations / function-local imports for runtime use), the `-X importtime` command to verify, and the byte-identical equivalence check for `PARALLELIZE_FUNCTIONS` and `_get_model_layer_group_specs`. It's additive to the root `AGENTS.md`; `test_claude_review_policy.py` pins `AGENTS_PATH` to the repo root so a nested file doesn't affect it.

/ok to test 0376468710cba99b6a847be1f6f7646ea92db2c3

akoumpa:
Note: plan is to deprecate the mega-centralized-parallelizer and move it per-model.

akoumpa:
I have a follow-up PR to make the parallelizer infra only

HuiyingLi (approved):

Issue #3882, opened 2026-09-14
LR scheduler resume applies the old weight-decay schedule and treats num_steps as a delta


kabirvashisht4-glitch:
## Describe the bug

`OptimizerParamScheduler.load_state_dict` writes the resolved learning rate and
weight decay into `optimizer.param_groups` **before** it has restored the
weight-decay fields. On resume the learning rate comes from the checkpoint's
schedule while the weight decay comes from the constructor's, and nothing
recomputes it until the next `scheduler.step()`.

The training loop applies the optimizer update first
(`nemo_automodel/recipes/llm/train_ft.py`: `opt.step()` at L1275,
`scheduler.step(1)` at L1284), so the stale weight decay is used for a real
parameter update.

The ordering in `nemo_automodel/components/optim/scheduler.py`:

```python
self.lr_decay_style = self._check_and_set(...)   # LR fields restored

num_steps = state_dict["num_steps"]
self.step(increment=num_steps)                   # <- resolves + WRITES lr and wd

if "start_wd" in state_dict:                     # <- wd fields restored AFTER
    self.start_wd = self._check_and_set(...)
    self.end_wd = self._check_and_set(...)
    self.wd_incr_steps = self._check_and_set(...)
    self.wd_incr_style = self._check_and_set(...)
```

Every learning-rate field is restored before `step()`; every weight-decay field
is restored after it. The asymmetry looks unintentional.

This matters precisely when the checkpoint and the constructor disagree, which
is the case `use_checkpoint_opt_param_scheduler=True` (the default) exists to
serve — `_check_and_set` is written to let checkpoint values win.

## Second problem in the same call: the step count is folded in, not set

```python
self.step(increment=num_steps)
```

`step()` does `self.num_steps += increment`, but `num_steps` in the state dict
is an absolute position. That is only correct when the scheduler has counted
nothing yet. Loading twice, or loading a scheduler that has already advanced,
lands past the checkpoint's step and silently shifts the rest of the schedule.

## Steps/Code to reproduce bug

```python
import torch
from nemo_automodel.components.optim.scheduler import OptimizerParamScheduler

def build(**overrides):
    p = [torch.nn.Parameter(torch.zeros(1))]
    opt = torch.optim.SGD(p, lr=1.0, weight_decay=0.0)
    kw = dict(optimizer=opt, init_lr=0.0, max_lr=1.0, min_lr=0.0,
[...]

PR #3883, opened 2026-09-14
fix(optim): apply checkpoint weight decay and step count on scheduler resume

kabirvashisht4-glitch:
# What does this PR do ?

Makes `OptimizerParamScheduler.load_state_dict` restore all of its state before
any of it is applied, so a resumed run gets the checkpoint's weight-decay
schedule and lands on the checkpoint's step.

`load_state_dict` restored the learning-rate fields, called `step()` — which
resolves the learning rate and weight decay and writes both into
`optimizer.param_groups` — and only then restored `start_wd`, `end_wd`,
`wd_incr_steps` and `wd_incr_style`:

```python
self.lr_decay_style = self._check_and_set(...)   # LR fields restored

num_steps = state_dict["num_steps"]
self.step(increment=num_steps)                   # <- resolves + WRITES lr and wd

if "start_wd" in state_dict:                     # <- wd fields restored AFTER
    self.start_wd = self._check_and_set(...)
    ...
```

So the learning rate came from the checkpoint and the weight decay came from the
constructor, and nothing recomputed it until the next `scheduler.step()`. The
training loop calls `opt.step()` (`recipes/llm/train_ft.py` L1275) before
`scheduler.step(1)` (L1284), so that stale weight decay is applied to a real
parameter update. Every LR field is restored before `step()` and every WD field
after it; the asymmetry looks unintentional.

It only diverges when the checkpoint and the constructor disagree — which is the
case `use_checkpoint_opt_param_scheduler=True` (the default) exists to serve,
since `_check_and_set` is written to let checkpoint values win.

Measured on `main` @ `8a646a73`, CPU, resuming a step-50 checkpoint
(`end_wd=0.1, wd_incr_steps=100`) into a scheduler constructed with
`end_wd=0.9, wd_incr_steps=1000`:

| | learning rate | weight decay |
|---|---:|---:|
| reference at step 50 | 0.586824 | 0.05 |
| after `load_state_dict` | 0.586824 | **0.045** |
| what the restored schedule implies | 0.586824 | 0.05 |

`0.045` is `50/1000 * 0.9` from the constructor. `scheduler.get_wd()` already
returns `0.05` after the load — only the value handed to the optimizer is wrong.

## Second defect in the same call

```python
self.step(increment=num_steps)
```

`step()` does `self.num_steps += increment`, but `num_steps` in the state dict is
an absolute position. That is correct only when the scheduler has counted nothing
yet:

| | `num_steps` after |
|---|---:|
| load once | 50 |
| load twice | **100** |
| `step(20)` then load | **70** |

## The fix

Move the weight-decay restore above the `step()` call so `step()` is the last
thing `load_state_dict` does, and replay the absolute count from zero:

```python
self.num_steps = 0
self.step(increment=num_steps)
```

No behaviour changes for a freshly-constructed scheduler whose config matches its
checkpoint, which is the common resume path.

# Changelog

- `OptimizerParamScheduler.load_state_dict` now restores `start_wd`, `end_wd`,
  `wd_incr_steps` and `wd_incr_style` before the `step()` call that writes the
  resolved values into `optimizer.param_groups`, so the learning rate and the
  weight decay both come from the checkpoint's schedule.
- The checkpoint's `num_steps` is applied as an absolute position rather than an
  increment, making `load_state_dict` idempotent.
- The method docstring now states both contracts.
- `tests/unit_tests/optim/test_scheduler.py`: added assertions on what the load
  writes into `optimizer.param_groups` (weight decay and, as a regression guard
  against fixing this the other way round, learning rate), plus idempotence and
  load-after-progress cases for the step count.

**Pre checks**:

The only documentation of this contract was the method's one-line docstring,
which is expanded here. No user-facing docs or example configs describe the
restore ordering, so nothing else needed updating.

Verified locally: `pytest tests/unit_tests/optim/test_scheduler.py` -> 60 passed.
Three of the four new tests fail on `main` (`assert 0.045 == 0.05`, `assert 100
== 50`, `assert 70 == 50`); the learning-rate one passes on `main` too, since it
guards behaviour this PR must not break. `ruff format --check` / `ruff check`
clean. CPU only, no GPU needed.

# Additional Information

- Related to #3882


yuhezhang-ai:
/ok to test [d923493](https://github.com/NVIDIA-NeMo/Automodel/pull/3883/commits/d92349350a0e17109233438c9776d824bce6a09a)

yuhezhang-ai (approved):
The fix looks correct.

For clarity, the weight-decay issue affects resumes where the config differs from the checkpoint while checkpoint scheduler values are selected. Same-config resumes and explicit config overrides are unaffected. The step-count issue concerns restoring an already-advanced scheduler.

This is a useful correctness fix with limited impact on typical training workflows.

akoumpa:
ci passes https://github.com/NVIDIA-NeMo/Automodel/actions/runs/34921370616/job/104543525128?pr=3883 ; dist errors are fixed in upstream main.

PR #3891, opened 2026-09-15
fix(nemotron_v3): keep the generation config on consolidated exports

### LINKED ISSUE #3890: nemotron-3 full-weight exports write an empty generation_config.json, so generate() never stops
**the bug**

when you train a nemotron-3 model with full weights and export it, the exported folder gets a `generation_config.json` that is just `{}`.

transformers always trusts that file over `config.json` when it exists. so the reloaded model has no eos token, and `generate()` keeps going until it hits `max_new_tokens`.

**why**

our `NemotronHForCausalLM` starts with an empty `GenerationConfig()`. the export just writes whatever the model has, so the empty one ends up on disk.

fixing the model init alone is not enough. the hub checkpoint has `eos_token_id: [2, 11]` in `generation_config.json` but only `2` in `config.json`. the loader has to read the checkpoint's own file, same as `PreTrainedModel.from_pretrained` does.

**who hits it**

every full-weight nemotron-3 config we ship (11 of them, including both customizer full-sft recipes), plus the two nemotron omni recipes that reuse the same llm class. peft runs are fine since adapters do not write this file.

**repro**

```
model = NemotronHForCausalLM(config)          # what the registry path does
model.generation_config                       # empty, eos is None
_save_generated_hf_assets(...)                # export
cat out/generation_config.json                # {}
AutoModelForCausalLM.from_pretrained("out")   # eos_token_id is None
```

**fix**

#<PR NUMBER>

stanley1208:
Fixes #3890.

nemotron-3 full-weight exports were writing an empty `generation_config.json`, so reloaded models had no eos token and never stopped generating. details and repro in the issue.

**what changed**

- the model builds its generation config from the model config instead of starting empty, same as `PreTrainedModel.__init__`
- after the loader builds a model from a checkpoint, it restores that checkpoint's generation settings: `generation_config.json` if present, otherwise the generation fields in `config.json`. this is what `PreTrainedModel.from_pretrained` does. it applies to any registry model with a real `GenerationConfig`, not just nemotron
- `NemotronHForCausalLM.from_pretrained` does the same, so calling it directly matches the loader

**tests**

- model: config seeds the generation config, `from_pretrained` restores the checkpoint file, export writes it back out
- loader: restores the file, falls back to `config.json`, ignores a checkpoint sitting at `config.name_or_path` on from_config builds, leaves models without a generation config alone
- helpers: seeding and the two-step restore

all of them fail on main except the "leave alone" cases.


nemo_automodel/_transformers/model_init.py
```diff
@@ -1304,7 +1305,15 @@ def __init_model(
...
             with local_torch_dtype(torch_dtype, model_cls.__name__):
-                return True, model_cls(hf_config, *model_args, **kwargs)
+                model = model_cls(hf_config, *model_args, **kwargs)
+            if is_pretrained_init:
+                # Custom constructors only see the config. Restore the checkpoint's
+                # generation settings the way PreTrainedModel.from_pretrained does:
+                # generation_config.json carries stop tokens and sampling defaults that
+                # the config lacks, and the consolidated export writes
+                # model.generation_config back out.
+                restore_pretrained_generation_config(model, pretrained_model_name_or_path)
```
yuhezhang-ai: **[P2] Preserve checkpoint-loading options**

Could we retain the checkpoint-loading options before constructor-argument filtering and pass them through both generation-config loading attempts? This call currently forwards only the path, dropping `subfolder`, `revision`, `token`, `cache_dir`, and `local_files_only`.

In a tiny real Qwen3 reproduction, loading with `subfolder="child"` correctly read the model config from that child, but this restore step selected the parent folder's generation EOS `7` instead of the child's `[2, 11]`. This can silently give the model the wrong stop tokens. A regression test with different generation configs in the parent and child directories would cover the case.

nemo_automodel/components/models/common/utils.py
```diff
@@ -83,6 +85,65 @@ def get_is_first_microbatch() -> bool | None:
...
+        return GenerationConfig.from_pretrained(pretrained_model_name_or_path)
+    except OSError:
+        logger.info(
+            "No generation_config.json in %s, using the generation fields of config.json.",
+            pretrained_model_name_or_path,
+        )
+    try:
+        return GenerationConfig.from_pretrained(
+            pretrained_model_name_or_path, config_file_name="config.json", _from_model_config=True
+        )
```
yuhezhang-ai, line 160: **[P2] Preserve explicit token overrides during fallback**

Could the fallback preserve explicit overrides already applied to `hf_config` while recovering legacy generation settings from disk? When `generation_config.json` is absent, the config returned here replaces the model's config-derived generation settings wholesale.

With a tiny real Qwen3 model and an explicit `eos_token_id=9` override, `model.config.eos_token_id` remains `9`, but `model.generation_config.eos_token_id` becomes the disk value `2`. If raw `config.json` omits EOS, it becomes `None`, losing the stopping token entirely. Disabling only the added restore call preserves generation EOS `9` in both cases. Tests for a conflicting disk value and an omitted EOS field would help protect the existing override behavior.

nemo_automodel/components/models/common/utils.py
```diff
@@ -83,6 +85,65 @@ def get_is_first_microbatch() -> bool | None:
...
+    """
+    if not isinstance(config, PretrainedConfig):
+        return GenerationConfig()
+    try:
+        return GenerationConfig.from_model_config(config)
+    except NotImplementedError:
+        return GenerationConfig()
+
+
+def load_pretrained_generation_config(pretrained_model_name_or_path) -> GenerationConfig | None:
```
yuhezhang-ai: **[P3] Add parameter types to the new public helpers**

Small API-contract follow-up: could we annotate the supported parameter types for all three new public helpers? `config` at line 88 and `pretrained_model_name_or_path` at lines 106 and 133 are currently untyped. The repository requires annotations on new public APIs; declaring the accepted config and path types would make these helper contracts explicit.

yuhezhang-ai (commented):
Thanks for addressing the generation-config export issue. I found two loading edge cases and one small API-contract issue, detailed inline.

Validation: all 15 focused generation-config tests passed, as did Ruff lint and formatting checks for the changed production files. Both loading issues below reproduced with a tiny real Qwen3 model. The local tests used Transformers 5.12.1, while this PR pins 5.15.1; I also inspected the relevant generation-config loading and fallback code in the 5.15.1 source. These results are focused local validation, not a full CI run.

stanley1208:
@yuhezhang-ai thanks, all three are fixed.

nemo_automodel/components/models/common/utils.py
```diff
@@ -83,6 +86,105 @@ def get_is_first_microbatch() -> bool | None:
...
+            pretrained_model_name_or_path, config_file_name="config.json", _from_model_config=True, **loading_kwargs
+        )
+    except (OSError, TypeError):
+        # transformers resolves a missing non-default config file to None and then
+        # fails to open it (TypeError) instead of raising OSError.
+        return None
+    if config is not None:
+        # The raw file only contributes the legacy sampling fields; the token ids and
+        # anything else set on the in-memory config, overrides included, stay on top.
+        generation_config.update(**generation_config_from_model_config(config).to_diff_dict())
```
yuhezhang-ai, line 168: **[P2] Preserve explicit default-valued EOS overrides**

Could we also preserve explicit overrides when their values equal the GenerationConfig defaults? `to_diff_dict()` omits those values, so an explicit `eos_token_id=None` is still lost during this merge.

With EOS `0` in `config.json` and no `generation_config.json`, a tiny real Qwen3 model loaded with `eos_token_id=None` correctly retains `None` in `model.config`, but its generation config restores EOS `0`. With the output head set to predict token `0` deterministically and `max_new_tokens=3`, generation stops after one token. Disabling only the added restore call preserves `None` and generates all three tokens.

Preserving explicit overrides independently of the default-eliding serialization, with a regression test for `eos_token_id=None`, would cover this remaining case.

yuhezhang-ai (commented):
Thanks for the updates. The earlier subfolder and eos_token_id=9 reproductions now pass, and the missing parameter annotations have been added. All 22 focused generation-config tests passed, along with Ruff lint and formatting checks. I found one remaining override edge case, detailed inline.

Validation used Transformers 5.12.1 locally; I also checked that the pinned 5.15.1 source has the same default-removal behavior in to_diff_dict().

stanley1208:
@yuhezhang-ai thanks for the catch, I just had another update on it, could you have a look when you got time?

nemo_automodel/_transformers/model_init.py
```diff
@@ -1299,10 +1301,14 @@ def __init_model(
...
             kwargs.pop("trust_remote_code", None)
+            # Keep the hub loading options before the constructor-argument filter drops
+            # them: the generation-config restore below has to read from the same
+            # subfolder/revision the weights came from.
+            loading_kwargs = {key: kwargs[key] for key in HUB_LOADING_KWARGS if key in kwargs}
             # Treat config-related kwargs as config overrides (HF behavior) and
             # avoid forwarding them into model __init__.
             init_param_names = _get_init_param_names(model_cls)
-            _consume_config_overrides(hf_config, kwargs, init_param_names=init_param_names)
+            config_overrides = _consume_config_overrides(hf_config, kwargs, init_param_names=init_param_names)
```
yuhezhang-ai: **Non-blocking follow-up: include dictionary config overrides**

Could we also retain the keys from a dictionary-style `config` override before it is removed from `kwargs`, and merge them into this set? This appears to be a small extension of the current fix.

`config={"eos_token_id": None}`—the form used by `model.config` YAML/CLI overrides—is applied by `get_hf_config` and popped earlier in this function, so `_consume_config_overrides` never records that `eos_token_id` was explicitly set. With EOS `0` in `config.json` and no `generation_config.json`, the fallback consequently restores EOS `0`. In a tiny real Qwen3 reproduction, this form generated one token instead of the requested three, while the direct `eos_token_id=None` keyword now correctly generated all three.

A regression test that passes the dictionary through the real config-loading path would cover this. The original export fix and the previously tested cases are working; I would not hold up the PR on this edge case alone.

yuhezhang-ai (commented):
Thanks for the update—the direct eos_token_id=None case now works, along with the earlier subfolder and EOS=9 cases. All 24 focused generation-config tests and the production-file lint/format checks passed. There is one remaining dictionary-override edge case below; I would treat it as non-blocking for the original Nemotron export fix. Local runtime validation used Transformers 5.12.1; the PR pins 5.15.1.

stanley1208:
@yuhezhang-ai thank you again, the issue is also fixed as well!

yuhezhang-ai (approved):
Thanks! LGTM.

yuhezhang-ai:
/ok to test [bbe0141](https://github.com/NVIDIA-NeMo/Automodel/pull/3891/commits/bbe01412013ce4a29ae0df942d994fd3bd2c1254)

PR #3897, opened 2026-09-15
fix(loss): align MTP ignore index with configured loss

akoumpa:
# What does this PR do ?

Align dataset padding, Masked CE, MTP masking, and train/validation normalization with the configured loss `ignore_index`.

# Changelog

- Normalize canonical dataset padding (`-100`) to the configured loss sentinel at the loss boundary.
- Derive MTP masking from the configured loss and reject mismatched overrides.
- Count supervised labels consistently in LLM and VLM train/validation paths.
- Preserve ignore-index and reduction settings when falling back to Masked CE.
- Add regressions for non-default `ignore_index`, MTP, and token denominators.

**Pre checks**:

# Additional Information

- Split from the ignore-index review discussion in #3303 so that PR remains scoped to domain-mixture datasets.


nemo_automodel/components/loss/masked_ce.py
```diff
@@ -80,7 +83,7 @@ def forward(
         if isinstance(labels, DTensor):
             labels = labels.full_tensor()
 
-        loss = F.cross_entropy(logits, labels, reduction=self.reduction)
+        loss = F.cross_entropy(logits, labels, ignore_index=self.ignore_index, reduction=self.reduction)
```
yuhezhang-ai: [P2] Align KD token counts with the new CE masking

The LLM and VLM knowledge-distillation recipes also use this loss, but their training and validation paths still count `labels != -100` (`recipes/llm/kd.py`: 792, 885, 1090; `recipes/vlm/kd.py`: 416, 540). With `ignore_index=0` and labels `[[0,1],[0,2]]`, they pass `num_label_tokens=4` although CE supervises only two labels. I reproduced CE loss and gradients at half the correctly normalized values in both recipes.

Could we update the KD normalization paths to respect the configured masking, or explicitly reject non-default `ignore_index` there, and add a regression test?

akoumpa: Addressed in 39218874b8fb9844b3a42851b694cd340b1fee59. LLM non-PP train, LLM PP train, LLM validation, VLM train, and VLM validation now derive num_label_tokens from the configured loss ignore_index via the shared counting helper. Added ignore_index=0 regressions using labels [[0,1],[0,2]], including the real LLM gradient-scaling and optimizer-update assertions. Validation: 157 passed, 8 skipped across the adjacent masked CE, KD loss, KD recipe, and base finetune suites; Ruff format/check pass.

yuhezhang-ai: Thanks, this fixes the CE denominator. I found one remaining P2 issue in the KD term at the current head (`10b8db1`): the shared token count now follows `self.loss_fn.ignore_index`, but `self.kd_loss_fn` still receives the original labels and applies its independently configured mask. This affects both LLM and VLM KD.

With `loss_fn.ignore_index=0`, `KDLoss` using its default `ignore_index=-100`, and labels `[[0,1],[0,2]]`, the new count is 2 while KD includes all 4 tokens. Using the recipe's counting code and the real `KDLoss`, I reproduced a KD loss of 1.18945 instead of 0.59473, with gradients exactly doubled. This also affects `kd_ratio=1`, where CE is unused. An all-zero-label batch produces a zero denominator and an infinite KD loss.

The default configuration with both ignore indices at `-100` is unaffected; this is specific to the non-default configuration this PR is adding support for. Could we align the KD mask with its denominator, use separate counts if different masks are intentional, or explicitly reject unsupported combinations? A regression exercising the real `KDLoss` would catch this; the new recipe tests stub the forward/loss computation, so they verify the denominator without checking the KD numerator.

yuhezhang-ai: @akoumpa see above comment

akoumpa: Thanks @yuhezhang-ai  wip

yuhezhang-ai (approved):

PR #3903, opened 2026-09-16
feat(config): opt in to typed nested dataclass construction

hasuoshenyun:
# What does this PR do ?

Allow dataclass fields to explicitly opt into typed nested YAML construction at the existing ConfigNode boundary.

# Changelog

- Support `field(metadata={"instantiate": True})` for unambiguous typed nested dataclasses and homogeneous sequences.
- Opt in an entire supported typed subtree: plain nested dataclass fields do not need additional metadata, and homogeneous list/tuple fields follow their annotations.
- Preserve existing list/dictionary behavior for unmarked fields on the explicitly instantiated target, outside opted-in subtrees.
- Do not coerce runtime overrides or resolve their contents as environment variables.
- Preserve construction behavior when runtime type hints cannot resolve TYPE_CHECKING-only references. Resolution is per class, not per field: a `NameError` or `TypeError` disables annotation-based field conversion for that class, so even marked fields can retain plain dictionaries. Existing explicit nested `_target_` behavior is unchanged.
- Add regression tests independent of QAT imports, including three-level construction, unmarked sibling preservation, repeated instantiation, runtime override identity, and unchanged source configuration.

This is a configuration-construction convenience, not a training or quantization algorithm change. Explicit nested `_target_` configurations remain supported; this opt-in lets annotations supply the nested dataclass types instead of repeating class paths in YAML.

- [x] Read the contributor guidelines and signed off commits.
- [x] Added focused compatibility tests.
- [x] Documented the explicit opt-in contract.

# Additional Information

This PR is now independent of #3902 and rebased onto upstream `main` at `44cf34834`. Its diff contains only `nemo_automodel/components/config/loader.py` and `tests/unit_tests/config/test_loader.py`; no expert LoRA or QAT implementation changes are included. The two config commits are `718ab3a00` and `10f9d6c6d`. Rebase range-diff confirms both patches are unchanged from their previously published versions.

The implementation deliberately does not enable type coercion globally: differential checks against the original loader confirm unchanged list/dict construction for ordinary dataclasses.

Validation rerun on the independent head `10f9d6c6de23444bd08f70c48e30e3dc0ccde5b2`: **97 passed** across `tests/unit_tests/config/`, including the three-level subtree regression and existing unresolved-hint fallback tests. Loader Ruff lint/format and `git diff --check` pass. Upstream CI for this SHA still requires maintainer authorization; local results are not a substitute for upstream CI.


hasuoshenyun:
Hi @akoumpa, this PR is now ready for review and independent of #3902, rebased onto main. The diff contains only the config loader and its tests. The recursive opt-in/fallback contract is documented, and all 97 config tests pass locally on head 10f9d6c6de23444bd08f70c48e30e3dc0ccde5b2. Could you please review and authorize CI for this SHA when convenient? Thank you!

PR #3918, opened 2026-09-16
fix(loss): stop masked losses from overwriting the caller's labels

### LINKED ISSUE #3917: Masked CE losses overwrite the caller's labels in place, zeroing a reused batch's loss
## Describe the bug

`MaskedCrossEntropy`, `ChunkedCrossEntropy` and `TEParallelCrossEntropy` all apply their
optional `mask` with an **in-place** `labels.masked_fill_`:

```python
logits = logits.view(-1, logits.size(-1))
labels = labels.view(-1)
if mask is not None:
    with torch.no_grad():
        ...
        labels.masked_fill_(mask.view(-1) == 0, self.ignore_index)
```

`labels.view(-1)` does not copy — it aliases the caller's storage. So the fill writes
`ignore_index` (`-100`) directly into the caller's label tensor. In
`TEParallelCrossEntropy` the aliasing is worse: `labels` has already been through
`labels.to_local()`, and that local shard aliases the DTensor's storage, so the write
lands in the caller's distributed tensor.

Nothing in the signature or (until now) the docstring says the argument is consumed.
`calculate_loss`, the wrapper most callers go through, documents the opposite:
*"The caller's mapping and tensors are not mutated."*

| file | line |
|---|---|
| `nemo_automodel/components/loss/masked_ce.py` | 72 |
| `nemo_automodel/components/loss/chunked_ce.py` | 197 |
| `nemo_automodel/components/loss/te_parallel_ce.py` | 174 |

```python
import torch, torch.nn.functional as F
from nemo_automodel.components.loss.masked_ce import MaskedCrossEntropy

torch.manual_seed(0)
logits = torch.randn(6, 5)
labels = torch.randint(0, 5, (6,))

mask_a = torch.tensor([1, 1, 1, 0, 0, 0])   # first half supervised
mask_b = torch.tensor([0, 0, 0, 1, 1, 1])   # second half supervised

def reference(mask):
    t = labels.clone(); t[mask == 0] = -100
    return F.cross_entropy(logits, t, reduction="sum")

loss_fn = MaskedCrossEntropy(reduction="sum")
shared = labels.clone()
a = loss_fn(logits, shared, mask=mask_a)
b = loss_fn(logits, shared, mask=mask_b)    # same tensor reused

print(a.item(), reference(mask_a).item())   # 4.717452 4.717452
print(b.item(), reference(mask_b).item())   # 0.000000 3.509581   <-- wrong
print(shared.tolist())                      # [-100, -100, -100, -100, -100, -100]
```

## Expected behavior

Each call scores the positions its own mask selects, and the caller's `labels` tensor is
unchanged afterwards.

## Actual behavior

The first call destroys the supervision. The second call sees a tensor that is already
all `ignore_index`, so **every position is ignored and the loss is exactly `0.0`** —
which backprops a zero gradient. There is no exception and no warning; the step simply
contributes nothing.

Any second consumer of the same tensor is affected the same way: another loss term over
the same batch, a token-accuracy or perplexity metric computed after the loss, a cached
or memory-mapped dataset that hands back the same underlying tensor on the next epoch.

## Why it has gone unnoticed

No recipe in-tree currently passes `mask=` (`calculate_loss` forwards only `logits`,
`labels` and `num_label_tokens`), so today this is a latent defect on a public API rather
than an active training-corruption bug. It is reachable by anyone calling these loss
modules directly, which the `mask` parameter exists to support.

The existing coverage cannot catch it. `test_masked_cross_entropy_with_mask` builds its
reference *after* calling the loss:

```python
loss_custom = MaskedCrossEntropy()(logits, targets, mask=mask)
targets_ref = targets.clone()      # already mutated to -100 here
targets_ref[mask == 0] = -100      # re-applies an applied mask: a no-op
```

so it re-applies a mask that is already applied and passes either way.

## Environment overview

- Reproduced on CPU, `torch 2.x`, current `main` (`4a334b644`). No GPU or distributed
  setup required.

## Proposed fix

Use the out-of-place `masked_fill` in all three losses, document that `labels` is left
intact, and snapshot `targets` before the call in the existing test so it can observe a
regression. PR to follow.

kabirvashisht4-glitch:
# What does this PR do ?

Fixes #3917: `MaskedCrossEntropy`, `ChunkedCrossEntropy` and `TEParallelCrossEntropy`
applied their optional `mask` with an in-place `labels.masked_fill_`, writing
`ignore_index` into the **caller's** tensor.

`labels.view(-1)` aliases the caller's storage rather than copying it, so the fill is
visible to whoever owns the batch. `TEParallelCrossEntropy` is worse: `labels` has
already been through `labels.to_local()`, and that local shard aliases the DTensor's
storage, so the write lands in the caller's distributed tensor.

The failure is silent and it zeroes a loss. Reusing the tensor for a second mask scores
the intersection of the two masks, not the second one:

| | `mask_a` = first half | `mask_b` = second half (same tensor) |
|---|---|---|
| expected | 4.717452 | 3.509581 |
| before | 4.717452 | **0.000000** |
| after | 4.717452 | 3.509581 |

A loss of exactly `0.0` backprops a zero gradient with no exception and no warning. The
same applies to any second consumer of the batch: another loss term, a token-accuracy or
perplexity metric computed after the loss, or a cached dataset that returns the same
underlying tensor on the next epoch.

`calculate_loss` — the wrapper most callers go through — documents the opposite
guarantee: *"The caller's mapping and tensors are not mutated."*

## Scope

No in-tree recipe passes `mask=` today (`calculate_loss` forwards only `logits`, `labels`
and `num_label_tokens`), so this is a latent defect on a public API rather than an active
training-corruption bug. It is reachable by anyone calling these modules directly, which
is what the `mask` parameter is for.

## Why the existing test could not catch it

`test_masked_cross_entropy_with_mask` built its reference *after* the loss call:

```python
loss_custom = MaskedCrossEntropy()(logits, targets, mask=mask)
targets_ref = targets.clone()      # already mutated to -100 here
targets_ref[mask == 0] = -100      # re-applies an applied mask: a no-op
```

It re-applied an already-applied mask, so it passed with or without the bug. This PR
snapshots `targets` beforehand, which makes the existing assertion able to fail.

# Changelog

- Replace the in-place `labels.masked_fill_` with the out-of-place `masked_fill` in
  `masked_ce.py`, `chunked_ce.py` and `te_parallel_ce.py`.
- Document on all three that `labels` is not modified in place. `chunked_ce.py`
  previously documented the opposite ("ignored positions are replaced with
  `ignore_index` in this tensor"); that line is now correct.
- Snapshot `targets` before the loss call in `test_masked_cross_entropy_with_mask` so it
  can observe an in-place write.
- Add, for `MaskedCrossEntropy` and `ChunkedCrossEntropy`: a test that the caller's
  `labels` is unchanged, and a test that two disjoint masks over one tensor each score
  their own positions. Add the corresponding no-mutation test for
  `TEParallelCrossEntropy`, guarded on `HAVE_TE_PARALLEL_CE` and CUDA.

**Pre checks**:

Verified locally:

- `pytest tests/unit_tests/loss/` -> 313 passed, 37 skipped.
- The four new CPU tests fail on `main` and pass here. Reverting only the
  `nemo_automodel/` changes and keeping the tests reproduces:
  `second mask scored 0.000000, expected 6.012943`.
- `ruff format --check` and `ruff check` clean on every changed file.
- CPU only; the TE test is skipped without Transformer Engine and CUDA.

# Additional Information

- The three call sites are the only in-place writes to `labels` under
  `nemo_automodel/components/loss/`.


akoumpa:
Hi @kabirvashisht4-glitch can you take a look at the conflicts when you get the chance? thank you

akoumpa (approved):

PR #2119, opened 2026-05-03
feat(config): add example YAML linter

zeel2104:
# What does this PR do ?

Adds a YAML linter for `examples/` configs and moves recipe target validation out of pytest into the lint stage.

# Changelog

- Added `tools/lint_example_yamls.py` to lint example YAML files.
- Validates YAML parseability, duplicate top-level keys, required recipe targets, recipe target resolution, `recipe` as the first top-level section, and `ci` as the last top-level section.
- Added the YAML linter to the recipe config validation GitHub workflow.
- Added a local pre-commit hook for changed example YAML files.
- Replaced the previous recipe target pytest scan with focused unit tests for the new linter.
- Updated existing example YAMLs that violated the new top-level ordering rules.

# Before your PR is "Ready for review"

**Pre checks**:

- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [x] Did you add or update any necessary documentation?

# Additional Information

- Related to #1643

Validation run:

```powershell
python tools\lint_example_yamls.py --automodel-dir .
python -m pytest tests\unit_tests\config\test_example_yaml_linter.py
python -m ruff format --preview --no-cache tools\lint_example_yamls.py tests\unit_tests\config\test_example_yaml_linter.py
python -m ruff check --preview --no-cache tools\lint_example_yamls.py tests\unit_tests\config\test_example_yaml_linter.py

PR #3938, opened 2026-09-18
refactor(distributed): replace globals() monkey-patch, add no-globals lint gate

akoumpa:
# What does this PR do ?

Removes the `globals()` monkey-patch from `components/distributed/parallelizer.py`, and adds a lint gate (pre-commit + CI) so dynamic module-namespace mutation cannot come back.

`Qwen3_5ParallelizationStrategy` needed `DefaultParallelizationStrategy.parallelize` to use a dtype-aware FSDP walk instead of `apply_fsdp2_sharding_recursively`. It got there by rebinding the module global for the duration of the `super()` call and restoring it in a `finally` — a process-wide mutation that any concurrent or nested `parallelize` of another model would have picked up, as the code comment itself acknowledged:

```python
# Temporarily swap the global — safe because model init is single-threaded
# (one model is parallelized at a time). Not safe under concurrent calls.
original_fn = globals().get("apply_fsdp2_sharding_recursively")
```

It is now an ordinary override point.

# Changelog

**Refactor (`parallelizer.py`)**
- `DefaultParallelizationStrategy._apply_fsdp_sharding()` wraps the module-level `apply_fsdp2_sharding_recursively`; `parallelize()` calls it through `self`. This is the override point for strategies deriving from the default one, including those registered via `register_parallel_strategy`.
- `Qwen3_5ParallelizationStrategy` overrides that method with the former `_fsdp_by_dtype` closure, moved out verbatim (recursion via `self._apply_fsdp_sharding`); the closed-over `fp32_compute_module_names` becomes the class attribute `_fp32_compute_module_names`. Its `parallelize()` is now `super().parallelize(...)` plus the CP-mesh installation — the `try`/`finally` is gone.
- The Qwen3.5 override now forwards `fully_shard_fn` to `fully_shard_by_dtype` and through its recursion instead of `del`-ing it. Previously a caller passing a model-specific primitive would have had the root and untied-embedding units wrapped by it but the decoder layers wrapped by torch's `fully_shard`. No in-tree caller passes `fully_shard_fn` to Qwen3.5, so this is latent; the two axes (which primitive wraps a unit vs. how units are grouped) now compose.

**Lint gate**
- New `tools/lint_no_globals.py`, an AST check. Ruff has no equivalent rule — `PLW0603` only covers the `global` statement and `TID251` cannot ban a bare builtin call. It rejects `globals()` and the spellings someone reaches for when `globals()` is refused: `setattr`/`delattr(sys.modules[__name__], ...)`, `sys.modules[__name__].__dict__`, and module-scope `vars()`.
- The one exemption is the PEP 562 lazy-import cache, `globals()[name] = attr` written directly inside a module-level `def __getattr__(name)` — the key must be the function's parameter, and nested scopes / `async def` / class `__getattr__` do not qualify. The eight existing caches all match: `nemo_automodel/__init__.py` (×2), `_diffusers/__init__.py`, `_transformers/__init__.py`, `components/datasets/diffusion/__init__.py`, `components/flow_matching/__init__.py`, `recipes/diffusion/__init__.py`, `recipes/multimodal/__init__.py`. Zero baseline suppressions.
- Scope matches bandit: `app.py`, `nemo_automodel/`, `examples/`, `scripts/`, `tools/`, `tutorials/`. `tests/` is excluded, as it is for ruff and bandit (three test files use `globals()` for fixtures; they are out of scope by design).
- Wired into `.pre-commit-config.yaml` (incremental, changed files) and the `linting` CI job (whole tree) next to the bandit step. Recorded in `.agents/contributor-skills/linting-and-formatting/SKILL.md` (the lint source of truth) and `CONTRIBUTING.md`.
- Files are read as bytes so BOMs and `# -*- coding -*-` cookies are honored; unparseable files are reported rather than skipped; a missing path, a non-checkout `--automodel-dir`, or zero collected files exits 2 instead of passing silently.

Run locally with `python tools/lint_no_globals.py`.

**Pre checks**:

# Additional Information

**Behavior**: unchanged for every strategy, with one deliberate exception — `Qwen3_5ParallelizationStrategy.parallelize(..., fully_shard_fn=custom)` now applies `custom` to the decoder layers too (no caller does this today). The default hook still resolves `apply_fsdp2_sharding_recursively` as a module global at call time, so the existing `mock.patch("...parallelizer.apply_fsdp2_sharding_recursively")` tests keep exercising the same seam.

**Testing**
- `test_dtype_sharding_does_not_mutate_module_globals`: asserts the module global is stable *while* Qwen3.5 shards (the old `finally` restored it, so an after-only check passes either way). Fails on the pre-refactor file, passes here.
- `test_fsdp_sharding_hook_forwards_the_positional_contract`: pins the hook's exact argument shape for the default strategy — subclasses override with the same signature, so a reordered positional argument would silently misconfigure them.
- `test_dtype_walk_honors_fully_shard_fn`: fails against the previous `del`, passes here.
- `tests/unit_tests/tools/test_no_globals_linter.py`: 25 tests covering the exemption boundary, each banned spelling, BOM / latin-1 / null-byte / syntax-error inputs, `main()` exit codes, and a guard that the shipped tree stays clean so a reintroduced `globals()` fails unit tests as well as lint.
- `tests/unit_tests/distributed/`: 935 passed, 13 skipped. The 6 failures under `distributed/pipelining/test_functional.py` reproduce unchanged on stock `main`.
- The linter flags all three former sites when run against the pre-refactor file; the current tree is clean (827 files).

This gate is red on `main` without the refactor, which is why both land in one PR (same pattern as #3240, #2119, #3535).

Observation, not addressed here: the `linting` CI job pins `ruff~=0.12.0` via the uv group while `.pre-commit-config.yaml` pins ruff v0.16.4; the two disagree on 13 files' formatting today.


HuiyingLi (approved):
TYSM

Issue #3867, opened 2026-09-11
Unify PEFT export naming and require real HF reload tests


yuhezhang-ai:
PEFT export failures keep recurring because wrapper prefixes, model-specific namespace conversion, adapter tensor layouts, and target metadata are handled in separate paths. AutoModel save/reload round trips can agree with each other while the exported artifact fails to load into the actual HF PEFT model.

Examples include [AMINT-330](https://linear.app/nvidia/issue/AMINT-330/missing-peft-adapter-weights-in-checkpoint-causing-assertionerror) (Nemotron backbone/model namespace mismatch), NVIDIA-NeMo/Automodel#3656 (Omni wrapper skips conversion for PEFT-prefixed keys), NVIDIA-NeMo/Automodel#3595 (thinker prefix order), and NVIDIA-NeMo/Automodel#3592 (expert ParamWrapper layout compatibility). The narrow [AMINT-330](https://linear.app/nvidia/issue/AMINT-330/missing-peft-adapter-weights-in-checkpoint-causing-assertionerror) fix is PR #3866; this follow-up stays in Backlog and does not block that fix.

## Scope

* Extend the existing tiny-model HF PEFT reload/merge tests into a shared suite for supported export formats. Establish coverage before changing the common export boundary.
* Give the shared checkpoint boundary sole ownership of the PEFT outer prefix, so model adapters convert model paths without independently adding or stripping base_model.model. Apply the same contract to bulk and streaming export, and preserve loading of existing supported checkpoints.
* Keep architecture-specific renaming, multimodal nesting, and expert tensor conversion in the owning model adapter. Use the same model-owned naming rules for exported tensor keys and target_modules/target_parameters metadata.
* Make the receiving HF implementation and supported PEFT layout explicit through the existing export compatibility choices. Avoid inferring the PEFT target namespace solely from the original base checkpoint's serialized names.
* Add appropriate checks for duplicate output keys and unresolved targets. Keep exceptions for unsupported components explicit and narrow; do not load a full second model on every production save.

## Acceptance criteria

* Representative tiny dense, MoE, and multimodal models export through the real metadata and tensor save path and load with the actual HF PEFT consumer.
* Tests initialize nonzero adapter weights and check every expected tensor's presence, shape, and value, plus forward or merged-weight equivalence against the native adapter behavior.
* Bulk and streaming exports agree. Explicit legacy exports and loading existing supported adapter checkpoints retain coverage.
[...]

PR #3866, opened 2026-09-11
fix(peft): export Nemotron adapters in the Transformers v5 namespace

yuhezhang-ai:
# What does this PR do ?

Fixes [AMINT-330: Missing PEFT adapter weights in checkpoint causing AssertionError](https://linear.app/nvidia/issue/AMINT-330/missing-peft-adapter-weights-in-checkpoint-causing-assertionerror).

Fix Nemotron PEFT exports from base checkpoints that use `backbone.*` names so they load into the built-in Transformers v5 model. Previously, the adapter weights and `target_modules` retained `backbone.*`, while the receiving model exposed `model.*`; PEFT loaded only the matching `lm_head` adapters and omitted the affected Mamba, latent-projection, and shared-expert adapters.

# Changelog

- Export modern Nemotron adapter weights and target metadata under `model.*`, independently of the original base checkpoint namespace.
- Preserve the existing namespace for full checkpoints and explicit `v4_compatible=True` adapter exports. Forward that existing option through the model-owned target-module mapping hook.
- Add a real tiny-model HF PEFT reload regression checking all saved tensors exactly, matching forward logits, bulk/streaming export agreement, and legacy export preservation. Exercise the existing Nemotron expert-merge test without forcing the source checkpoint prefix.
- Select the upstream PyTorch Mamba reference functions within the CPU regression test so installed optional CUDA kernels do not override its CPU execution.

# Before your PR is "Ready for review"

**Pre checks**:

- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Automodel/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [x] Did you add or update any necessary documentation? Updated API docstrings; no user configuration change.

# Additional Information

Validation: **82 passed, no skips** across:

```
tests/unit_tests/models/test_moe_peft_v5_state_dict_adapters.py
tests/unit_tests/models/nemotron_v3/test_nemotron_v3_state_dict_adapter.py
tests/unit_tests/checkpoint/test_addons.py
tests/unit_tests/models/kimi_k3/test_state_dict_adapter.py
```

[...]

PR #3945, opened 2026-09-19
test(peft): add a shared HF PEFT reload suite

### LINKED ISSUE #3944: qwen3_omni_moe adapter_config target_modules miss the thinker. namespace, so peft also adapts the talker
**what happens**

a lora adapter trained on qwen3_omni_moe saves tensor names under `thinker.` but `adapter_config.json` target_modules without it:

```
adapter_model.safetensors : base_model.model.thinker.model.layers.0.self_attn.q_proj.lora_A.weight
adapter_config.json       : "model.layers.0.self_attn.q_proj"
```

peft matches target_modules by suffix, so on the full omni model that entry also matches the talker. the talker gets adapter modules the file has no weights for, so they stay randomly initialized. you serve a model with untrained adapters on a tower you never fine-tuned, with no warning.

**why**

`to_hf` and `convert_single_tensor_to_hf` both add the namespace via `_add_thinker_prefix`, but the adapter never overrides `map_peft_target_module_to_hf`, so target_modules go out unchanged. qwen2_5_omni already handles this, its override docstring describes the same failure.

**repro**

save a lora from qwen3_omni_moe, load it onto `Qwen3OmniMoeForConditionalGeneration`, and compare `get_peft_model_state_dict` against the file. 8 extra keys under `talker.`.

**fix**

pr: #3945

stanley1208:
First deliverable for #3867. Fixes #3944.

**the suite**

one registry of families, each a tiny transformers model plus the automodel adapter that owns its naming. per family: save through `Checkpointer.save_model` with `is_peft=True`, reload with `PeftModel.from_pretrained`, check tensors and logits match. a second test checks bulk and per-tensor exports agree.

covers llama (dense control), nemotron_v3, qwen3_moe, minimax_m2, qwen3_omni_moe.

i attach the adapter to the transformers model rather than building the native class, since the native moe classes call `torch.cuda.current_device()` unconditionally and cannot be built on cpu. both routes give identical output.

**the bug it found**

qwen3_omni_moe exported tensors under `thinker.` but target_modules without it, so peft also adapted `talker.model.*` and left those adapters random. `map_peft_target_module_to_hf` was never overridden; qwen2_5_omni already does this. detail in #3944.

reverting the override fails only omni. reverting the nemotron fix from #3866 fails only nemotron.

**not covered**

fused expert lora, since only nn.Linear gets wrapped here. that stays with `test_moe_peft_v5_state_dict_adapters.py`.


tests/unit_tests/checkpoint/test_peft_hf_reload_suite.py
```diff
@@ -0,0 +1,454 @@
...
+    from safetensors.torch import load_file
+
+    model, reference, peft_config = _adapted_model(family, tmp_path / "source")
+    adapter_dir = _save_through_checkpointer(model, peft_config, tmp_path)
+
+    exported = load_file(str(adapter_dir / "adapter_model.safetensors"))
+    assert exported, f"{family.id}: the export wrote no adapter tensors"
+    assert len(set(exported)) == len(exported), f"{family.id}: duplicate keys in the export"
+
+    loaded = PeftModel.from_pretrained(reference, str(adapter_dir), key_mapping={}, autocast_adapter_dtype=False).eval()
```
yuhezhang-ai, line 442: **[P1] Resolve the PEFT/Transformers reload incompatibility**

Four new reload cases fail with the repository's declared Transformers 5.15.1 and minimum PEFT 0.19.1 versions: `nemotron_v3`, `qwen3_moe`, `minimax_m2`, and `qwen3_omni_moe`. PEFT's automatic v5 conversion passes `distributed_operation` to `WeightConverter`, whose constructor no longer accepts it, raising `TypeError: WeightConverter.__init__() got an unexpected keyword argument 'distributed_operation'` before the assertions. `key_mapping={}` does not disable this conversion.

Please resolve the dependency compatibility before adding these as unconditional unit tests. The suite passes with the older Transformers 5.12.1 / PEFT 0.18.1 environment, but that does not validate the declared dependency combination.

nemo_automodel/components/models/qwen3_omni_moe/state_dict_adapter.py
```diff
@@ -84,6 +84,24 @@ def from_hf(
...
+        has no layout in which the thinker's modules are addressed without it.
+
+        Args:
+            name: A target-module name in native layout.
+            v4_compatible: Legacy export selection; omni uses one module layout for both.
+
+        Returns:
+            Target-module name in the HF omni layout.
+        """
+        return self._add_thinker_prefix(name)
```
yuhezhang-ai: **[P2] Honor the detected thinker namespace**

`from_hf()` supports standalone-thinker checkpoints and sets `_uses_thinker_prefix=False` when their expert keys omit `thinker.`. Both `to_hf()` and `convert_single_tensor_to_hf()` then preserve that layout, but this hook adds `thinker.` unconditionally. Consequently, `adapter_config.json` targets `thinker.model.layers...` while the exported tensor keys remain `base_model.model.model.layers...`, and real PEFT rejects the standalone-thinker reload with `Target modules ... not found`.

Please honor `_uses_thinker_prefix` here, matching both tensor exporters, and add a standalone-thinker reload case. I reproduced the failure through `Checkpointer.save_model`; restoring only the inherited pre-PR hook allowed reload with exact adapter tensor equality. That control used Transformers 5.12.1 / PEFT 0.18.1 to isolate this regression from the separate dependency incompatibility above.

yuhezhang-ai (commented):
Thanks, @stanley1208, for adding the shared reload suite and catching the Omni namespace mismatch. Testing the exported artifact with the real PEFT consumer is valuable coverage.

I found two issues in this revision, detailed inline. The focused CPU run of the new suite and existing Qwen3 Omni prefix tests reported **4 failed, 13 passed, 1 skipped** with Transformers 5.15.1, PEFT 0.19.1, and TorchAO 0.16.0.

yuhezhang-ai:
/ok to test [a279c5e](https://github.com/NVIDIA-NeMo/Automodel/pull/3945/commits/a279c5ef3bae31212220319251475c6895d1a2da)

yuhezhang-ai (approved):
Thanks for addressing both review findings. The target-module mapping now follows the detected full-Omni or standalone-thinker layout, and PEFT 0.20.0 resolves the Transformers 5.15.1 reload incompatibility.

I reran the shared reload suite, Omni prefix tests, and existing fused-MoE PEFT tests after merging current main: 28 passed, 7 skipped (six existing GPU-only cases and the inapplicable dense-model adapter comparison). Ruff format/check also pass. The generated dependency files have been refreshed to resolve the CI installation failures. Approving the code; full CI remains the merge gate.

yuhezhang-ai:
Approved after verifying both fixes. I merged current main and refreshed the generated dependency files on this branch; the dependency check now passes. The focused tests still report 28 passed and 7 expected skips.

Started maintainer CI proxy [#3965](https://github.com/NVIDIA-NeMo/Automodel/pull/3965) at the identical source commit `e7a0008c03ea64121e0ea81fe93202cbbde3b264`: [CI run](https://github.com/NVIDIA-NeMo/Automodel/actions/runs/35624707882). This PR remains the merge target.

ko3n1g (approved):

PR #3946, opened 2026-09-19
fix(dflash): apply activation checkpointing to the draft

kashif:
# What does this PR do ?

DFlash never wired up `distributed.activation_checkpointing` for the draft model, so the flag was silently ignored. DSpark already had this. Moved DSpark's helper into the shared `_spec_train_utils` and call it from DFlash too.

# Changelog

- Add `apply_draft_activation_checkpointing` to `_spec_train_utils.py`, shared by DSpark and DFlash.
- Wire it into `TrainDFlashRecipe.setup()` before the draft's DDP wrap (covers DFlash2, Domino, and JetSpec via inheritance).
- Move DSpark's activation checkpointing tests next to the shared helper in `test_spec_train_utils.py`.

**Pre checks**:

# Additional Information


tests/unit_tests/recipes/llm/test_spec_train_utils.py
```diff
@@ -14,17 +14,20 @@
 
 """Unit tests for the shared speculative-recipe training utilities.
 
-These pin the grad-accumulation bookkeeping and the warmup+cosine LR schedule
-that EAGLE-1/2, EAGLE-3, and DFlash now share, so a change is caught for all of
-them at once (the drift these utilities exist to prevent). Recipe-level
-integration of the same logic lives in each recipe's own grad-accum tests.
+These pin the grad-accumulation bookkeeping, the warmup+cosine LR schedule, and
+the draft activation-checkpointing wiring that EAGLE-1/2, EAGLE-3, DFlash, and
```
khazic: Only DFlash and DSpark call apply_draft_activation_checkpointing; the EAGLE-1/2 and EAGLE-3 recipes still build their draft without any AC wiring (nothing in train_eagle*.py reads distributed.activation_checkpointing). Listing them here reads as if the flag were covered there too, which is the same silent no-op this PR fixes for DFlash.

kashif: good catch, fixed the docstring to only claim DFlash and DSpark

nemo_automodel/recipes/llm/_spec_train_utils.py
```diff
@@ -64,6 +72,32 @@ def apply_draft_compile(draft_model: nn.Module, cfg_compile: Any) -> None:
...
+def apply_draft_activation_checkpointing(draft_model: nn.Module, mode: bool | str) -> None:
+    """Optionally wrap the draft's transformer layers with activation checkpointing, in place.
+
+    ``mode`` is the recipe's top-level ``distributed.activation_checkpointing``
+    YAML value (bool or ``"selective"``); no-op when falsy. These recipes wrap
+    the draft with plain ``DistributedDataParallel`` / ``fully_shard`` instead of
+    the FSDP2Manager/DDPManager path that applies AC automatically, so it must be
+    requested explicitly here, before the draft is wrapped.
+    """
+    if not mode or (isinstance(mode, str) and mode.lower() == "false"):
```
khazic: Now that this is the shared helper, it is worth matching `_normalize_activation_checkpointing` in `recipes/_dist_utils.py`, which also treats `off`, `none`, `disabled` and `no` as disabled. Here only `false` is recognized, so `distributed: {activation_checkpointing: off}` leaves the target without AC (normalized to False on its path) but fully checkpoints the draft, since `is_selective_activation_checkpointing("off")` is False and the else branch runs. Passing the normalized value in, or reusing that function here, keeps the two in step.

HuiyingLi (approved):

PR #3954, opened 2026-09-20
perf(kimi_k3): opt-in fused KDA kernels (kda_chunk_impl=fused)

yisongbetter:
# What does this PR do ?

Adds an opt-in implementation of the Kimi-K3 KDA (Kimi Delta Attention) chunked delta rule: `KimiK3TextConfig.kda_chunk_impl: "fla" | "fused"`. `"fla"` (default) is the current flash-linear-attention `chunk_kda` path, byte-for-byte unchanged. `"fused"` routes the same call to two kernels written for it: a fused CUDA forward (one CTA per head × document walking the 64-token chunks with the [K, V] state on-chip) and a Triton backward that recomputes every intermediate from the raw inputs, so the autograd function saves only q, k, v, g, beta and cu_seqlens. On the K3 SFT benchmark at 64 × GB200 this is **+2.04 % tok/s/GPU** against a same-job control (A/A −0.02 %), with the loss trajectory inside the A/A band, and 1.5 GiB less activation memory held per KDA layer-microbatch.

# Changelog

- `nemo_automodel/components/models/kimi_k3/config.py`: `kda_chunk_impl` ("fla" | "fused", validated; `"fused"` also requires `kda_mode="chunk"` and in-kernel q/k l2norm).
- `nemo_automodel/components/models/kimi_k3/model.py`: in `KimiDeltaAttention._kda_core` the single kernel call becomes a two-way branch; the FLA branch is the previous code verbatim. The fused branch first asks `fused_kda_unsupported_reason(...)` and raises with the first violated precondition (chunk mode, no context parallelism, in-kernel l2norm, bounded gate, head dims K = V = 128, bf16 q/k/v, fp32 g/beta, packed layout) instead of silently running FLA.
- `nemo_automodel/components/models/kimi_k3/kda_fused/__init__.py`: `fused_chunk_kda` (drop-in `(o, None)` return), the autograd function, the precondition check, and the extension loader: the CUDA forward is JIT-built with `torch.utils.cpp_extension.load_inline` once per build directory (`NEMO_KDA_FUSED_BUILD_DIR`, default `~/.cache/nemo_automodel/kda_fused`, keyed by sources + torch version + device arch, pinned to the current device's `-gencode`) and afterwards imported straight from the built `.so`; a fresh build lock defers to `load_inline`, a stale one (> 10 min) is removed with a warning. Dense equal-length batches (`cu_seqlens=None`) are viewed as packed documents with host-side offsets, so the dense path has no device→host sync (FLA's own `cu_seqlens_cpu` contract); packed input syncs once in the backward, as FLA's varlen path does.
- `nemo_automodel/components/models/kimi_k3/kda_fused/chunk_kda_fwd_cuda.py`: the forward kernel, its header and its torch binding as the source strings `load_inline` builds (no `.cu`/`.cpp` files in the package; device/shape/dtype/contiguity checks, `C10_CUDA_KERNEL_LAUNCH_CHECK`, the one-time dynamic-shared-memory attribute checked).
- `nemo_automodel/components/models/kimi_k3/kda_fused/chunk_kda_bwd_triton.py`: the backward kernel (`run(q, k, v, g, beta, cu_seqlens, do, dq, dk, dv, dg, dbeta, cu_seqlens_cpu=None)`).
- `pyproject.toml`: the Triton module is coverage-omitted like the other kernel modules.
- Tests: `tests/unit_tests/models/kimi_k3/test_kda_fused.py` (CPU, kernels stubbed) — config default and validation, every precondition message, routing to the fused path and the raise-instead-of-fallback rule, the autograd wrapper (saved tensors, `do` dtype, host offsets for dense batches, grads returned in the input dtypes), the prebuilt-import and stale-lock logic.

# Validation

**Measured at 64 × GB200 NVL72 (256 GPUs, EP32 × PP8 × DP32, Kimi-K3 2k-token rows, mock dataset, `nemo_automodel/recipes/llm/benchmark.py`; one job, three rungs on the same node set, job 3114806):**

| config (all on top of the current K3 GB200 benchmark stack incl. the KDA knobs of #3932) | tok/s/GPU | step s | recipe MFU | vs control |
|---|---|---|---|---|
| control (`kda_chunk_impl: fla`) | 822.8 | 39.827 | 20.7 % | — |
| `kda_chunk_impl: fused` | **839.6** | 39.030 | 21.1 % | **+2.04 %** |
| control repeated (A/A) | 822.6 | 39.836 | 20.7 % | −0.02 % |

Loss over the 12 measured steps: `fused` vs control max |Δ| 0.0006, final −0.0002; A/A vs control max |Δ| 0.0004, final 0.0000 (first step 12.4943 on all three, last 12.4540 / 12.4542 / 12.4542).

**Kernel level (one GB200, the 64-node layer-microbatch shape: q/k/v bf16 [1, 8192, 96, 128], 4 documents × 2048; CUDA events, median of 20; jobs 3109822 / 3109874 / 3109958 / 3114624 / 3114635):** forward 0.78–0.80 ms vs FLA 0.4.2 `chunk_kda` 2.07–2.19 ms, max |diff| 1–2 bf16 ulp, matched ratio 0.99926 at atol 1e-4 / rtol 1e-2 and 0.999998 against the captured training activations; backward 4.25–4.39 ms vs FLA's autograd backward 4.89–4.95 ms (FLA keeping its saved intermediates), every gradient within the per-output tolerances (dq/dk/dv/dg atol 2e-2 rtol 2e-2, dbeta atol 2e-1 rtol 5e-2, ≥ 0.99 matched), identical numerics on document layouts the kernels were not tuned on (1, 8 and 32 documents, unequal lengths).

**Layer level (one GB200, the real `KimiDeltaAttention` layer, identical weights and inputs, jobs 3114711 / 3114790 / 3114842):** output matched 1.000000 (max |diff| 1.6e-2 on |o| rms 0.39, cos 0.99998), input gradient 0.999994; a 20-step single-layer AdamW run on a fixed target tracks FLA's loss to 2.7e-4 relative. Activation-checkpointed layer-microbatch (2 forwards + backward): 32.7 → 29.3 ms with the #3932 knobs on (held after the forward 4.09 → 2.59 GiB); on this branch alone over plain `main` (FLA recomputing in its backward): 33.419 → 29.792 ms, held 2.78 → 2.59 GiB (job 3114950). Parameter gradients differ from FLA at the 0.4–1 % rms level (dt_bias 4 %) — the backward keeps its intermediates in bf16; the 64-node and 8-node loss curves above are the evidence that training is unaffected.

**Mini-scale gate (8 × GB200, 16-layer Kimi-K3 twin, job 3114745):** +2.64 % vs control (A/A +0.11 %); loss max |Δ| 0.0004 vs A/A 0.0005.

**Tests:** `tests/unit_tests/models/kimi_k3/test_kda_fused.py` (18 tests), run together with the other 15 `tests/unit_tests/models/kimi_k3/` files including #3932's `test_kda_knobs.py` on one GB200 after the rebase onto `main` a29b10155: 105 passed in 20.0 s (kernels stubbed, cold Triton cache; job 3115456 on this branch's tree b4c4b0ff2596), including an import of the Triton module with Triton unavailable (the macOS installation check); ruff clean; NVIDIA copyright headers on the new files.

**Pre checks**:

- [x] Did you write any new necessary tests? (config default/validation, every precondition message, routing, raise-instead-of-fallback, the autograd wrapper incl. dense-batch packing and host offsets, prebuilt import / stale lock — all CPU with the kernels stubbed)
- [x] Did you add or update any necessary documentation? (module docstring of `kda_fused` explaining the contract, the build/import policy, the supported shapes and that `kda_transpose_state_layout` / `kda_disable_recompute` are FLA-only; the `_validate` error messages name each precondition; inline comments at the kernel call. `KimiK3TextConfig` has no per-argument docstring section, so none was added.)

# Additional Information

- Rebased onto `main` after #3932 (the FLA KDA knobs) merged: the fused branch sits right before the FLA kernel call, so `kda_disable_recompute` and `kda_transpose_state_layout`, which configure FLA's `chunk_kda`, are not consulted on the fused path, while `kda_conv_backend` (the short convolution before the kernel) applies to both. The 64-node number above was measured with the #3932 knobs on for both control and treatment.
- The kernels are specialised to the K3 call (head dims 128, packed or dense equal-length layout, in-kernel l2norm, bounded gate, chunk mode, no CP). Any other call raises with the violated condition; `"fla"` stays the default everywhere.
- The CUDA forward needs `nvcc` in the container for its one-time build (~15 s; present in the NGC/NeMo images); the built `.so` is reused across processes and jobs through the shared build directory.
- Related to # (issue)


akoumpa:
Hi @yisongbetter can you use the load_inline facility of pytorch for the c++ code? https://docs.pytorch.org/docs/2.14/cpp_extension.html

yisongbetter:
> akoumpa

Sure, will do. Thanks.

akoumpa (dismissed):

HuiyingLi (approved):

chtruong814 (approved):

